PR #489, opened 2024-03-10
RFC: support dynamic taxonomies

booxter:
For some taxonomies, it makes more sense to generate them programmatically, which allows to generate a larger number of samples as well as control their correctness without relying on humans.

For example, generating samples of simple rule application to an input, that can be expressed as a simple pure function, e.g. transliteration rules, is more robust and correct when taxonomies are not human written. Some other taxonomies that lean more into knowledge territory may also benefit from this.

This patch implements dynamic taxonomies as Containerfile definitions that can generate the seed on invocation.

### DIFF
--- a/cli/generator/generate_data.py
+++ b/cli/generator/generate_data.py
@@ -1,9 +1,11 @@
 # Standard
 from datetime import datetime
 from functools import partial
+from itertools import chain
 from multiprocessing import Pool
 from os.path import splitext
 from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import Optional
 import json
 import os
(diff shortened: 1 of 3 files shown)


nathan-weinberg:
@bonzini @booxter It was a GitHub bug having to do with the open-sourcing, no action was taken on my part, feel free to reopen

bonzini:
I can't but @booxter can. :)

booxter:
@bonzini I will rebase the old PR and post a new one the next week. I was hoping to get initial feedback on the approach from the core team so far before I do rebase / un-draft it, but looks like perhaps I may have better luck doing it upfront.

PR #544, opened 2024-03-12
Add support for custom per-seed allow/deny word lists

n1hility:
Some taxonomy qna subjects involve usage of words currently on the deny list. Add support for (optional) allow_words and deny_words yaml values, which augment the constant default for the batch currently being processed. Seed instructions are grouped by policy to prevent cross-contamination of instructions in the same batch with different word allow/deny policies.

In my case I was looking at authoring a skill to improve/correct the models understanding of file permissions directly and in ls directory output format. `lab generate` was running forever, and it was because the word `file` was on the default deny list. 

With this change you can do something like below:

```yaml
- answer:   
  question:
  context: 
  allow_words:
  - file
  - files
  deny_words:
  - bad
 ```

### DIFF
--- a/cli/generator/generate_data.py
+++ b/cli/generator/generate_data.py
@@ -6,6 +6,7 @@ from os.path import splitext
 from pathlib import Path
 from typing import Optional
 import json
+import logging
 import os
 import random
 import re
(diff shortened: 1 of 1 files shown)


bbrowning:
For what it's worth, I'm constantly running into problems with file, files, image, and images being on the deny list when trying to generate synthetic data for anything in the Kubernetes space, or even computing space (with file on there). In my case, I end up forking InstructLab to make the source code edits to adjust the deny list, but it doesn't feel great needing to carry a fork for this. At a minimum I need a way to override this during data generation, but could also see how adjusting it per-skill may be needed with a more advanced taxonomy tree.

PR #580, opened 2024-03-13
 WIP: Communicate with the model using JSON

derekhiggins:
The text based communication with the model can be
inconsistent, resulting in much of the results being
dropped. See https://github.com/instruct-lab/cli/issues/466
    
This is an attempt to see if instructing it to generate
JSON is more consistent. Anecdotal evidence on small
instructions suggest there are less format misses then
the current regex.split approach. But this might not
be as reliable in instructions with more tokens as we
might be more likely to get un parsable JSON. More testing
needed here.

I've pushed this up early (more work is needed on this) to get opinions on
weither this idea is worth pursuing? If so I'll work on it further.

An alternative approach might be to teach the model to give more
consistent responses (see https://github.com/instruct-lab/taxonomy/pull/300 )
or putting some example consistent prompt request/responses the knowledge 
section of taxonomy?

Fixes #https://github.com/instruct-lab/cli/issues/466

### DIFF
--- a/cli/generator/generate_data.py
+++ b/cli/generator/generate_data.py
@@ -151,6 +151,15 @@ def post_process_gpt3_response(num_prompt_instructions, response, discarded_file
         f"{num_prompt_instructions + 1}. Instruction:" + response.message.content
     )
     raw_instructions = re.split("###", raw_instructions)
+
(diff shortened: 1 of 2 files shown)


derekhiggins:
Got closed in error, leaving closed in favour of https://github.com/instructlab/instructlab/pull/919

Issue #485, opened 2024-03-10
Prompt batching isn't batching


derekhiggins:
The prompt generation code is batching prompts into batches
The loop that sends those batches only seems to be using 1 prompt from each batch

It doesn't seem correct as most of the instructions in the batch aren't used

Each iteration for this loop takes a batch of prompts
https://github.com/instruct-lab/cli/blob/881b8fafbc8537d2be87f930900794bd0dde5789/cli/generator/utils.py#L95

But then only using one prompt from that batch
https://github.com/instruct-lab/cli/blob/881b8fafbc8537d2be87f930900794bd0dde5789/cli/generator/utils.py#L108

alimaredia:
@derekhiggins do you have an update on this?

derekhiggins:
> @derekhiggins do you have an update on this?

There is been a PR waiting for review that I've been rebasing for the last few weeks 
https://github.com/instructlab/instructlab/pull/688

It got closed with the repo move, so I opened a new one
https://github.com/instructlab/instructlab/pull/931

I see it needs another rebase, I'll do it now.

nathan-weinberg:
SDG code has moved to a dedicated repo: https://github.com/instructlab/sdg

PR #688, opened 2024-03-20
Simplify generation code

derekhiggins:
The post_process_gpt3_response function takes a lot of params that we don't use and returns a format we need to convert. Because of this we're looping on
datastructures for no reason. Remove it to simplify generation code.

Fixes: #485


derekhiggins:
Re-opened here https://github.com/instructlab/instructlab/pull/931

PR #823, opened 2024-04-06
Ensure the CUDA container image works on most CPUs

### LINKED ISSUE #822: CUDA container image dumps core during ilab serve on some CPU families
When the cuda Containerfile recompiles llama-cpp-python, that binary gets compiled based on the specific processor instruction set that does the container build. If you then attempt to run that container on a different enough type of processor (Intel vs AMD or older vs newer), `ilab serve` will error out with `Illegal instruction (core dumped)`. I ran into this myself when building a container image on my laptop from this Containerfile. That image ran fine on some AWS nodes with A100 GPUs but dumped core when trying to run the same image on some AWS nodes with A10 GPUs, and it was due to the different CPU generations in use.

bbrowning:
This disables AVX2, FMA, and F16C in the CMAKE_ARGS when rebuilding llama_cpp_python to ensure that the container build from this Containerfile will work on any CPU architecture as long as it has the basic `avx` instruction set.

Fixes #822


bbrowning:
Reopened as #918

PR #854, opened 2024-04-12
Restore the note about regen.json

### LINKED ISSUE #156: How to stop `lab generate` and save output for continuation later?
I may have misunderstood / misheard, but I thought it was possible to continue from an incomplete generate run given a regen.jsonl in the directory.

However, when we tried this just now via ctrl+c the generate process, no regen.jsonl file (or any file as we can see at all) was output. 

Is there a way to quit the generate process and get a WIP regen.jsonl written out for later continuing?

derekhiggins:
This was removed weeks ago during an update of the README. The code still exists and it still works.

Fixes #156

-
we could add code to remove regen.json after its used but minimising new code at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,8 @@ Detailed contribution instructions can be found in the [taxonomy repository](htt
 
    The synthetic data set will be three files in the newly created `generated` directory named `generated*.json`, `test*.jsonl`, and `train*.jsonl`.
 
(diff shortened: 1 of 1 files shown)


derekhiggins:
Re-opened here https://github.com/instructlab/instructlab/pull/929

PR #868, opened 2024-04-15
Update documented training durations

### LINKED ISSUE #867: Documented training durations are too high
the cli `generate` command up until a recent fix (https://github.com/instruct-lab/cli/pull/763 )
created a train_merlinite-7b-Q4_K_M_.....jsonl file with duplicate instructions, this meant that
instead of the default 100 instructions, training was happening on 1000's in instructions.

When training with 100 instructions as intended the process is significantly faster, we should update the docs to
represent this.

I've verified that following the colab notebook with 100 items in the trainset takes 5 minutes 
and training on a Linux Server (not a laptop) with no GPU and enough RAM takes  50 minutes  (although num epochs defaults to 1 vs 5 in the notebook)

These times are only from a single run through but indicate that training duration's  are much lower when they were documented.

I havn't tested on a MAC or Kaggle but assume I can scale the number down also

derekhiggins:
These were written when generation had a bug causing many duplicate instructions, training is signifigantly faster when using 100 instructions.

Fix #867

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ There are three options to train the model on your synthetic data-enhanced datas
 ilab train
 ```
 
-> **NOTE:** ⏳ This step can take **several hours** to complete depending on your computing resources.
+> **NOTE:** ⏳ This step can potentially take **several hours** to complete depending on your computing resources.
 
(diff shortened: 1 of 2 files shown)


derekhiggins:
Re-opened here https://github.com/instructlab/instructlab/pull/930

PR #884, opened 2024-04-17
fix serve.model_path usage

### LINKED ISSUE #880: `ilab serve` + `ilab chat` model-path behavior is wrong
a few things I noticed in serve+chat

1. in config.yaml, `chat.model` shouldn't exist. `serve.model_path` should be the only option and chat should pull from it if the user does not specify `--model`
2. `--model` doesn't seem to do anything nor is it validated (we have an issue for the validation)
3. `--model-path` in `ilab serve` should update `config.serve.model_path` so that the next chat uses the proper value

cdoern:
the config on disk has config.model and serve.model_path.

This commit fixes two things:
1. when `--model-path` is specified in `ilab serve`, the config on disk is updated to have this path
2. `ilab chat` uses `config.serve.model_path` as opposed to `config.chat.model` so now we will always chat by default with the same model that is being served

resolved #880

### DIFF
--- a/cli/chat/chat.py
+++ b/cli/chat/chat.py
@@ -442,6 +442,7 @@ def chat_cli(
     api_base,
     api_key,
     config,
+    serverconfig,
     question,
     model,
     context,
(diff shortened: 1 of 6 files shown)


cdoern:
I can't seem to re-open this... anyone else have the button?

nathan-weinberg:
@cdoern you might need to delete/recreate your fork, there's some guidance in Slack for this

PR #891, opened 2024-04-17
Enable editable for tox environments

booxter:

Enable editable installation for the cli package. This allows a developer to iterate on a test failure without pip installing from the local checkout after every change to the code base.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,7 @@ envlist = fmt, lint, unit, functional
 # are huge. This reduces venv from 5.7 GB to 1.5 GB.
 setenv =
     PIP_EXTRA_INDEX_URL=https://download.pytorch.org/whl/cpu
+use_develop = true
 
 [testenv:unit]
 deps = -r requirements-dev.txt


booxter:
@tiran I don't think it's the job for the project to document how tox works. If the team doesn't believe this should be a default behavior, so be it.

I am not sure how CI could run on a package that is significantly different from non-editable version, since presumably it's run from a PR - meaning that e.g. the issue of missing files that were forgotten to be added in git should be a non-issue. But maybe I miss some other scenario where this is still a risk.

nathan-weinberg:
@booxter Btw this was closed due to a GitHub bug having to do with the open-sourcing, no action was taken on my part, feel free to reopen

tiran:
Tox uses pyproject_api hooks to build and install packages. 

## regular

A regular installations first create an sdist (source distribution). In our case, the build system backend is `setuptools.build_meta`. Then it  use `pip install` to install the `instructlab` sources and `instructlab*.dist-info` from the wheel file into the virtual env.

```shell
$ tox -e unit
.pkg: _optional_hooks> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_sdist> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_wheel> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: prepare_metadata_for_build_wheel> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: build_sdist> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
unit: install_package> python -I -m pip install --force-reinstall --no-deps .../.tox/.tmp/package/2/instructlab-0.13.1.dev50+g6fcceed.tar.gz
```

```shell
ls .tox/unit/lib/python3.12/site-packages/ | grep instructlab
instructlab
instructlab-0.13.1.dev50+g6fcceed.dist-info
```

## editable

An editable installation creates a special wheel file that does not contain any source file . It just has a `instructlab*.dist-info` metadata directory and a `.pth` file. The `.pth file contains the absolute path to the `src` directory. It is parsed by `site` module during interpreter and has the same effect as `sys.path.append(".../src")`.

```shell
.pkg: _optional_hooks> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_sdist> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_editable> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: install_requires_for_build_editable> python -I -m pip install wheel
.pkg: build_editable> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
unit: install_package> python -I -m pip install --force-reinstall --no-deps .../.tox/.tmp/package/3/instructlab-0.13.1.dev50+g6fcceed-0.editable-py3-none-any.whl
```

```
$ unzip -l .tox/.tmp/package/5/instructlab-0.13.1.dev50+g6fcceed-0.editable-py3-none-any.whl
Archive:  .tox/.tmp/package/5/instructlab-0.13.1.dev50+g6fcceed-0.editable-py3-none-any.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
       51  04-20-2024 05:16   __editable__.instructlab-0.13.1.dev50+g6fcceed.pth
    11359  04-20-2024 05:16   instructlab-0.13.1.dev50+g6fcceed.dist-info/LICENSE
    22792  04-20-2024 05:16   instructlab-0.13.1.dev50+g6fcceed.dist-info/METADATA
       92  04-20-2024 05:16   instructlab-0.13.1.dev50+g6fcceed.dist-info/WHEEL
       45  04-20-2024 05:16   instructlab-0.13.1.dev50+g6fcceed.dist-info/entry_points.txt
       12  04-20-2024 05:16   instructlab-0.13.1.dev50+g6fcceed.dist-info/top_level.txt
      708  04-20-2024 05:16   instructlab-0.13.1.dev50+g6fcceed.dist-info/RECORD
---------                     -------
    35059                     7 files
```

## Why is this a problem?

There are scenarios were a file can be present in `src` but missing from sdist and wheel. Right now we are *probably* safe, because we are using `setuptools-scm`, use implicit `include-package-data=true`, and don't have a `MANINFEST.in` with excludes.

Python packaging is finicky and I don't trust the status quo. I have been burned by packaging issues and respins with `fix MANIFEST.in` in the past. Lot's of people get real angry when this happens to a package with > [1.5 million downloads a day](https://pypistats.org/packages/defusedxml). I'm sorry if I'm coming over as pedantic. I want to spare this project any packaging pain.

booxter:
@tiran thanks for detailed answer. This is not pedantry (or at least - it's a useful flavor of pedantry!)

What I hear is that we should be cautious when merging PRs, and I agree with that. What I slightly disagree with is that `tox` should be the tool used to validate / enforce proper packaging. The project could have it both ways - `tox` set up for simple out-of-the-box developer workflow; AND proper packaging CI job to validate (pre-)release tarballs.

At the end of the day, I myself know how to deal with this locally; I just want to spare some pain for new contributors. But I'm subjective and may misinterpret what may surprise new contributors.

PR #898, opened 2024-04-17
Remove the interaction test about "Barter Town"

### LINKED ISSUE #897: linux train spending time talking about mad max
linux_train.py is testing the model by talking about Barter Town before doing any training,
this came from the Training_a_LoRA_With_Instruct_Lab.ipynb notebook and was there was a sanity test

in linux train we are not using this output so it shouldn't be needed, it should save over a  minute (on my HW)

derekhiggins:
Remove this interaction from linux train, it was part of the Notebook that linux_train.py was based on, we're not looking at the output so it shouldn't be needed.

Fixes #897

### DIFF
--- a/cli/train/linux_train.py
+++ b/cli/train/linux_train.py
@@ -196,9 +196,6 @@ def linux_train(
         )
         return tokenizer.batch_decode([o[:-1] for o in outputs])[0]
 
-    model_generate(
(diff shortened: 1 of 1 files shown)


derekhiggins:
reopened here https://github.com/instructlab/instructlab/pull/927

PR #900, opened 2024-04-17
Refactor the CUDA container spec

### LINKED ISSUE #899: Container ships build dependencies

fabiendupont:
Modify the Containerfile for NVIDIA CUDA to:
- use the CUDA runtime image for the final stage
- use a cache mount for pip wheel cache
- install the Python packages from the pip wheel cache
- make the final container rootless

Propose a shell alias that should simplify the user experience when using the container image, avoiding using long `podman run` commands.

Modify the containerization documentation to explain this and provide steps for the user.

Update the `cuda` target in the Makefile to reflect the changes above.

Resolves #899


fabiendupont:
Closing in favor of #932

PR #916, opened 2024-04-18
Add option for knowledge yaml with self-contained document

dhilung:
# Changes

Enables knowledge yaml files with self-contained document. This will restore compatibility with `instructlab-ui` generated files.

### DIFF
--- a/src/instructlab/schema/v1/knowledge.json
+++ b/src/instructlab/schema/v1/knowledge.json
@@ -56,13 +56,30 @@
         "document": {
             "description": "The knowledge documents.",
             "type": "object",
-            "required": [
-                "repo",
-                "commit",
-                "patterns"
+            "oneOf": [
+                {
+                    "required": [
+                        "content"
+                    ]
+                },
+                {
+                    "required": [
+                        "repo",
+                        "commit",
+                        "patterns"
+                    ]
+                }
             ],
             "unevaluatedProperties": false,
             "properties": {
+                "content": {
+                    "description": "Document content.",
+                    "type": "string",
+                    "minLength": 1,
+                    "examples": [
+                        "This is a text content."
+                    ]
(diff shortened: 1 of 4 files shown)


hickeyma (changes requested):
@dhilung Do you mind rebasing?

hickeyma:
@bjhargrave Do you mind looking at this?

bjhargrave (changes requested):
It has been previously decided by project leadership that knowledge documents are not to be hosted in the taxonomy repo. Putting knowledge content in qna.yaml's would of course be counter to this decision. 

And a change like this would also need to be coordinated with the backend SDG processes since they will also need to understand that knowledge can be in the qna.yaml. 

So before this change could be considered, you would need to get buy in from project leadership and the backend team. It is not enough to change the schema and the cli.

hickeyma:
Thanks @bjhargrave for your response.

@dhilung I am putting this on hold as pers @bjhargrave response in https://github.com/instructlab/instructlab/pull/916#pullrequestreview-2031256255.

dhilung:
@bjhargrave I totally agree with the strict schema restriction for content allowed on the public [taxonomy repo](https://github.com/instructlab/taxonomy), which can be separately enforced with separate schema document. However, I think `ilab` as a versatile tool, we may support options for various ways of getting the knowledge documents. 

Users of `ilab` may not always intend to contribute to the public taxonomy repo. E.g., train a custom model for private documents. In the long term

leseb:
What's the latest on this one? Based on comments, I'd lean towards closing @dhilung thoughts? Thanks!

nathan-weinberg:
I am going to go ahead and close this for now as it's been about a month since the most recent engagement - that said please feel free to reopen this if you wish to revisit it in the future!

PR #918, opened 2024-04-19
Ensure the CUDA container image works on most CPUs

### LINKED ISSUE #822: CUDA container image dumps core during ilab serve on some CPU families

bbrowning:
This disables AVX2, FMA, and F16C in the CMAKE_ARGS when rebuilding llama_cpp_python to ensure that the container build from this Containerfile will work on any CPU architecture as long as it has the basic `avx` instruction set.

Fixes #822

This is a reopened version of #823  which accidentally got closed.

### DIFF
--- a/containers/cuda/Containerfile
+++ b/containers/cuda/Containerfile
@@ -14,7 +14,7 @@ RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cu
     && export PATH="/usr/local/cuda/bin:$PATH" \
     && export XLA_TARGET=cuda120 \
     && export XLA_FLAGS=--xla_gpu_cuda_data_dir=/usr/local/cuda
-RUN --mount=type=ssh,id=default CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install --force-reinstall --no-cache-dir llama-cpp-python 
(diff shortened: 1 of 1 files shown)


tiran (approved):

tuhinsharma121:
this works fine in my system.

bbrowning:
Due to other changes in the CUDA Containerfile, this may now actually be obsolete. I see `-mno-avx` was merged, so will see if I can reproduce any of the crashes with the current state of our CUDA Containerfile. If not, will close this PR.

hickeyma (commented):
@bbrowning Let us know how the investigations turn out and if this PR is still required or not.

bbrowning:
I have not yet been able to get reliable access to an A100 EC2 node to test again, but testing on an A10 EC2 node (which used to fail), it now works. So, I'm going to close this and will reopen it if I manage to get our container failing again on a node.

Issue #883, opened 2024-04-17
"ilab chat" temp server crashes on ctrl+C during response


leseb:
See:

```
(venv) [leseb@tarox~/cli/test_run_RNR30OPb][fix-sigint-on-temp-server] ilab chat
DEBUG 2024-04-17 09:44:02,214 server.py:62 Trying to connect to http://127.0.0.1:8000/v1...
DEBUG 2024-04-17 09:44:05,196 server.py:77 Trying port 43275...
DEBUG 2024-04-17 09:44:05,196 server.py:97 Port 43275 is available.
DEBUG 2024-04-17 09:44:05,196 server.py:101 Connection to http://127.0.0.1:8000/v1 failed. Starting a temporary server at http://127.0.0.1:43275/v1...
╭────────────────────────────────────────────────────────── system ──────────────────────────────────────────────────────────╮
│ Welcome to Chat CLI w/ MERLINITE-7B-Q4_K_M (type /h for help)                                                              │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
>>> hello                                                                                                         [S][default]
╭─────────────────────────────────────────────────── merlinite-7b-Q4_K_M ────────────────────────────────────────────────────╮
╭─────────────────────────────────────────────────── merlinite-7b-Q4_K_M ────────────────────────────────────────────────────╮
│ Hello! How can I                                                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────── elapsed 0.684 seconds ─╯
>>> disconnected                                                                                                  [S][default]
Disconnected from client (via refresh/close) Address(host='127.0.0.1', port=36076)
Process Process-1:
Traceback (most recent call last):
  File "/home/leseb/cli/test_run_RNR30OPb/venv/lib64/python3.10/site-packages/httpx/_transports/default.py", line 69, in map_httpcore_exceptions
    yield
  File "/home/leseb/cli/test_run_RNR30OPb/venv/lib64/python3.10/site-packages/httpx/_transports/default.py", line 233, in handle_request
    resp = self._pool.handle_request(req)
  File "/home/leseb/cli/test_run_RNR30OPb/venv/lib64/python3.10/site-packages/httpcore/_sync/connection_pool.py", line 216, in handle_request
    raise exc from None
  File "/home/leseb/cli/test_run_RNR30OPb/venv/lib64/python3.10/site-packages/httpcore/_sync/connection_pool.py", line 196, in handle_request
[...]

PR #925, opened 2024-04-19
fix: ignore SIGINT when running the temp server

leseb:
When we simply run `ilab chat`, a temporary server will run in the background and we can enjoy a great chat. However, if the chat response is interrupted with ctrl+C, the signal will be caught by uvicorn and then the server will be killed making further chat impossible and printing a large backtrace.

Another option was to run the temp server in a dedicated thread but this requires a bit more changes as well as the difficulty of handling the thread termination. This would have been handled by uvicorn natively since it is not setting up any signal handler but would still require us to handle it manually in our code.

Resolves: https://github.com/instructlab/instructlab/issues/883

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -246,6 +246,24 @@ test_temp_server(){
     '
 }
 
+test_temp_server_sigint(){
+    expect -c '
+        set timeout 30
+        spawn ilab chat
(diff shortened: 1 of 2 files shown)


russellb (approved):
makes sense to me - thanks for the fix!

Issue #907, opened 2024-04-18
`ilab init --non-interactive` asks interactive question if config.yaml already exists


bbrowning:
If config.yaml already exists in the current directory, `ilab init --non-interactive` interactively asks if I want to continue.

```
ilab init --non-interactive
Found config.yaml in the current directory, do you still want to continue? [y/N]:
```

It shouldn't ask any interactive questions when I give `--non-interactive` (such as in my case, where I'm running this in an automated pipeline). Instead, it should either init anyway and overwrite the current config.yaml, or error. I have a slight preference for overwriting the current one, or at least allowing an option to overwrite any existing one in a non-interactive way.

PR #926, opened 2024-04-19
fix: 'ilab init --non-interactive' does not ask questions anymore

leseb:
Previously when running `ilab init --non-interactive`, if a `config.yaml` was present it would ask whether it should be overwritten or not.
Now if the non-interactive mode is enabled the `config.yaml` will be overwritten.

Resolves https://github.com/instructlab/instructlab/issues/907
Signed-off-by: Sébastien Han [seb@redhat.com](mailto:seb@redhat.com)

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -131,15 +131,15 @@ def init(
     min_taxonomy,
 ):
     """Initializes environment for InstructLab"""
-    if exists(config.DEFAULT_CONFIG):
-        overwrite = click.confirm(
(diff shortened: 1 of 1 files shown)


alimaredia (approved):
Lgtm! Let me figure out if we want this PR in the 5/2 bugfix release or the release after that  on5/30. If we want the PR in the 5/2 release we'll merge the PR ASAP, if not we'll hold off for a couple of weeks.

nathan-weinberg (approved):

PR #933, opened 2024-04-19
ci(lint): add shell linter - Differential ShellCheck

jamacku:
## Description

Addition of a new job in CI workflow that scans all shell scripts in the repo using ShellCheck and reports only newly added defects.

## Motivation and Context

Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR.

It produces reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme).

## Screenshots

image

image

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,9 +1,17 @@
 name: Differential ShellCheck
 on:
   push:
-    branches: [ main ]
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - ''**.sh'
   pull_request:
-    branches: [ main ]
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - ''**.sh'
 
 permissions:
   contents: read


jamacku:
@russellb It is expected. 

On PR, it checks only changed shell scripts and reports only newly added defects by comparing two ShellCheck outputs (with and without changes introduced in PR).

On commit, it does the same (so it fails only when new defects are introduced) but uploads all detected defects to the Security Dashboard so you can have a look at them at some point.

russellb:
> @russellb It is expected.
> 
> On PR, it checks only changed shell scripts and reports only newly added defects by comparing two ShellCheck outputs (with and without changes introduced in PR).
> 
> On commit, it does the same (so it fails only when new defects are introduced) but uploads all detected defects to the Security Dashboard so you can have a look at them at some point.

What about an initial scan for all of the code already in the repo?

jamacku:
Here is a run of ShellCheck in main branch of my fork:
https://github.com/jamacku/instructlab/actions/runs/8750859113/job/24015248562#step:4:1

>List of shell scripts for scanning
>  ./containers/rocm/gfx-version.sh ./containers/rocm/remove-gfx.sh ./scripts/fmt.sh ./scripts/functional-tests.sh

The initial scan that you can see happens once the workflow is committed to the branch. But in my fork 19 defects were detected. Here is the Security Dashboard screenshot:

image

russellb:
I guess I’d prefer that this work like other lint jobs where we introduce the tool with a clean full run and then any further issues are blocked at PR time when new code is introducing something that is flagged.

jamacku:
> I guess I’d prefer that this work like other lint jobs where we introduce the tool with a clean full run and then any further issues are blocked at PR time when new code is introducing something that is flagged.

With this GitHub Action, you don't have to fix old defects. PRs adding new defects will still be blocked (it does differential scans). That doesn't mean that you should not look at them, but you don't have to fix them. Also, you can mark them in the Security Dashboard as false positives (once reviewed), and they won't be shown in the list.

containers/rocm/.shellcheckrc
```diff
@@ -0,0 +1 @@
+disable=SC2086
```
tiran, line 1: #980 fixes the problem. I can remove the ignore file after I have rebased my PR.

russellb:
> > I guess I’d prefer that this work like other lint jobs where we introduce the tool with a clean full run and then any further issues are blocked at PR time when new code is introducing something that is flagged.
> 
> With this GitHub Action, you don't have to fix old defects. PRs adding new defects will still be blocked (it does differential scans). That doesn't mean that you should not look at them, but you don't have to fix them. Also, you can mark them in the Security Dashboard as false positives (once reviewed), and they won't be shown in the list.

Yeah, I understand. I'm saying my preference would be:

1. fix them instead of not
2. just see what gets flagged without having to go through the security dashboard.

leseb:
@jamacku are you interested in pursuing this work? I opened https://github.com/instructlab/instructlab/pull/1211 to fix all the shellcheck issues in the codebase. It'll be great to get https://github.com/instructlab/instructlab/pull/933/commits/f1148a0edf169f9d12637869392fb1f89fadc618 as well. Thanks.

.github/workflows/differential-shellcheck.yml
```diff
@@ -0,0 +1,27 @@
...
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
```
leseb, line 12:
```suggestion
  shellcheck-lint:
```

leseb (changes requested):
We need to update the `mergify.yaml` as well, please look into this PR https://github.com/instructlab/instructlab/pull/1202/files. We need something similar for shell scripts:

```
    - or:
      - and:
        - or:
          - files~='.*\.sh$'
        - check-success=shellcheck-lint
      - and:
        - -files~='.*\.sh$'
 ```

Thanks!

File: .github/workflows/differential-shellcheck.yml [hunk 3]
```diff
@@ -0,0 +1,27 @@
+name: Differential ShellCheck
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
leseb, line 6:
```suggestion
on:
  push:
    branches:
      - "main"
      - "release-**"
    paths:
      - ''**.sh'
  pull_request:
    branches:
      - "main"
      - "release-**"
    paths:
      - ''**.sh'
```

leseb (approved):

jamacku:
Then drop the token and permissions - https://github.com/redhat-plumbers-in-action/differential-shellcheck?tab=readme-ov-file#token

Issue #911, opened 2024-04-18
ilab chat displays wrong model in the header


gshipley:
Given the following 'ilab serve' command:

`ilab serve --model-path ./ibm-merlinite-7b-mlx-q-fused-pt/ggml-model-f16.gguf`

It is expected that the 'ilab chat' command would display the correct value in the header. However, it displays the following:

`
╭───────────────────────────────────────────────╮

│ Welcome to Chat CLI w/ MERLINITE-7B-Q4_K_M (type /h for help)                                                                                                                                                   │
`

leseb:
As far as I can tell the client and serve do not negotiate the name of the model provided by the server.
I can work on this, can someone assign it to me? 

Thanks!

PR #935, opened 2024-04-19
fix: return model name from the server when chatting

leseb:
When executing in client/server mode (`ilab serve` + `ilab chat`) we now print the model that the server uses instead of printing the value from the config.

Fixes: https://github.com/instructlab/instructlab/issues/911


src/instructlab/lab.py
```diff
@@ -625,6 +626,42 @@ def chat(
...
+            models = list_models(
+                api_base=api_base,
+                tls_insecure=tls_insecure,
+                tls_client_cert=tls_client_cert,
+                tls_client_key=tls_client_key,
+                tls_client_passwd=tls_client_passwd,
+            )
+
+            # Currently, we only present a single model so we can safely assume that the first model
+            server_model = models.data[0].id if models is not None else None
```
russellb, line 643: I don't think this is a safe assumption, since you can point `lab chat` at any OpenAI compatible API endpoint.

russellb (changes requested):

src/instructlab/lab.py
leseb: That makes sense, I overlooked that. That complicates the fix a little. Do you have an alternative in mind?

russellb: I would think that if the model option was provided to `ilab chat`, that has to be included in the OpenAI request (pretty sure it is, because I know it works). If that's the case, then we can always use the one from the command line arg if present.

If none is specified, we need a sane default. That could either be from config, or just default to the first one in whatever we get back from the API. I don't have a strong opinion, really.

I think in all cases, we should be displaying whatever value we're sending in the API request.

leseb: I believe that this is what the proposed code does already, I'll add an extra line and more comments to make it clear. Thanks!

leseb:
@russellb ready for review. Thanks!

scripts/functional-tests.sh
```diff
@@ -293,6 +295,42 @@ test_temp_server_ignore_internal_messages(){
...
+        sed -i "s/merlinite-7b-lab-Q4_K_M/baz/g" config.yaml
+        spawn ilab chat
+        expect {
+            "Welcome to Chat CLI w/ BAZ" { exit 0 }
+            timeout { exit 1 }
+        }
+        sed -i "s/baz/merlinite-7b-lab-Q4_K_M/g" config.yaml
+    '
+
+    # validate that we print the model the server since it is different from the config
```
cdoern, line 324:
```suggestion
    # validate that we print the model from the server since it is different from the config
```

cdoern (approved):
looks good. I like the level of comments you added to lab.py.

one small nit, but I won't block on it.

russellb:
I'm having trouble validating this in a manual test, but it's probably my fault. Reaching out to @leseb on slack about it

anik120:
> I'm having trouble validating this in a manual test, but it's probably my fault. Reaching out to @leseb on slack about it

The functional tests were passing before @russellb tested them manually, it'll be good to have an analysis of why the functional tests say something, whereas manual tests reveal something else.

russellb:
> I'm having trouble validating this in a manual test, but it's probably my fault. Reaching out to @leseb on slack about it

It didn't behave how I expected, but it was because of contents in my `config.yaml`.

However, even in my config, I'm not sure it behaved how I would expect. It's not a new bug, though, so I could write it up in a follow-up issue.

1. I had a non-default model in my `config.yaml`
2. When I ran `ilab serve` I used `--model-path path/to/another-model`.
3. When I ran `ilab chat`, I got the model name from my config file in the `ilab chat` CLI, even though that was NOT what was being used.

I think the case not covered is:

`ilab chat` is using some value of `--model`, either via command line argument or config, that is NOT included in the list of models served by the model endpoint we hit. In the case of `ilab serve`, the server doesn't care. It's hosting one model and is going to respond regardless of what model we ask for. I think there are two improvements to consider here:

1. If it's going to proceed and work, we should probably be showing the model name we pulled from the server. If the `models` API only tells us one, that's the only one we're going to get a response from.
2. What's probably even better is that if what we have for `--model` is NOT found in the `models` endpoint, `ilab chat` should exit with an error, including the list of available models at the API endpoint.

With that change in place, we should know that whatever we're specifying for the model is what should be shown in the CLI, as we've done all the validation to ensure that's the model we're talking to.

russellb:
> @russellb , @jaideepr97 and I are working on a follow up that will change the `config.serve.model_path` when the user specifies `--model-path` in `ilab serve`. I think that will solve the unexpected behavior you found

Editing the config file as the result of a one-time configuration option doesn't sound like a good direction to me. What I would expect.

1. Use the command line arg if specified
2. Use the config if specified
3. Use a default value in the code if it's not in the config either

leseb:
Now I need https://github.com/instructlab/instructlab/pull/1095 before this one can be merged. Thanks!

anik120 (approved):

leseb:
> Now I need #1095 before this one can be merged. Thanks!

Actually, I'll revisit this again to make it smarter but it's good to go as is.

leseb:
Need this to fix the CI https://github.com/instructlab/instructlab/pull/1100

russellb:
@mergifyio rebase

src/instructlab/lab.py
```diff
@@ -667,6 +668,50 @@ def chat(
...
+        # server if they are different else let's use what the user provided
+        #
+        # 'model' will always get a value and never be None so it's hard to distinguish whether
+        # the value came from the user input or the default value.
+        # We can only assume that if the value is the same as the default value and the value
+        # from the config is the same as the default value, then the user didn't provide a value
+        # we then compare it with the value from the server to see if it's different
+        if (
+            model == config.DEFAULT_MODEL
+            and ctx.obj.config.chat.model == config.DEFAULT_MODEL
```
russellb, line 685: I think this condition needs to include verifying that the API endpoint URL is the same as the one configured for `ilab serve`. We don't want this logic running if someone did `ilab chat --endpoint-url ${SOMETHING_REMOTE}

russellb (changes requested):

russellb:
It looks like CI is still broken, as well, after rebasing on top of `main` which should have included the fix from the last comment about CI

> Need this to fix the CI https://github.com/instructlab/instructlab/pull/1100

scripts/functional-tests.sh
```diff
@@ -312,6 +314,42 @@ wait_for_server(){
...
+    ilab serve &
+    PID_SERVE=$!
+    mv models/merlinite-7b-lab-Q4_K_M.gguf models/foo.gguf
+
+    # validate that we print the model from the CLI
+    expect -c '
+        set timeout 30
+        spawn ilab chat -m bar
+        expect {
+            "Welcome to Chat CLI w/ BAR" { exit 0 }
```
bjhargrave, line 327: I think this needs to be "Welcome to InstructLab Chat"

scripts/functional-tests.sh
```diff
@@ -312,6 +314,42 @@ wait_for_server(){
...
+            timeout { exit 1 }
+        }
+    '
+
+    # validate that we print the model from the config
+    expect -c '
+        sed -i "s/merlinite-7b-lab-Q4_K_M/baz/g" config.yaml
+        spawn ilab chat
+        expect {
+            "Welcome to Chat CLI w/ BAZ" { exit 0 }
```
bjhargrave, line 337: I think this needs to be "Welcome to InstructLab Chat"

scripts/functional-tests.sh
```diff
@@ -312,6 +314,42 @@ wait_for_server(){
...
+            timeout { exit 1 }
+        }
+        sed -i "s/baz/merlinite-7b-lab-Q4_K_M/g" config.yaml
+    '
+
+    # validate that we print the model from the server since it is different from the config
+    expect -c '
+        spawn ilab chat
+        expect {
+            "Welcome to Chat CLI w/ MODELS/FOO.GGUF" { exit 0 }
```
bjhargrave, line 347: I think this needs to be "Welcome to InstructLab Chat"

bjhargrave (changes requested):

cdoern:
@leseb can we get this merged? we are looking to cut this release. Please fix the issues in the reviews above, thanks!

russellb:
> @leseb can we get this merged? we are looking to cut this release. Please fix the issues in the reviews above, thanks!

He's in Europe. We'll have to make changes if we want it in today. It shouldn't be much...

src/instructlab/lab.py
russellb: i fixed this

scripts/functional-tests.sh
russellb: i fixed these, thanks!

scripts/functional-tests.sh
```diff
@@ -324,7 +324,7 @@ test_model_print(){
         set timeout 30
         spawn ilab chat -m bar
         expect {
-            "Welcome to Chat CLI w/ BAR" { exit 0 }
+            "Welcome to InstructLab Chat w/ BAR" { exit 0 }
```
bjhargrave, line 327: "Welcome to InstructLab Chat CLI w/" still need the CLI

russellb: Ah, thanks @bjhargrave !

russellb: fixed for real this time, i hope ...

russellb:
```
╭─────────────────────────────────── system ───────────────────────────────────╮
  │ Welcome to InstructLab Chat w/ BAR (type /h for help)                        │
  ╰──────────────────────────────────────────────────────────────────────────────╯
```

This is what I'm seeing in the log, and the code ... where did you get "InstructLab Chat CLI" ?

bjhargrave (approved):

Issue #438, opened 2024-03-08
llamacpp/quantize binary in git repo


giuseppe:
currently there is a binary file distributed as part of the git repository.

I understand it simplifies the distribution, so I am opening the issue just to not track it somewhere. 

What is the plan for it?  Should it be a git submodule and built from the llama.cpp repository?

mairin:
@tiran Wondering if you might have cycles to work on this for the May 30 release. Let us know and we can pop it onto the 5/30 milestone, or we can keep it on the backlog for now.

tiran:
End of May is doable. I'm going to need somebody with a Mac book to test the new package.

markstur:
we shouldn't put a binary in git.

Ideally if we need it, we'd build it and attach it to releases.  Also dev instructions.  

Do we need a fork of llama.cpp to have fixes for quantize and build that in our build?

Also I can test on Mac

tiran:
The idea is to move the `quantize` binary into a separate project and Python package. This allows us to have a conditional dependency on the binary. MacOS users will then get the binary automatically. I have created https://github.com/tiran/instructlab-quantize/ .

```
# requirements.txt
instructlab-quantize; sys_platform == 'darwin' and platform_machine == 'arm64'
```

```python
# Python code
if sys.platform == "darwin" and  platform.machine() == "arm64":
    from instructlab_quantize import run_quantize

    run_quantize("llama.cpp/models/ggml-vocab-llama.gguf", "Q4_K_M")
```

hickeyma:
@tiran This looks like a nice approach. I can test as well once you are ready.

tiran:
Go ahead, the package is on PyPI and the PR is ready to test.

tiran:
I have extended my instructlab-quantize package for @cdoern . It now comes with Linux binaries for x86_64 and aarch64 with RHEL 8 ABI. The binaries work on all Linux distros with glibc ABI 2.27 or newer.

PR #941, opened 2024-04-19
removes os.system in lab.py

### LINKED ISSUE #530: Replace os.system("python ...") with subprocess.check_call([sys.executable, ...])
The `cli.lab` uses `os.system()` to run sub processes. The use of `os.system` is problematic for two major reasons:

* The call does not raise an exception if a sub process fails. The main program just continues to run.
* The function requires a properly shell-escaped argument. If the argument is not correctly escaped for the user's current shell, then the command can fail in all sorts of funny ways, e.g. with space or non-ASCII directory name.

I highly recommend to replace any use of `os.system()` with `subprocess.check_call()`

Instead of
```python
cmd = f"{script} --gguf {gguf_model_path} --repo {tokenizer_dir} --mlx-path {dest_model_dir}"
os.system("python {}".format(cmd))
```
do
```python
subprocess.check_call([sys.executable, script, "--gguf", gguf_model_path, "--repo", tokenizer_dir, "--mlx-path", dest_model_dir])
```

`check_call()` checks the exit status of a command and fails with an exception if the program fails.

wgahnagl:
# Changes

Resolves #530

removes the os.system in lab.py! These changes are untested since we don't have a working mac set up for this! If someone with a mac could test before merging it would be much appreciated!

re-open of closed PR #862

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -9,6 +9,7 @@ import json
 import logging
 import os
 import shutil
+import subprocess
 import sys
 import typing
 
@@ -1152,4 +1153,6 @@ def convert(model_dir, adapter_file, skip_de_quantize, skip_quantize):
         gguf_model_q_dir = f"{model_dir_fused_pt}/ggml-model-Q4_K_M.gguf"
         script = os.path.join(cli_dir, "llamacpp/quantize")
         cmd = f"{script} {gguf_model_dir} {gguf_model_q_dir} Q4_K_M"
-        os.system("{}".format(cmd))
+        subprocess.check_call(
+            [sys.executable, script, gguf_model_dir, gguf_model_q_dir, "Q4_K_M"]
+        )


nathan-weinberg:
@wgahnagl Not necessarily opposed to the current implementation but can I ask why you chose `check_call` instead of `run`? https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module

nathan-weinberg:
After reading a little closer I realize now this was called for explicitly by the original issue - @tiran could you answer the above q?

tiran:
This PR conflicts with the proposed fix for #438 . I plan to move the quantize binary into a separate package that provides a `run_quantize` function:

```
from instructlab_quantize import run_quantize
run_quantize("llama.cpp/models/ggml-vocab-llama.gguf", "Q4_K_M")
```

nathan-weinberg:
Given @tiran's comment above, I'm going to close this in favor the fix for #438

Issue #947, opened 2024-04-20
Functional tests fail with `Aborted!` message when calling `ilab chat` (RHEL 8.6, Python 3.9)


booxter:
Functional tests fail for me while waiting for the context length error in chat log.

```
+2024-04-19 22:53:59 ./scripts/functional-tests.sh:126: test_ctx_size(): timeout 100 bash -c '
        until grep -q "Message too large for context size." "$TEST_CTX_SIZE_LAB_CHAT_LOG_FILE"; do
        echo "waiting for chat error"
        sleep 1
    done
'
waiting for chat error
waiting for chat error
waiting for chat error
waiting for chat error
waiting for chat error
```

When I check the chat command log, I only see:

```
INFO 2024-04-19 22:06:54,675 _client.py:1026 HTTP Request: GET http://127.0.0.1:8000/v1/models "HTTP/1.1 200 OK"
Warning: Input is not a terminal (fd=0).
╭────────────────────────────────────────────────────── system ──────────────────────────────────────────────────────╮
│ Welcome to Chat CLI w/ MERLINITE-7B-LAB-Q4_K_M (type /h for help)                                                  │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

Aborted!
```

I disabled `standalone_mode` for `click` command for `chat`, and here is the full exception hidden by the library:

```
Traceback (most recent call last):
  File "/usr/lib64/python3.9/asyncio/selector_events.py", line 256, in _add_reader
    key = self._selector.get_key(fd)
  File "/usr/lib64/python3.9/selectors.py", line 193, in get_key
    raise KeyError("{!r} is not registered".format(fileobj)) from None
KeyError: '0 is not registered'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ihrachys/instruct-lab-gitlab/cli/.tox/functional/lib/python3.9/site-packages/prompt_toolkit/input/vt100.py", line 165, in _attached_input
    loop.add_reader(fd, callback_wrapper)
  File "/usr/lib64/python3.9/asyncio/selector_events.py", line 331, in add_reader
    self._add_reader(fd, callback, *args)
  File "/usr/lib64/python3.9/asyncio/selector_events.py", line 258, in _add_reader
    self._selector.register(fd, selectors.EVENT_READ,
  File "/usr/lib64/python3.9/selectors.py", line 360, in register
    self._selector.register(key.fd, poller_events)
[...]

PR #949, opened 2024-04-20
chat: fix AttributeError crash on -qq with short context

booxter:
Closes #948

Always pass logger to `start_prompt`.

PR #950, opened 2024-04-20
tests: use `ilab chat -qq` in func tests when sending prompts (fixes func test run for rhel8.x env)

booxter:
This is not strictly required (ilab chat may also work), but:

- it's a bit redundant when all we need is triggering the model api
- `ilab chat` has some issue with some terminals handling in dependency libraries that may affect older systems (e.g. rhel8.x)

With this patch, I was able to progress in func test execution in my rhel8 environment.

Related: https://github.com/instructlab/instructlab/issues/947

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -21,6 +21,8 @@ done
 PID_SERVE=
 PID_CHAT=
 
+chat_shot="ilab chat -qq"
+
 cleanup() {
     set +e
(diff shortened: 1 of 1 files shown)


booxter:
Requires https://github.com/instructlab/instructlab/pull/949 to work.

booxter:
Was merged as part of https://github.com/instructlab/instructlab/pull/949

PR #951, opened 2024-04-20
Don't dump test data on each generate iteration

booxter:

The series makes generate function no longer write test data over and over on each iteration. Also consolidates some common code to dump jsonl data into different files.

NOTE: this PR does not change behavior for train_* and generated_* and discarded_* files, which are still dumped on every iteration. Only test_* (seed examples from qna taxonomy) is optimized.

This is https://github.com/instructlab/instructlab/pull/510 reposted.

### DIFF
--- a/src/instructlab/generator/generate_data.py
+++ b/src/instructlab/generator/generate_data.py
@@ -297,6 +297,34 @@ def get_instructions_from_model(
     return instruction_data, discarded
 
 
+def unescape(s):
+    return bytes(s, "utf-8").decode("utf-8")
+
+
+def dump_test_data(fname, data):
+    test_data = []
+    for seed_example in data:
+        user = seed_example["instruction"]
+
+        if len(seed_example["input"]) > 0:
+            user += "\n" + seed_example["input"]
+        try:
+            test_data.append(
+                {
+                    "system": utils.get_sysprompt(),
+                    "user": unescape(user),
+                    "assistant": unescape(seed_example["output"]),
+                }
+            )
+        except TypeError as exc:
+            click.secho(
+                f"Error reading seed examples: {exc}. Please make sure your answers are verbose enough.",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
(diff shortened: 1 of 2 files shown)


tiran:
I find it useful to have the data flushed to disk on every iteration. Last week `ilab generate` on my failed after an hour and almost 90 iterations, because the internal server crashed. The result was still good enough for testing. Also it's not as bad as it may look at first. Each iteration takes tenth of seconds to minutes even on machines with an accelerator. Each write is in the order of several MB.

My test system has 16 cores and an AMD Radeon 7900XT. `ilab generate` failed at `83/100 [1:06:41<13:39, 48.21s/it]` and produced a 6 MB JSON file.

booxter:
@tiran I don't think test_* file ever changes - it contains seed examples from taxonomy, and only that; (other files - like train_* or generated_* - do, and this PR does not touch these.)

What you are interested in is the data generated by the model - which is in train_* file. This won't change.

nathan-weinberg:
@booxter @tiran can one of you post this PR to the `#frontend` upstream Slack channel so we can socialize the discussion a bit more?

booxter:
@nathan-weinberg I don't think this PR is that important to raise it for discussion, it's just a code cleanup with no functional impact except saving some jsonl file writes and - what I find more interesting - some code deduplication in generation code. I am surprised it turned out controversial (?) but I understand the project has more pressing needs than landing it.

leseb:
I believe we should decide what to do with this. @tiran could perhaps revisit your review based on https://github.com/instructlab/instructlab/pull/951#issuecomment-2070943984? Thanks!

russellb:
Apologies this sat for so long.

Now we're moving this code into its own library. My preference is to only merge bug fixes for this code in this repo in the meantime. If you'd like, you can submit this over to https://github.com/instructlab/sdg

booxter:
@russellb I will move this to sdg; please just don't close this PR so that I don't forget about it. :)

booxter:
Not worth it. Would take time to re-plant it elsewhere.

Issue #814, opened 2024-04-05
unit tests fail with python < 3.11


markstur:
tox -e unit passes on MacOS with python3.11, but with 3.9 or 3.10 the following error happens:

```
>           self.assertEqual(result.exit_code, 0)
E           AssertionError: 1 != 0

tests/test_lab_train.py:341: AssertionError
```

booxter:
@tiran let's continue the discussion here since the PR was closed and I am not sure we want to proceed with the code as was posted.

You suggested that instead of moving imports around, we can adopt `mock_mlx` or something similar. I don't think this mock function solves the same problem though. See below.

When I run test cases on python 3.9, I see that 1) the very first test case that mocks out a function, .e.g `make_data`, indeed sees MagicMock() in lab.train function (I can validate it with pdb / print / write to file); but 2) any consequent test cases that mock the same function (`make_data`) see unpatched version of the function in the same lab.train code. This looks like the first test case unpatches a global state that is reused for the second test case. `pytest` docs suggest that some class level state is shared between test cases of the same class, which suggests to me that global state is also shared (same interpreter.)

I also suspect there was some fix in unittest.mock (?) or CPython interpreter that unshared some of it, which makes the problem show for py3.10 and lower only. But I couldn't find anything of interest in 3.11 release notes.

BTW I also think that `mock_mlnx` is invalid - it executes mock.patch.dict when the decorator is applied, not when the test case is actually executed. I believe the correct implementation would look like:

```
def mock_mlx(f):
    """mlx is not available on Linux"""
    mlx_modules = {
        name: MagicMock()
        for name in ["mlx", "mlx.core", "mlx.nn", "mlx.optimizers", "mlx.utils"]
    }

    def func(*args, **kwargs):
        with patch.dict(sys.modules, mlx_modules):
            return f(*args, **kwargs)

    return func
```

(Side note: I couldn't make much progress here, partially because I struggle with the "unit" tests using complete `click` library to run functions. It would - in my mind - be better if unit tests actually trigger "units" - functions under test.)

tiran:
+1 for continuing the discussion here.

That's an great finding. You are a genius! Yes, the mock function is wrong. I made a mistake when I designed the mock helper for `mlx` package. It doesn't work as intended. On Python 3.11 and 3.12, it kinda works and tests are passing. I have a local fix with a better and simpler implementation that uses a pytest fixture instead.

[...]

PR #952, opened 2024-04-20
Use pytest fixture to mock mlx package

tiran:
# Changes

See #814

Tests now use a pytest fixture to mock-patch the mlx package. This resolves a problem with my old `mock_mlx` wrapper, which broke testing with Python 3.9 and 3.10

Kudos to Ihar Hrachyshka for discovering the problem and reporting it in issue #814.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# Standard
+from unittest import mock
+import sys
+
 # Third Party
 import pytest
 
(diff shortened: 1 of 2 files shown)


booxter:
Confirming this works. I could (with https://github.com/instructlab/instructlab/pull/962) pass test run on python3.9 / rhel8.4 machine.

bjhargrave (approved):
I can also confirm this PR allows unit tests to pass on python 3.9 on macOS. Thanks!

tiran:
> Perhaps we should now add 3.9 to the python test matrix?

My PR #954 makes it easier to test with multiple Python versions. Once the PR has landed, you can do `py39-unit` to run unit tests with Python 3.9.

bjhargrave:
> Once the PR has landed, you can do `py39-unit` to run unit tests with Python 3.9.

That is very useful for local testing if developers think to do it. But I still think that if we claim to support python 3.9, then our ci build needs to confirm that via regular testing. One could add code which runs fine on 3.11 but not on 3.9 and we may not find this until much later than we could have known if the ci build tests on 3.9.

tiran:
> That is very useful for local testing if developers think to do it. But I still think that if we claim to support python 3.9, then our ci build needs to confirm that via regular testing. One could add code which runs fine on 3.11 but not on 3.9 and we may not find this until much later than we could have known if the ci build tests on 3.9.

You also need the change for CI to run tests with multiple Python versions. A common approach is [tox-gh-action](https://pypi.org/project/tox-gh-actions/) with a Python matrix and mapping of matrix env to tox env. The project documentation for tox-gh-action has documentation how to set up a test matrix.

russellb (approved):

PR #954, opened 2024-04-22
Use matrix configuration for tox

tiran:
# Changes

The previous `tox.ini` used fixed testenvs that used the default `basepython`, which typically points to `python3`. It was not easily possible to run tests with another Python version.

The new configuration uses tox's matrix configuration feature. The default targets for tests are now `py3-unit`, `py3-unitcov`, and `py3-functional`. These targets use `python3` as base Python. With the new configurion it's now possible to run tests with different Python versions. For example `tox -e py39-unit` runs unit tests with `python3.9`.

Tox is also configured to use wheel with a common wheel env for sdist and binary wheel dist. This speeds up purelib wheel creation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,7 +74,7 @@ jobs:
         run: |
           tox \
             -x testenv:unitcov.system_site_packages=True \
-            -e unitcov
+            -e py3-unitcov
 
       - name: Install the expect package
         run: |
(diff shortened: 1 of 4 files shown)


tiran:
The PR is a prerequisite to test InstructLab with multiple Python versions. A common approach is [tox-gh-actions](https://pypi.org/project/tox-gh-actions/), Python matrix in GHA, and a mapping of GHA env to tox env in `tox.ini`.

russellb (approved):
lgtm!

PR #962, opened 2024-04-22
tests: fix test file system not isolated

booxter:
The test case was reading from `generated` directory in the current path, not a temporary directory dedicated for the test case.


booxter:
Maybe we should build a fixture to apply this across the board to all test cases...

russellb (approved):

Issue #782, opened 2024-04-03
Document knowledge doc size limit 


anik120:

mairin:
@kelbrown20 complete for taxonomy, working on CLI today

mingxzhao:
While I understand limiting the question to the token max for merlinite, why is the qna pair be limited as well? My understanding is that it is passed to the teacher model as you say but can the context limit of the teacher model not be upped to accommodate this? If the teacher model is a different model than the merlinite model is the max context length still the same for both?

bbrowning:
A larger teacher model can have a larger context window, but the default experience with the `ilab` CLI uses the merlinite-7b-Q4_K_M model that only has a 4096 context size. The `--server-context-size` parameter to `ilab generate` allows you to override this from 4096 to something larger.

And, from a documentation standpoint, you're right that the values here really depend on the model in use and its maximum context size.

mingxzhao:
Thank you for answering. To follow up, I understand that the ilab cli has this max in place, but is this max also set for the teacher model, which based on the documentation is the mixtral 8x7B? My confusion is that even if the ilab has this max, it is never passed the question and answer anyways, it is only ever passed the question with generate, so having the max be a combination of question and answer doesn't quite make sense to me.

bbrowning:
The teacher model is the same merlinite model unless the user changes the default to point to another teacher model they host themselves. And we do pass two question and answer pairs to the teacher model (as examples in the prompt we give it), which by default is the merlinite model with the 4096 max context size.

mingxzhao:
Ahhhh I see I apologize I did not realize this was for the quantized model for the cli. Is there any reason these maxes would need to be implemented in actual pr submissions?

bbrowning:
My personal opinion is that PR submissions to the taxonomy repo should be limited to whatever works by default with the `ilab` CLI. I'm not sure others agree, and there's probably a broader discussion around what role `ilab` plays in taxonomy PR submissions.

PR #971, opened 2024-04-23
Doc: Document key pair size in cli

kelbrown20:
**Description:** Size limit was documented in taxonomy in https://github.com/instructlab/taxonomy/blob/main/README.md?plain=1#L49

Adding same text to CLI readme

#782

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,8 @@ Contribute new knowledge or compositional skills to your local [taxonomy](https:
 
 Detailed contribution instructions can be found in the [taxonomy repository](https://github.com/instructlab/taxonomy/blob/main/README.md).
 
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -249,6 +249,8 @@ Contribute new knowledge or compositional skills to your local [taxonomy](https:
 
 Detailed contribution instructions can be found in the [taxonomy repository](https://github.com/instructlab/taxonomy/blob/main/README.md).
 
+> [!IMPORTANT]
+> There is a limit to how much content can exist in the question/answer pairs for the model to process. Due to this, only add a maximum of around 2300 words to your question and answer seed example pairs in the qna.yaml file.
```
russellb, line 253: This should resolve the spellcheck job failure

```suggestion
> There is a limit to how much content can exist in the question/answer pairs for the model to process. Due to this, only add a maximum of around 2300 words to your question and answer seed example pairs in the `qna.yaml` file.

```

russellb (changes requested):

russellb (approved):

Issue #977, opened 2024-04-24
nc invalid argument in functional test


leseb:
Error from the tests:

```
nc -l 8000 --keep-open 
nc: invalid option -- '-'
```

I believe the runner is running an old version of `nc` where the `--keep-open` does not exist.
Apart from the error, it's interesting to see that `nc` seems to return 0 on that error...

PR #978, opened 2024-04-24
fix: remove useless nc command

leseb:
The test does not need to listen on 8000 to active the temp server. The
temp server will run if 8000 is not present and choose a random port to
listen on. The `nc` line was not needed.

Fixes: https://github.com/instructlab/instructlab/issues/977

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -28,7 +28,8 @@ cleanup() {
     fi
     for pid in $PID_SERVE $PID_CHAT; do
         if [ -n "$pid" ]; then
-            kill $pid
+            kill "$pid"
+            wait "$pid"
         fi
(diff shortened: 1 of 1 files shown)


anik120 (changes requested):
I'm a little confused about this, what's the purpose of `temp_test_server` again? Looks like `nc` was returning a 0 exit code, so anything that was being run after that was passing regardless, but the tests kept passing anyway?

leseb:
@anik120 see this commit message that explains the reason why the test was passing. The test has always been valid btw. https://github.com/instructlab/instructlab/pull/978/commits/bc13873a4817b0bbf3b0c5fa092e7d51949a3bfd

anik120 (approved):

PR #984, opened 2024-04-24
Test with multiple Python versions

### LINKED ISSUE #983: CI: validate unit tests with python3.9
The python version is set as the minimal supported version for the package; but CI validates using python3.11. We should either have only python3.9; or both. But the main point is - CI should validate with the official dependency version.

tiran:
# Changes

Resolves #983

- use `tox-gh` to run tests for each Python env
- use `setup-python`'s pip cache instead of our own home-grown solution. It's nearly as fast as caching a tox env and less brittle.
- cache 4GB model from huggingface
- test on Ubuntu / Python 3.9, 3.10, 3.11, and 3.12
- test on macOS ARM64 / Python 3.11 (only unit tests, functional shell script does not work on macOS)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,12 +25,24 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    name: "${{ matrix.python }} on ${{ matrix.platform }}"
+    runs-on: "${{ matrix.platform }}"
     strategy:
       matrix:
         python:
+          - "3.9"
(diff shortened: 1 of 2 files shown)


booxter (approved):

russellb (approved):
This looks great. It just needs a rebase.

russellb:
I did the rebase. Now CI just needs to run again to make sure it's all still green.

russellb:
all green!

PR #922, opened 2024-04-19
Enable PyPI publishing and sigstore signing

tiran:
# Changes

**Which issue is resolved by this Pull Request:**
See #972

**Description of your changes:**

Packages are now automatically published to PyPI using GitHub Actions and PyPI's [Trusted
Publisher](https://docs.pypi.org/trusted-publishers/) workflow.

On PR merge to main branch event, a dev package like `instructlab-0.13.1.dev47+g666a84f` is pushed to Test PyPI. The version number means "47 commits after 0.13.0 tag at git commit 666a84f". This job lets us test the release and publishing workflow. It gives our users an easy way to install in-development wheels from https://test.pypi.org.

On release publish event, a new release version is signed with sigstore and pushed to production [PyPI](https://pypi.org).

I have already configured the `instructlab` project on PyPI and Test PyPI for trusted publishing.

PR #985, opened 2024-04-24
Document release strategy for Instructlab CLI

### LINKED ISSUE #958: Draft document detailing CLI release strategy
Write a document that proposes the release management process for this repo

alimaredia:
Resolves #958

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -29,7 +29,7 @@ Finetuning
 GFX
 GGUF
 GiB
-Github
+GitHub
 Gmail
 gpu
 hipBLAS
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -7,6 +7,7 @@ arge
 arXiv
 backend
 backends
+backported
 Bhandwaldar
 CLI
 cli
@@ -28,6 +29,7 @@ Finetuning
 GFX
 GGUF
 GiB
+Github
 Gmail
 gpu
 hipBLAS
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -7,6 +7,7 @@ arge
 arXiv
 backend
 backends
+backport
 backported
 Bhandwaldar
 CLI
--- a/docs/release-strategy.md
+++ b/docs/release-strategy.md
@@ -8,61 +8,58 @@ This document discusses the release strategy and processes for the
 
 Releases use a `X.Y.Z` numbering scheme.
 
-X-stream release are for major releases. At this stage in the project a major release has not been cut and we expect each release (2-3 weeks) to be a new Y-stream.
+X-stream release are for major releases. At this stage in the project a major release has not been cut and we expect each release to be a new Y-stream.
 
-Backwards compatibility will exist within a Y-stream for example within 0.1.8 and 0.1.9 and between Y-stream releases for update purposes. For example between 0.1.9 and 0.2.0.
-
-Z-stream releases are meant for critical bug and documentation fixes. Backward compatibility is guaranteed between Z-stream releases. Z-stream releases are cut as maintainers see fit.
+Z-stream releases are meant for critical bug and documentation fixes. Z-stream releases are cut as maintainers see fit.
 
 ## Schedule
 
 The project currently operates on a time-based release schedule.
 The goal is to release a new Y-stream every 2-3 weeks though the time span remains flexible based on the discretion of the maintainers team.
 
-The cadence for major releases starting from 1.0 and onward will be determined as the project matures.
+The cadence for major releases starting from 1.0 onward will be determined as the project matures.
 
 A schedule will be updated in a markdown file on the <https://github.com/instructlab/instructlab> GitHub repository.
 
 ## Release Tracking
 
-Release planning is tracked via milestones in GitHub. Milestones on GitHub will exist for the next two releases.
+Release planning is tracked via [milestones](https://github.com/instructlab/instructlab/milestones) in GitHub. Milestones on GitHub will exist for the next two releases.
 
-PRs and issues associated with the next two milestones will be prioritized for review and merging into the main branch of instructlab.
+PRs and Issues associated with the next two milestones will be prioritized for review and merging into the `main` branch of instructlab.
 
 ## Git Branches and Tags
 
 Every `X.Y` release stream gets a new branch.
 
-Each release, `X.Y.Z`, exists as a tag named `vX.Y.Z` on the `release-vX.Y` branch.
+Each release, `X.Y.Z`, exists as a tag named `vX.Y.Z`.
 
 ## Release Branch Maintenance
 
-Maintenance efforts are primarily focused on the `main` branch.
-Occasionally, critical bug fixes may be backported to the most recent
-`X.Y` branch for a Z-stream bug-fix release. Bug fixes in general are not
-actively backported to release branches.
+Maintenance efforts are only on the most recent Y-stream.
+Critical bug fixes are backported to the most recent `X.Y` branch that contains the `stable` tag for a Z-stream release.
 
 ## Release Mechanics
 
-The following are the steps for how a release gets cut.
+Release mechanics are done by a Release Manager identified for that release.
+The Release Manager is a member of the CLI Maintainers team that has agreed to take on these responsibilities.
+The Release Manager can change on a per-release basis.
+The Release Manager for each release is identified in the Description of the Milestone used to plan and track that release on GitHub.
+
+The following are the steps for how Y-stream and Z-stream releases gets cut.
 
 1. Maintainers determine a commit on the main branch that will serve as the basis for the next release.
     - For a Z-stream release skip this step.
 2. For a Y-Stream release: create a new branch in the format `release-vX.Y`.
     - For a Z-Stream release skip this step.
-3. Create a commit for the release.
-    - The subject of the commit will be `vX.Y.Z release` and there will be no body in the commit.
-4. Open up a PR with the commit for the release.
-5. Maintainers Merge PR.
-    - The PR will require 2 maintainer approvals.
-6. Create a new release on GitHub. The following is automated:
+3. For a Z-Stream release: backport all relevant commits from `main` to the `release-X.Y` branch.
+    - For a Y-Stream release skip this step.
+4. Create a new release on GitHub. The following is automated:
     - Tagging the branch on GitHub
     - Creating a change log on GitHub
     - Version number of the Python package is derived from the tag name
-7. Validate the tag release (manual testing).
-8. Publish the release on GitHub. The following is automated:
-    - Publishing to PyPI. Once published, a release cannot be overridden. Bad releases can be yank by a PyPI maintainer.
-9. Announce release on the following channels:
-    - Community Slack
-10. Create a milestone on GitHub for the next release without a milestone.
-    - The date for this milestone is set to 2-3 weeks after the next milestone. This date can be tweaked at the next refinement session.
+5. Validate the release (manual testing).
+6. Move the `stable` tag to the newest release.
+7. Announce release on the following channels:
+    - InstructLab Slack
+8. Create a milestone on GitHub for the next release without a milestone.
+    - For a Z-Stream release skip this step.
--- a/docs/release-strategy.md
+++ b/docs/release-strategy.md
@@ -1,0 +1,69 @@
+# InstructLab CLI Release Strategy
+
+This document discusses the release strategy and processes for the
+`instructlab` Python package built from the
+<https://github.com/instructlab/instructlab> git repository.
+
+## Versioning Scheme
+
+Releases use a `X.Y.Z` numbering scheme.
+
+X-stream release are for major releases. At this stage in the project a major release has not been cut and we expect each release (2-3 weeks) to be a new Y-stream.
+
+Backwards compatibility will exist within a Y-stream for example within 0.1.8 and 0.1.9 and between Y-stream releases for update purposes. For example between 0.1.9 and 0.2.0.
+
+Z-stream releases are meant for critical bug and documentation fixes. Backward compatibility is guaranteed between Z-stream releases. Z-stream releases are cut as maintainers see fit.
+
+## Schedule
+
+The project currently operates on a time-based release schedule.
+Before releasing version 1.0, the goal is to release a new Y-stream every 2-3 weeks though the time span remains flexible based on the discretion of the maintainers team.
+
+The cadence for major releases starting from 1.0 and onward will be determined as the project matures.
+
+A schedule will be updated in a markdown file on the <https://github.com/instructlab/instructlab> Github repository.
+
+## Release Tracking
+
+Release planning is tracked via milestones in Github. Milestones on Github will exist for the next two releases.
+
+Issues and PRs for milestones are set, monitored, and changed at weekly grooming sessions.
+Only PRs and issues associated with the next two milestones will be prioritized and merged into the main branch of instructlab
+
+## Git Branches and Tags
+
+Every `X.Y` release stream gets a new branch.
+
+Each release, `X.Y.Z`, exists as a tag on the `X.Y` branch.
+
+## Release Branch Maintenance
+
+Maintenance efforts are primarily focused on the `main` branch.
+Occasionally, critical bug fixes may be backported to the most recent
+`X.Y` branch for a Z-stream bug-fix release. Bug fixes in general are not
+actively backported to release branches.
+
+We will not have a X-stream release unless a total re-architecture of instructlab is occurring.
+
+## Release Mechanics
+
+The following are the steps for how a release gets cut.
+
+1. Maintainers determine a commit on the main branch that will serve as the basis for the next release.
+    - For a Z-stream release skip this step.
+2. For a y-Stream release: create a new branch in the format release-X.YY.0.
+    - For a Z-Stream release skip this step
+3. Create a commit for the release
+4. Open up a PR
+5. Maintainers Merge PR
+    - The PR will require 2 maintainer approvals.
+7. Create a new release on Github. The following is automated
+    - Publishing to PyPI
+    - Tagging the branch on Github
+    - Creating a change log on Github
+8. Announce release on the following channels
+    - Community Slack
+    - The Instructlab community mailing list
+    - Update the releases page on the Instructlab website
+9. Create a milestone on Github for the next release without a milestone
+    - The date for this milestone is set to 2-3 weeks after the next milestone. This date can be tweaked at the next refinement session
--- a/docs/release-strategy.md
+++ b/docs/release-strategy.md
@@ -17,24 +17,23 @@ Z-stream releases are meant for critical bug and documentation fixes. Backward c
 ## Schedule
 
 The project currently operates on a time-based release schedule.
-Before releasing version 1.0, the goal is to release a new Y-stream every 2-3 weeks though the time span remains flexible based on the discretion of the maintainers team.
+The goal is to release a new Y-stream every 2-3 weeks though the time span remains flexible based on the discretion of the maintainers team.
 
 The cadence for major releases starting from 1.0 and onward will be determined as the project matures.
 
-A schedule will be updated in a markdown file on the <https://github.com/instructlab/instructlab> Github repository.
+A schedule will be updated in a markdown file on the <https://github.com/instructlab/instructlab> GitHub repository.
 
 ## Release Tracking
 
-Release planning is tracked via milestones in Github. Milestones on Github will exist for the next two releases.
+Release planning is tracked via milestones in GitHub. Milestones on GitHub will exist for the next two releases.
 
-Issues and PRs for milestones are set, monitored, and changed at weekly grooming sessions.
-Only PRs and issues associated with the next two milestones will be prioritized and merged into the main branch of instructlab
+PRs and issues associated with the next two milestones will be prioritized for review and merging into the main branch of instructlab.
 
 ## Git Branches and Tags
 
 Every `X.Y` release stream gets a new branch.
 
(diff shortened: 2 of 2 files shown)


docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+2. For a y-Stream release: create a new branch in the format release-X.YY.0.
+    - For a Z-Stream release skip this step
+3. Create a commit for the release
+4. Open up a PR
+5. Maintainers Merge PR
+    - The PR will require 2 maintainer approvals.
+7. Create a new release on Github. The following is automated
+    - Publishing to PyPI
+    - Tagging the branch on Github
+    - Creating a change log on Github
```
tiran, line 63: Publishing to PyPI happens when the release is published, not when it is created. This allows us to run some checks on a release build before it gets pushed to PyPI. PyPI does not permit users to override files. A PyPI maintainer can yank a release, but they cannot override a release.

```suggestion
7. Create a new release on Github. The following is automated
    - Tagging the branch on Github
    - Creating a change log on Github
8. Validate the tag release (manual testing)
9. Publish the release on Github. The following is automated
    - Publishing to PyPI. Once published, a release cannot be overridden. Bad releases can be yank by a PyPI maintainer.
```

tiran: The document does not define how a release tag should look like. We currently prefix the release tag with `v`, so it's `vX.Y.Z`. That's a recommended schema for setuptools-scm, too.

.spellcheck-en-custom.txt
```diff
@@ -28,6 +29,7 @@ Finetuning
 GFX
 GGUF
 GiB
+Github
```
bjhargrave, line 32:
```suggestion
GitHub
```

docs/release-strategy.md
khaledsulayman: @tiran is the creation and publishing of releases a definitively separate process? I'm not sure, I haven't published a release on GH before.

If so, I agree with your suggestion.

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+Release planning is tracked via milestones in Github. Milestones on Github will exist for the next two releases.
+
+Issues and PRs for milestones are set, monitored, and changed at weekly grooming sessions.
+Only PRs and issues associated with the next two milestones will be prioritized and merged into the main branch of instructlab
+
+## Git Branches and Tags
+
+Every `X.Y` release stream gets a new branch.
+
+Each release, `X.Y.Z`, exists as a tag on the `X.Y` branch.
```
bjhargrave, line 37: https://github.com/instructlab/instructlab/blob/bc0bdf33f1e78876e109911f535d1d89918f63e3/.github/workflows/pypi.yaml#L9

assumes tags start with `v`.

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+`instructlab` Python package built from the
+<https://github.com/instructlab/instructlab> git repository.
+
+## Versioning Scheme
+
+Releases use a `X.Y.Z` numbering scheme.
+
+X-stream release are for major releases. At this stage in the project a major release has not been cut and we expect each release (2-3 weeks) to be a new Y-stream.
+
+Backwards compatibility will exist within a Y-stream for example within 0.1.8 and 0.1.9 and between Y-stream releases for update purposes. For example between 0.1.9 and 0.2.0.
```
alimaredia, line 13: @cdoern I should probably make the wording say subsequent Y-stream releases right? We're not ensuring compatibility between 0.1.9 and 0.3.X but we are 0.2.9 and 0.3.X?

anik120 (dismissed):
One thing that isn't clear to me is, why do we need "backporting" strategies/cadance established? 

It's an open source, upstream project. When I as a user use upstream project, I'm always told "use latest". 

Same thing happens for Openshift projects. If it'a project that's under "Openshift", then only we do backporting. 

If project "my-project" is instead an open source upstream project under "my-org", then there's no backporting happening there. 

Instead, the following happens: 

There's upstream and downstream. 
Upstream: my-org/my-project 
Downstream: Openshift/my-project 

Let's say Openshift 4.14 was released with my-project v0.12.0. 
A bug was discovered, and Openshift's 4.15 branch is open now. 
`my-project` fixes the bug, cuts a v0.12.1 release. 
`Openshift/my-project` v4.15 is then updated to v0.12.1, and that update is back ported to Openshift v4.14 

`my-org/my-project` is never subjected to backporting, mainly because upstream projects are supposed to be consumable by other projects (with appropriate permissions), and every project will need to set their own release cadence. Setting a release cadence for upstream projects that includes backporting is not the norm as far as I know. 

That said, maybe someone else has seen popular upstream projects that do back ports, and can come up with those examples as a counter argument?

docs/release-strategy.md
russellb: I would take out the mention of update compatibility. I'm not totally sure what that means. I think you just mean you can `pip upgrade` or whatever? I'm not sure how to NOT allow that and I don't want anyone to think we're doing anything extra clever.

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+
+Z-stream releases are meant for critical bug and documentation fixes. Backward compatibility is guaranteed between Z-stream releases. Z-stream releases are cut as maintainers see fit.
+
+## Schedule
+
+The project currently operates on a time-based release schedule.
+Before releasing version 1.0, the goal is to release a new Y-stream every 2-3 weeks though the time span remains flexible based on the discretion of the maintainers team.
```
russellb, line 20:
```suggestion
The goal is to release a new Y-stream every 2-3 weeks though the time span remains flexible based on the discretion of the maintainers team.
```

I don't think it needs a qualifier about 1.0

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+
+The cadence for major releases starting from 1.0 and onward will be determined as the project matures.
+
+A schedule will be updated in a markdown file on the <https://github.com/instructlab/instructlab> Github repository.
+
+## Release Tracking
+
+Release planning is tracked via milestones in Github. Milestones on Github will exist for the next two releases.
+
+Issues and PRs for milestones are set, monitored, and changed at weekly grooming sessions.
```
russellb, line 30: I don't think this should be here until such a thing is established in a public time and place

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+Only PRs and issues associated with the next two milestones will be prioritized and merged into the main branch of instructlab
```
russellb, line 31: "Only" seems too strong here. If as a maintainer, I jump on the PR queue and see a good bug fix that was just posted, merging it should be fine, even if it hasn't gone through a grooming session. It just seems like overhead that shouldn't be strictly always required.

```suggestion 
PRs and issues associated with the next two milestones will be prioritized for review and merging into the main branch of instructlab.
```

docs/release-strategy.md
russellb:
```suggestion
Each release, `X.Y.Z`, exists as a tag named `vX.Y.Z` on the `X.Y` branch.
```

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+Each release, `X.Y.Z`, exists as a tag on the `X.Y` branch.
+
+## Release Branch Maintenance
+
+Maintenance efforts are primarily focused on the `main` branch.
+Occasionally, critical bug fixes may be backported to the most recent
+`X.Y` branch for a Z-stream bug-fix release. Bug fixes in general are not
+actively backported to release branches.
+
+We will not have a X-stream release unless a total re-architecture of instructlab is occurring.
```
russellb, line 46: Well, I assume 1.0 would happen someday without a re-architecture.

The line could probably be removed ...

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Release Mechanics
+
+The following are the steps for how a release gets cut.
+
+1. Maintainers determine a commit on the main branch that will serve as the basis for the next release.
+    - For a Z-stream release skip this step.
+2. For a y-Stream release: create a new branch in the format release-X.YY.0.
+    - For a Z-Stream release skip this step
+3. Create a commit for the release
```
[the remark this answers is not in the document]
russellb: What would the contents of this commit include?

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+4. Open up a PR
+5. Maintainers Merge PR
+    - The PR will require 2 maintainer approvals.
```
russellb, line 59: What would the PR include?

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+7. Create a new release on Github. The following is automated
+    - Publishing to PyPI
+    - Tagging the branch on Github
+    - Creating a change log on Github
+8. Announce release on the following channels
+    - Community Slack
+    - The Instructlab community mailing list
```
russellb, line 66: I would drop this since it doesn't exist (yet, anyway)

docs/release-strategy.md
```diff
@@ -0,0 +1,69 @@
...
+    - Update the releases page on the Instructlab website
```
russellb, line 67: This doesn't exist either, right?

russellb (changes requested):

docs/release-strategy.md
anik120: Also fyi industry is moving away from using the term `grooming` to use `refinement` instead

docs/release-strategy.md
alimaredia: @tiran Do you know if there are any expectations the python community has around `pip upgrade` or if you have any comments update compatibility in general?

alimaredia:
@russellb @russellb @anik120 @markstur I've updated the release strategy document after all of your comments. The next round of discussion is now open, I'm hoping merge this by Tuesday or Wednesday.

docs/release-strategy.md
tiran: No, projects are free to define their own compatibility and update policy. People often assume CalVer or SemVer these days, but nobody is forcing projects to adopt SemVer or CalVer. Also we are < 1.0, so there is no backwards compatibility promise anyway.

russellb: To @tiran's point about pre-1.0, I think I would remove the comment about compatibility of upgrades. If we say that, I think we need to get more detailed about what that means. For example, there's a proposal about completely changing the CLI subcommand layout (a good thing, I like it to be clear!) but that breaks compatibility from a user perspective. I don't think we're quite far enough along to require maintaining backwards compatibility with all of the old commands.

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+
+The cadence for major releases starting from 1.0 and onward will be determined as the project matures.
+
+A schedule will be updated in a markdown file on the <https://github.com/instructlab/instructlab> GitHub repository.
+
+## Release Tracking
+
+Release planning is tracked via milestones in GitHub. Milestones on GitHub will exist for the next two releases.
+
+PRs and issues associated with the next two milestones will be prioritized for review and merging into the main branch of instructlab.
```
russellb, line 30:
```suggestion
PRs and issues associated with the next two milestones will be prioritized for review and merging into the `main` branch of instructlab.
```

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+
+1. Maintainers determine a commit on the main branch that will serve as the basis for the next release.
+    - For a Z-stream release skip this step.
+2. For a Y-Stream release: create a new branch in the format `release-vX.Y`.
+    - For a Z-Stream release skip this step.
+3. Create a commit for the release.
+    - The subject of the commit will be `vX.Y.Z release` and there will be no body in the commit.
+4. Open up a PR with the commit for the release.
+5. Maintainers Merge PR.
+    - The PR will require 2 maintainer approvals.
```
russellb, line 57: I don't think you can create a completely empty commit. Is the intent to put tagging of a release through a review process? I think it is, but let me know if I'm wrong ...

An alternative to having that review is to create the branch and then create a PR on that release branch that updates a ChangeLog file or something like that with a version and some changelog text (could just be a link to the github milestone? tbd ...)

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+6. Create a new release on GitHub. The following is automated:
+    - Tagging the branch on GitHub
+    - Creating a change log on GitHub
+    - Version number of the Python package is derived from the tag name
+7. Validate the tag release (manual testing).
+8. Publish the release on GitHub. The following is automated:
+    - Publishing to PyPI. Once published, a release cannot be overridden. Bad releases can be yank by a PyPI maintainer.
```
russellb, line 64: Normally the automation I would expect is that this would happen immediately and automatically after step 6 above (creating the tag in git). You have it listed here as after some manual testing. Either way is OK, but this would require a different github workflow configuration and set of steps for how to run the automation. I want to make sure we're clear on the intended workflow so I can help ensure it's configured accordingly.

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+9. Announce release on the following channels:
+    - Community Slack
```
russellb, line 66:
```suggestion
    - InstructLab Slack
```

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+Each release, `X.Y.Z`, exists as a tag named `vX.Y.Z` on the `release-vX.Y` branch.
+
+## Release Branch Maintenance
+
+Maintenance efforts are primarily focused on the `main` branch.
+Occasionally, critical bug fixes may be backported to the most recent
+`X.Y` branch for a Z-stream bug-fix release. Bug fixes in general are not
+actively backported to release branches.
+
+## Release Mechanics
```
russellb, line 45: One topic not covered is identifying who is responsible for running all of these steps. Here's a proposal:
```suggestion
## Release Mechanics

Release mechanics are done by a Release Manager identified for that release. The Release Manager is a member of the CLI Maintainers team that has agreed to take on these responsibilities. The Release Manager can change on a per-release basis. The Release Manager for each release is identified in the Description of the Milestone used to plan and track that release on GitHub.
```

russellb (changes requested):
Getting very close from my perspective, thank you!

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+Every `X.Y` release stream gets a new branch.
+
+Each release, `X.Y.Z`, exists as a tag named `vX.Y.Z` on the `release-vX.Y` branch.
+
+## Release Branch Maintenance
+
+Maintenance efforts are primarily focused on the `main` branch.
+Occasionally, critical bug fixes may be backported to the most recent
+`X.Y` branch for a Z-stream bug-fix release. Bug fixes in general are not
+actively backported to release branches.
```
leseb, line 43: I found this section to be blurry. I would be in favor of a stronger commitment on whether we do backport or not.
Also bringing clarity on the stable branches we support is important. Let's say we support the last two stable versions and everything below is not maintained anymore.

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+The project currently operates on a time-based release schedule.
+The goal is to release a new Y-stream every 2-3 weeks though the time span remains flexible based on the discretion of the maintainers team.
+
+The cadence for major releases starting from 1.0 and onward will be determined as the project matures.
+
+A schedule will be updated in a markdown file on the <https://github.com/instructlab/instructlab> GitHub repository.
+
+## Release Tracking
+
+Release planning is tracked via milestones in GitHub. Milestones on GitHub will exist for the next two releases.
```
leseb, line 28:
```suggestion
Release planning is tracked via [milestones](https://github.com/instructlab/instructlab/milestones) in GitHub. Milestones on GitHub will exist for the next two releases.
```

docs/release-strategy.md
tiran: A git tag can also be created locally with `git tag` and pushed to GitHub with `git push --tags` / `git push upstream refs/tags/tagname`. Or are you going to prevent tag pushes from users? A GH release can either use an existing tag or create a new tag on `publish`.

My #922 uses the event `github.event.action == 'published'` to push a release to PyPI and to attach the release files and sigstore signatures to the GH release. The event is fired when a privileged user clicks the green `publish` button on a GH release.

The release publish event is my favorite. We can use different approach, e.g. publish event with [deployment environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) or manual workflow dispatch. Environments can be used for additional protection, to delay release jobs, and to enforce reviews. I have never used deploy environment for any of my personal projects and I'm not familiar how they work in detail.

docs/release-strategy.md
nathan-weinberg, line 28: Why two?

docs/release-strategy.md
nathan-weinberg, line 30:
```suggestion
PRs and Issues
```

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+
+PRs and issues associated with the next two milestones will be prioritized for review and merging into the main branch of instructlab.
+
+## Git Branches and Tags
+
+Every `X.Y` release stream gets a new branch.
+
+Each release, `X.Y.Z`, exists as a tag named `vX.Y.Z` on the `release-vX.Y` branch.
```
nathan-weinberg, line 36: IMHO, we should delete all branches except for `main` and `release-0.14` currently present - none of the existing release branches follow this scheme and it's confusing frankly

docs/release-strategy.md
nathan-weinberg: I'd go further and say we start off _only_ supporting the latest `stable` branch - currently `0.14.0` (soon to be `0.14.1`) - I don't see any advantage to supporting more than one release at this stage of the project (early, fast dev cycle, etc)

docs/release-strategy.md
nathan-weinberg: the way I understand it was this:

so everytime we do a new Y-stream:
1. new branch off main, tag X.Y.0 and stable

new Z stream:
1. backport commits to existing release branch
3. new tag for Z stream off latest commit to release branch
4. move stable tag to the same commit (should match new tag in step 2)

docs/release-strategy.md
nathan-weinberg: Public email lists as well should be included here

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+    - Tagging the branch on GitHub
+    - Creating a change log on GitHub
+    - Version number of the Python package is derived from the tag name
+7. Validate the tag release (manual testing).
+8. Publish the release on GitHub. The following is automated:
+    - Publishing to PyPI. Once published, a release cannot be overridden. Bad releases can be yank by a PyPI maintainer.
+9. Announce release on the following channels:
+    - Community Slack
+10. Create a milestone on GitHub for the next release without a milestone.
+    - The date for this milestone is set to 2-3 weeks after the next milestone. This date can be tweaked at the next refinement session.
```
nathan-weinberg, line 68: Again, let's have timeline/schedule stuff be in one place - repeating ourselves just means more to maintain/more chances for these docs to be inconsistent

docs/release-strategy.md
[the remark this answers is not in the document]
nathan-weinberg: Why do we need a commit at all?

docs/release-strategy.md
nathan-weinberg: Agreed and also don't think we need to stick to a schedule here, there's a separate section in the doc for that

docs/release-strategy.md
nathan-weinberg: @cdoern and I will be doing a Z-stream release today, I will update here based on our experience

nathan-weinberg: This is what we did for `0.14.1`
1. `git checkout --track upstream/release-0.14`
2. `git tag -a v0.14.1` (create new release tag)
3. `git tag -fa stable` (move `stable` tag)
4. `git push upstream --tags --force` (force needed for moving `stable` tag)
5. Draft new release in the GitHub UI **with the existing new tag** (`v0.14.1` in this case)

nathan-weinberg:
It's worth merging or even just cross-referencing this doc with https://github.com/instructlab/instructlab/blob/main/MAINTAINERS/STABLE.md

docs/release-strategy.md
alimaredia: @lhawthorn Are the email lists active yet?  I'd be happy to include them and a release page on the website on this list in a future release.

docs/release-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+3. For a Z-Stream release: backport all relevant commits from `main` to the `release-X.Y` branch.
+    - For a Y-Stream release skip this step.
+3. Create a new release on GitHub. The following is automated:
+    - Tagging the branch on GitHub
+    - Creating a change log on GitHub
+    - Version number of the Python package is derived from the tag name
+4. Validate the release (manual testing).
+5. Move the `stable` tag to the newest release.
+6. Publish the release on GitHub. The following is automated:
+    - Publishing to PyPI. Once published, a release cannot be overridden. Bad releases can be yank by a PyPI maintainer.
```
russellb, line 64: Are you sure PyPI publishing won't happen until this step? If the tag is created at step 4 above, I think that's when PyPI will happen automatically.

https://github.com/instructlab/instructlab/blob/main/.github/workflows/pypi.yaml

russellb (approved):
I'm generally good with this -- some very minor comments inline

docs/release-strategy.md
russellb: nevermind, this is wrong.  THe latest workflow that I linked to limits publishing to the release publish event

nathan-weinberg (changes requested):

nathan-weinberg (approved):

PR #986, opened 2024-04-24
ci: Update labeler.yml with additional files/labels

nathan-weinberg:
Auto-labeler hasn't been updated in a bit, making some additions here with existing files/labels for maintainer ease

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,16 +1,40 @@
 # SPDX-License-Identifier: Apache-2.0
 
+CI/CD:
+  - changed-file:
+      - any-glob-to-any-file:
+          - .github/labeler.yml
+          - .github/workflows/**/*
+          - scripts/ruff.sh
+          - .pre-commit-config.yaml
+          - .pylintrc
+          - .spellcheck-en-custom.txt
+          - .spellcheck.yml
+          - tox.ini
+
+container:
+  - changed-file:
+      - any-glob-to-any-file:
+          - containers/**/*
+          - .dockerignore
+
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -1,16 +1,40 @@
...
           - CODE_OF_CONDUCT.md
+          - CONTRIBUTOR_RULES.md
+          - governance.md
+          - MAINTAINERS.md
           - README.md
+          - SECURITY.md
           - TROUBLESHOOTING.md
           - CONTRIBUTING/**/*
           - docs/**/*
+          - MAINTAINERS/**/*
```
russellb, line 33: It looks like this supports globbing, so `*.md` may be better for the top level?

.github/labeler.yml
```diff
@@ -1,16 +1,40 @@
 # SPDX-License-Identifier: Apache-2.0
 
+CI/CD:
+  - changed-file:
+      - any-glob-to-any-file:
+          - .github/labeler.yml
```
russellb, line 6: How about `.github/*`

nathan-weinberg: `PULL_REQUEST_TEMPLATE.md` shouldn't be included IMO

.github/labeler.yml
nathan-weinberg: Thought about that, but some Markdown files aren't docs necessarily, e.g. `PULL_REQUEST_TEMPLATE.md`

russellb (approved):

PR #987, opened 2024-04-24
Add ARM64 support to the Containerfile example

nerdalert:
- Fixes ARM64 build for the containerfile
- Bump cuda:12.4.1-devel-ubi9 to the latest

### DIFF
--- a/containers/cuda/Containerfile
+++ b/containers/cuda/Containerfile
@@ -1,10 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
-FROM nvcr.io/nvidia/cuda:12.3.2-devel-ubi9
-RUN dnf install -y python3.11 git python3-pip make automake gcc gcc-c++
+FROM nvcr.io/nvidia/cuda:12.4.1-devel-ubi9
+RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc gcc-c++
 WORKDIR /instructlab
 RUN python3.11 -m ensurepip
-RUN dnf install -y gcc
 RUN rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
 RUN dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo && dnf repolist && dnf config-manager --set-enabled cuda-rhel9-x86_64 && dnf config-manager --set-enabled cuda && dnf config-manager --set-enabled epel && dnf update -y
 RUN dnf install -y libcudnn8 nvidia-driver-NVML nvidia-driver-cuda-libs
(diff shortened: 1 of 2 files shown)


russellb:
#988 includes the `https` change, but applies it in a bunch more places, so I think we should apply that one first

nerdalert:
@russellb rolled back the https change. Noticed another issue while looking at it. The order of the current containerfile won't use the GPU unless you swap the install order that I updated the PR with.

russellb:
@nerdalert bonus points if you add some build automation for the image

russellb (approved):
lgtm, but looks like @cdoern wanted to delay merging it based on the milestone setting

leseb:
@Mergifyio rebase

leseb:
@nerdalert please rebase so we can move forward with this, thanks!
@cdoern are we good to go or do you still want to delay this?

cdoern:
@leseb looks good, just rebase and this should be good to do! cc @nerdalert

nerdalert:
@leseb @cdoern done! On rebase it picked up the x86 build arg `-mno-avx`. I just added a conditional for `aarch64`. Alternative is `Containerfile.ARM` but that's a UX decision. Be nice to still keep it simple imo. Double checked the build still works on a  Graviton ARM instance. I think I saw a big refactor in the pipe anyways. Thanks!

containers/cuda/Containerfile
```diff
@@ -16,7 +15,12 @@ RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cu
     && export XLA_FLAGS=--xla_gpu_cuda_data_dir=/usr/local/cuda
 
 ARG GIT_TAG=stable
-RUN CMAKE_ARGS="-DLLAMA_CUBLAS=on" CFLAGS="-mno-avx" python3.11 -m pip install -r https://raw.githubusercontent.com/instructlab/instructlab/${GIT_TAG}/requirements.txt --force-reinstall --no-cache-dir llama-cpp-python
+RUN if [ "$(uname -m)" = "aarch64" ]; then \
+        export CFLAGS=""; \
```
leseb, line 19: Can't we just go with:

```bash
if [[ "$(uname -m)" != "aarch64" ]]; then
  export CFLAGS="-mno-avx"
fi
```

Or is `CFLAGS` always expected?j

leseb (commented):
Small nit

containers/cuda/Containerfile
nerdalert: @leseb done! double checked on a graviton and amd64 instances.

containers/cuda/Containerfile
```diff
@@ -1,10 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
-FROM nvcr.io/nvidia/cuda:12.3.2-devel-ubi9
-RUN dnf install -y python3.11 git python3-pip make automake gcc gcc-c++
+FROM nvcr.io/nvidia/cuda:12.4.1-devel-ubi9
+RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc gcc-c++
```
leseb, line 4: Why do we need `python3.11-devel` all of the sudden?

leseb (approved):
Just left a small question but not a blocker, thanks!

containers/cuda/Containerfile
leseb: Thanks!

containers/cuda/Containerfile
nerdalert: @leseb  Good question, I should have mentioned it! aarch64 was wanting Python.h to build. So adding the headers with `python3.11-devel` fixes the dependency. Like a 300k package. ubi9 defaults to 3.9 so had to make it explicit to match paytohn3.11.

leseb: Thanks for the clarification!

leseb:
It's strange it looks like Mergify doesn't see the two approvals, can we merge this? @russellb @cdoern? Thanks!

nathan-weinberg:
@Mergifyio rebase

nathan-weinberg (approved):

PR #988, opened 2024-04-24
Prefer https over git and ssh protocols for repo cloning.

igsilya:
# Changes

HTTPS has much lower chances to be blocked in certain internal networks as well as more resilient to having issues with changing ssh keys on a GitHub server side.  HTTPS is also the default method recommended by GitHub.

Passing around ssh keys is also not needed anymore.

Resolves #

SSH access replaced with HTTPS everywhere in docs and the code.

This was originally submitted as https://github.com/instructlab/instructlab/pull/663 , but since the org and the repo got renamed I couldn't re-open that PR, so sending a new one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
-BUILD_ARGS = --ssh=default
+BUILD_ARGS =
 CENGINE ?= podman
 CONTAINER_PREFIX ?= localhost/instructlab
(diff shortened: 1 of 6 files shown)


russellb (approved):
agree - this made more sense with private repos

just approved CI to run, so need to let that finish ...

cdoern (approved):
merging this instead of https://github.com/instructlab/instructlab/pull/1004 but this has been deemed necessary ASAP.

Issue #647, opened 2024-03-15
CLI flags/configuration and examples for Linux GPU training


bbrowning:
The exact settings that need to change to successfully train with a Linux GPU can vary quite a lot by system. AMD cards vs Nvidia cards, memory available on the card, age of card, etc.

A list of the settings I've found so far that a user may want to tweak to get training working at all or to make tradeoffs in overall training speed versus resources used, nonexhaustive:

- device(s) to use
- fp16 vs bf16 precision
- quantitization, specifically using 4bit BitsAndBytes vs not
- gradient accumulation steps or disabled entirely
- gradient checkpointing enabled/disabled
- per-device training batch size
- distributed training across multiple GPUs/CPUs (may be out of scope for just config, as that's more work to setup)

Some of these are configurable by CLI flags today. Can we expose all the needed parameters via CLI flags? Do we need configuration files? Which of these are also applicable to other lab commands, such as serve, generate, test, convert?

However we expose the necessary configuration, a list of example configuration/flags to use for different setups would be nice. Show people the things to tweak to lower memory usage at the expense of speed. Perhaps some guidance on the options needed to reduce GPU memory required under popular thresholds, like 8GB, 16GB, 24GB.

My assumption here is a goal would be to give people enough knobs to turn that they can get training going on their machine without having to change the actual Python code to do it. Perhaps others disagree with that assumption? All opinions are welcome!

nathan-weinberg:
@bbrowning is this issue still relevant or can it be closed?

bbrowning:
We still don't expose all the necessary knobs so that users can get the most out of the simple Linux training in this repo without editing the source code, so this issue is still relevant in that sense. How much work we want to do to enable this depends on how far we want to push training here versus focusing on instructlab/training repo for Linux training work.

nathan-weinberg:
@RobotSail @Maxusmusti @JamesKunstle same question for you fellas here

bbrowning:
Given all the changes since this was opened, I don't think it's still exactly relevant today. Closing this as outdated, and we can open specific issues based on the current CLI workflow as needed.

PR #993, opened 2024-04-25
Auto-detect bf16 support for CUDA

### LINKED ISSUE #1006: Optimize CPU training on Linux
Currently both inference and training are multithreading. This significantly hampers performance. Other settings appear to be tuned for GPU usage, even when device_type is set to CPU.

I plan to work on a PR for this.

tiran:
# Changes

See #647

bf16 (bfloat16) is not available on older CUDA versions < 11.0 as well as devices with CUDA support level < 8.0. linux_train now detects and reports bf16 support. Training on CUDA falls back to fp16 (half precision float).

---

also closes #1006

### DIFF
--- a/src/instructlab/config.py
+++ b/src/instructlab/config.py
@@ -32,7 +32,7 @@ DEFAULT_CHUNK_WORD_COUNT = 1000
 DEFAULT_NUM_INSTRUCTIONS = 100
 DEFAULT_PROMPT_FILE = "prompt.txt"
 DEFAULT_GENERATED_FILES_OUTPUT_DIR = "generated"
-DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=300.0)
+DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=30.0)
 # use spawn start method, fork is not thread-safe
 DEFAULT_MULTIPROCESSING_START_METHOD = "spawn"
 
--- a/src/instructlab/config.py
+++ b/src/instructlab/config.py
@@ -32,7 +32,7 @@ DEFAULT_CHUNK_WORD_COUNT = 1000
 DEFAULT_NUM_INSTRUCTIONS = 100
 DEFAULT_PROMPT_FILE = "prompt.txt"
 DEFAULT_GENERATED_FILES_OUTPUT_DIR = "generated"
-DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=30.0)
+DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=300.0)
 # use spawn start method, fork is not thread-safe
 DEFAULT_MULTIPROCESSING_START_METHOD = "spawn"
 
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -194,15 +194,17 @@ def linux_train(
     torch_dtype = "auto" if device.type == "cuda" else None
     if device.type == "cpu":
         total_memory = psutil.virtual_memory().total / (1024**3)
-        if total_memory < 64:
+        if total_memory < 60:
             # Using our default model, a system with 32 GB of RAM
             # will get OOM killed using torch_dtype=None, though we
             # seem to get much better performance with this setting
-            # where there's enough memory.
+            # where there's enough memory. Using `None` makes it
+            # use float32 as opposed to float16 or bf16.
             #
-            # There's more going on here and needs deeper exploration to find
-            # the right parameters to be checking for choosing the best
-            # configuration.
+            # Anecdotally, 64 GB seems to be enough, but this calculation
+            # may come out to be slightly less than 64 GB, so we just check
+            # for 60 GB. It would be better to do a smarter calculation on
+            # the actual memory requirement here.
             torch_dtype = "auto"
 
     # torch compile fails to build, see PyTorch #124707
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -94,6 +94,7 @@ def report_cuda_device(args_device: torch.device, min_vram: int = 0) -> None:
     """Report CUDA/ROCm device properties"""
     print(f"  NVidia CUDA version: {torch.version.cuda or 'n/a'}")
     print(f"  AMD ROCm HIP version: {torch.version.hip or 'n/a'}")
+    print(f"  Supports bf16: {torch.cuda.is_bf16_supported()}")
 
     def _gib(size: int) -> str:
         return "{:.1f} GiB".format(size / 1024**3)
@@ -173,6 +174,36 @@ def linux_train(
         hpu.init()
         report_hpu_device(device)
 
+    # device register a module, e.g. torch.cpu or torch.cuda
+    device_module = getattr(torch, device.type, None)
+    # bfloat16 is not supported on older CUDA versions and devices
+    # with CUDA support level < 8.0.
+    if hasattr(device_module, "is_bf16_supported"):
+        use_bf16 = device_module.is_bf16_supported()
+        use_fp16 = not use_bf16
+    elif device.type == "cpu":
+        # TODO: check if Torch and CPU support AVX2, F16C, AVX512
+        use_bf16 = False
+        use_fp16 = False
+    else:
+        # assume bf16 supported unless device says otherwise
+        use_bf16 = True
+        use_fp16 = False
+
+    torch_dtype = "auto" if device.type == "cuda" else None
+
+    # torch compile fails to build, see PyTorch #124707
+    # scaled_dot_product_attention(): argument 'is_causal' must be bool, not SymBool
+    use_torch_compile = False
+    # if device.type == "cuda" and torch.version.cuda is not None:
+    #     # check for NVIDIA V100, A100, or H100
+    #     cap = torch.cuda.get_device_capability(device)
+    #     use_torch_compile = cap in {(7, 0), (8, 0), (9, 0)}
+
+    print(
+        f"LINUX_TRAIN.PY: {use_bf16=}, {use_fp16=}, {torch_dtype=}, {use_torch_compile=}"
+    )
+
     print("LINUX_TRAIN.PY: LOADING DATASETS")
     # Get the file name
     train_dataset = load_dataset("json", data_files=train_file, split="train")
@@ -194,6 +225,7 @@ def linux_train(
 
     if four_bit_quant:
         print("LINUX_TRAIN.PY: USING 4-bit quantization with BitsAndBytes")
+        use_bf16 = False
         use_fp16 = True
         bnb_config = BitsAndBytesConfig(
             load_in_4bit=True,
@@ -203,7 +235,6 @@ def linux_train(
         )
     else:
         print("LINUX_TRAIN.PY: NOT USING 4-bit quantization")
-        use_fp16 = False
         bnb_config = None
 
     # Loading the model
@@ -214,7 +245,7 @@ def linux_train(
 
     model = AutoModelForCausalLM.from_pretrained(
         model_name,
-        torch_dtype="auto",
+        torch_dtype=torch_dtype,
         quantization_config=bnb_config,
         config=config,
         trust_remote_code=True,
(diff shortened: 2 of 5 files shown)


tiran:
@mergifyio rebase

tiran:
@mergifyio rebase

leseb:
@tiran what's the status on this? Thanks!

tiran:
@leseb I have rebased the PR. Let's see if tests are now passing.

src/instructlab/config.py
```diff
@@ -30,7 +30,7 @@
 DEFAULT_NUM_INSTRUCTIONS = 100
 DEFAULT_PROMPT_FILE = "prompt.txt"
 DEFAULT_GENERATED_FILES_OUTPUT_DIR = "generated"
-DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=30.0)
+DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=300.0)
```
leseb, line 33: That's going to cause issue if the client tries to connect to the server (during a chat session) and the latter crashed, users will wait for 5min... Can we change the client timeout for the user case (train?) that needs 5min?

leseb (changes requested):
Holding off based on https://github.com/instructlab/instructlab/pull/993/files#r1611590560. Thanks

src/instructlab/config.py
russellb: are you sure? I would expect the connection to be immediately rejected if there's not a process listening on that port anymore.

russellb:
@tiran have you been able to test this with CPU based training? On my linux laptop with 32 GB RAM, I haven't been able to get it to work yet.

When testing #1225 and #1222, `ilab` got OOM killed.  When running this PR, the laptop is completely unresponsive after a much longer time period, so I'm not positive what happened this time yet.

russellb:
> @tiran have you been able to test this with CPU based training? On my linux laptop with 32 GB RAM, I haven't been able to get it to work yet.
> 
> When testing #1225 and #1222, `ilab` got OOM killed. When running this PR, the laptop is completely unresponsive after a much longer time period, so I'm not positive what happened this time yet.

After the laptop recovered, I was able to see what happened.  It's failing to load the model.

```
LINUX_TRAIN.PY: NOT USING 4-bit quantization
LINUX_TRAIN.PY: LOADING THE BASE MODEL
Loading checkpoint shards:  67%|██████████████████████████████████████████████████████████                             | 2/3 [00:04<00:02,  2.49s/it]Killed
```

I think we're going to have to look at how much system memory is present and adjust settings further.

russellb:
I just pushed a change to limit the `dtype` optimization for when the system has >= 64 GB RAM. That gets it working on my CPU + 32 GB system again.

Next I'll validate the performance difference of CPU training on a system with more RAM.

src/instructlab/config.py
russellb: either way it's not clear why this was necessary, so I'll drop it from the PR for now

russellb (approved):
I've tested this on a couple of CPU environments and it looks good.

On the 32 GB env, I validated that it doesn't get OOM killed and works as it did before.

I tried on a CPU-only system with 128 GB RAM and see a drastic improvement in `ilab train` performance. I did generate with only 5 instructions and ran `ilab train` with a single epoch for testing purposes. That config with this PR had `ilab train` finish in about 12 minutes. Without the PR, it's still running and on track to be $N hours (not sure, but much, much longer).

I'm going to kick off the e2e CI workflow next to do one more validation of the cuda code path.

russellb:
I placed a `hold` on this while we wait for the e2e workflow to run which will run the cuda code path: https://github.com/instructlab/instructlab/actions/runs/9309130759

russellb:
Also I'm not sure my approval should count since the PR now includes a change I added.  I'll leave the `hold` on it for that reason, as well, to make sure it gets adequate review.

src/instructlab/train/linux_train.py
```diff
@@ -194,6 +242,7 @@ def linux_train(
 
     if four_bit_quant:
         print("LINUX_TRAIN.PY: USING 4-bit quantization with BitsAndBytes")
+        use_bf16 = False
```
cdoern, line 245: yeah, I was thinking this should go here. I noticed we were doing this already by setting it to !fp16 below (I think?)

cdoern (commented):
This won't harm my pytorch rewrite. If people want to get this in first I am fine with that, though I'd prefer the other way around

russellb:
> This won't harm my pytorch rewrite. If people want to get this in first I am fine with that, though I'd prefer the other way around

For clarity, you're referring to #1157. I'm OK with that going in first if it's ready. It looks like it may still need more review. At least I don't think it'll be hard to apply these changes to it if this is ready first.

cdoern:
@tiran if you had a chance could you give #1157 a look and see if it's ready? I believe I addressed all your comments

src/instructlab/train/linux_train.py
```diff
@@ -173,6 +175,52 @@ def linux_train(
...
+    elif device.type == "cpu":
+        # TODO: check if Torch and CPU support AVX2, F16C, AVX512
+        use_bf16 = False
+        use_fp16 = False
+    else:
+        # assume bf16 supported unless device says otherwise
+        use_bf16 = True
+        use_fp16 = False
+
+    torch_dtype = "auto" if device.type == "cuda" else None
```
markmc, line 194: I think it would be super useful to document why torch_dtype=None is faster on CPUs 

So on CPUs, `dtype=float32` will be faster than `dtype=bfloat16`. Why? A link to a nice explanation of that would be great

I guess `torch_dtype=None` gives a `dtype=float32` on CPUs? Why? I can't quickly find any docs that explains that - does None just mean that we use whatever torch.get_default_dtype() gives, which is float32 by default? 

What dtype does `torch_dtype=auto` give on CPUs? Sounds like `dtype=bfloat16`? Why? It's detecting that this particular model was saved with that dtype?

Can't seem to find the docs on hugginface.co/docs so see: https://github.com/huggingface/transformers/blob/f3f640dce14bee3b3930a774c3dfac92977eee7f/src/transformers/modeling_utils.py#L2878-L2898

We don't have a dtype in the model's [config.json](https://huggingface.co/instructlab/merlinite-7b-lab-GGUF/blob/main/config.json) so that doesn't seem to be a factor, but it could be

If the torch_dtype=auto behavior is model-specific, but we know we want float32 except on low-memory systems ... maybe for CPUs, we should just explicitly set torch_dtype to either float32 or bfloat16?

tiran: It very much depends on the hardware and compiler. In general, x86_64 CPUs have support for standard precision and double precision floats (fp32, fp64). Half precision instructions (fp16) were added in ISA level x86_64-v3 and brain float (bf16) SIMD instruction were added in x86_64-v4. The document https://pytorch.org/docs/stable/amp.html#ops-that-can-autocast-to-float16 explains autocasting.

AFAIK we want to use

* *bf16* when the CPU or hardware accelerator supports it
* fall back to *fp16* on Nvidia hardware with CUDA compute level < 8.0
* fall back to *fp32* on CPUs

src/instructlab/train/linux_train.py
```diff
@@ -173,6 +175,52 @@ def linux_train(
...
+            # We know 64 GB is enough to work. If we have less memory,
+            # go back to letting PyTorch decide the dtype.
+            #
+            # Yes, it does seem like "auto" could use some improvement,
+            # but right now "auto" does not choose the best performing
+            # type for CPU based training.
+            torch_dtype = "auto"
+            print(
+                "LINUX_TRAIN.PY: NOTE - Adjusting configuration for system with less than 64 GB RAM. A system with at least 64 GB RAM will have better performance."
+            )
```
russellb, line 210: I need to remove this message, because it doesn't seem to be true in all cases. 

My laptop with 32 GB of RAM is performing similar as the much larger server -- it finished in about 13 minutes compared to 12.  With `torch_dtype=None` it will run out of memory.

Meanwhile, on that powerful server, using `torch_dtype="auto"` absolutely kills performance.  I know it would have run for many, many hours.

There's definitely more going on here. Maybe it's a difference of instructions available on the different CPUs in these environments.

src/instructlab/train/linux_train.py
russellb: Based on @tiran 's comment, it's possible that my laptop has this support while my server does not:

> * **bf16** when the CPU or hardware accelerator supports it

which could explain why `"auto"` gets the best performance on my laptop.

If `"auto"` isn't doing an adequate job of checking if that is supported and is choosing it even if it's not actually supported on my server, maybe that's killing the performance. I'll have to keep digging here ...

russellb:
> I placed a `hold` on this while we wait for the e2e workflow to run which will run the cuda code path: https://github.com/instructlab/instructlab/actions/runs/9309130759

This passed.

src/instructlab/train/linux_train.py
tiran: What is the output of `torch.backends.cpu.get_cpu_capability()` on your server and your laptop?

russellb: On my laptop -- 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz

```
>>> torch.backends.cpu.get_cpu_capability()
'AVX512'
```

On the server -- AMD EPYC 7R32

```
>>> torch.backends.cpu.get_cpu_capability()
'AVX2'
```

leseb (changes requested):
Kindly request to edit the newly added `CHANGELOG.md` file, since this looks like a nice improvement for CPU-only system. Thanks!

russellb:
I spoke with @leseb on Slack and we determine that the memory check came out to `62` on his 64 GB system, so I've changed the rough check in the code to now be `< 60` instead of `< 64`. I'd like to see if that now gets him a boost, as his system should work with dtype=None (using float32).

src/instructlab/train/linux_train.py
```diff
@@ -173,6 +175,50 @@ def linux_train(
...
+        use_fp16 = False
+    else:
+        # assume bf16 supported unless device says otherwise
+        use_bf16 = True
+        use_fp16 = False
+
+    torch_dtype = "auto" if device.type == "cuda" else None
+    if device.type == "cpu":
+        total_memory = psutil.virtual_memory().total / (1024**3)
+        if total_memory < 60:
```
leseb, line 197:
```suggestion
        if total_memory < 62:
```

A system with 64GB of RAM, will report:

```
>>> import psutil
>>> mem = psutil.virtual_memory()
>>> mem
svmem(total=67228049408, available=31099351040, percent=53.7, used=35383861248, free=468701184, active=27983499264, inactive=37159084032, buffers=1079336960, cached=30296150016, shared=2109440, slab=1340628992)
```

And we have. `67228049408` Bytes converted to GiB gives us `67228049408 / 1024 ** 3` gives us `62.6 GiB`

src/instructlab/train/linux_train.py
```diff
@@ -194,15 +194,17 @@ def linux_train(
...
             # will get OOM killed using torch_dtype=None, though we
             # seem to get much better performance with this setting
-            # where there's enough memory.
+            # where there's enough memory. Using `None` makes it
+            # use float32 as opposed to float16 or bf16.
             #
-            # There's more going on here and needs deeper exploration to find
-            # the right parameters to be checking for choosing the best
-            # configuration.
+            # Anecdotally, 64 GB seems to be enough, but this calculation
```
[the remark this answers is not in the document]
russellb: Since it's such a rough guess, 60 still seems fine? We need to actually do some math at some point ...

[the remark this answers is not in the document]
leseb: I'll share my math in a few :) stay tuned!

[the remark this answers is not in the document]
leseb: Screenshot 2024-06-07 at 14 23 04

Some more numbers:

* The training part take ~30GB of RAM to process, there is a very small chance that this could work on very minimal Linux installation, by minimal I mean, only system critical services run and nothing else.
* The inference part takes ~35GB of RAM

Essentially a system with 48GB of RAM should be able to run both training and inferencing. Although 48 GB of RAM is not very common.

courtneypacheco:
Hi @tiran! Are you still working on this PR? We're looking to do some housekeeping and close out stale PRs, including drafts.

If we don't hear back within 7 days, we will close this PR, but please know that you are more than welcome to reopen it if you'd like! Thank you!

PR #994, opened 2024-04-25
Refactor CUDA Container file

### LINKED ISSUE #899: Container ships build dependencies

tiran:
# Changes

Resolves #899
Resolves #932

Modify the Containerfile for NVIDIA CUDA to:

* use multi-stage build with CUDA runtime image for the final stage
* update to CUDA 12.4.1
* install the Python packages in a virtualenv that is copied in the final container
* mimic the layout and behavior of `ubi9/python-311` container
* make the final container rootless
* install `flash-attn` and `bitesandbytes`

Propose a shell alias that should simplify the user experience when using the container image, avoiding using long podman run commands.

Modify the containerization documentation to explain this and provide steps for the user.

Update the `cuda` target in the Makefile to reflect the changes above.

The changeset is co-authored by Fabien Dupont and is built on top of previous work by Charlie Doern and Fabien Dupont.


containers/cuda/ilab-cuda.alias
```diff
@@ -0,0 +1,22 @@
...
+ilab-cuda() {
+    declare INSTRUCTLAB_IMAGE=${INSTRUCTLAB_IMAGE:-localhost/instructlab:cuda}
+    declare GPU_INDEX=${GPU_INDEX:-0}
+    declare ILABDIR="${HOME}/.config/instructlab"
+
+    mkdir -p "${ILABDIR}"
+    podman unshare chown 1001:1001 -R "${ILABDIR}"
+
+    if [ ! -f "${ILABDIR}/config.yaml" ] && [ "$1" != "init" ]; then
+        echo "Error: \`config.yaml\` does not exists, please run \`ilab init\` or point to a valid configuration file using \`--config=<path>\`."
```
leseb, line 12:
```suggestion
        echo "Error: \`config.yaml\` does not exist, please run \`ilab init\` or point to a valid configuration file using \`--config=<path>\`."
```

docs/containerization.md
```diff
@@ -6,61 +6,98 @@ using dedicated GPUs. This guide shows you how to put the `ilab` CLI, all of its
...
-    && export PATH="/usr/local/cuda/bin:$PATH" \
-    && export XLA_TARGET=cuda120 \
-    && export XLA_FLAGS=--xla_gpu_cuda_data_dir=/usr/local/cuda
-RUN --mount=type=ssh,id=default CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install --force-reinstall --no-cache-dir llama-cpp-python 
-RUN --mount=type=ssh,id=default python3.11 -m pip install git+ssh://git@github.com/instructlab/instructlab.git@stable
-CMD ["/bin/bash"]
+## Build the `ilab` container image
+
+We encourage you to read the [Containerfile](../containers/cuda/Containerfile).
+understand the following explanation.
```
leseb, line 12: I think there is something wrong with both sentences, especially the second one. Was the sentence cut or something?

docs/containerization.md
```diff
@@ -6,61 +6,98 @@ using dedicated GPUs. This guide shows you how to put the `ilab` CLI, all of its
...
+
+For convenience, we have created a `cuda` target in the Makefile, so building is
+as simple as `make cuda`.
+
+The default image name is `localhost/instructlab`, but you can override it with
+the `CONTAINER_PREFIX` environment variable.
+
+## Configure Podman with NVIDIA container runtime
+
+To configure your machine running RHEL 9.4+, you can follow NVIDIA's documentation to
```
leseb, line 31: Why are we pointing to RHEL and not Fedora? This is a community repository.

containers/cuda/ilab-cuda.alias
```diff
@@ -0,0 +1,22 @@
+# shellcheck shell=bash
+
+ilab-cuda() {
+    declare INSTRUCTLAB_IMAGE=${INSTRUCTLAB_IMAGE:-localhost/instructlab:cuda}
+    declare GPU_INDEX=${GPU_INDEX:-0}
+    declare ILABDIR="${HOME}/.config/instructlab"
+
+    mkdir -p "${ILABDIR}"
+    podman unshare chown 1001:1001 -R "${ILABDIR}"
```
leseb, line 9: Should we validate the presence of podman first?

containers/cuda/ilab-cuda.alias
```diff
@@ -0,0 +1,22 @@
+# shellcheck shell=bash
+
```
leseb, line 2: Should we have an OS check and fail if not Linux? Does this work on MacOS?

containers/cuda/ilab-cuda.alias
tiran: The alias file was originally written by Fabien. I could drop it from my PR and ask Fabien to submit it in a separate PR when is is available.

docs/containerization.md
tiran: Let me double check. There was a merge conflict and I might have picked the wrong resolution strategy,

containers/cuda/ilab-cuda.alias
tiran: I have removed the alias file from the PR and its documentation. We can add it later.

docs/containerization.md
tiran: I have removed the reference to RHEL. I also added a comment that the documentation assumes Fedora-like Linux distribution with Podman and SELinux + some hints how to make it work on distros with Docker.

Makefile
```diff
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
-BUILD_ARGS = --ssh=default
+BUILD_ARGS =
```
leseb, line 3: This is being fixed in https://github.com/instructlab/instructlab/pull/988, it'd be great to align :)

cdoern: yep, this makes sense now too since the repos are open :)

cdoern: yep this makes sense too since the repos are open :)

containers/cuda/Containerfile
```diff
@@ -1,20 +1,91 @@
 # SPDX-License-Identifier: Apache-2.0
+#
+# This container file requires a RHEL subscription to build and
```
cdoern, line 3: this is a good note to add.

@tiran do you think this is a barrier and we should use the cuda ubuntu img instead?

containers/cuda/Containerfile
```diff
@@ -1,20 +1,91 @@
...
+    chown -R 1001:0 "${VIRTUAL_ENV}/src" && \
+    useradd -u 1001 -g 0 -c "Default Application User" -d /opt/app-root/src -s /sbin/nologin default
+
+USER 1001
+ENV HOME="${VIRTUAL_ENV}/src" \
+    PS1="(app-root) \w\$ "
+WORKDIR "${HOME}"
+VOLUME ["/opt/app-root/src"]
+ENTRYPOINT []
+CMD ["/opt/app-root/bin/ilab"]
```
cdoern, line 91: what is the point in having ilab be the cmd? users will need to `ilab init` `ilab download` etc anyway, may as well just have it bin/bash

docs/containerization.md
```diff
@@ -1,66 +1,100 @@
-# Putting `ilab` in a Container AND making it go fast
+# Putting InstructLab in a Container AND making it go fast on
```
cdoern, line 1:
```suggestion
# Putting InstructLab in a Container AND making it go fast
```

containers/cuda/Containerfile
```diff
@@ -1,20 +1,91 @@
...
+    find ${VIRTUAL_ENV} -name __pycache__ | xargs rm -rf && \
+    chown -R 1001:0 ${VIRTUAL_ENV}
+
+COPY --chown=0:0 . /tmp/instructlab/
+RUN ${VIRTUAL_ENV}/bin/pip install --no-deps /tmp/instructlab && \
+    find ${VIRTUAL_ENV} -name __pycache__ | xargs rm -rf && \
+    chown -R 1001:0 ${VIRTUAL_ENV}
+
+
+FROM nvcr.io/nvidia/cuda:${CUDA_VERSION}-runtime-ubi9
```
cdoern, line 56: nice diff between the devel and runtime imgs. I was wondering what the best way to do this was!

docs/containerization.md
```diff
@@ -1,66 +1,100 @@
...
+sudo nvidia-ctk cdi generate --output=/etc/cdi/nvidia.yaml
 ```
 
-Voila! You now have a container with CUDA and GPUs enabled!
+### Run the CUDA-accelerated InstructLab container
 
-### Sources
+When running our model, we want to create some paths that will be mounted in
+the container to provide data persistence. As an unprivileged user, you will
+run a rootless container, so you need to let the internal user (UID 1001)
```
cdoern, line 75: `podman run --privileged` might be easier for users to digest here. Though, not sure it accomplishes 100% of what you are trying to do...

docs/containerization.md
```diff
@@ -1,66 +1,100 @@
...
+podman run --rm -it --user 1001 --device nvidia.com/gpu=0 --volume $HOME/.config/instructlab:/opt/app-root/src:z localhost/instructlab:cuda
+```
+
+The above command will print the help, as we didn't pass any argument.
 
-Thanks to Taj Salawu for figuring out how to pass the git ssh keys properly!
+Let's initialize our configuration, download the model and start the chatbot.
+
+```shell
+podman run --rm -it --device nvidia.com/gpu=0 --volume $HOME/.config/instructlab:/opt/app-root/src:z localhost/instructlab:cuda init
```
cdoern, line 97: I don't know if I like this flow, I think we should have a way as well for users to run all of these in a ctr, rather than passing in the command they want to run.

Maybe we can have a different build target for this?

cdoern (commented):
nice cleanup! One thing I am unsure of is the new flow in which you `podman run` for each command you want to run. I would be ok with this as long as there was another build target that installed `ilab` but the entrypoint was just `bin/bash`

containers/cuda/Containerfile
tiran: Actually I was wrong and could build the container without an active subscription. I must have confused this container with another container that pulls in dependencies that are not part of public UBI9.

I'm removing the comment again.

containers/cuda/Containerfile
tiran: That was Fabien's idea. I guess he wanted to save one `ilab` on the command line, so you can do `podman run ... instructlab:cuda init`  instead of `podman run ... instructlab:cuda ilab init`.

I don't mind to use a shell as entry point.

docs/containerization.md
tiran: This is also Fabien's idea. I'm not entirely sure why he has chosen this approach.

We should not recommend `--privileged` unless it is really, really required to run a container. `podman run --userns keep-id:uid=1001` should work, shouldn't it?

docs/containerization.md
tiran: It's coming from Fabien as well.

For my ROCm container, I suggest that the user enters the container and uses it as an interactive shell.

docs/containerization.md
```diff
@@ -1,61 +1,97 @@
...
+mkdir -p $HOME/.config/instructlab
+```
 
-[Nvidia Container Toolkit Install Guide](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html)
+Then, we can run the container, mounting the above folder and using the first
+NVIDIA GPU.
 
-[Podman Support for Container Device Interface](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html)
+```shell
+podman run --rm -it --userns keep-id:uid=1001 --device nvidia.com/gpu=0 --volume $HOME/.config/instructlab:/opt/app-root/src:z localhost/instructlab:cuda
```
luis5tb, line 87: this will return:
Error: no command or entrypoint provided, and no CMD or ENTRYPOINT from image

tiran: Append `CMD ["/bin/bash"]` to the container file. Does that solve the problem?

luis5tb: I used --entrypoint /bin/bash in the podman run, and works fine

docs/containerization.md
```diff
@@ -1,61 +1,97 @@
...
+### Run the CUDA-accelerated InstructLab container
+
+When running our model, we want to create some paths that will be mounted in
+the container to provide data persistence. As an unprivileged user, you will
+run a rootless container and map your UID to UID `1001` inside the container.
+The volume mount will contain InstructLab's configuration, models, taxonomy,
+and Hugging Face download cache.
+
+```shell
+mkdir -p $HOME/.config/instructlab
```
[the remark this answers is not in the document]
tiran: That should not be necessary with `--userns keep-id:uid=1001`.

docs/containerization.md
```diff
@@ -1,61 +1,97 @@
...
-[Podman Support for Container Device Interface](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html)
+```shell
+podman run --rm -it --userns keep-id:uid=1001 --device nvidia.com/gpu=0 --volume $HOME/.config/instructlab:/opt/app-root/src:z localhost/instructlab:cuda
+```
+
+The above command will give you an interactive shell inside the container.
+Let's initialize our configuration, download the model and start the chatbot.
+
+```shell
+(app-root) lab init
```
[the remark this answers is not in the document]
tiran: fixed, thanks!

`lab` was the old name before we switched to `ilab`. Old habits die hard.

stefwalter:
@tiran FYI, I also needed this package ```nvidia-driver-cuda-libs``` ... not sure if that's accounted for here.

tiran:
@stefwalter In the final image or on the host system?

stefwalter:
> @stefwalter In the final image or on the host system?

In the image. Without ```nvidia-driver-cuda-libs```

```
[root@8a706bc26838 instructlab]# podman3.11
bash: podman3.11: command not found
[root@8a706bc26838 instructlab]# python3.11
Python 3.11.5 (main, Sep  7 2023, 00:00:00) [GCC 11.4.1 20230605 (Red Hat 11.4.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
torc>>> torch.cuda.device_count()
1
>>> torch.cuda.is_available()
False
```

With the library installed then ```is_available()``` returns True.

docs/containerization.md
```diff
@@ -1,61 +1,97 @@
...
+```shell
+sudo nvidia-ctk cdi generate --output=/etc/cdi/nvidia.yaml
+```
+
+### Run the CUDA-accelerated InstructLab container
+
+When running our model, we want to create some paths that will be mounted in
+the container to provide data persistence. As an unprivileged user, you will
+run a rootless container and map your UID to UID `1001` inside the container
+with `--userns keep-id:uid=1001`. The volume mount will contain InstructLab's
```
n1hility, line 76: IMO it would be better to drop the USER= assign and just let this default to the container root user, then we don't need users to manually specify a userns.

tiran: The container is now using `USER 0` again. We can switch the container to rootless at a later point in time.

cdoern:
@tiran does this still only have the capability to only run 1 cmd per container run? if so, I would really vote to add back the old workflow of exec'ing into a running ctr with ilab installed.

I may have lost context here so feel free to correct me :)

tiran:
Fabien is taking care of CUDA container. Closing this stale PR.

PR #995, opened 2024-04-25
fix: force aspell spellchecker

leseb:
pyspelling defaults to aspell, although it supports both 'aspell' and 'hunspell'. However 'hunspell' reports a lot more issues than 'aspell'. It will be hard to maintain both, so let's enforce 'aspell'. Also, the GitHub action uses 'aspell'.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -57,10 +57,12 @@ commands =
 allowlist_externals = ./scripts/ruff.sh
 
 [testenv:spellcheck]
-description = spell check (needs 'aspell' or 'hunspell' command)
+description = spell check (needs 'aspell' command)
 skip_install = true
 skipsdist = true
 deps =
     pyspelling
 commands =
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -57,10 +57,12 @@ commands =
...
 [testenv:spellcheck]
-description = spell check (needs 'aspell' or 'hunspell' command)
+description = spell check (needs 'aspell' command)
 skip_install = true
 skipsdist = true
 deps =
     pyspelling
 commands =
-    {envpython} -m pyspelling --config {toxinidir}/.spellcheck.yml
+    bash -c 'command -v aspell || (echo "aspell is not installed. Please install it." && exit 1)'
```
tiran, line 66:
```suggestion
    sh -c 'command -v aspell || (echo "aspell is not installed. Please install it." && exit 1)'
```
+1 for using `command` instead of `which`

nathan-weinberg (approved):

Issue #379, opened 2024-03-07
Expose additional parameters for `lab generate` and `lab train` to end users


jaideepr97:
There are a few parameters we should consider exposing to end users in order to enable them to improve the quality of their models' outputs

This list is not exhaustive to the follwing:
`--num-examples` in `lab generate` to allow users to dictate how many of their provided examples are used to generate synthetic data 

`--learning-rate` and `--batch-size` to influence how the model trains in `lab train`

any others folks can think of

khaledsulayman:
@mbestavros has agreed to take this on, though I'm currently unable to assign him.

nathan-weinberg:
@cdoern @mbestavros any update on this?

mbestavros:
I did some exploratory work here, but ended up pivoting to other stuff (signing, mainly). It also seemed like there was going to be a lot of overlap with @cdoern's redesign PRs.

ktam3:
can someone confirm if this is GA blocker? I moved it to .19 but if this belong is .18 lmk

jaideepr97:
@Maxusmusti @JamesKunstle @RobotSail 
This is a really old issue but I want to check with yall to see if there are any new training args/parameters we might want to make configurable to keep this around? 
I see learning rate for one is still not a field in the config.yaml - is that something (among other things) that should be exposed to users?

jaideepr97:
> @jaideepr97 this was already implemented by @cdoern with his `additional_args` config field. He can validate whether or not this is still a relevant issue.

Thanks oleg
@cdoern what do you think? I guess that opens up the question about whether any of the args passed through additional args should be promoted to be first class args in the config/get a flag of their own?
if not I would propose closing this issue

jaideepr97:
for reference: https://github.com/instructlab/instructlab/pull/1634

ktam3:
outdated

PR #997, opened 2024-04-25
cli: Support batch size on Linux.

tyll:
# Changes

This solves one aspect of #379

Training speed can increase significantly with a higher batch size. Support to easily specify it on the command-line.

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -1003,9 +1003,9 @@ TORCH_DEVICE = TorchDeviceParam()
 @click.option(
     "--batch-size",
     "batch_size",
-    default=1,
+    default=None,
     help=(
-        "Batch size for training (only supported on Linux). Increase this if "
+        "Batch size for training Increase this if "
         "the GPU is underused."
     )
 )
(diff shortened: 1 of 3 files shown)


tyll:
@russellb @hickeyma what's needed to get a review for this?

src/instructlab/lab.py
```diff
@@ -885,6 +885,14 @@ def convert(self, value, param, ctx) -> "torch.device":
...
     help="model name to use in training",
 )
+@click.option(
+    "--batch-size",
+    "batch_size",
+    default=1,
+    help=(
+        "Batch size for training. Increases GPU usage."
+    )
+)
```
russellb, line 895: It looks like this is Linux-only? If so, could you update the text to indicate that?

Also, if you think about a user trying to use the tool, how should they decide if they should use this option? Is there text you could add to give any guidance?

Maybe the guidance should go in here?  https://github.com/instructlab/instructlab/blob/main/docs/gpu-acceleration.md

russellb (changes requested):

leseb:
@tyll do you mind addressing this review https://github.com/instructlab/instructlab/pull/997#pullrequestreview-2036839019 so we can move forward? Thanks!

leseb (commented):
Apologies, we lost track of this and missed your latest update. If you're still interested in continuing this work, could you please rebase your PR? We’ll ensure it gets the attention it deserves. Thank you!

PR #1004, opened 2024-04-25
containers: Fix the CUDA build to not use an SSH URL

stefwalter:
This was failing during the build process:

    Cloning ssh://****@github.com/instructlab/instructlab.git (to revision stable) to /tmp/pip-req-build-en5sp88r
    Running command git clone --filter=blob:none --quiet 'ssh://****@github.com/instructlab/instructlab.git' /tmp/pip-req-build-en5sp88r
    git@github.com: Permission denied (publickey).

Switch to an HTTP URL instead fixes the problem.

# Changes

Resolves #

### DIFF
--- a/containers/cuda/Containerfile
+++ b/containers/cuda/Containerfile
@@ -15,6 +15,6 @@ RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cu
     && export XLA_TARGET=cuda120 \
     && export XLA_FLAGS=--xla_gpu_cuda_data_dir=/usr/local/cuda
 RUN --mount=type=ssh,id=default CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install --force-reinstall --no-cache-dir llama-cpp-python 
-RUN --mount=type=ssh,id=default python3.11 -m pip install git+ssh://git@github.com/instructlab/instructlab.git@stable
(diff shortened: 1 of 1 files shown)


containers/cuda/Containerfile
```diff
@@ -15,6 +15,6 @@ RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cu
     && export XLA_TARGET=cuda120 \
     && export XLA_FLAGS=--xla_gpu_cuda_data_dir=/usr/local/cuda
 RUN --mount=type=ssh,id=default CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install --force-reinstall --no-cache-dir llama-cpp-python 
-RUN --mount=type=ssh,id=default python3.11 -m pip install git+ssh://git@github.com/instructlab/instructlab.git@stable
+RUN --mount=type=ssh,id=default python3.11 -m pip install git+https://github.com/instructlab/instructlab@stable
```
Gregory-Pereira, line 18: Should we use `python3` instead of `python3.11`? I have the same issue locally with multiple versions of python + another corresponding default version of pip, but I think the standard in our codebase is to just use `python3`.

derekhiggins, line 18: is the --mount=type=..... still required ?

stefwalter:
This is now fixed bi #988 which was merged. Closing.

PR #1005, opened 2024-04-25
RFC: support dynamic taxonomies

booxter:
For some taxonomies, it makes more sense to generate them programmatically, which allows to generate a larger number of samples as well as control their correctness without relying on humans.

For example, generating samples of simple rule application to an input, that can be expressed as a simple pure function, e.g. transliteration rules, is more robust and correct when taxonomies are not human written. Some other taxonomies that lean more into knowledge territory may also benefit from this.

This patch implements dynamic taxonomies as Containerfile definitions that can generate the seed on invocation.

# Changes

Resolves #

### DIFF
--- a/docs/qna.md
+++ b/docs/qna.md
@@ -1,0 +1,30 @@
+# Taxonomy input formats
+
+## Static `qna.yaml`
+
+The CLI supports a number of formats for taxonomy definitions. The default mode
+is a static `qna.yaml` file, populated by contributors of a new taxonomy with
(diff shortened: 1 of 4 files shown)


nathan-weinberg:
@booxter maybe this could be a PR in https://github.com/instructlab/dev-docs?

booxter:
Not something I'm interested in pursuing. Maybe even out of scope for what the project is trying to achieve. I will close, and we can revisit if needed.

PR #1008, opened 2024-04-25
Allow arbitary trainging args to be overridden

### LINKED ISSUE #1007: Training options only allow for well known/tested HW
The instructlab userbase will have hardware that require various training configuration options

e.g. to train on colab I needed to disable bf16 and set 
gradient_checkpointing=true
gradient_accumulation_steps=8

we should add a hidden, (perhaps temporary) argument to allow users to try various training options without
having to change the code

this could eventual be replaced with a happy set of arguments we know would cover all the cases
or code to auto sense the HW and set options based on what it finds.

derekhiggins:
Adding as a hidden argument to allow experimentation on various devices. Eventually once we know whats needed we can add something more permanent.

Fixes #1007

With this PR and  https://github.com/instructlab/instructlab/pull/1012 , running ilab e2e including training works on colab
> !ilab train --device cuda --override-training-args '{"bf16":false, "gradient_checkpointing":true, "gradient_accumulation_steps":8}'

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -977,6 +977,17 @@ TORCH_DEVICE = TorchDeviceParam()
     show_default=True,
     help="model name to use in training",
 )
+@click.option(
+    "--override-training-args",
+    default="{}",
+    show_default=True,
+    hidden=True,
+    help=(
+        "Additional arguments for linux training (json string)"
+        "e.g. '--override-training-args {\"gradient_accumulation_steps\"=8}'"
+        "e.g. '--override-training-args \"$(< override_train_args.json)\"'"
+    ),
+)
 @click.pass_context
 def train(
     ctx,
@@ -993,6 +1004,7 @@ def train(
     device: "torch.device",
     four_bit_quant: bool,
     model_name: str,
+    override_training_args: str,
 ):
     """
     Takes synthetic data generated locally with `ilab generate` and the previous model and learns a new model using the MLX API.
@@ -1005,6 +1017,11 @@ def train(
 
     if four_bit_quant and device.type != "cuda":
         ctx.fail("--4-bit-quant option requires --device=cuda")
+
+    try:
+        override_training_args_dict = json.loads(override_training_args)
+    except json.decoder.JSONDecodeError as e:
+        ctx.fail("Parsing override trainign args: " + str(e))
 
     # NOTE: If given a data_dir, input-dir is ignored in favor of existing!
     if data_dir is None:
@@ -1058,6 +1075,7 @@ def train(
             num_epochs=num_epochs,
             device=device,
             four_bit_quant=four_bit_quant,
+            override_training_args=override_training_args_dict,
         )
 
         training_results_dir = "./training_results"
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -150,6 +150,7 @@ def linux_train(
     num_epochs: Optional[int] = None,
     device: torch.device = torch.device("cpu"),
     four_bit_quant: bool = False,
+    override_training_args: dict = {},
 ):
     """Lab Train for Linux!"""
     print("LINUX_TRAIN.PY: NUM EPOCHS IS: ", num_epochs)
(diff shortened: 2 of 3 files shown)


src/instructlab/train/linux_train.py
```diff
@@ -237,16 +238,21 @@ def model_generate(user):
     output_dir = "./training_results"
     per_device_train_batch_size = 1
 
+    training_args = {}
+    training_args["per_device_train_batch_size"] = per_device_train_batch_size
+    training_args["fp16"] = use_fp16
+    training_args["bf16"] = not use_fp16
```
tiran, line 244: The bf16 issue is addressed in #993

derekhiggins: Sounds good, but this PR isn't really intended to deal with any specific training options, the point is to allow the advanced user to override any of them without needing to make changes to the code.

maxamillion:
I tested this patch and the following worked for me using my RTX A4000 GPU with 16G of VRAM:
```sh
$ ilab train --device cuda --override-training-args '{"bf16":false, "gradient_checkpointing":true, "gradient_accumulation_steps":8}'
```

TY!

src/instructlab/lab.py
```diff
@@ -878,6 +878,16 @@ def convert(self, value, param, ctx) -> "torch.device":
...
     help="model name to use in training",
 )
+@click.option(
+    "--override-training-args",
+    default="{}",
+    show_default=True,
+    hidden=True,
+    help=(
+        "Additional arguments for training (json string)"
+        "e.g. '--override-training-args {\"gradient_accumulation_steps\"=8}'"
```
leseb, line 888:
```suggestion
        "e.g. '--override-training-args {\"gradient_accumulation_steps\"=8}'"
         "e.g. '--override-training-args $(< override_train_args.json)'"
```

Depending on the complexity and length I believe it's important to show user that a file can be used too.

leseb (commented):
Can we have functional test coverage for this?

derekhiggins:
> Due to the complexity of the data, it seems this is better suited to be added to `config.yaml` instead of passing it on the command line.

I've added a example of how to use this from a json file e.g. --override-training-args "$(< override_train_args.json)"
would this be enough? As there is no training section currently in the config.yaml and I'm not sure this is a good reason to add one?

> Can we have functional test coverage for this?
If this is merged I'll update the e2e tests which should cover it (e.g. https://github.com/instructlab/instructlab/pull/1111 )

leseb (approved):

src/instructlab/lab.py
```diff
@@ -1006,6 +1018,11 @@ def train(
     if four_bit_quant and device.type != "cuda":
         ctx.fail("--4-bit-quant option requires --device=cuda")
 
+    try:
+        override_training_args_dict = json.loads(override_training_args)
+    except json.decoder.JSONDecodeError as e:
+        ctx.fail("Parsing override trainign args: " + str(e))
```
JamesKunstle, line 1024: "trainign" nit on spelling.

I think the command fail (CLI exits) if the input is malformed, too, rather than proceeding and making the user ctl-c and reload.

JamesKunstle (commented):
This functionality is super desirable. In the churn of designing the CLI in the context of other pillars of the project (SDG, evaluation, publishing), it's become clear that we need more wide-spread and type-checked configuration for _everything_. @cdoern's inbound "profiles" PR accounts for this, taking a first step toward application-wide default and override configuration support.

@derekhiggins your PR is very very appreciated, we'd love your input on @cdoern's work as well.

russellb:
> This functionality is super desirable. In the churn of designing the CLI in the context of other pillars of the project (SDG, evaluation, publishing), it's become clear that we need more wide-spread and type-checked configuration for _everything_. @cdoern's inbound "profiles" PR accounts for this, taking a first step toward application-wide default and override configuration support.
> 
> @derekhiggins your PR is very very appreciated, we'd love your input on @cdoern's work as well.

@JamesKunstle can you provide a link (or links) to the work you're referring to and requesting feedback on?

derekhiggins:
Closing this, a lot has changed since it was created and its probably no longer relevant

PR #1010, opened 2024-04-25
#1006 - Optimize CPU training on Linux

jsight:
- Timeout for chat mode increased to 300 seconds. Without this, it fails in a nearly silent manner in 30 seconds. This causes issues for CPU chat
 - Return the filename when converting the final resulting file and use this for copying. This is important when using f32, as llamam_to_gguf will have a different filename from expected (f16).
 - Determine the system memory available and use this to determine the batch size in cpu mode. This enables more parallelization on smaller systems.
 - Disable fp16 and bf16 in cpu mode. This also enables paralleization. Without it, training is unusable on my i7 with 64GB ram. With it, training takes <30 minutes for one epoch of ~26 iterations.
 - Turn off auto for dtype when loading the model. This greatly improved inference performance on my i7 as well. It went from single threaded to fully utilizing the CPU.
 - Updated the logging to be 1 based instead of zero based :)

# Changes

Resolves #

### DIFF
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+import psutil # to determine system memory
 from typing import Optional
 
 # Third Party
@@ -17,6 +18,7 @@ from transformers import (
 )
 from trl import DataCollatorForCompletionOnlyLM, SFTTrainer
 import torch
+from tqdm import tqdm
 
 # Local
 from ..chat.chat import CONTEXTS
@@ -100,6 +102,17 @@ def report_cuda_device(args_device, min_vram=0):
             "stopping the server to free up about 5 GiB of GPU memory."
         )
 
+def determine_batch_size(device):
+    if device.type == "cpu":
+        # Gets the total system memory, excluding swap
+        mem_bytes = psutil.virtual_memory().total
+        mem_gb = mem_bytes / 1024 / 1024 / 1024
+        return 2 if mem_gb < 60 else 4
+    else:
+        # Defaulting for now... should check and return
+        # more for very large GPUs
+        return 1
+
 
 def linux_train(
     train_file: str,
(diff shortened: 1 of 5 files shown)


jsight:
Resolves: https://github.com/instructlab/instructlab/issues/1006

src/instructlab/train/linux_train.py
```diff
@@ -235,14 +253,24 @@ def model_generate(user):
...
-    per_device_train_batch_size = 1
+    per_device_train_batch_size = determine_batch_size(device)
+    print(f"LINUX_TRAIN.PY: Training with batch size: {per_device_train_batch_size}")
+
+    use_bf16 = not use_fp16
 
+    if device.type == "cpu":
+        print("LINUX_TRAIN.PY: Disabling fp16 and bf16 for cpu training")
+        # CPU performance is very bad with fp16 or bf16, so disable both
+        use_fp16 = use_bf16 = False
```
bbrowning, line 264: The best option to use likely depends on specific CPU used here. Xeon CPUs with AVX512 and ipex may actually get better performance leaving bf16 enabled. For the sake of better CPU defaults, this is probably fine. But this is another example where the work @derekhiggins is doing around exposing all of these training arguments would be useful so that users could ultimately tweak these for their specific system until or unless we get smart enough to make optimal decisions in most cases.

derekhiggins: I'll test this to compare on my HW but it will be early next week before I get a chance

Here is the PR I've been testing to allow arbitrary training options 
https://github.com/instructlab/instructlab/pull/1008

jsight: Agreed. IMO, it'd be nice to have something like "profiles" and have it automatically select a reasonable default if none is specified.

src/instructlab/train/linux_train.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+import psutil # to determine system memory
```
tiran, line 4: It's a 3rd party import.

Please add the requirement to `requirements.txt`, too. Right now it's a transient dependency from `peft`.

src/instructlab/train/linux_train.py
tiran: I have been facing a similar problem for Intel Gaudi (Habana Labs HPUs and SynpseAI) and created an abstract layer for Torch device properties. I just ripped the code out of my experimental branch, cleaned it up, and published it as draft PR #1015.

src/instructlab/train/linux_train.py
jsight: Thanks, I have added it to requirements.txt

jsight:
Closing... will followup in 1109

jsight:
https://github.com/instructlab/instructlab/pull/1109

PR #1012, opened 2024-04-26
More efficiently use diskspace

### LINKED ISSUE #1011: Training uses up all available disk space in colab
The disk space available in colab is limited to 78G, some already in used by the system
During training we copy various large files around using up more diskspace then is necessary.
training eventually runs out of space (on colab) and fails
we should clean up as we go along to allow training to complete

derekhiggins:
In order to allow ilab to run e2e on colab where diskspace is limited we move rather then copy large files around, also rm files we're done with to save space for the next step.

Fix #1011

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -1052,10 +1052,17 @@ def train(
         )
 
         training_results_dir = "./training_results"
-        os.makedirs(training_results_dir, exist_ok=True)
 
         final_results_dir = training_results_dir + "/final"
         os.makedirs(final_results_dir, exist_ok=True)
+
+        gguf_models_dir = "./models"
+        if not os.path.isdir(gguf_models_dir):
+            os.mkdir(gguf_models_dir)
+        gguf_models_file = os.path.join(gguf_models_dir, "ggml-model-f16.gguf")
+        # Remove previously trained model, its taking up space we may need in the next step
+        if os.path.isfile(gguf_models_file):
+            os.remove(gguf_models_file)
 
         # TODO: Figure out what to do when there are multiple checkpoint dirs.
         # Right now it's just copying files from the first one numerically not necessarily the best one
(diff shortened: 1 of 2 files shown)


tests/test_lab_train.py
```diff
@@ -321,6 +318,11 @@ def test_train_linux(
         linux_train_mock,
         is_macos_with_m_chip_mock,
     ):
+        # pylint: disable=consider-using-with
```
wking, line 321: nit: you may be able to avoid arguing with the linter about this if you declare a helper function instead of using a `lambda`?

derekhiggins: thanks, the lambda has been replaced

src/instructlab/lab.py
```diff
@@ -1052,9 +1052,10 @@ def train(
         )
 
         training_results_dir = "./training_results"
-        os.makedirs(training_results_dir, exist_ok=True)
 
         final_results_dir = training_results_dir + "/final"
+        if os.path.exists(final_results_dir):
+            shutil.rmtree(final_results_dir)
```
russellb, line 1058: I'm a bit uncomfortable blindly removing this whole directory. I'd rather only operate on exactly the files we know are relevant to the current run of `ilab` and not make assumptions about what can be removed otherwise. I just want to avoid any surprising destructive behavior.

russellb (changes requested):

src/instructlab/lab.py
derekhiggins: Actually looking at this again I think I can just remove these two lines. The large model files that it contained is now moved instead of copied so this directory isn't taking up a lot of space.

src/instructlab/lab.py
```diff
@@ -1103,12 +1110,13 @@ def train(
...
-            os.mkdir(gguf_models_dir)
-        shutil.copy(final_results_dir + "/ggml-model-f16.gguf", gguf_models_dir)
-        # cleanup original copy of model
-        os.remove(final_results_dir + "/ggml-model-f16.gguf")
+        # Remove safetensors files to save space, were done with them here
+        # and the huggingface lib has them cached
+        for file in glob(final_results_dir + "/*.safetensors"):
+            os.remove(file)
+
+        shutil.move(final_results_dir + "/ggml-model-f16.gguf", gguf_models_file)
```
russellb, line 1118: only if you have to make another change anyway, would be good to make this `os.path.join()`. I know this isn't new code though, so it's not worth blocking over

russellb (approved):

nathan-weinberg (approved):

PR #1013, opened 2024-04-26
ci: add markdown linter

nathan-weinberg:
Adds a Markdown Linter - needs some testing before merge - I'll review the `hold` label and add reviewers once I feel it's ready

### DIFF
--- a/ .markdownlint-cli2.yaml
+++ b/ .markdownlint-cli2.yaml
@@ -1,0 +1,15 @@
+config:
+  line-length: false
+  no-emphasis-as-header: false
+  first-line-heading: false
+  code-block-style: false
+  no-duplicate-header: false
+  single-trailing-newline: false
+globs:
+  - "**/*.md"
+ignores:
+  - "pkg"
+  - "**/node_modules/*"
+  - "**/node_modules/**"
+  - ".tox/**"
+  - "venv/**"
--- a/Makefile
+++ b/Makefile
@@ -127,3 +127,24 @@ verify: ## Run tox -e fmt,lint,spellcheck against code
 .PHONY: spellcheck-sort
 spellcheck-sort: .spellcheck-en-custom.txt
 	sort -d -f -o $< $<
+
+##@ Linting
+
+#
+# If you want to see the full commands, run:
+#   NOISY_BUILD=y make
+#
+ifeq ($(NOISY_BUILD),)
+    ECHO_PREFIX=@
+    CMD_PREFIX=@
+    PIPE_DEV_NULL=> /dev/null 2> /dev/null
+else
+    ECHO_PREFIX=@\#
+    CMD_PREFIX=
+    PIPE_DEV_NULL=
+endif
+
+.PHONY: md-lint
+md-lint: ## Lint markdown files
+	$(ECHO_PREFIX) printf "  %-12s ./...\n" "[MD LINT]"
+	$(CMD_PREFIX) podman run --rm -v $(CURDIR):/workdir --security-opt label=disable docker.io/davidanson/markdownlint-cli2:v0.6.0 > /dev/null
--- a/Makefile
+++ b/Makefile
@@ -147,4 +147,4 @@ endif
 .PHONY: md-lint
 md-lint: ## Lint markdown files
 	$(ECHO_PREFIX) printf "  %-12s ./...\n" "[MD LINT]"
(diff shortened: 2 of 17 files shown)


 .markdownlint-cli2.yaml
```diff
@@ -0,0 +1,15 @@
...
+  no-duplicate-header: false
+  single-trailing-newline: false
+globs:
+  - "**/*.md"
+ignores:
+  - "pkg"
+  - "**/node_modules/*"
+  - "**/node_modules/**"
+  - ".tox/**"
+  - "venv/**"
```
tiran, line 15: We don't need node stuff. It's common to use either `.venv` or `venv` for a local development virtual env.
```suggestion
  - ".tox/**"
  - "venv/**"
  - ".venv/**"
```

Makefile
```diff
@@ -127,3 +127,24 @@ verify: ## Run tox -e fmt,lint,spellcheck against code
...
+else
+    ECHO_PREFIX=@\#
+    CMD_PREFIX=
+    PIPE_DEV_NULL=
+endif
+
+.PHONY: md-lint
+md-lint: ## Lint markdown files
+	$(ECHO_PREFIX) printf "  %-12s ./...\n" "[MD LINT]"
+	$(CMD_PREFIX) podman run --rm -v $(CURDIR):/workdir --security-opt label=disable docker.io/davidanson/markdownlint-cli2:v0.6.0 > /dev/null
```
tiran, line 150: userns keep-id is better than disabling SELinux:

```suggestion
	$(CMD_PREFIX) podman run --rm -v $(CURDIR):/workdir:ro :ro --userns=keep-id docker.io/davidanson/markdownlint-cli2:v0.6.0 > /dev/null
```

tiran (commented):
I like the idea to link markdown files, too.

Makefile
nathan-weinberg: I get the following error when trying this on Fedora
```bash
[nathan@nathan-redhat instructlab (md-lint)]$ make md-lint 
  [MD LINT]    ./...
Error: invalid reference format
make: *** [Makefile:150: md-lint] Error 125
```

russellb: same error on mac fwiw

nathan-weinberg:
This should be good to go

notebooks/README.md
```diff
@@ -48,9 +49,7 @@ and the NVIDIA P100 provided in the free tier, we will fine tune a LoRA.
 1. You'll need a Kaggle account, which you can create by visiting [Kaggle's Sign-up Page](https://www.kaggle.com/account/login?phase=startRegisterTab&returnUrl=%2F).
 1. To use Kaggle's accelerators, you'll have to verify your account with a phone number. Visit the [account settings page](https://www.kaggle.com/settings) and select "Phone Verification".
 
-
-**NOTE: At present, you'll need to download the notebook and upload it to Kaggle. The following steps will walk you through uploading the notebook. Once this repository is open sourced, we will make an 'Open in Colab' button**
-
+##### NOTE: At present, you'll need to download the notebook and upload it to Kaggle. The following steps will walk you through uploading the notebook. Once this repository is open sourced, we will make an 'Open in Colab' button**
```
russellb, line 52: It can be another PR, but this formatting is nicer:

https://github.com/orgs/community/discussions/16925

src/instructlab/generator/README.md
```diff
@@ -1,7 +1,8 @@
 ### Notice
-This folder uses code from the `stanford_alpaca` project developed by Tatsu Lab, available at https://github.com/tatsu-lab/stanford_alpaca. The original code is licensed under the Apache License:
 
-```
+This folder uses code from the `stanford_alpaca` project developed by Tatsu Lab, available [here](https://github.com/tatsu-lab/stanford_alpaca). The original code is licensed under the Apache License:
+
+```license
```
russellb, line 5: Do you know for sure this is valid?  I don't think it'll hurt if it's not recognized. I would have just done `text` probably. I won't block on it since I don't think it breaks anything

nathan-weinberg: I see it being used other places in the repo, so I assume it's valid

notebooks/README.md
nathan-weinberg: simple enough to change, will do that as part of this PR

russellb (approved):
One follow-up is to block merging on this job passing in the mergify config

Makefile
nathan-weinberg: We can address this in a followup

russellb:
@mergifyio refresh

PR #1017, opened 2024-04-26
Use quantize binary from instructlab_quantize

### LINKED ISSUE #438: llamacpp/quantize binary in git repo

tiran:
# Changes

Resolves #438

The `quantize` binary is a macOS ARM64 program. It is part of llama.cpp and is needed on macOS. The binary has been moved out of the main project into a separate `instructlab-quantize` package. The package provides quantize for macOS ARM64, Linux x86_64, and Linux aarch64.


russellb:
Has anyone looked at moving the packaging of this into llama-cpp-python ? This doesn't seem like a thing we should be doing long-term. This is an improvement over the current hack (thank you!), but ideally, this goes away from instructlab completely. Maybe it's something to look at in the context of llama-cpp-python? If not, what's our desired end game here?

tiran:
Some people have been looking into vLLM (not to confused with llvm) as a replacement for llama-cpp-python. Would that solve our problem?

bbrowning:
I don't think vLLM has Mac support, so I don't think it would help this specific problem.

tiran:
> I don't think vLLM has Mac support, so I don't think it would help this specific problem.

It looks like you are right. I don't see any support for Apple MPS in vLLM's documentation or roadmap for Q2 2024.

tiran:
@cdoern @russellb @nathan-weinberg Can we get this PR merged and the repo moved before we land Charlie's PR #1157? The latest version of the instructlab-quantize package contains binaries:

- quantize-aarch64-linux
- quantize-arm64-darwin
- quantize-x86_64-linux

The repo has infrastructure to build binaries for ppc64le and s390x on Linux, too.

cdoern (approved):
I think this is the best step forward. LGTM

jaideepr97 (approved):
LGTM

do we have follow up issues to move `instructlab_quantize` into the instructlab org? And would the Pypi package need to be updated to point to the new repo then?

khaledsulayman (approved):

PR #990, opened 2024-04-24
`ilab` command redesign

cdoern:
add the following groups to ilab:

model
config
data
taxonomy

inside of these commands, there are groups commands that relate to the sub-command. For example `ilab model list` lists all models known to the system.

a full list of the commands in this PR is:

model
 - convert
 - download
 - train
 - chat
 - serve
 - evaluate
 
data
 - generate

config
 - init

taxonomy
 - diff

the goal of this is to make `ilab` commands more logical. `ilab generate` for example is vague. also, adding structure allows for natural expansion of the project and makes the repo more digestible to new contributors.

Grouping ilab commands under common resources which the cli deals with makes sense as the role of instructlab grows. This is the first step in making `ilab` a model engine in which the cli will be a tool for full blown LLM management and deployment

PR #1022, opened 2024-04-26
feat: Support creating a workspace with `lab init`

### LINKED ISSUE #729: support creating a workspace folder directly with `lab init`
current
```
mkdir ws
cd ws
lab init
```

to-support
```
lab init ws
cd ws
```

rafvasq:
# Changes
- Adds an option `--workspace-path` to the `init` command
  - If passed, config.yaml is created there (the workspace directory is created if it doesn't exist)
- Adds support to pass a workspace-path as an env var `ILAB_WORKSPACE_PATH`

Resolves #729 

Allows for a user to pass a path to `ilab init` for config initialization in that given directory.


src/instructlab/lab.py
```diff
@@ -126,16 +126,26 @@ def cli(ctx, config):
     "Please do not use this option if you are planning to contribute back "
     "using the same taxonomy repository. ",
 )
+@click.option(
+    "--workspace",
```
russellb, line 130:
```suggestion
    "--workspace-path",
    type=click.Path(),
```

`workspace-path` would be more consistent with how the `taxonomy-path` option is named. It would also be good to set the type, since click has a path type.

russellb (changes requested):
I have one minor suggestion inline.

I also wonder about the most desirable UX here. It seems kind of weird to do:

```
ilab init --workspace-path myworkspace
cd myworkspace
ilab download
ilab ...
```

A slight improvement would be to allow `--workspace-path` to every command.

```
ilab init --workspace-path myworkspace
ilab download --workspace-path myworkspace
...
```

This next part could be a separate PR, because the implementation would apply to more than just this use case, but I want to call it out as it helps the UX further once the workspace option is available for all commands.

On top of that, it would be great if instead of only allowing the `--workspace-path` option if it also supported an environment variable. More on that is in "Values from Environment variables" here https://click.palletsprojects.com/en/8.1.x/options/

Now we have ...

```
export ILAB_WORKSPACE_PATH=/path/to/workspace
ilab init
ilab download
ilab serve
```

What is super nice here is it no longer matters what directory you run `ilab` from. Once you have a workspace set and you point to it in the env, you can use the `ilab` commands anywhere.

I think that'd be a pretty nice UX.

rafvasq:
Thanks @russellb, I made some updates and opened an issue #1034 to capture the changes to the other commands. I could take those on too.

src/instructlab/lab.py
```diff
@@ -126,16 +126,29 @@ def cli(ctx, config):
...
     repository,
     min_taxonomy,
+    workspace_path,
 ):
     """Initializes environment for InstructLab"""
 
+    if workspace_path:
+        if not exists(workspace_path):
+            os.mkdir(workspace_path)
+        os.chdir(workspace_path)
```
russellb, line 150: I had another thought about the UX for a workspaces concept. I think the use of `os.chdir()` results in some surprising behavior.  Take this for example:

```
ilab init --taxonomy-path relative/path/to/taxonomy --workspace-path path/to/workspace
```

Any time I specify a relative path, I would expect it to be interpreted relative to the working directory where I'm running the command.

For my example, I would expect `ilab` to take my relative path and write an absolute path into the config file in the workspace that points it to the right place. I think other surprises could happen when the tool is switching its working directory out by surprise.

In other words, while it's more code, I think less surprising behavior would be to not change working directories, but instead, do more careful handling of paths throughout using that workspace path value.

nathan-weinberg:
@Mergifyio rebase

russellb:
just want to apologize for the long delay in coming back to this. I have been traveling the last 1.5 weeks or so.

src/instructlab/config.py
```diff
@@ -8,12 +8,13 @@
...
 import yaml
 
+DEFAULT_WORKSPACE_PATH = "workspace"
 DEFAULT_API_KEY = "no_api_key"
 DEFAULT_CONFIG = "config.yaml"
 DEFAULT_MODEL = "merlinite-7b-lab-Q4_K_M"
 DEFAULT_MODEL_PATH = f"models/{DEFAULT_MODEL}.gguf"
 DEFAULT_TAXONOMY_REPO = "https://github.com/instructlab/taxonomy.git"
-DEFAULT_TAXONOMY_PATH = "taxonomy"
+DEFAULT_TAXONOMY_PATH = f"{DEFAULT_WORKSPACE_PATH}/taxonomy"
```
russellb, line 17: I think I'd prefer not to use the workspace feature by default. Most instructions tell people to do a similar thing manually, I believe.

cdoern:
my work in #990 intends to also create a centralized dir in ~/.config for the ilab config and convert to a .conf file type. 

I am unsure of how this would interact with this PR. However, I am totally open to taking whichever approach people are more comfortable with

russellb:
> my work in #990 intends to also create a centralized dir in ~/.config for the ilab config and convert to a .conf file type.
> 
> I am unsure of how this would interact with this PR. However, I am totally open to taking whichever approach people are more comfortable with

I think the ideas are complementary.

The idea in this PR would be to configure a workspace directory so that it didn't matter what directory you were in to run `ilab`, it would operate on the files in that workspace dir. I was just thinking we should make sure we don't go really far in this PR in any way that might conflict with what you've got planned.

I think it may be mostly a new config file option (and option to `init`) plus an optional arg to (almost?) all other commands. I think that probably carries over OK.

nathan-weinberg:
@rafvasq we have this as part of our 5/30 milestone, do you think it's something we could get done before then or should we push it out to later?

rafvasq:
@nathan-weinberg I'll be travelling until mid-next week but based on @russellb and @cdoern's comments, most of the work is done and I should be able to complete it as soon as I'm back. I'm pretty confident it can get into the 5/30 milestone!

russellb:
Apologies again for the delay ... 

The current implementation is a good start. I see that `ilab init` takes a workspace path and when specified, it will ensure that the directory exists. I would like to get a more complete UX for the feature before we merge anything, though.  I think it would be easiest to write out a design for the spec of the changes to CLI commands before implementing it. We can try doing that just in comments here, or if it gets too difficult, it can be a design doc posted to https://github.com/instructlab/dev-docs.

A starting point would be my original comment here: https://github.com/instructlab/instructlab/pull/1022#pullrequestreview-2027422480

To summarize - once a workspace is created and configured, I'd like to see *everything* done relative to that workspace instead of the current working directory. Otherwise, it feels incomplete and not enough of an improvement over `mkdir workspace ; cd workspace`.

leseb:
@rafvasq did you get a chance to propose a small design doc like highlighted [here](https://github.com/instructlab/instructlab/pull/1022#issuecomment-2140541840). Thanks!

rafvasq:
@leseb I've now created an issue https://github.com/instructlab/dev-docs/issues/100 but sadly I don't have time to commit to seeing this work past what was originally scoped and done in this PR so far.

leseb:
> @leseb I've now created an issue [instructlab/dev-docs#100](https://github.com/instructlab/dev-docs/issues/100) but sadly I don't have time to commit to seeing this work past what was originally scoped and done in this PR so far.

Thanks for your effort on this and sorry for the delay. I think we should probably close this and wait for someone else to pick up. Do you mind? Thanks

PR #1025, opened 2024-04-27
Fix llama_cpp cuda usage in Containerfile

n1hility:
# Problem

The current cuda builds of instructlab are unaccelerated for serve/generate

# Changes

Sync requirements.txt of  llama_cpp_python with instructlab

The step which pip installs llama_cpp_python is overwritten by the instructlab install since the instructlab has a pinned version that is lesser than the preceding llama_cpp_python. Since the instructlab install does not specify the cuda CMAKE args it gets rebuilt without cuda support. By aligning requirements we ensure that  llama_cpp_python will not be rebuilt.

Before this change:

```
>>> llama_cpp.llama_supports_gpu_offload()      
False
```

After:

```      
>>> llama_cpp.llama_supports_gpu_offload()      
True
```

### DIFF
--- a/containers/cuda/Containerfile
+++ b/containers/cuda/Containerfile
@@ -14,7 +14,7 @@ RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cu
     && export PATH="/usr/local/cuda/bin:$PATH" \
     && export XLA_TARGET=cuda120 \
     && export XLA_FLAGS=--xla_gpu_cuda_data_dir=/usr/local/cuda
-RUN CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install --force-reinstall --no-cache-dir llama-cpp-python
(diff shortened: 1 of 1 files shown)


russellb (approved):

tiran:
@n1hility @russellb 
The workaround in this PR is problematic. It doesn't use the current `requirements.txt` but the version from the main branch. This is likely to cause issues when we create a stable branch and modify requirements in main. Also `pip install -r ...` will install all dependencies in the requirements file.

I'm using this approach in my container files, which I believe is better:
1) copy requirements file from context into the container
2) convert the requirements file to a constraints file by stripping off optional dependencies
3) pip install with `-c` instead of `-r`

```Dockerfile
COPY requirements.txt /tmp
RUN sed 's/\[.*\]//' /tmp/requirements.txt >/tmp/constraints.txt
RUN CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install -c /tmp/constraints.txt --force-reinstall --no-cache-dir llama-cpp-python
```

russellb:
Thanks @tiran -- good feedback. you're right.

I think there's a simpler change needed at the moment. This Containerfile is not building instructlab from the current source tree. It's explicitly installing `stable`. We should just change the URL to use stable, as well.  I think this should still be improved to be more flexible with what it's installing, but that's at least a quick fix improvement.

PR #1027, opened 2024-04-29
feat: ping PR author on merge conflicts

leseb:
From now on, if a pull request has merge conflicts and needs a rebase, the author will be pinged.


russellb (commented):
I like this, thanks. A couple things I’d like to add

- add a needs-rebase label
- Another rule that removes that label when a PR is not in conflict

I have config for this in another repo I can add here later, but if you’d like to add that’s even better!

leseb:
> I like this, thanks. A couple things I’d like to add
> 
> * add a needs-rebase label
> * Another rule that removes that label when a PR is not in conflict
> 
> I have config for this in another repo I can add here later, but if you’d like to add that’s even better!

Can do!

leseb:
@russellb the 'needs-rebase' label needs to be created. Additionally, should we remove 'DRAFT - DO NOT MERGE'?

.github/mergify.yml
```diff
@@ -8,6 +30,7 @@ pull_request_rules:
     - base=main
     - label!=hold
     - label!=do-not-merge
+    - label!=needs-rebase
```
russellb, line 33: This doesn't hurt, but I don't think it's necessary. Or is there a case in which you think this would prevent a merge that otherwise would happen?

russellb:
> @russellb the 'needs-rebase' label needs to be created. 

Done.

> Additionally, should we remove 'DRAFT - DO NOT MERGE'?

I'm not sure what this is referring to. Can you clarify?

leseb:
> > @russellb the 'needs-rebase' label needs to be created.
> 
> Done.
> 
> > Additionally, should we remove 'DRAFT - DO NOT MERGE'?
> 
> I'm not sure what this is referring to. Can you clarify?

I saw that label, and I was wondering if we should keep it or not.

russellb:
> > > @russellb the 'needs-rebase' label needs to be created.
> > 
> > 
> > Done.
> > > Additionally, should we remove 'DRAFT - DO NOT MERGE'?
> > 
> > 
> > I'm not sure what this is referring to. Can you clarify?
> 
> I saw that label, and I was wondering if we should keep it or not.

Oh, yeah, that's probably not necessary. There's also 'hold'. It looks like someone renamed `do-not-merge` to that, which breaks the mergify configuration.

russellb (approved):

PR #1031, opened 2024-04-29
github: Add a test-workflow-complete job

russellb:
Instead of creating dependencies on specific test jobs, make a final job that depends on the others that can be referenced instead. This will simplify mergify configuration.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -87,3 +87,10 @@ jobs:
       - name: Run unit tests with tox (macOS)
         if: startsWith(matrix.platform, 'macos')
         run: tox -e py-unit
+
+  test-workflow-complete:
+    needs: ["test"]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Test Workflow Complete
+        run: echo "Test Workflow Complete"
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -13,6 +13,7 @@ on:
       - 'requirements*.txt'
       - 'tox.ini'
       - 'scripts/*.sh'
+      - .github/workflows/test.yml'
```
nathan-weinberg, line 16: @russellb mismatched `'` here

.github/workflows/test.yml
```diff
@@ -85,3 +87,10 @@ jobs:
       - name: Run unit tests with tox (macOS)
         if: startsWith(matrix.platform, 'macos')
         run: tox -e py-unit
+
+  test-workflow-complete:
+    needs: ["test"]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Test Workflow Complete
+        run: echo "Test Workflow Complete"
```
nathan-weinberg, line 96: nit: add newline

nathan-weinberg (changes requested):

.github/workflows/test.yml
russellb: i need to figure out how to make vscode do this automatically like I have vim configured to do ...

bjhargrave:
```json
"files.insertFinalNewline": true
```

nathan-weinberg (approved):

Issue #491, opened 2024-03-10
Support or meet 'man lab' requirement?


yangboroy:
Run `man lab`
```
[rhel@bootp-73-75-178 ~]$ man lab
No manual entry for lab
```

If possible, should support `lab version` `lab -V` `lab --version`.

PR #1032, opened 2024-04-29
tox: add -e docs target to generate man pages

booxter:
Use click-man to generate manual pages.

Fixes: #491


tox.ini
```diff
@@ -67,6 +67,16 @@ commands =
     "{envpython} -m pyspelling --config {toxinidir}/.spellcheck.yml" --spellchecker aspell
 allowlist_externals = sh
 
+[testenv:docs]
+description = docs
+skipsdist = true
```
tiran, line 72: `skipsdist` is only useful in combination with `skip_install`. Otherwise tox still creates a package.
```suggestion
```

.github/workflows/test.yml
```diff
@@ -85,3 +85,21 @@ jobs:
       - name: Run unit tests with tox (macOS)
         if: startsWith(matrix.platform, 'macos')
         run: tox -e py-unit
+
+  man:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
tiran, line 92: All workflows now check out the full history + submodules.
```suggestion
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
          submodules: true
```

.github/workflows/test.yml
```diff
@@ -85,3 +85,21 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install tox tox-gh>=1.2
```
tiran, line 96: You are not using `tox-gh` in the workflow. `tox-gh` only affects bare `tox` calls without an `-e` argument.

tiran (commented):
Thanks for taking care of the man page idea. I wasn't aware that the `hidden` bug was fixed in upstream.

tox.ini
```diff
@@ -67,6 +67,16 @@ commands =
...
+deps =
+    # TODO: switch to a release tag when it's cut
+    # We need https://github.com/click-contrib/click-man/pull/62 to handle hidden options
+    git+https://github.com/click-contrib/click-man@24ec8377e3c24378417f6fc4f571b4f585d7df6b
```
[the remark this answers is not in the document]
booxter: I will this as-is; if we want to remove the flag, we can clean up here later.

booxter:
@mergifyio rebase

russellb (approved):
This lgtm and worked when I tested it locally.

A follow-up would be to add the man pages it generates to the Python packaging so they get installed.

PR #1033, opened 2024-04-29
Make `ilab` able to download safetensors

cdoern:
when downloading from HF, check check if the repo has .safetensor files. If it does, and `--filename` is default, download the entire repo and put it into a <repo_name> folder inside of models

Signed-off-by: Charlie Doern <cdoern@redhat.com>

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -15,7 +15,7 @@ import typing
 # Third Party
 from click_didyoumean import DYMGroup
 from git import GitError, Repo
-from huggingface_hub import hf_hub_download
+from huggingface_hub import hf_hub_download, list_repo_files
 from huggingface_hub import logging as hf_logging
 import click
 import yaml
@@ -712,15 +712,36 @@ def chat(
 def download(ctx, repository, release, filename, model_dir):
     """Download the model(s) to train"""
     click.echo(f"Downloading model from {repository}@{release} to {model_dir}...")
+    if (
+        "HF_TOKEN" not in os.environ
+        and repository != "instructlab/merlinite-7b-lab-GGUF"
+    ):
+        raise ValueError(
+            "HF_TOKEN var needs to be set in your environment to download HF Model. The HF Token is used to authenticate your identity to the Hugginface Hub."
+        )
     try:
         if ctx.obj is not None:
             hf_logging.set_verbosity(ctx.obj.config.general.log_level.upper())
-        hf_hub_download(
-            repo_id=repository,
-            revision=release,
-            filename=filename,
-            local_dir=model_dir,
-        )
+        files = list_repo_files(repo_id=repository, token=os.getenv("HF_TOKEN"))
+        if any(".safetensors" in string for string in files):
+            if not os.path.exists(os.path.join(model_dir, repository)):
+                os.makedirs(name=os.path.join(model_dir, repository), exist_ok=True)
+            for f in files:
+                hf_hub_download(
+                    token=os.getenv("HF_TOKEN"),
+                    repo_id=repository,
+                    revision=release,
+                    filename=f,
+                    local_dir=os.path.join(model_dir, repository),
+                )
+        else:
+            hf_hub_download(
+                token=os.getenv("HF_TOKEN"),
+                repo_id=repository,
+                revision=release,
+                filename=filename,
+                local_dir=model_dir,
+            )
     except Exception as exc:
         click.secho(
             f"Downloading model failed with the following Hugging Face Hub error: {exc}",


src/instructlab/lab.py
```diff
@@ -712,15 +712,37 @@ def chat(
...
-            repo_id=repository,
-            revision=release,
-            filename=filename,
-            local_dir=model_dir,
-        )
+        files = list_repo_files(repo_id=repository, token=os.getenv("HF_TOKEN"))
+        if any(".safetensors" in string for string in files):
+            folder_name = repository.replace("/", "-")
+            if not os.path.exists(os.path.join(model_dir, folder_name)):
+                os.mkdir(os.path.join(model_dir, folder_name))
```
tiran, line 729:
```suggestion
            os.makedirs(os.path.join(model_dir, folder_name), exist_ok=True)
```

src/instructlab/lab.py
```diff
@@ -712,15 +712,37 @@ def chat(
 def download(ctx, repository, release, filename, model_dir):
     """Download the model(s) to train"""
     click.echo(f"Downloading model from {repository}@{release} to {model_dir}...")
+    if (
+        "HF_TOKEN" not in os.environ
+        and repository != "instructlab/merlinite-7b-lab-GGUF"
+    ):
+        raise ValueError(
+            "HF_TOKEN var needs to be set in your environment to download HF Model"
```
oindrillac, line 720: Should we also update the readme user docs which point to the additional steps (this and stuff like model-family etc)? They can be a separate PR in the interest of time. For now maybe we can make the error message more descriptive

```suggestion
            "HF_TOKEN var needs to be set in your environment to download HF Model. The HF Token is used to authenticate your identity to the Hugginface Hub."
```

alinaryan: small nit in case you commit this: 
's/Hugginface/Huggingface/'

russellb (approved):
lgtm!  I know @alimaredia is testing this, so I requested his review. It should block merging until he approves as well.

src/instructlab/lab.py
```diff
@@ -712,15 +712,36 @@ def chat(
 def download(ctx, repository, release, filename, model_dir):
     """Download the model(s) to train"""
     click.echo(f"Downloading model from {repository}@{release} to {model_dir}...")
+    if (
+        "HF_TOKEN" not in os.environ
+        and repository != "instructlab/merlinite-7b-lab-GGUF"
```
jaideepr97, line 717: do we need this check? I think it might be best not to hardcode merlinite for such checks - it would also be good practice to always have an HF token exposed and would make the code easier and cleaner to maintain (even if technically you dont need a token for the instructlab org)

just me 0.02

russellb: I assume this was so at least the default config didn't require it, which is nice.

An even better UX would determine whether the token is required for a given model before then erroring out because it's not set, but that could be a future improvement ...

jaideepr97: right - then for now should we at least only check if "instructlab" is part of the repo string, and we allow the download without token if true? since we would want to at least have uniform behavior for all the models in the `instructlab` HF org

tiran: How about you use clicks' feature to read [values from env vars](https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables)?
`@click.option('--hf-token', envvar='HF_TOKEN')`

cdoern: merl is the default. This is my way of checking `If NO HF TOKEN and THE USER DIDN'T SET --FILENAME`

src/instructlab/lab.py
```diff
@@ -712,15 +712,36 @@ def chat(
...
     try:
         if ctx.obj is not None:
             hf_logging.set_verbosity(ctx.obj.config.general.log_level.upper())
-        hf_hub_download(
-            repo_id=repository,
-            revision=release,
-            filename=filename,
-            local_dir=model_dir,
-        )
+        files = list_repo_files(repo_id=repository, token=os.getenv("HF_TOKEN"))
```
alimaredia, line 725: @cdoern why not download the entire repository via `snapshot_download()` instead of going file by file? https://huggingface.co/docs/huggingface_hub/en/guides/download#download-an-entire-repository

src/instructlab/lab.py
russellb: @tiran yes! I was just looking at that yesterday when looking at https://github.com/instructlab/instructlab/pull/1022

src/instructlab/lab.py
russellb: This looks like a good idea to me +1

russellb: Apologies @alimaredia -- once you posted this review, that made mergify happy that you looked at it.  If you had done "request changes" it would have blocked it, just FYI.

I probably should just set a `hold` label and let you remove it

cdoern:
uh oh, seems mergify has gone rogue!!!! I will open another PR to address folks' comments

src/instructlab/lab.py
bbrowning: I was surprised by it asking me for a HF_TOKEN when I did `ilab download --repository instructlab/merlinite-7b-lab` as that repository does not need one. In my case, I specifically wanted it to download the non-gguf version of this model, but perhaps the logic around HF_TOKEN usage should be a bit broader?

russellb: This was improved further in https://github.com/instructlab/instructlab/pull/1035

PR #1036, opened 2024-04-29
Use mergify for automating GitHub PR labels

russellb:
commit 3b48af1f7d52023e3fd438433e60793a08ee7ae5
Date:   Mon Apr 29 16:19:25 2024 -0400

    Use mergify to automate github labels on PRs
    
    Mergify can handle label automation, so make use of that instead of
    the separate labeler workflow.
    

.github/mergify.yml
```diff
@@ -68,3 +68,63 @@ pull_request_rules:
     merge:
       method: merge
     delete_head_branch:
+
+- name: label-cicd
+  description: Automatically apply CI/CD label
+  conditions:
+    - or:
+      - files=.github/labeler.yml
```
nathan-weinberg, line 76: Can be removed

.github/mergify.yml
```diff
@@ -68,3 +68,63 @@ pull_request_rules:
...
+      - files=.github/mergify.yml
+      - files=.github/workflows/**/*
+      - files=scripts/ruff.sh
+      - files=.pre-commit-config.yaml
+      - files=.pylintrc
+      - files=.spellcheck-en-custom.txt
+      - files=.spellcheck.yml
+      - files=tox.ini
+  actions:
```
nathan-weinberg, line 85: Should we add https://github.com/instructlab/instructlab/blob/main/.markdownlint-cli2.yaml to this this?

nathan-weinberg (commented):
I was going to suggest adding https://github.com/instructlab/instructlab/blob/main/.isort.cfg but it seems like it isn't being used anymore?

.github/mergify.yml
russellb: done!

nathan-weinberg (approved):

Issue #802, opened 2024-04-04
functional tests don't run on macOS


markstur:
Running into problems with:

* need to install utils
* printf incompatible
* sed incompatible

PR #1039, opened 2024-04-30
feat: run functional test on MacOS

leseb:
The CI will now run the functional test on a MacOS runner. The testing script was changed to accommodate MacOS.

Fixes: #802

Co-authored-by: Mark Sturdevant <mark.sturdevant@ibm.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,6 +60,11 @@ jobs:
         run: |
           sudo apt-get install -y expect
 
+      - name: Install tools on MacOS
+        if: startsWith(matrix.platform, 'macos')
+        run: |
+          brew install expect coreutils bash
+
       - name: Setup Python ${{ matrix.python }}
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -58,6 +58,16 @@ jobs:
         run: |
           sudo apt-get install -y expect
 
+      - name: Install homebrew package manager
```
bjhargrave, line 61: Do you need to reinstall brew? https://github.com/actions/runner-images/blob/macos-14-arm64/20240422.3/images/macos/macos-14-arm64-Readme.md#package-management says that brew is present on the image.

If brew was not already present, this step would be incomplete since it does not add brew to the path for the following step.

`eval "$(/opt/homebrew/bin/brew shellenv)"`

.github/workflows/test.yml
```diff
@@ -60,6 +60,11 @@ jobs:
         run: |
           sudo apt-get install -y expect
 
+      - name: Install tools on MacOS
+        if: startsWith(matrix.platform, 'macos')
+        run: |
+          brew install expect coreutils
```
chmouel, line 66: what about installing bash from brew as well so you don't have a bash old version compatibility issue in the future

leseb: Thanks for the pointer, but it looks like the runner is already running a recent version `5.2.26(1)-release`. So we should be fine.

chmouel (approved):

markstur (approved):
lgtm

leseb:
@russellb PTAL

hickeyma (approved):
Great to get this working. Thanks @leseb and @markstur.

PR #1045, opened 2024-04-30
Disable push to Test PyPI on merge

tiran:
# Changes

Test PyPI does not accept `dev` builds with local version. Disable push on PR merge for now.

> The use of local versions in <Version('0.14.1.dev11+g12e0386')> is not
> allowed. See https://packaging.python.org/specifications/core-metadata


russellb (approved):
thanks!

PR #1057, opened 2024-04-30
docs: update `STABLE.md` location and content

nathan-weinberg:
Followup from Slack conversation around the `v0.14.1` release


docs/STABLE.md
```diff
@@ -1,18 +1,18 @@
...
-This can be done by:
+This can be done via the following steps:
 
 Get current:
 
 ```ShellSession
 $ git switch main
-$ git fetch
-$ git pull
+$ git fetch upstream
```
bjhargrave, line 14: The use of `upstream` is sort of developer dependent. That is, the convention of naming the remote for the main repo to `upstream` is not universal. I call it `origin`. So you should probably state that the remote name `upstream` used in this doc refers to the main repo.

An example below still pushes to origin :-)

bjhargrave (changes requested):

docs/STABLE.md
nathan-weinberg: Fair enough! I'll go back to what was there :+1:

docs/STABLE.md
```diff
@@ -47,7 +49,7 @@ Verify the tag SHA hashes look correct:
 $ git show-ref --tags
 ```
 
-Push the new tag to remote (origin) with `-f` (force) flag.
+Push the new tag to your fork (origin) with `-f` (force) flag.
```
bjhargrave, line 52: This section seems not useful. It assumes my fork has a stable tag. But forking in the github UI only copies the main branch (by default) to the fork. My fork has no tags since I never pushed any to my fork.

nathan-weinberg: @markstur any thoughts on that? Happy to remove this section but not sure if you had a specific intention for having it in here I'm not aware of

markstur: In this case "origin" is upstream.  So it has tags.   I guess this needs to be clarified for folks that use a fork as origin.

I have only been tagging directly on the upstream repo.  Never tagged my fork.  So there are no instructions here to tag a fork and push from there somehow.

bjhargrave (approved):

nathan-weinberg:
@Mergifyio rebase

nathan-weinberg:
@russellb is this not auto-merging due to the requested reviews?

markstur (commented):
I don't use "upstream" on this project.   So it's hard for me to "like" this, but I think you did a good job using upstream to clarify because apparently origin if ambiguous.

Issue #579, opened 2024-03-13
llamacpp_convert_to_gguf does not support 4-bit quantied models (KeyError: 'U8')


tiran:
I have GPU accelerated training with CUDA / ROCm and BitsAndBytes 4-bit quantization working. See https://github.com/instruct-lab/cli/pull/520#issuecomment-1993645744 for more information. However `lab train` fails to convert the trained model, because the output has different tensors and `llamacpp_convert_to_gguf` does not support that kind of model.

First of all, it fails because `SAFETENSORS_DATA_TYPES` does not have dtype `U8`. When I ignore tensors with dtype `U8`, then the code fails in `pick_output_type()`. The function expects `gguf.TENSOR_NAMES[gguf.MODEL_TENSOR.ATTN_Q].format(bid=0) + ".weight"` (`blk.0.attn_q.weight`), but the model has `ATTN_NORM` (`blk.0.attn_norm.weight`). The converter also complains about a lot of unknown tensors.

I'm not familiar with gguf format and don't know how to address the problem myself. To reproduce the problem, use my PR #520 and run training with `lab train --device cuda --4-bit-quant`. You need a supported GPU with at least 11 GB memory, a PyTorch build with CUDA or ROCm bindings, and BitsAndBytes for CUDA or ROCm.

If it's of any help, here is a list of tensors from the input model with their dtype and shape:

```
model.layers.0.input_layernorm.weight, dtype: F32, shape: [4096]
model.layers.0.mlp.down_proj.weight.nested_absmax, dtype: F32, shape: [3584]
model.layers.0.mlp.down_proj.weight.nested_quant_map, dtype: F32, shape: [256]
model.layers.0.mlp.down_proj.weight.quant_map, dtype: F32, shape: [16]
model.layers.0.mlp.gate_proj.weight.nested_absmax, dtype: F32, shape: [3584]
model.layers.0.mlp.gate_proj.weight.nested_quant_map, dtype: F32, shape: [256]
model.layers.0.mlp.gate_proj.weight.quant_map, dtype: F32, shape: [16]
model.layers.0.mlp.up_proj.weight.nested_absmax, dtype: F32, shape: [3584]
model.layers.0.mlp.up_proj.weight.nested_quant_map, dtype: F32, shape: [256]
model.layers.0.mlp.up_proj.weight.quant_map, dtype: F32, shape: [16]
model.layers.0.post_attention_layernorm.weight, dtype: F32, shape: [4096]
[...]

PR #1058, opened 2024-04-30
linux-train: Skip convert to gguf when --4-bit-quant is used

russellb:
This is known to be broken, per issue #579. Instead of failing,
exit cleanly before we know it will fail, but print a message
noting what was skipped, along with a link to the issue.

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -1002,6 +1002,13 @@ def train(
             shutil.copy(file, final_results_dir)
             print("Copied ", file, "to ", final_results_dir)
 
+        if four_bit_quant:
+            print(
+                "SKIPPING CONVERSION to gguf. This is unsupported with --4-bit-quant. "
+                + "See https://github.com/instructlab/instructlab/issues/579."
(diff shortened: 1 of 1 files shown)


tiran:
It's not the cleanest approach, but the entire `train` method is a mess anyway. Go for it!

src/instructlab/lab.py
```diff
@@ -1002,6 +1002,13 @@ def train(
             shutil.copy(file, final_results_dir)
             print("Copied ", file, "to ", final_results_dir)
 
+        if four_bit_quant:
+            print(
+                "SKIPPING CONVERSION to gguf. This is unsupported with --4-bit-quant. "
+                + "See https://github.com/instructlab/instructlab/issues/579."
+            )
+            return
```
leseb, line 1010: Why return and not make this part of the condition?

russellb: mainly because I prefer reducing levels of nesting -- style choice

nathan-weinberg (approved):

PR #1059, opened 2024-05-01
generate: warn when llama_cpp has not hardware acceleration

tiran:
# Changes

`ilab generate` without hardware acceleration is very slow. Notify the user when llama_cpp is not compiled with GPU or other accelerators. `llama_supports_gpu_offload` does not check whether a device is available or GPU offloading actually works.

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -484,6 +484,18 @@ def generate(
     if endpoint_url:
         api_base = endpoint_url
     else:
+        # Third Party
+        import llama_cpp
+
+        if not llama_cpp.llama_supports_gpu_offload():
+            # TODO: check for working offloading. The function only checks
+            # for compile time defines like `GGML_USE_CUDA`.
+            click.secho(
+                "llama_cpp_python is built without hardware acceleration. "
+                "ilab generate will be very slow.",
+                fg="red",
+            )
+
         try:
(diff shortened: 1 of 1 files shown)


src/instructlab/lab.py
```diff
@@ -484,6 +484,16 @@ def generate(
     if endpoint_url:
         api_base = endpoint_url
     else:
+        # Third Party
+        import llama_cpp
+
+        if not llama_cpp.llama_supports_gpu_offload():
+            click.secho(
+                "NOTE: llama_cpp is built without hardware acceleration. "
```
leseb, line 492:
```suggestion
                "llama_cpp is built without hardware acceleration. "
```

Is mentioning llama_cpp relevant for the end user?

tiran: It should be `llama_cpp_python`, not `llama_cpp`. It's relevant because that's what they have to rebuild and reinstall with hardware acceleration.

leseb: It's just "sad" that users have to have that sort of knowledge, this complicates the user experience. 🤷🏻

tiran: I hear you and agree with you. #1072 should make it a bit easier on our users.

leseb (approved):

russellb (approved):
nice improvement, thanks!

PR #1066, opened 2024-05-01
containers/cuda: Allow specifying git branch to build from

russellb:
This PR adds a new arg to the cuda Containerfile to specify which
instructlab git branch to build from. The default stays the same,
using `stable`.


leseb (commented):
I don't see any CI that builds this container (perhaps I missed it?), it'd be nice to add it in the future to make sure changes like this don't break anything :). Otherwise LGTM!

russellb:
> I don't see any CI that builds this container (perhaps I missed it?), it'd be nice to add it in the future to make sure changes like this don't break anything :). Otherwise LGTM!

Yes, we do need test builds in this repo. There is some automation elsewhere,  but we need it in this repo, too.

leseb (approved):

russellb:
This went in with #1086

PR #1069, opened 2024-05-01
Enforce maximum length in generate seed examples

### LINKED ISSUE #1024: Generate needs to consider trainer model context window length
When running `ilab generate` on a qna.yaml where questions/answers are long, the command can fail due to the trainer model's context window limitations.

`Generating dataset failed with the following error: There was a problem connecting to the server Error code: 400 - {'error': {'message': "This model's maximum context length is 4096 tokens. However, you requested 9360 tokens (6288 in the messages, 3072 in the completion). Please reduce the length of the messages or completion.", 'type': 'invalid_request_error', 'param': 'messages', 'code': 'context_length_exceeded'}}`

To reproduce this, use the base instructions for creating a custom model (instructlab/README.md), create a qna.yaml where the answers are really verbose, which should encourage the model to replicate the verbose pattern, and run `ilab generate`. After a few synthetic pairs you should see a failure.

If normal q/a pairs don't work, try replicating the same word over and over for multiple lines for each answer. This seems to encourage the model to be extremely verbose (and the content doesn't really matter for this test).

Also, don't forget to include the prompt.txt in the context size as I believe it also gets sent along with the qna content.

bbrowning:
This calculates a maximum number of characters per seed instruction example based on the configured server context size as passed to `ilab generate`. Before starting the generation, we verify that none of the examples we're going to use exceed this limit. If they do, we fail generation with a helpful error message.

When adding generated responses to our list of seed examples, we do a similar check of the response length. If the response is too long, we skip adding that particular response to our list of seed examples so that we don't end up adding really long machine-generated responses to our seed examples and blowing our context size that way.

Helps #1024 but doesn't entirely fix all sources of exceeding our context.

### DIFF
--- a/src/instructlab/generator/generate_data.py
+++ b/src/instructlab/generator/generate_data.py
@@ -21,7 +21,12 @@ import tqdm
 
 # Local
 from ..config import DEFAULT_MULTIPROCESSING_START_METHOD, get_model_family
-from ..utils import chunk_document, read_taxonomy
+from ..utils import (
+    chunk_document,
+    max_seed_example_tokens,
+    num_chars_from_tokens,
+    read_taxonomy,
+)
 from . import utils
 from .utils import GenerateException
 
@@ -362,6 +367,7 @@ def generate_data(
     tls_client_passwd: Optional[str] = None,
 ):
     seed_instruction_data = []
+    machine_seed_instruction_data = []
     generate_start = time.time()
 
     if not os.path.exists(output_dir):
@@ -376,6 +382,22 @@ def generate_data(
         )
     else:
         raise SystemExit(f"Error: taxonomy ({taxonomy}) does not exist.")
+
+    prompt_template = check_prompt_file(
+        prompt_file_path, get_model_family(model_family, model_name)
+    )
+    max_seed_tokens = max_seed_example_tokens(server_ctx_size, len(prompt_template))
+    max_seed_chars = num_chars_from_tokens(max_seed_tokens)
+    for seed_example in seed_instruction_data:
+        if (
+            len(seed_example["instruction"])
+            + len(seed_example["input"])
+            + len(seed_example["output"])
+            >= max_seed_chars
+        ):
+            raise SystemExit(
+                f"Error: An example in the taxonomy path {seed_example['taxonomy_path']} is too long for the server context size of {server_ctx_size}. Ensure the total number of characters across the combined question, answer, and context is less than {max_seed_chars} for each example or use a server with a larger context size."
+            )
 
     seeds = len(seed_instruction_data)
     logger.debug(f"Loaded {seeds} human-written seed instructions from {taxonomy}")
@@ -449,9 +471,6 @@ def generate_data(
         scorer._tokenizer.tokenize(inst) for inst in all_instructions
     ]
 
-    prompt_template = check_prompt_file(
-        prompt_file_path, get_model_family(model_family, model_name)
-    )
     if console_output:
         print(
             "Synthesizing new instructions. If you aren't satisfied with the generated instructions, interrupt training (Ctrl-C) and try adjusting your YAML files. Adding more examples may help."
@@ -471,7 +490,7 @@ def generate_data(
         # Filter the pool
         instruction_data_pool = [
             e
-            for e in seed_instruction_data + machine_instruction_data
+            for e in seed_instruction_data + machine_seed_instruction_data
             if e["taxonomy_path"] == selected_taxonomy
         ]
         instruction_data, discarded = get_instructions_from_model(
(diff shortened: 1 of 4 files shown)


bbrowning:
Marking as draft based on some conversations in Slack where at minimum more documented is needed here on the what/why we're doing this, where these calculations come from, and improvements we need to keep making.

bbrowning:
Ok, unmarking this as draft as this should substantially reduce how often we exceed the server's context size. And, at least with the default `ilab generate` that uses llama_cpp_python, this should entirely eliminate any fatal generation errors from exceeding context size and instead we'll just log a warning when that happens but keep generating.

Things are not perfect in the calculations here, and I tried to leave enough breadcrumbs in the added code comments so that as we continue to improve this based on user feedback/experience others will understand how and why we're calculating the maximum seed example length today.

src/instructlab/generator/generate_data.py
```diff
@@ -298,18 +303,35 @@ def get_instructions_from_model(
...
+        #
+        # Both llama_cpp_python and vllm use this exact string in their error
+        # responses when exceeding the model's max content length. Other
+        # OpenAI-compatible servers may as well, but no guarantees.
+        if "model's maximum context length" in str(exc):
+            logger.warn(
+                "Generated prompt exceeded the server's maximum context length. "
+                "If you see this warning many times during generation, lower "
+                "the length of your example question and answers or raise the "
+                "server's maximum context size."
```
leseb, line 331:
```suggestion
                "server's maximum context size using `max_ctx_size`."
```

bbrowning: Good idea - I was unable to sign off and commit your suggestion via the GitHub UI for some reason, but pushed another commit with it.

leseb: No worries,  you can probably squash and don't need a new commit for this. Unless you want to squash at the end of the reviews :)

jeremyeder:
is this a GA blocker for RHEL AI?

bbrowning:
For some additional context on why I use this, I cannot successfully complete any sufficiently long (say 500+ instructions) `ilab generate` run without this patch. Ultimately, the model ends up exceeding server's context window at some point during the run and the process aborting. With this change, it exceeds the generation window less and recovers gracefully when that happens, as long as the backend server hosting the teacher model is llama.cpp or vLLM.

Theoretically this isn't needed if your teacher model's context window is large enough that you never exceed it. In practice, with 4k and 8k context window teacher models I need this because inevitably one of the generated teacher question and answer pairs is really long and when that randomly gets selected back into my prompt for a future iteration of the generation loop the context window gets exceeded and the client aborts the generation loop.

lhawthorn:
> is this a GA blocker for RHEL AI?

Ibid [comment](https://github.com/instructlab/instructlab/pull/1107) in #1107

bbrowning:
This will need changes if #1107 merges, since that greatly expands our system prompt that we'll need to now account for in these calculating when determining the maximum length of seed examples.

russellb (approved):
I appreciate the detailed comments!

I understand there's some estimating going on here, but it seems like a nice improvement to what's there.

tests/test_lab_generate.py
```diff
@@ -162,6 +162,34 @@ def test_OpenAI_server_error(self, get_instructions_from_model):
...
+    def test_new_data_too_long(self):
+        runner = CliRunner()
+        with open("tests/testdata/skill_too_long_answer.yaml", "rb") as qnafile:
+            with runner.isolated_filesystem():
+                mt = MockTaxonomy(pathlib.Path("taxonomy"))
+                mt.create_untracked(
+                    "compositional_skills/tracked/qna.yaml", qnafile.read()
+                )
+                result = runner.invoke(
+                    lab.generate,
```
leseb, line 174: Where do you set the small context size?

leseb (changes requested):
Can you squash f4f5bb234129185c30c686380c9ca29c6d7334dd into 42900e5e44872807397bdd53fd153ab4930c127b. We don't need a dedicated commit for this suggestion.

I like the amount of comments and details in the code as well as the commit messages. Very nice work and way better than what we have at the moment (for the estimate piece). I just have one small comment around the test to overall LGTM! Thanks!

tests/test_lab_generate.py
bbrowning: I do not lower the context size there, but instead use a yaml file that is long enough to cause this problem - that's the added `tests/testdata/skill_too_long_answer.yaml` file. However, this does rely on the default context length of lab generate not changing from the 4096 it has today, so I'll figure out how to amend this to explicitly set a context length on the server-side in this test to ensure even if the default changes from 4096 the test still does a proper context length check explicitly.

leseb: Ohhh ok, thanks for clarifying. Right, amending by forcing `max_ctx_size` might make the test more resilient. If you want to address it soon I'll maintain my request change, but I don't see this as a blocker so if you want to do this as a followup I'm fine with that and I'll approve it. Just let me know.

bbrowning: I'll go ahead and fix this as well, since I have to make other changes in the test after rebasing on top of the last month's worth of commits.

bbrowning: ca5a9666b59324e0abe9ead65af0523e87a114ab sets a small context size explicitly, so marking this comment thread as resolved.

leseb (commented):
@bbrowning thanks for the last update. Do you have changes planned or can we go ahead? Thanks!

bbrowning:
I don't have anything else planned here, and the requested changes (squashing that 1 commit into another plus explicitly specifying a small context size in the test vs relying on the default to be small forever) have been made. So, please proceed :)

leseb (approved):

PR #1074, opened 2024-05-02
feat: use labels to trigger backports

leseb:
When setting a label to a PR, if it matches 'backport-release-v014' Mergify will automatically create a backport PR in the configured stable branch.
I've found this approach easier than writing a comment to a PR, also we can visually see if a PR made it in a release branch by looking at the label.
The only condition is to create the label.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -145,3 +145,11 @@ pull_request_rules:
     label:
       add:
         - backport
+
+- name: backport release-v0.14
+  actions:
+    backport:
+      branches:
+        - release-v0.14
+  conditions:
+    - label=backport-release-v0.14


leseb:
@russellb 👋🏻

nathan-weinberg:
Right now the flow is as follows:
1. comment to create a backport PR to a branch
2. backport PRs get the `backport` label automatically when they're created
So three open questions:
1. Do we make this an addition to the comment functionality, or replace it?
2. Do we keep the `backport` label being applied to backport PRs? i want a way to identify/batch all backport PRs
3. Will a backport PR be created automatically when the label is applied, even if the PR isn't merged? will it update automatically to match the open PR if that's the case?

PR #1083, opened 2024-05-02
Update the --model arg

spzala:
The arg `--model` is not supported. The correct arg is `--model-path`.

A related error,
`Error: No such option: --model`


spzala:
cc @markstur

leseb (approved):

hickeyma (approved):

PR #1086, opened 2024-05-02
Do test builds of the cuda image in CI

russellb:
- **containers/cuda: Allow specifying git branch to build from**
- **ci: Add a job that does test builds of the cuda image**

commit 971994d1df8a12e47cc00d3b94c11e8c0961054f
Date:   Sat Apr 27 21:07:59 2024 -0400

    containers/cuda: Allow specifying git branch to build from
    
    This PR adds a new arg to the cuda Containerfile to specify which
    instructlab git branch to build from. The default stays the same,
    using `stable`.
    
commit d6134622e3a64e215e45a39afd23c48723d31d70
Date:   Thu May 2 16:53:20 2024 -0400

    ci: Add a job that does test builds of the cuda image
    
    If the file `containers/cuda/Containerfile` changes, do a test build
    of it against `main` and `stable`. I don't have it doing test builds
    for all changes, because this isn't written to test building an image
    with the code in the current source tree.
    
---

Includes #1066

### DIFF
--- a/.github/workflows/image-cuda.yml
+++ b/.github/workflows/image-cuda.yml
@@ -1,0 +1,109 @@
+name: Test build cuda container image
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'containers/cuda/Containerfile'
+      - '.github/workflows/image-cuda.yml'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'containers/cuda/Containerfile'
+      - '.github/workflows/image-cuda.yml'
+
+# Note that the current containerfile builds against a git ref.
+# It is not built against the current source tree. So, we test
+# build the image against `stable` and `main` if the file changes.
+jobs:
+  build_cuda_image_stable:
+    name: Build CUDA image for stable
+    runs-on: ubuntu-latest
+    steps:
+      - name: Free disk space
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/image-cuda.yml
```diff
@@ -0,0 +1,109 @@
...
+
+# Note that the current containerfile builds against a git ref.
+# It is not built against the current source tree. So, we test
+# build the image against `stable` and `main` if the file changes.
+jobs:
+  build_cuda_image_stable:
+    name: Build CUDA image for stable
+    runs-on: ubuntu-latest
+    steps:
+      - name: Free disk space
```
tiran, line 25: I noticed that GHA have a large, mostly unused `/mnt` mount point with over 65GB of free disk space. Try this hack: https://github.com/tiran/instructlab-containers/blob/cbe5957fd94a375dad3c93c53264cf55a05a6a3a/.github/workflows/containers.yaml#L70-L82

russellb: nice.

The current changes are enough for these jobs, at least for now.

I also found this github action which should work if we configure it for our use case -- https://github.com/marketplace/actions/maximize-build-disk-space

nathan-weinberg (approved):

PR #1089, opened 2024-05-03
docs: add inline mermaid diagram of basic workflow

makelinux:
[preview](https://github.com/RedHat-Israel/instructlab/tree/diargam?tab=readme-ov-file#-what-is-ilab)

image


hickeyma (commented):
@makelinux What is the benefit over the current diagram: https://github.com/instructlab/instructlab/blob/main/docs/workflow.png?

russellb (approved):
lgtm, thanks!

Issue #976, opened 2024-04-24
ilab chat tries to talk to a local server that isn't ready yet


derekhiggins:
Using a newly trained model in a colab VM, I've observed that the model take several minutes to load when running the `ilab serve` command, it eventually works if you wait long enough

But when using chat with a local server, if you try to chat before the local model is ready it will fail with no indication as why

```
$ sed -i -e 's%model_path:.*%model_path: models/ggml-model-f16.gguf%g;' config.yaml
$ ilab chat -qq "what does the 'ilab download' command do?"
INFO 2024-04-23 22:58:48,351 _base_client.py:1039 Retrying request to /models in 0.978751 seconds
INFO 2024-04-23 22:58:49,332 _base_client.py:1039 Retrying request to /models in 1.523653 seconds
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:   no
ggml_cuda_init: CUDA_USE_TENSOR_CORES: yes
ggml_cuda_init: found 1 CUDA devices:
  Device 0: Tesla T4, compute capability 7.5, VMM: yes
INFO 2024-04-23 22:58:52,161 _base_client.py:1039 Retrying request to /chat/completions in 0.897866 seconds
INFO 2024-04-23 22:58:53,061 _base_client.py:1039 Retrying request to /chat/completions in 1.840679 seconds
Connection error, try again...
```

tiran:
Also see #889 for a related problem. There is no debug logging when the server dies or does not come up.

leseb:
This is likely fixed already, we have code to wait for the server to be ready before doing anything.

Issue #889, opened 2024-04-17
`ilab chat` and `generate` do not report server errors


tiran:
The `ilab chat` command does not wait until the server is up and running. It also does not forward error messages when `ilab serve` process crashes. I have had the issue a couple of times due to miscompiled llama-cpp-python, issues with GPU drivers, or a missing model file.

```shell
$ ilab chat
INFO 2024-04-17 21:10:01,265 _base_client.py:1040 Retrying request to /models in 0.867120 seconds
INFO 2024-04-17 21:10:02,133 _base_client.py:1040 Retrying request to /models in 1.706371 seconds
╭──────────────────────────────────────────────────────────────────────────────────── system ────────────────────────────────────────────────────────────────────────────────────╮
│ Welcome to Chat CLI w/ MERLINITE-7B-Q4_K_M (type /h for help)                                                                                                                  │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
>>> hello                                                                                                                                                             [S][default]
INFO 2024-04-17 21:10:06,540 _base_client.py:1040 Retrying request to /chat/completions in 0.965782 seconds
INFO 2024-04-17 21:10:07,507 _base_client.py:1040 Retrying request to /chat/completions in 1.638098 seconds
Connection error, try again...
>>> hello                                                                                                                                                             [S][default]
INFO 2024-04-17 21:10:10,205 _base_client.py:1040 Retrying request to /chat/completions in 0.783466 seconds
INFO 2024-04-17 21:10:10,990 _base_client.py:1040 Retrying request to /chat/completions in 1.800137 seconds
Connection error, try again...
>>> hello                                                                                                                                                             [S][default]
INFO 2024-04-17 21:10:13,607 _base_client.py:1040 Retrying request to /chat/completions in 0.934567 seconds
INFO 2024-04-17 21:10:14,543 _base_client.py:1040 Retrying request to /chat/completions in 1.842860 seconds
Connection error, try again...
>>> hello                                                                                                                                                             [S][default]
INFO 2024-04-17 21:10:17,541 _base_client.py:1040 Retrying request to /chat/completions in 0.855725 seconds
INFO 2024-04-17 21:10:18,398 _base_client.py:1040 Retrying request to /chat/completions in 1.944861 seconds
Connection error, try again...
>>>                                                                                                
```

Server crash with missing model:
```shell
$ ilab serve
[...]

PR #1090, opened 2024-05-03
feat: better troubleshooting on internal server error

leseb:
We try to surface as much information as we can using DEBUG level logging when the temporary server crashes. This is important when running the chat without a server or a generation.

Fixes: https://github.com/instructlab/instructlab/issues/976 and https://github.com/instructlab/instructlab/issues/889


leseb:
Will revisit later.

Issue #1091, opened 2024-05-03
ilab stack traces if an incorrect log level is set


leseb:
```
Traceback (most recent call last):
  File "/Users/leseb/Documents/AI/instructlab/venv/bin/ilab", line 8, in <module>
    sys.exit(cli())
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.9/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.9/site-packages/click/core.py", line 1077, in main
    with self.make_context(prog_name, args, **extra) as ctx:
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.9/site-packages/click/core.py", line 943, in make_context
    self.parse_args(ctx, args)
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.9/site-packages/click/core.py", line 1644, in parse_args
    rest = super().parse_args(ctx, args)
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.9/site-packages/click/core.py", line 1408, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.9/site-packages/click/core.py", line 2400, in handle_parse_result
    value = self.process_value(ctx, value)
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.9/site-packages/click/core.py", line 2362, in process_value
    value = self.callback(ctx, self, value)
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.9/site-packages/instructlab/lab.py", line 72, in configure
    ctx.obj = Lab(filename)
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.9/site-packages/instructlab/lab.py", line 46, in __init__
    self.logger.setLevel(self.config.general.log_level.upper())
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 1421, in setLevel
    self.level = _checkLevel(level)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 198, in _checkLevel
    raise ValueError("Unknown level: %r" % level)
[...]

PR #1092, opened 2024-05-03
fix: handle incorrect log level

leseb:
If the specified log level in the config file is incorrect we return an
error and print the allowed levels.
Sadly, only Python 3.11+ supports logging.getLevelNamesMapping that
could be used to get the exhaustive list of supported level names. So we
maintain our own for the time being.

Fixes: https://github.com/instructlab/instructlab/issues/1091
Co-authored-by: Ihar Hrachyshka <ihrachys@redhat.com>

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,6 +5,7 @@
 pre-commit>=3.0.4,<4.0
 pydeps>=1.12.12,<2
 pylint>=2.16.2,<4.0
+pylint-pydantic
 pytest
 pytest-asyncio
 pytest-cov
--- a/src/instructlab/config.py
+++ b/src/instructlab/config.py
@@ -4,7 +4,14 @@
 from typing import Optional
 
 # Third Party
-from pydantic import BaseModel, ConfigDict, PositiveInt, StrictStr, ValidationError
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    PositiveInt,
+    StrictStr,
+    ValidationError,
+    field_validator,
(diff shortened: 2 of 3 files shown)


nathan-weinberg:
Is the the config coming from the `config.yaml` file @leseb?

I wonder if instead of setting to `INFO`, we should instead exit and have the user go and fix the incorrect log level in the config file itself - but tbh I'm open to either approach

booxter:
We can validate it with `pydantic`, same way other config fields are already enforced, see `config.py`. Something like:

```
    @validator("log_level")
    def validate_log_level(cls, v):
        if not hasattr(logging, v):
            raise ValueError(f"{v} is not a valid log level name")
        return v
```

Sadly, only Python 3.11+ supports `logging.getLevelNamesMapping` that could be used to get the exhaustive list of supported level names. We can maintain our own list until we bump to 3.11 (it can also be extracted from `_nameToLevel` from `logging` module, but that's accessing an internal structure, so probably not a good idea.)

leseb:
@booxter PTAL, thanks!

russellb (approved):
lgtm, but I placed a `hold` on the PR to allow @booxter to review first.

This also looks like an easy one to add a functional test for, but I don't feel strongly enough to block over it.

booxter (approved):
lgtm; a func test could be useful, agreed

leseb:
@booxter @russellb unit tests added as part of https://github.com/instructlab/instructlab/pull/1155

PR #1093, opened 2024-05-03
fix: update config file and chat ctx while serving model

jaideepr97:
# Changes
Updates logic in `serve` and `chat` to use the `config.yaml` to determine model name to be displayed in the chat window

Resolves #

use model path provided to serve, identify model name from given path, update `chat.model` field in `config.yaml`, read the `chat.model` field and populate the chat ctx before displaying the name during `ilab chat`

### DIFF
--- a/src/instructlab/chat/chat.py
+++ b/src/instructlab/chat/chat.py
@@ -181,7 +181,7 @@ class ConsoleChatBot:  # pylint: disable=too-many-instance-attributes
             raise KeyboardInterrupt
         context = cs[1]
         if context not in CONTEXTS:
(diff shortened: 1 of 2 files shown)


russellb:
I will review this -- related to #935 that I was reviewing earlier

russellb (changes requested):
I would prefer we don't automatically edit any configuration.

I also think this doesn't take into account the possibility of running `ilab chat` against a remote endpoint.

jaideepr97:
closing since https://github.com/instructlab/instructlab/pull/935 was merged

PR #1109, opened 2024-05-03
Issue 1006 cpu optimization

### LINKED ISSUE #1006: Optimize CPU training on Linux

jessesightler-redhat:
# Changes

Resolves #1006

 Tweaks the data types used for cpu training. Speedup of only ~98% on a typical Red Hat laptop..

### DIFF
--- a/src/instructlab/config.py
+++ b/src/instructlab/config.py
@@ -23,7 +23,7 @@ DEFAULT_CHUNK_WORD_COUNT = 1000
 DEFAULT_NUM_INSTRUCTIONS = 100
 DEFAULT_PROMPT_FILE = "prompt.txt"
 DEFAULT_GENERATED_FILES_OUTPUT_DIR = "generated"
-DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=30.0)
+DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=300.0)
 # use spawn start method, fork is not thread-safe
 DEFAULT_MULTIPROCESSING_START_METHOD = "spawn"
 
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+import psutil # to determine system memory
 from typing import Optional
 
 # Third Party
@@ -17,6 +18,7 @@ from transformers import (
 )
 from trl import DataCollatorForCompletionOnlyLM, SFTTrainer
 import torch
+from tqdm import tqdm
 
 # Local
 from ..chat.chat import CONTEXTS
@@ -96,6 +98,17 @@ def report_cuda_device(args_device, min_vram=0):
             "stopping the server to free up about 5 GiB of GPU memory."
         )
 
+def determine_batch_size(device):
+    if device.type == "cpu":
+        # Gets the total system memory, excluding swap
+        mem_bytes = psutil.virtual_memory().total
+        mem_gb = mem_bytes / 1024 / 1024 / 1024
+        return 2 if mem_gb < 60 else 4
+    else:
+        # Defaulting for now... should check and return
+        # more for very large GPUs
+        return 1
+
 
 def linux_train(
     train_file: str,
@@ -155,9 +168,14 @@ def linux_train(
         model_name, torchscript=True, trust_remote_code=True
     )
 
+    model_torch_dtype = "auto"
+    if device.type == "cpu":
+        # Setting to None to improve inference performance on cpu
+        model_torch_dtype = None
+
     model = AutoModelForCausalLM.from_pretrained(
         model_name,
-        torch_dtype="auto",
+        torch_dtype=model_torch_dtype,
         quantization_config=bnb_config,
         config=config,
         trust_remote_code=True,
@@ -198,7 +216,7 @@ def linux_train(
 
     assistant_old_lst = [
         model_generate(d["user"]).split(response_template.strip())[-1].strip()
-        for d in test_dataset
+        for d in tqdm(test_dataset)
     ]
     attention_layers = [
         module for module in model.modules() if "attention" in str(type(module)).lower()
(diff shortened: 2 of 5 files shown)


src/instructlab/config.py
```diff
@@ -23,7 +23,7 @@
 DEFAULT_NUM_INSTRUCTIONS = 100
 DEFAULT_PROMPT_FILE = "prompt.txt"
 DEFAULT_GENERATED_FILES_OUTPUT_DIR = "generated"
-DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=30.0)
+DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=300.0)
```
leseb, line 26: We need to think about this, waiting up to 5min seems too long. If the server crashes the chat prompt will hang for 5min, this degrades the user experience.
How about tweaking the value for the CPU use case only?

src/instructlab/train/linux_train.py
```diff
@@ -96,6 +98,17 @@ def _gib(size: int) -> str:
             "stopping the server to free up about 5 GiB of GPU memory."
         )
 
+def determine_batch_size(device):
+    if device.type == "cpu":
+        # Gets the total system memory, excluding swap
+        mem_bytes = psutil.virtual_memory().total
+        mem_gb = mem_bytes / 1024 / 1024 / 1024
```
leseb, line 105: How about `mem_gb = mem_bytes / 1024**3`?

src/instructlab/train/linux_train.py
```diff
@@ -96,6 +98,17 @@ def _gib(size: int) -> str:
...
+        return 2 if mem_gb < 60 else 4
```
leseb, line 106: What's the rationale behind a batch size of 2 or 4 for the CPU case?

src/instructlab/train/linux_train.py
```diff
@@ -96,6 +98,17 @@ def _gib(size: int) -> str:
...
+    else:
```
leseb, line 107: Is the `else` always "gpu"?

src/instructlab/train/linux_train.py
```diff
@@ -155,9 +168,14 @@ def linux_train(
         model_name, torchscript=True, trust_remote_code=True
     )
 
+    model_torch_dtype = "auto"
+    if device.type == "cpu":
+        # Setting to None to improve inference performance on cpu
```
leseb, line 173: Can we expand that comment and explain why this is helping?

src/instructlab/train/linux_train.py
```diff
@@ -155,9 +168,14 @@ def linux_train(
         model_name, torchscript=True, trust_remote_code=True
     )
 
+    model_torch_dtype = "auto"
```
leseb, line 171: So IIUC, the "auto" is not capable of tuning the behavior based on CPU or GPU? Do you know what does it do then?

src/instructlab/train/linux_train.py
```diff
@@ -231,14 +249,24 @@ def model_generate(user):
     )
 
     output_dir = "./training_results"
-    per_device_train_batch_size = 1
+    per_device_train_batch_size = determine_batch_size(device)
+    print(f"LINUX_TRAIN.PY: Training with batch size: {per_device_train_batch_size}")
```
leseb, line 253: Do we need to print the file name?

src/instructlab/train/linux_train.py
```diff
@@ -231,14 +249,24 @@ def model_generate(user):
...
+
+    use_bf16 = not use_fp16
 
+    if device.type == "cpu":
+        print("LINUX_TRAIN.PY: Disabling fp16 and bf16 for cpu training")
```
leseb, line 258: ditto

leseb (commented):
Can we squash 5a850ea20219e4588ba5cc68c6b266297664ab82?

src/instructlab/train/linux_train.py
jessesightler-redhat: That's really for the project to decide, as it is only following the convention used by other print statements within the file. There are some exceptions as the file is not consistent.

src/instructlab/train/linux_train.py
jessesightler-redhat: It looks like it selects bfloat16. This is a terrible choice for my laptop and I'd guess most non-server CPUs as well as it triggers single threaded execution.

src/instructlab/train/linux_train.py
jessesightler-redhat: Larger batch sizes use more memory. It is attempting to fit as large of a size as possible.

src/instructlab/train/linux_train.py
jessesightler-redhat: Likely yes.

src/instructlab/config.py
jessesightler-redhat: There's little point to a lower timeout. If the server crashes, you'll get a broken pipe. If it doesn't crash and the timeout is too low, you'll get a weird network error while the server chugs along in the background. If you try it again, you'll be even more likely to get an error, since it didn't actually stop processing the previous request.

The only downside here is if your machine is too slow even for five minutes, you'll wait even longer before getting that error.

The short timeout is a real usability problem.

src/instructlab/train/linux_train.py
leseb: Alright, if this is the convention followed by the file already I'm fine with it. Thanks!

src/instructlab/config.py
leseb: This argument is true when running `ilab serve` + `ilab chat`, however when simply running `ilab chat`, a temporary server runs and if it crashes you would get the broken pipe on an incoming request, you just wait to 5 minutes.

cdoern:
This is going to be clobbered by #1157 lets hold this until that merges

jessesightler-redhat:
@cdoern I'll just close it. This took the training process from unusable to usable on typical laptops, but the review attention received was weird and largely unhelpful.

I have no interest in pursuing it further, but you are welcome to pick up where it left off if you'd like.

makelinux:
I would like to pull PR, but it unfortunately it doesn't work:
```
gh pr checkout 1109 
GraphQL: Could not resolve to a PullRequest with the number of 1109. (repository.pullRequest)
```

bjhargrave:
> I would like to pull PR, but it unfortunately it doesn't work:

This PR is closed which is probably why it down not let you check it out.

russellb:
I wanted to try this out, so I rebased it and posted it as a draft here.  https://github.com/instructlab/instructlab/pull/1222

I'll make updates in that PR based on what I find. Input welcome from others too, of course.

makelinux:
I've reviewed this PR. One change is already merged: #1193.

PR #1114, opened 2024-05-04
E2E tests on macOS

### LINKED ISSUE #1112: Provision a MacOS box for running unit and e2e tests
IIUC, the current actions workflow only runs on Ubuntu. Certain jobs such as the [e2e job](https://github.com/instructlab/instructlab/actions/workflows/e2e.yml) may behave differently between MacOS and Linux and this could lead to us missing breaking changes. 

Provision a box running MacOS that we can run unit and e2e tests on. Hopefully it wouldn't be too much more costly since it doesn't need to run all actions. Certain actions like lint and format as they shouldn't differ between systems.

tiran:
Resolves #1112

Use macos-latest runner to run E2E tests with mlx training.

### DIFF
--- a/.github/workflows/e2e-macos.yml
+++ b/.github/workflows/e2e-macos.yml
@@ -1,0 +1,86 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: E2E test on macOS ARM64
+
+on:
+    workflow_dispatch:
+    push:
+        branches:
+            - "main"
+            - "release-**"
+        paths:
+            - "**.py"
+            - "pyproject.toml"
+            - "requirements*.txt"
+            - "tox.ini"
+            - "scripts/*.sh"
+            - ".github/**"
+    pull_request:
+        branches:
+            - "main"
+            - "release-**"
+        paths:
+            - "**.py"
+            - "pyproject.toml"
+            - "requirements*.txt"
+            - "tox.ini"
+            - "scripts/*.sh"
+            - ".github/**"
+
+jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-macos.yml
```diff
@@ -0,0 +1,80 @@
...
+                  # config contains DEFAULT_MODEL
+                  key: huggingface-${{ hashFiles('src/instructlab/config.py') }}
+
+            - name: Install ilab
+              run: |
+                  python3 -m venv venv
+                  . venv/bin/activate
+                  sed 's/\[.*\]//' requirements.txt > constraints.txt
+                  python3 -m pip cache remove llama_cpp_python
+                  CMAKE_ARGS="-DLLAMA_METAL=on" python3 -m pip install --no-binary llama_cpp_python -c constraints.txt llama_cpp_python
```
leseb, line 74:
```suggestion
                  CMAKE_ARGS="-DLLAMA_METAL=off" python3 -m pip install --no-binary llama_cpp_python -c constraints.txt llama_cpp_python
```

Did you mean to disable it instead since the runner does not support it? I've seen the error and this is how I worked around it https://github.com/instructlab/instructlab/blob/main/scripts/functional-tests.sh#L60-L71.

tiran: I was hoping to get CI working with hardware acceleration.

Let's give your suggestion a try!

leseb:
@tiran why did you close this?

PR #1050, opened 2024-04-30
Don't fail if multiprocessing is already set to spawn; fix issues from lack of cleanup

tiran:
# Changes

**Description of your changes:**

Python multiprocessing does not allow th change the default start method when another code path has already configured a default start method. Even `get_start_method` or `get_context` sets a default context.

I was running into this exception when doing training with some hardware accelerator frameworks like Habana.

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/multiprocessing/context.py", line 247, in set_start_method
    raise RuntimeError('context has already been set')
RuntimeError: context has already been set
```

---

After fixing the issue above, we were still seeing multiprocessing-related bugs exposed via CI. We went through and made several other changes ensuring that resources were cleaned up appropriately. Those fixes are included, as well.

PR #1118, opened 2024-05-06
fix: re-add 'spawn' method for the multiprocessing task on the temp server

leseb:
With the help of https://github.com/instructlab/instructlab/pull/1050 we
have a working implementation of multiprocessing. We can then add again
the 'spawn' method when running the temporary server.

### DIFF
--- a/src/instructlab/server.py
+++ b/src/instructlab/server.py
@@ -20,7 +20,7 @@ import uvicorn
 
 # Local
 from .client import ClientException, list_models
-from .config import get_api_base
+from .config import DEFAULT_MULTIPROCESSING_START_METHOD, get_api_base
 
 templates = [
     {
@@ -71,7 +71,7 @@ def ensure_server(
         return (None, None, None)
     except ClientException:
         tried_ports = set()
-        mpctx = multiprocessing.get_context(None)
+        mpctx = multiprocessing.get_context(DEFAULT_MULTIPROCESSING_START_METHOD)
         # use a queue to communicate between the main process and the server process
         queue = mpctx.Queue()
         port = random.randint(1024, 65535)


src/instructlab/server.py
```diff
@@ -71,7 +71,7 @@ def ensure_server(
         return (None, None, None)
     except ClientException:
         tried_ports = set()
-        mpctx = multiprocessing.get_context(None)
+        mpctx = multiprocessing.get_context(DEFAULT_MULTIPROCESSING_START_METHOD)
```
russellb, line 74: Are you sure this changes behavior?  When passing `None`, it should use the currently set default context. In other words, I don't think this change does anything.  Can you let me know if I'm wrong?

russellb (changes requested):

src/instructlab/server.py
leseb: That's correct `None` will point to the default context. On Darwin, it's `spawn`, on Linux it's `fork`. We want to force `spawn` on all platforms.

@tiran please correct me if I'm wrong.

tiran: Multiprocessing has two different types of defaults. There is the initial default setting for the context and there is the global default context. On Linux, the initial default is `fork` for Python <= 3.13. Python 3.14 will use spawn. We force-set `spawn` everywhere and override the value for `get_context(None)`:

```python
>>> import multiprocessing
>>> multiprocessing.get_context(None)
<multiprocessing.context.ForkContext object at 0x7f40a5b4bf50>
>>> multiprocessing.set_start_method("spawn", force=True)
>>> multiprocessing.get_context(None)
<multiprocessing.context.SpawnContext object at 0x7f40a5b4bfe0>
```

leseb: Oh ok so this is the override:

https://github.com/instructlab/instructlab/blob/167ccd81f83dafbeb58d5f48fa2e87b2b7d45a6d/src/instructlab/lab.py#L31-L33

I'm closing this PR then. Thanks all!

PR #1121, opened 2024-05-06
tests: Isolate file system for all test cases, implicitly

### LINKED ISSUE #963: Isolate unit test file system for each test case using an implicit fixture
There's a patch to fix one of test cases that missed isolated_filesystem() context manager: https://github.com/instructlab/instructlab/pull/962

The patch fixes the immediate failure, but leaves the window open for the situation to happen in the future for another new test case. Ideally, the test suite would enforce a new file system structure for every test case in the test class(es). This issue is to build such mechanism, as a fixture or otherwise.

booxter:
# Changes

Resolves #963

Instead of applying the isolation decorator to each test case - and risking to miss it for some test cases - automatically isolate the file system using `autouse=True` fixture in global `conftest`.

### DIFF
--- a/tests/test_lab_diff.py
+++ b/tests/test_lab_diff.py
@@ -33,102 +33,91 @@ class TestLabDiff(unittest.TestCase):
         tracked_file = "compositional_skills/tracked/qna.yaml"
         self.taxonomy.create_untracked(untracked_file)
         self.taxonomy.add_tracked(tracked_file)
-        runner = CliRunner()
-        with runner.isolated_filesystem():
-            result = runner.invoke(
-                lab.diff,
-                [
-                    "--taxonomy-base",
-                    TAXONOMY_BASE,
-                    "--taxonomy-path",
-                    self.taxonomy.root,
-                ],
-            )
-            self.assertIn(untracked_file, result.output)
-            self.assertNotIn(tracked_file, result.output)
-            self.assertEqual(result.exit_code, 0)
+        result = CliRunner().invoke(
+            lab.diff,
+            [
+                "--taxonomy-base",
+                TAXONOMY_BASE,
+                "--taxonomy-path",
+                self.taxonomy.root,
+            ],
+        )
+        self.assertIn(untracked_file, result.output)
+        self.assertNotIn(tracked_file, result.output)
+        self.assertEqual(result.exit_code, 0)
 
     def test_diff_rm_tracked(self):
         tracked_file = "compositional_skills/tracked/qna.yaml"
         self.taxonomy.add_tracked(tracked_file)
         self.taxonomy.remove_file(tracked_file)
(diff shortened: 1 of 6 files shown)


booxter:
This PR will conflict on most test changes, so it would be nice to get it reviewed / merged quickly (assuming it's fitting of course).

tests/test_lab_diff.py
```diff
@@ -33,102 +33,91 @@ def test_diff(self):
...
-                    "--taxonomy-base",
-                    TAXONOMY_BASE,
-                    "--taxonomy-path",
-                    self.taxonomy.root,
-                ],
-            )
-            self.assertIn(untracked_file, result.output)
-            self.assertNotIn(tracked_file, result.output)
-            self.assertEqual(result.exit_code, 0)
+        result = CliRunner().invoke(
```
bjhargrave, line 36: Does it matter that there are now 2 CliRunners? One from the fixture for isolation and one in the test method for invoke. Perhaps the fixture could pass the runner to the test method so that it could be used for invoke?

booxter: I don't think it matters. For what I see in the context manager, it just changes the directory into a temp directory; then switches back on __exit__. In theory, we can even replace this isolated_filesystem context manager that depends on the runner with something else, runner is just a container for "useful utilities" in this context.

tiran (changes requested):
I find this approach too magic and confusing. With this patch, the auto-fixer applies to **all** test functions except a few that match a regular expression. It's going to apply to a bunch of functions that do not need the runner.

Can we follow the Zen of Python here, please? "

> Explicit is better than implicit."

I suggest that you create a regular fixture with name "clirunner", function scope, and no autouse. Yield a `CliRunner` instance from the fixture. Then pass `clirunner` to any function that needs access to a `CliRunner` object.

tiran:
PS: Pytest fixtures do not work with old-style unittest.TestCase classes. I have created #1135 to port the entire test suite to modern pytest-style testing. Draft PR #1134 has a preview of the new fixture.

booxter:
@tiran the source of this "magic" is that it was very easy to write a test case that doesn't isolate file system and "passes on my machine". (See the original issue.) This is a general solution for the suite (safe by default).

tiran:
An explicit fixture is a good approach to solve the issue and make the code more maintainable. A global auto-use fixture is problematic. The current fixture in this PR fires for every single test function, even for functions that do not touch the file system.

booxter:
We clearly have differents view on how test suites should be built and isolated.

If your concern would be just about the regex special casing, we could definitely switch these tests to not read files from current dir too. But seems like the disagreement is a lot deeper. Someone will have to post a patch with an alternative solution. This PR is not going anywhere.

PR #1125, opened 2024-05-07
Do not cache llama-cpp-python builds

tiran:
# Changes

Prevent any job from caching a llama-cpp-python wheel. It looks like a cached CUBLAS build from `e2e` test can leak into other tests that do not have the CUDA system libraries installed.

pip does not have an option to force a rebuild of a single package that has been locally cached. `--no-binary` applies to downloads but not to binary wheels in local cache. Purge `llama_cpp_python` wheels from cache after setup-python has populated pip's cache. Also remove the wheel before a job finishes.

Also refactor lint job to use the same caching and add constraints to a couple of packages to trigger the bug.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -50,11 +50,15 @@ jobs:
       - name: Setup Python 3.11
         uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
+          python-version: 3.11
           cache: pip
           cache-dependency-path: |
             **/pyproject.toml
             **/requirements*.txt
+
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e.yml
```diff
@@ -79,3 +83,8 @@ jobs:
         run: |
           . venv/bin/activate
           ./scripts/basic-workflow-tests.sh -cm
+
+      - name: Remove llama-cpp-python from cache
+        if: always()
```
russellb, line 88: This line probably isn't needed, but I guess it doesn't hurt?

russellb (approved):

PR #1126, opened 2024-05-07
e2e: Disable pip cache

russellb:
The caching from this job seems to be causing a problem with the rest
of the test jobs that do not install with cuda support. This is
causing CI failures where you see packages fail to load because NVIDIA
dependencies are not on the system.

The quick fix for the moment is to stop caching from this job. We'll
have to come back around and do a more focused cache specifically for
this.


russellb:
Nevermind, #1125 is a better solution

PR #1127, opened 2024-05-07
fix: make chat model detection from server more robust

leseb:
We added a few guardrails depending on the response from the server. Especially if there is less or more than one model being available on the server.
Followup of #911.

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -666,6 +666,8 @@ def chat(
     from .client import ClientException, list_models
     from .server import ensure_server, is_temp_server_running
 
+    logger = ctx.obj.logger
+
     if endpoint_url:
         api_base = endpoint_url
         server_process = None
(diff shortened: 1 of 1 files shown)


Agu1l4r821 (approved):

russellb (changes requested):
It looks like this change is handling a case that we know will never happen. I'm not sure if it's worth it. Do you see a case where the new code path could occur? Or are you thinking this is to protect against changes in our default serving support that might be serving multiple models in the future?

There is a related change that is needed - #1102. That PR is trying to add validation of the requested model name when something specific is requested.

leseb:
> It looks like this change is handling a case that we know will never happen. I'm not sure if it's worth it. Do you see a case where the new code path could occur? Or are you thinking this is to protect against changes in our default serving support that might be serving multiple models in the future?
> 
> There is a related change that is needed - #1102. That PR is trying to add validation of the requested model name when something specific is requested.



> It looks like this change is handling a case that we know will never happen. I'm not sure if it's worth it. Do you see a case where the new code path could occur? Or are you thinking this is to protect against changes in our default serving support that might be serving multiple models in the future?
> 
> There is a related change that is needed - #1102. That PR is trying to add validation of the requested model name when something specific is requested.

The way I see it is more of a future-proof code for upcoming features. This code is not changing the current behavior but it could catch potential errors or misconfiguration in the future.

I think #1102 is important.

russellb:
I feel a bit odd about merging this, knowing it'll never run and that we have no way to test it. It is certainly an area worth looking at if / when we change the server behind `ilab serve`, but I think right now my preference would be to leave the code as-is, focused on the currently expected behavior.

leseb:
Closing based on https://github.com/instructlab/instructlab/pull/1127#issuecomment-2110294586

PR #1135, opened 2024-05-08
Use pytest instead of unittest.TestCase

tiran:
# Changes

Tests now use modern pytest style instead of old `unittest.TestCase`. This has multiple benefits. Test code is easier to read. Pytest fixtures in functions now work. Test failures now show error context and local variables.

The code was automatically fixed by `pytestify==1.5.0`. ExceptionInfo was manually fixed to `str(exc.value)`.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,8 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
-# Standard
-import unittest
-
 # Third Party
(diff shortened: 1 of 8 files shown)


bjhargrave (commented):
+1 from me. I will let another maintainer review and approve.

russellb (approved):
thanks!

PR #1137, opened 2024-05-08
Refactor CliRunner tests

tiran:
# Changes

Functional tests now behave more like actual click CLI. Tests cases no longer run the subcommand function directly. Instead they invoke the cli main entry point and delegate the subcommands to click.

The config file is now parsed in the main cli entry point instead of the option's validation callback. The `ctx.obj` is created in the entry point, too. The change was necessary because click's CliRunner does neither update `sys.argv` or set `ctx.invoked_subcommand` for the callback. IMO it's also a nicer design.


bjhargrave (approved):
LGTM

PR #1111, opened 2024-05-03
Run e2e training test without 4 bit quant

derekhiggins:
Set training options to use GMU RAM to fit in a Tesla T4.

PR #1139, opened 2024-05-09
e2e: Add timestamps to the test script

russellb:
Every time we log a new step in the script, include a timestamp. This
should help with seeing how long each piece is taking within the
script.

This came up in discussion on #1111.

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -26,7 +26,7 @@ NC='\033[0m' # No Color
 SCRIPTDIR=$(dirname "$0")
 
 step() {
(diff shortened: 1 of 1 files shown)


russellb:
test e2e run here -- https://github.com/instructlab/instructlab/actions/runs/9013033848

Agu1l4r821 (approved):

derekhiggins:
lgtm 
although when I mentioned it I was more thinking of line by line timestamps, I now see that line by line would be redundant as github provides it (in the results output page, click settings icon in the top right and then show timestamps)

Agu1l4r821 (approved):

anik120 (approved):

PR #1015, opened 2024-04-26
[WIP] Torch device helpers to abstract device properties

tiran:
Torch devices and hadware accelerators have different properties and require special configuration. For Intel Gaudi HPUs, I started a new system that provices an abstracts layer for device properties. The system also maps 'cuda' to either NVIDIA CUDA or AMD ROCm properties.

# Changes

**Which issue is resolved by this Pull Request:**
Resolves #

**Description of your changes:**

PR #1157, opened 2024-05-13
train rework, introduce --backend and --dtype flags

### LINKED ISSUE #1108: ilab train pytorch support
instructlab-training does not have CPU support. It does now support the MPS device type. This means that once we add CPU support, we can remove `--legacy` and probably the custom MLX support we currently have in `ilab`. The overall functionality of the training library is better and is based off of pytorch.

cdoern:
# Changes

added `--backend` flag for `ilab train` allowing users to switch their training backend between pytorch and mlx. 

changed what was `linux_train.py` to work for both macos and linux depending on some new flags like fp16, bfp16, etc. 

resolves #1108 

ilab train on macos and linux are really different just for MLX support.

It turns out pytorch supports an mps device enabling hardware accleration on macos. This allows us to use 99% of the codepath for linux train with less storage and memory used. the mlx method requires adapters files, and a whole -fused dir that almost doubles the storage used in this whole process

This backend is also more maintainable as mlx is written in a way that lead us to need some heavy infrastructure to run it.

In testing, the pytorch train code takes roughly the same amount of time on comparable hardware and yields a better result with a model that appears to learn the new skill/knowledge in a better manner with less hallucinations. 

I will note, 18gb of ram is the minimum to run this backend. When you have 32/38 train takes less than a quarter of the amount of time that 18gb does, and produces better results. 

ilab train now defaults to --backend=pytorch but --backend=mlx is still available

added a few more enhancements:
- the --backend flag lets users choose pytorch or mlx
- I got rid of the multiple flags referencing "model" and just made a --model-repo flag since train should always pull the full safetensors from HF
- I got rid of a few bad flags people should probably never use like gguf-model-path gguf models for training yield horrible results most of the time
- --dtype takes fp16, bf16, ftp32, and auto. This sets the dtype to user in pytorch. bf16 and fp16 were hardcoded in the past.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,11 +29,11 @@ rouge-score>=0.1.2,<0.2.0
 sentencepiece>=0.2.0,<0.3.0
 tokenizers>=0.15.2,<0.16.0
 toml>=0.10.2,<0.11.0
-# Habana installer has 2.2.0a0+git8964477 with oneMKL
 torch>=2.2.0a0,<3.0.0 ; python_version == '3.10'
-torch>=2.2.1,<3.0.0 ; python_version != '3.10'
+torch>=2.3.0,<3.0.0 ; python_version != '3.10'
 tqdm>=4.66.2,<5.0.0
-transformers>=4.30.0,<=4.38.2
+transformers==4.38.1; sys_platform == 'darwin' and platform_machine == 'arm64'
+transformers>=4.37.0,<=4.38.2; sys_platform != 'darwin' and platform_machine != 'arm64'
 trl>=0.7.11,<0.8.0
 wandb>=0.16.4,<0.17.0
 langchain-text-splitters
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -80,18 +80,18 @@ def cli(ctx, config_file):
         if config_file == "DEFAULT":
             config_obj = config.get_default_config()
         elif not os.path.isfile(config_file):
-            config_obj = None
             ctx.fail(
                 f"`{config_file}` does not exists, please run `ilab init` "
                 "or point to a valid configuration file using `--config=<path>`."
             )
+            assert False, "unreachable"
         else:
             try:
                 config_obj = config.read_config(config_file)
             except config.ConfigException as ex:
                 raise click.ClickException(str(ex))
-
-        ctx.obj = Lab(config_obj)
+        if config_obj is not None:
+            ctx.obj = Lab(config_obj)
         # default_map holds a dictionary with default values for each command parameters
         ctx.default_map = config.get_dict(ctx.obj.config)
 
@@ -847,7 +847,7 @@ class TorchDeviceParam(click.ParamType):
     """
 
     name = "deviceinfo"
-    supported_devices = {"cuda", "cpu", "hpu"}
+    supported_devices = {"cuda", "cpu", "hpu", "mps"}
 
     def convert(self, value, param, ctx) -> "torch.device":
         # pylint: disable=C0415
@@ -860,7 +860,6 @@ class TorchDeviceParam(click.ParamType):
                 device = torch.device(value)
             except RuntimeError as e:
                 self.fail(str(e), param, ctx)
-
         if device.type not in self.supported_devices:
             supported = ", ".join(repr(s) for s in sorted(self.supported_devices))
             self.fail(
@@ -901,39 +900,22 @@ TORCH_DEVICE = TorchDeviceParam()
 @cli.command()
 @click.option("--data-dir", help="Base directory where data is stored.", default=None)
 @click.option(
-    "--input-dir",
-    type=click.Path(),
-    show_default=True,  # TODO: set to None and change help message
-    help="Path to generated files to use as input.",
-)
-@click.option(
     "--skip-preprocessing",
     is_flag=True,
 )
 @click.option(
-    "--tokenizer-dir",
-    help="Base directory where tokenizer is stored.",
-    default=None,
-    show_default=True,
+    "--input-dir",
+    type=click.Path(exists=True, file_okay=False, dir_okay=True),
+    show_default=True,  # TODO: set to None and change help message
+    help="Path to generated files to use as input.",
 )
 @click.option(
-    "--gguf-model-path",
-    help="Local directory where gguf model is stored.",
-    default=None,
-    show_default=True,
-)
-@click.option(
-    "--model-dir",
-    help="Base directory where model is stored.",
+    "--model-repo",
+    help="Base repository where model is stored.",
     default="instructlab/merlinite-7b-lab",
     show_default=True,
 )
 @click.option("--iters", help="Number of iterations to train LoRA.", default=100)
-@click.option(
-    "--local",
-    is_flag=True,
-    help="Whether or not `model_dir` is remote from HuggingFace.",
-)
 @click.option(
     "-sq",
     "--skip-quantize",
@@ -955,6 +937,7 @@ TORCH_DEVICE = TorchDeviceParam()
     help=(
         "PyTorch device for Linux training (default: 'cpu'). Use 'cuda' "
         "for NVidia CUDA / AMD ROCm GPU, 'cuda:0' for first GPU."
+        "Use 'mps' for Apple Silicon hardware."
     ),
 )
 @click.option(
@@ -972,33 +955,43 @@ TORCH_DEVICE = TorchDeviceParam()
     ),
 )
 @click.option(
-    "--model-name",
-    default="instructlab/merlinite-7b-lab",
+    "--dtype",
+    type=click.Choice(["bf16", "fp16", "fp32", "auto"]),
     show_default=True,
-    help="model name to use in training",
+    default="auto",
+    help=("Full half precision training. BF16 is more precise than FP16."),
+)
+@click.option(
+    "--backend",
+    type=click.Choice(["pytorch", "mlx"]),
+    hidden=True,
+    show_default=True,
+    default="pytorch",
+    help=("Backend to use for training. Options are 'mlx' or 'pytorch'."),
 )
 @click.pass_context
 def train(
     ctx,
     data_dir,
-    input_dir,
     skip_preprocessing,
-    tokenizer_dir,
-    gguf_model_path,
-    model_dir,
+    input_dir,
+    model_repo,
     iters,
-    local,
     skip_quantize,
     num_epochs,
     device: "torch.device",
     four_bit_quant: bool,
-    model_name: str,
+    dtype: str,
+    backend,
 ):
     """
     Takes synthetic data generated locally with `ilab generate` and the previous model and learns a new model using the MLX API.
     On success, writes newly learned model to {model_dir}/mlx_model, which is where `chatmlx` will look for a model.
     """
     # pylint: disable=C0415
+    # Third Party
+    from instructlab_quantize import run_quantize  # pylint: disable=import-error
+
     if not input_dir:
         # By default, generate output-dir is used as train input-dir
         input_dir = ctx.obj.config.generate.output_dir
@@ -1045,19 +1038,31 @@ def train(
             )
             raise click.exceptions.Exit(1)
 
-    if not utils.is_macos_with_m_chip():
+    if backend == "pytorch":
         # Local
         from .llamacpp.llamacpp_convert_to_gguf import convert_llama_to_gguf
-        from .train.linux_train import linux_train
+        from .train.pytorch_train import pytorch_train
 
-        linux_train(
-            ctx=ctx,
+        if device.type == "mps":
+            if (
+                all("PYTORCH_MPS_HIGH_WATERMARK_RATIO", "PYTORCH_ENABLE_MPS_FALLBACK")
+                not in os.environ
+            ):
+                click.secho(
+                    "To use MPS, PYTORCH_MPS_HIGH_WATERMARK_RATIO must be set to 0.0 and PYTORCH_ENABLE_MPS_FALLBACK must be set to 1",
+                    fg="red",
+                )
+                raise click.exceptions.Exit(1)
+        best_checkpoint = pytorch_train(
+            ctx,
             train_file=train_files[0],
             test_file=test_files[0],
-            model_name=model_name,
+            model_name=model_repo,
             num_epochs=num_epochs,
             device=device,
             four_bit_quant=four_bit_quant,
+            dtype=dtype,
+            train_style="full",
         )
 
         training_results_dir = "./training_results"
@@ -1073,42 +1078,58 @@ def train(
         if os.path.isfile(gguf_models_file):
             os.remove(gguf_models_file)
 
+        if best_checkpoint is None:
+            all_checkpoints = glob(training_results_dir + "/checkpoint-*")
+            best_checkpoint = all_checkpoints[0]
+            print(
+                f"Did not get best checkpoint, choosing most recent which is {best_checkpoint}"
+            )
         # TODO: Figure out what to do when there are multiple checkpoint dirs.
         # Right now it's just copying files from the first one numerically not necessarily the best one
-        added_tokens_file = glob(
-            training_results_dir + "/checkpoint-*/added_tokens.json"
-        )
-        special_tokens_map = glob(
-            training_results_dir + "/checkpoint-*/special_tokens_map.json"
-        )
-        tokenizer_json = glob(training_results_dir + "/checkpoint-*/tokenizer.json")
-        tokenizer_model = glob(training_results_dir + "/checkpoint-*/tokenizer.model")
-        tokenizer_config_json = glob(
-            training_results_dir + "/checkpoint-*/tokenizer_config.json"
-        )
+        added_tokens_file = os.path.join(best_checkpoint, "added_tokens.json")
+        special_tokens_map = os.path.join(best_checkpoint, "special_tokens_map.json")
+        tokenizer_json = os.path.join(best_checkpoint, "tokenizer.json")
+        tokenizer_model = os.path.join(best_checkpoint, "tokenizer.model")
+        tokenizer_config_json = os.path.join(best_checkpoint, "tokenizer_config.json")
         config_json = glob(training_results_dir + "/merged_model/config.json")
         generation_config_json = glob(
             training_results_dir + "/merged_model/generation_config.json"
         )
         safe_tensors = glob(training_results_dir + "/merged_model/*.safetensors")
 
-        shutil.copy(added_tokens_file[0], final_results_dir)
-        print("Copied ", added_tokens_file[0], "to ", final_results_dir)
-        shutil.copy(special_tokens_map[0], final_results_dir)
-        print("Copied ", special_tokens_map[0], "to ", final_results_dir)
-        shutil.copy(tokenizer_json[0], final_results_dir)
-        print("Copied ", tokenizer_json[0], "to ", final_results_dir)
-        shutil.copy(tokenizer_model[0], final_results_dir)
-        print("Copied ", tokenizer_model[0], "to ", final_results_dir)
-        shutil.copy(tokenizer_config_json[0], final_results_dir)
-        print("Copied ", tokenizer_config_json[0], "to ", final_results_dir)
-        shutil.copy(config_json[0], final_results_dir)
-        print("Copied ", config_json[0], "to ", final_results_dir)
-        shutil.copy(generation_config_json[0], final_results_dir)
-        print("Copied ", generation_config_json[0], "to ", final_results_dir)
+        try:
+            shutil.copy(added_tokens_file, final_results_dir)
+            print("Copied ", added_tokens_file, "to ", final_results_dir)
+            shutil.copy(special_tokens_map, final_results_dir)
+            print("Copied ", special_tokens_map, "to ", final_results_dir)
+            shutil.copy(tokenizer_json, final_results_dir)
+            print("Copied ", tokenizer_json, "to ", final_results_dir)
+            shutil.copy(tokenizer_model, final_results_dir)
+            print("Copied ", tokenizer_model[0], "to ", final_results_dir)
+            shutil.copy(tokenizer_config_json, final_results_dir)
+            print("Copied ", tokenizer_config_json, "to ", final_results_dir)
+            shutil.copy(config_json[0], final_results_dir)
+            print("Copied ", config_json[0], "to ", final_results_dir)
+            shutil.copy(generation_config_json[0], final_results_dir)
+            print("Copied ", generation_config_json[0], "to ", final_results_dir)
+        # pylint: disable=W0718
+        except Exception as exc:
+            click.secho(
+                f"Could not copy train files: {exc}",
+                fg="red",
+            )
+            return
         for file in safe_tensors:
-            shutil.move(file, final_results_dir)
-            print("Moved ", file, "to ", final_results_dir)
+            try:
+                shutil.copy(file, final_results_dir)
+                print("Copied ", file, "to ", final_results_dir)
+            # pylint: disable=W0718
+            except Exception as exc:
+                click.secho(
+                    f"Could not copy train files: {exc}",
+                    fg="red",
+                )
+                return
 
         if four_bit_quant:
             print(
@@ -1124,16 +1145,24 @@ def train(
         for file in glob(final_results_dir + "/*.safetensors"):
             os.remove(file)
 
-        shutil.move(gguf_file_path, gguf_models_file)
-
+        gguf_models_dir = "./models"
+        if not os.path.isdir(gguf_models_dir):
+            os.mkdir(gguf_models_dir)
+        gguf_model = os.path.join(final_results_dir, "ggml-model-f16.gguf")
+        gguf_model_q = "./models/ggml-model-Q4_K_M.gguf"
+        if not skip_quantize:
+            run_quantize(gguf_model, gguf_model_q, "Q4_K_M")
+        else:
+            shutil.move(gguf_file_path, gguf_models_file)
+        # cleanup original copy of model
+        if os.path.exists(final_results_dir + "/ggml-model-f16.gguf"):
+            os.remove(final_results_dir + "/ggml-model-f16.gguf")
         # cleanup checkpoint dir since it's name is unpredictable
         # TODO: figure out how checkpoint dirs should be cleaned up
-        # checkpoint_dirs = glob(training_results_dir + "/checkpoint*")
-        # shutil.rmtree(checkpoint_dirs[0])
+        checkpoint_dirs = glob(training_results_dir + "/checkpoint*")
+        shutil.rmtree(checkpoint_dirs[0])
     else:
         # Local
-        from .mlx_explore.gguf_convert_to_mlx import load
-        from .mlx_explore.utils import fetch_tokenizer_from_hub
         from .train.lora_mlx.convert import convert_between_mlx_and_pytorch
         from .train.lora_mlx.lora import load_and_train
         from .train.lora_mlx.make_data import make_data
@@ -1150,7 +1179,7 @@ def train(
 
         # NOTE we can skip this if we have a way ship MLX
         # PyTorch safetensors to MLX safetensors
-        model_dir_local = model_dir.replace("/", "-")
+        model_dir_local = model_repo.replace("/", "-")
         model_dir_mlx = f"{model_dir_local}-mlx"
         model_dir_mlx_quantized = f"{model_dir_local}-mlx-q"
 
@@ -1161,29 +1190,12 @@ def train(
             dest_model_dir = model_dir_mlx_quantized
             quantize_arg = True
 
-        if tokenizer_dir is not None and gguf_model_path is not None:
-            if not local:
-                tokenizer_dir_local = tokenizer_dir.replace("/", "-")
-                fetch_tokenizer_from_hub(tokenizer_dir, tokenizer_dir_local)
-
-            # no need to pass quantize_arg for now, script automatically detects if quantization is necessary based on whether gguf model is quantized or not
-            load(gguf=gguf_model_path, repo=tokenizer_dir, mlx_path=dest_model_dir)
-
-            for filename in os.listdir(model_dir_local):
-                shutil.copy(
-                    os.path.join(model_dir_local, filename),
-                    os.path.join(dest_model_dir, filename),
-                )
-            shutil.rmtree(model_dir_local, ignore_errors=True)
-
-        else:
-            # Downloading PyTorch SafeTensor and Converting to MLX SafeTensor
-            convert_between_mlx_and_pytorch(
-                hf_path=model_dir,
-                mlx_path=dest_model_dir,
-                quantize=quantize_arg,
-                local=local,
-            )
+        # Downloading PyTorch SafeTensor and Converting to MLX SafeTensor
+        convert_between_mlx_and_pytorch(
+            hf_path=model_repo,
+            mlx_path=dest_model_dir,
+            quantize=quantize_arg,
+        )
 
         adapter_file_path = f"{dest_model_dir}/adapters.npz"
         # train the model with LoRA
@@ -1197,9 +1209,6 @@ def train(
             save_every=10,
             steps_per_eval=10,
         )
-
-        # TODO copy some downloaded files from the PyTorch model folder
-        # Seems to be not a problem if working with a remote download with convert.py
 
 
 @cli.command()
@@ -1293,9 +1302,11 @@ def test(data_dir, model_dir, adapter_file):
     default=None,
     show_default=True,
 )
-@utils.macos_requirement(echo_func=click.secho, exit_exception=click.exceptions.Exit)
 @click.pass_context
-def convert(ctx, model_dir, adapter_file, skip_de_quantize, skip_quantize, model_name):
+@utils.macos_requirement(echo_func=click.secho, exit_exception=click.exceptions.Exit)
+def convert(
+    ctx, model_dir, adapter_file, skip_de_quantize, skip_quantize
+):
     """Converts model to GGUF"""
     # pylint: disable=C0415
     # Third Party
@@ -1340,7 +1351,6 @@ def convert(ctx, model_dir, adapter_file, skip_de_quantize, skip_quantize, model
     convert_llama_to_gguf(
         model=model_dir_fused_pt,
         pad_vocab=True,
-        skip_unknown=True,
         outfile=f"{model_dir_fused_pt}/{model_name}.gguf",
     )
 
--- a/src/instructlab/train/pytorch_train.py
+++ b/src/instructlab/train/pytorch_train.py
@@ -142,7 +142,7 @@ def report_hpu_device(args_device: torch.device) -> None:
             print(f'  {key}="{value}"')
 
 
-def linux_train(
+def pytorch_train(
     ctx: click.Context,
     train_file: str,
     test_file: str,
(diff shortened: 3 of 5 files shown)


tiran (changes requested):
Nice work, Charlie! I have left a few inline comments.

The current state of the PR breaks Intel Gaudi support. It may also cause problems for AMD ROCm. We don't have CI runners for these accelerators, so you probably haven't noticed it.

I also suggest to rename the module like the `--backend` argument. `from instructlab.train.pytorch import pytorch_train` is more descriptive than `linux_macos_train`.

cdoern:
> Nice work, Charlie! I have left a few inline comments.
> 
> The current state of the PR breaks Intel Gaudi support. It may also cause problems for AMD ROCm. We don't have CI runners for these accelerators, so you probably haven't noticed it.
> 
> I also suggest to rename the module like the `--backend` argument. `from instructlab.train.pytorch import pytorch_train` is more descriptive than `linux_macos_train`.

thanks for the review @tiran I had a feeling gaudi support might be impacted by this. I will try to clean everything up.

taxonomy_data/test_gen.jsonl
```diff
@@ -0,0 +1,6 @@
+{"system": "You are an AI language model developed by IBM Research. You are a cautious assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior.", "user": "David has three sisters. and he is their brother Each of them has one brother. How many brothers does David have?\n", "assistant": "David has three sisters, and each of them has one brother.\nDavid is their one brother. So, David has no brothers.\n"}
```
alimaredia, line 1: Are these files, accidental adds?

RobotSail (approved):
LGTM Witnessed this PR work end-to-end.

requirements.txt
```diff
@@ -30,14 +30,14 @@ rouge-score>=0.1.2,<0.2.0
...
-transformers>=4.30.0,<=4.38.2
+transformers==4.38.1
 trl>=0.7.11,<0.8.0
 wandb>=0.16.4,<0.17.0
 langchain-text-splitters
 # the below library should NOT be imported into any python files
 # it is for CLI usage ONLY
 yamllint>=1.35.1,<1.36.0
+bitsandbytes
+accelerate
```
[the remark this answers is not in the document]
cdoern: I see there is a `requirements-hpu.txt` I don't think we should have niche stack's in our main requirements.txt. 

Would having that other requirements.txt contain all gaudi reqs be ok?

src/instructlab/train/linux_macos_train.py
```diff
@@ -142,21 +145,26 @@ def report_hpu_device(args_device: torch.device) -> None:
             print(f'  {key}="{value}"')
 
 
-def linux_train(
-    ctx: click.Context,
+def train(
```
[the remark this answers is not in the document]
cdoern: done!

src/instructlab/lab.py
```diff
@@ -961,29 +944,62 @@ def convert(self, value, param, ctx) -> "torch.device":
         "Use BitsAndBytes for 4-bit quantization "
         "(reduces GPU VRAM usage and may slow down training)"
     ),
+    callback = utils.macosrequired,
 )
 @click.option(
-    "--model-name",
-    default="instructlab/merlinite-7b-lab",
+    "--fp16",
```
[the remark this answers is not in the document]
tiran: CPU and older CUDA devices also have quirks with bf16. Some settings are device and hardware specific. My PR #1015 is an attempt to provide a unified API around devices and device configuration.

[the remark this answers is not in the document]
cdoern: what i did is `bf16 = not fp16 and device.type != mps` since when bf16 is enabled torch tells me bf16 doesn't work with mps

requirements.txt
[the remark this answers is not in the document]
cdoern: @tiran ok, here are the requirements we have overlapping:

Gaudi: Python 3.10, torch 2.2.0a
MPS and all other Devices: 3.10+, torch 2.3.0+

Can you please advise the best way to merge these two requirements? 

I was thinking `pip install  -r requirements-hpu.txt -e .` would allow us to override the torch version in requirements.txt. 

One thing I was thinking of would be to have `make hpu` or `make mps` or `make cuda` targets that also build ilab to be able to handle these backends?

I am open to whatever way is best! Though, I don't think having this stuff in the base requirements file makes sense since conflicts like this are bound to happen a lot.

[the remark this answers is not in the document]
tiran: Optional dependencies cannot introduce conflicting dependencies. They are limited to additional dependencies or more restrictive dependencies. Pip flat-out refuses to resolve conflicts:

```
ERROR: Cannot install instructlab[hpu]==0.14.1.dev183+gdeee169.d20240514 because these package versions have conflicting dependencies.

The conflict is caused by:
    instructlab[hpu] 0.14.1.dev183+gdeee169.d20240514 depends on torch<3.0.0 and >=2.3.0
    instructlab[hpu] 0.14.1.dev183+gdeee169.d20240514 depends on torch==2.2.0a0; extra == "hpu"
```

Why do we require PyTorch 2.3.0 anyway? Is there a specific feature that is required?

pyproject.toml
```diff
@@ -54,7 +54,8 @@ where = ["src"]
 include = ["instructlab", "instructlab.chat", "instructlab.generator", "instructlab.train", "instructlab.train.lora_mlx", "instructlab.train.lora_mlx.models", "instructlab.llamacpp", "instructlab.mlx_explore", "instructlab.schema.v1", "instructlab.schema.v2"]
 
 [tool.setuptools.package-data]
-"instructlab.llamacpp" = ["quantize"]
+"instructlab.llamacpp.macos" = ["quantize"]
+"instructlab.llamacpp.linux" = ["quantize"]
```
tiran, line 58: The quantize binary is a ARM64 macOS program and does not work on Linux

```shell
$ file src/instructlab/llamacpp/quantize 
src/instructlab/llamacpp/quantize: Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|WEAK_DEFINES|BINDS_TO_WEAK|PIE>
```

cdoern: i have two, one built on linux one built on macos. They now live in `/llamacpp/linux/quantize` and `/llamacpp/macos/quantize`.

This needs a better solution but I believe is outside the scope of this PR

src/instructlab/train/pytorch_train.py
```diff
@@ -291,8 +317,53 @@ def model_generate(user, **kwargs):
...
+        load_best_model_at_end=load_best_model_at_end,
+        optim="adamw_torch",
+        evaluation_strategy=evaluation_strategy,
+        do_eval=do_eval,
+        bf16=bf16,
+        use_cpu=False,
+        save_strategy="epoch",
+        report_to="none",
+        torch_compile_backend="aot_eager",
+        torch_compile=True,
```
tiran, line 350: Last time I tried, torch compile did not work for me on AMD ROCm. You also cannot use it work HPUs. The Habana framework / SynapseAI uses a different approach to compile kernels for Intel Gaudi.

cdoern: interesting. This is rly only needed for mps. Will take this into account.

so many corner cases!!!

requirements.txt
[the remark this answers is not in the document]
cdoern: mps backend is broken below 2.3. I can test it out to give you the specific reason. However, I don't think it can compile with MPS below 2.3

pyproject.toml
cdoern: thanks for catching this. Will build a linux one now

src/instructlab/lab.py
```diff
@@ -85,8 +85,9 @@ def cli(ctx, config_file):
                 config_obj = config.read_config(config_file)
             except config.ConfigException as ex:
                 raise click.ClickException(str(ex))
-
-        ctx.obj = Lab(config_obj)
+        # pylint: disable=possibly-used-before-assignment
+        if config_obj is not None:
+            ctx.obj = Lab(config_obj)
```
tiran, line 90: This is fixed in #1162 .

src/instructlab/lab.py
```diff
@@ -450,6 +451,7 @@ def serve(ctx, model_path, gpu_layers, num_threads, max_ctx_size, model_family):
     help="model family to use when picking a generation template",
 )
 @click.pass_context
+# pylint: disable=possibly-used-before-assignment
```
tiran, line 454: Also fixed in 1162

src/instructlab/lab.py
```diff
@@ -639,6 +641,7 @@ def generate(
     help="model family to use when picking a chat template",
 )
 @click.pass_context
+# pylint: disable=possibly-used-before-assignment
```
tiran, line 644: Also fixed in 1162

src/instructlab/lab.py
```diff
@@ -891,40 +893,23 @@ def convert(self, value, param, ctx) -> "torch.device":
...
-    default=None,
-    show_default=True,
-)
-@click.option(
-    "--gguf-model-path",
-    help="Local directory where gguf model is stored.",
-    default=None,
-    show_default=True,
+    "--input-dir",
+    type=click.Path(),
```
tiran, line 902:
```suggestion
    type=click.Path(exists=True, file_okay=False, dir_okay=True),
```

src/instructlab/lab.py
```diff
@@ -961,29 +947,46 @@ def convert(self, value, param, ctx) -> "torch.device":
         "Use BitsAndBytes for 4-bit quantization "
         "(reduces GPU VRAM usage and may slow down training)"
     ),
+    callback=utils.macosrequired,
 )
 @click.option(
-    "--model-name",
-    default="instructlab/merlinite-7b-lab",
+    "--dtype",
+    type=click.Choice(["bf16", "fp16", "ftp32", "auto"]),
```
tiran, line 954: FTP is insecure :)

```suggestion
    type=click.Choice(["bf16", "fp16", "fp32", "auto"]),
```

requirements.txt
```diff
@@ -30,14 +30,13 @@ rouge-score>=0.1.2,<0.2.0
 sentencepiece>=0.2.0,<0.3.0
 tokenizers>=0.15.2,<0.16.0
 toml>=0.10.2,<0.11.0
-# Habana installer has 2.2.0a0+git8964477 with oneMKL
 torch>=2.2.0a0,<3.0.0 ; python_version == '3.10'
-torch>=2.2.1,<3.0.0 ; python_version != '3.10'
+torch>=2.3.0,<3.0.0 ; python_version != '3.10'
 tqdm>=4.66.2,<5.0.0
-transformers>=4.30.0,<=4.38.2
+transformers==4.38.1
```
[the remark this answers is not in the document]
cdoern: this is another mps specific. 4.38.1 is needed.

I can test again without it to show you the error.

cdoern:
@tiran if I ran `make quantize` on linux, that bin will be for linux, right? if so, I would really opt to get this in and then follow up with #1017

src/instructlab/lab.py
```diff
@@ -896,40 +895,23 @@ def convert(self, value, param, ctx) -> "torch.device":
...
+    help="Base repository where model is stored.",
     default="instructlab/merlinite-7b-lab",
     show_default=True,
 )
 @click.option("--iters", help="Number of iterations to train LoRA.", default=100)
-@click.option(
-    "--local",
-    is_flag=True,
-    help="Whether or not `model_dir` is remote from HuggingFace.",
-)
```
jaideepr97, line 932: @cdoern are we sure we want to remove this? maybe we need to have a larger conversation at some point about not forcing users to download a model all over again just for training and allow them to use a model they already have on their system

cdoern: i don't mind. I just found that it was not used as intended in most scenarios. I say we remove for now and add it back in properly

cdoern:
@tiran I don't think we should block on only having x86 linux support in this PR. as long as you have a follow up that fixes it, we should be fine. Blocking this PR on that is blocking some pretty urgent work for upstream users getting a demo.

It seems like #1017 has a bunch of sub tasks that may be a longer term thing. 

In the short term, I can build on a rpi if that counts as aarch? I can add that to the list of binaries lol

cdoern:
or @tiran I guess I could add some sort of check for "if not x86" then we can do the old convert method?

requirements.txt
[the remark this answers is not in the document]
jaideepr97: I don't like the idea of this but do we need to think about conditional imports based on environemnts?
or have separate requirement.txt files for separate HW dependencies

tiran:
@cdoern It's not just the lack of aarch64 support. As I mentioned before, it's more complicated than just CPU arch. Your Linux binary does not run on RHEL 9. I guess you have built it on Fedora 39 or 40.

```
./quantize 
./quantize: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by ./quantize)
./quantize: /lib64/libstdc++.so.6: version `CXXABI_1.3.15' not found (required by ./quantize)
./quantize: /lib64/libc.so.6: version `GLIBC_2.38' not found (required by ./quantize)
```

cdoern:
@tiran I approved your PR, it looks really clean. Thanks for doing that.

Will see if we can get it in before I get this working!

requirements.txt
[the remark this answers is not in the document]
nathan-weinberg: Agreed with @jaideepr97 - don't see any other path forward here

src/instructlab/lab.py
```diff
@@ -86,8 +86,8 @@ def cli(ctx, config_file):
                 config_obj = config.read_config(config_file)
             except config.ConfigException as ex:
                 raise click.ClickException(str(ex))
-
-        ctx.obj = Lab(config_obj)
+        if config_obj is not None:
```
nathan-weinberg, line 89: What if `config_obj` is `None`?

src/instructlab/lab.py
```diff
@@ -896,40 +895,23 @@ def convert(self, value, param, ctx) -> "torch.device":
...
-    show_default=True,
-)
-@click.option(
-    "--gguf-model-path",
-    help="Local directory where gguf model is stored.",
-    default=None,
-    show_default=True,
+    "--input-dir",
+    type=click.Path(exists=True, file_okay=False, dir_okay=True),
+    show_default=True,  # TODO: set to None and change help message
```
nathan-weinberg, line 905: Does this `TODO` need to be addressed?

src/instructlab/lab.py
nathan-weinberg: How was it being used?

src/instructlab/lab.py
```diff
@@ -968,39 +951,58 @@ def convert(self, value, param, ctx) -> "torch.device":
...
+    default="auto",
+    help=("Full half precision training. BF16 is more precise than FP16."),
+)
+@click.option(
+    "--backend",
+    type=click.Choice(["pytorch", "mlx"]),
     show_default=True,
-    help="model name to use in training",
+    default="pytorch",
+    help=("backend to use for training. Options are mlx or pytorch"),
```
nathan-weinberg, line 965:
```suggestion
    help=("Backend to use for training. Options are 'mlx' or 'pytorch'."),
```

src/instructlab/lab.py
```diff
@@ -968,39 +951,58 @@ def convert(self, value, param, ctx) -> "torch.device":
...
+)
+@click.option(
+    "--train-style",
+    type=click.Choice(["quick", "full"]),
+    show_default=True,
+    default="quick",
```
nathan-weinberg, line 971: Should we default to `full` instead of  `quick`? In case users aren't aware of this flag

src/instructlab/lab.py
```diff
@@ -968,39 +951,58 @@ def convert(self, value, param, ctx) -> "torch.device":
...
+    from instructlab_quantize import run_quantize  # pylint: disable=import-error
+
     if not input_dir:
         # By default, generate output-dir is used as train input-dir
         input_dir = ctx.obj.config.generate.output_dir
 
-    if four_bit_quant and device.type != "cuda":
-        ctx.fail("--4-bit-quant option requires --device=cuda")
+    # if four_bit_quant and device.type != "cuda":
+    #    ctx.fail("--4-bit-quant option requires --device=cuda")
```
nathan-weinberg, line 1005: Is this meant to be commented out?

src/instructlab/lab.py
```diff
@@ -1041,19 +1043,31 @@ def train(
...
-        from .train.linux_train import linux_train
+        from .train.pytorch_train import pytorch_train
 
-        linux_train(
-            ctx=ctx,
+        if device.type == "mps":
+            if (
+                "PYTORCH_MPS_HIGH_WATERMARK_RATIO" not in os.environ
+                or "PYTORCH_ENABLE_MPS_FALLBACK" not in os.environ
+            ):
```
nathan-weinberg, line 1055:
```suggestion
            if all("PYTORCH_MPS_HIGH_WATERMARK_RATIO", "PYTORCH_ENABLE_MPS_FALLBACK") not in os.environ:
```

src/instructlab/lab.py
```diff
@@ -1111,17 +1137,20 @@ def train(
...
-        shutil.copy(final_results_dir + "/ggml-model-f16.gguf", gguf_models_dir)
+        gguf_model = os.path.join(final_results_dir, "ggml-model-f16.gguf")
+        gguf_model_q = "./models/ggml-model-Q4_K_M.gguf"
+        if not skip_quantize:
+            run_quantize(gguf_model, gguf_model_q, "Q4_K_M")
+            # return gguf_model_q
+        # shutil.copy(final_results_dir + "/ggml-model-f16.gguf", gguf_models_dir)
         # cleanup original copy of model
-        os.remove(final_results_dir + "/ggml-model-f16.gguf")
+        # os.remove(final_results_dir + "/ggml-model-f16.gguf")
```
nathan-weinberg, line 1147: Commented out lines needed? If not let's remove them

src/instructlab/train/pytorch_train.py
```diff
@@ -193,40 +201,45 @@ def linux_train(
...
-        torch_dtype="auto",
+        torch_dtype=dtype,
         quantization_config=bnb_config,
         config=config,
         trust_remote_code=True,
         low_cpu_mem_usage=True,
     )
+    #  if model.dtype == torch.bfloat16:
+    # bfloat16 slows down training... alot.
+    #      model.dtype = torch.float16
```
nathan-weinberg, line 231: Same comment about commented code - needed? or can we can remove?

src/instructlab/train/pytorch_train.py
```diff
@@ -254,15 +267,28 @@ def model_generate(user, **kwargs):
...
 
-    print("LINUX_TRAIN.PY: GETTING THE ATTENTION LAYERS")
+    # TODO: investigate if we can use CPU for the sanity check and GPU for train.
+    # original_device = model.device
+    # if model.device != device:
+    #     print(f"here {device}")
+    #     model = model.to(device)
+    # print(f"PYTORCH_TRAIN.PY: Model device {model.device}")
+    # if model.device.type == "cuda":
+    #     print(torch.cuda.memory_summary())
```
nathan-weinberg, line 289: Commented code?

src/instructlab/train/pytorch_train.py
```diff
@@ -339,21 +387,29 @@ def model_generate(user, **kwargs):
...
+            torch_compile_backend="aot_eager",
+            torch_compile=torch_compile,
+            # half_precision_backend = "cpu_amp",
+            # use_ipex=True,
+            # TODO CPU test this possible optimization
+            #  gradient_accumulation_steps=4,
+            #  gradient_checkpointing=True,
+            # eval_accumulation_steps=1,
+            # per_device_eval_batch_size=1,
+            # bf16_full_eval=True,
```
nathan-weinberg, line 409: Commented out code?

requirements.txt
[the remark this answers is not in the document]
tiran: With the current limitations of Python's package metadata, there is only one way forward:
- leave the version pins in `requirements.txt` as relaxed as possible
- introduce an optional requirement `mps` with more strict requirements
- have users install MPS support with `instructlab[mps]` or `git+https://github.com/instructlab/instructlab.git@stable[mps]`

src/instructlab/lab.py
tiran: `config_obj` cannot be None. `get_default_config()` and `read_config` always return a config object or raise an exception. The `not os.path.isfile` case always raises an exception. `ctx.fail` is markes as `NoReturn`, but pylint does not understand the typing marker.

I recommend that you replace the block with this code to make it obvious that ctx.fail always raises an exception:

```python
         elif not os.path.isfile(config_file):
-            config_obj = None
             ctx.fail(
                 f"`{config_file}` does not exists, please run `ilab init` "
                 "or point to a valid configuration file using `--config=<path>`."
             )
+            assert False, "unreachable"
```

requirements.txt
```diff
@@ -31,11 +31,10 @@ rouge-score>=0.1.2,<0.2.0
...
```
tiran, line 37: The optional extra dependencies are a clean, maintainable solution. A quick hack is a platform-specific version pin. It abuses the fact that MPS is only available on macOS ARM64:

```suggestion
transformers==4.38.1; sys_platform == 'darwin' and platform_machine == 'arm64'
transformers>=4.37.0,<=4.38.2; sys_platform != 'darwin' and platform_machine != 'arm64'
```

I'll try to come up with a better solution next week.

src/instructlab/lab.py
```diff
@@ -1041,19 +1043,31 @@ def train(
...
         from .llamacpp.llamacpp_convert_to_gguf import convert_llama_to_gguf
-        from .train.linux_train import linux_train
+        from .train.pytorch_train import pytorch_train
 
-        linux_train(
-            ctx=ctx,
+        if device.type == "mps":
+            if (
+                "PYTORCH_MPS_HIGH_WATERMARK_RATIO" not in os.environ
+                or "PYTORCH_ENABLE_MPS_FALLBACK" not in os.environ
```
leseb, line 1054: Why are these needed as env variables? Why do users have to explicitly set them? Why can't we handle that on our own?

src/instructlab/lab.py
```diff
@@ -1041,19 +1043,31 @@ def train(
...
+            ):
+                click.secho(
+                    "To use MPS, PYTORCH_MPS_HIGH_WATERMARK_RATIO must be set to 0.0 and PYTORCH_ENABLE_MPS_FALLBACK must be set to 1",
```
leseb, line 1057: What if there are set but don't have the expected values?

src/instructlab/lab.py
```diff
@@ -1064,40 +1078,52 @@ def train(
...
-        print("Copied ", tokenizer_model[0], "to ", final_results_dir)
-        shutil.copy(tokenizer_config_json[0], final_results_dir)
-        print("Copied ", tokenizer_config_json[0], "to ", final_results_dir)
-        shutil.copy(config_json[0], final_results_dir)
-        print("Copied ", config_json[0], "to ", final_results_dir)
-        shutil.copy(generation_config_json[0], final_results_dir)
-        print("Copied ", generation_config_json[0], "to ", final_results_dir)
+        try:
+            shutil.copy(added_tokens_file, final_results_dir)
+            print("Copied ", added_tokens_file, "to ", final_results_dir)
```
leseb, line 1094: I know it was there before but if someone has the answer I'll be happy to hear it, why do we print and not user a logger here? Additionally, those look like DEBUG-level information to me.

src/instructlab/lab.py
```diff
@@ -1064,40 +1078,52 @@ def train(
...
+            print("Copied ", config_json[0], "to ", final_results_dir)
+            shutil.copy(generation_config_json[0], final_results_dir)
+            print("Copied ", generation_config_json[0], "to ", final_results_dir)
+        # pylint: disable=W0718
+        except Exception as exc:
+            print(f"Error copying train files to final dir: {exc}")
+            click.secho(
+                f"Could not copy train files: {exc}",
+                fg="red",
+            )
```
leseb, line 1113: Why both `print` and `click.secho`?

src/instructlab/lab.py
```diff
@@ -1064,40 +1078,52 @@ def train(
...
+            try:
+                shutil.copy(file, final_results_dir)
+                print("Copied ", file, "to ", final_results_dir)
+            # pylint: disable=W0718
+            except Exception as exc:
+                print(f"Error copying train files to final dir: {exc}")
+                click.secho(
+                    f"Could not copy train files: {exc}",
+                    fg="red",
+                )
```
leseb, line 1125: ditto

tests/test_lab_train.py
```diff
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# pylint: disable=R0801
```
leseb, line 3: Please use human-readable `duplicate-code` version instead of the code.

tests/test_lab_train.py
```diff
@@ -245,146 +266,64 @@ def test_skip_preprocessing(
...
-    @patch("instructlab.lab.utils.is_macos_with_m_chip", return_value=True)
-    @patch("instructlab.mlx_explore.utils.fetch_tokenizer_from_hub")
-    @patch("instructlab.mlx_explore.gguf_convert_to_mlx.load")
-    @patch("instructlab.train.lora_mlx.make_data.make_data")
-    @patch("instructlab.train.lora_mlx.convert.convert_between_mlx_and_pytorch")
-    @patch("instructlab.train.lora_mlx.lora.load_and_train")
-    def test_load_local(
+    @patch("instructlab.train.pytorch_train.pytorch_train")
+    @patch("instructlab.llamacpp.llamacpp_convert_to_gguf.convert_llama_to_gguf")
+    # pylint: disable=W0613
```
leseb, line 281: ditto

leseb (commented):
Very nice work! Small nits from me :)

requirements.txt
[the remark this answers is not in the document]
cdoern: @tiran how do I specify the dependencies in requirements.txt so that one can do`@stable[mps]`? Also, would I keep the current transformers version as well?

src/instructlab/lab.py
cdoern: well, when I fixed up the code so that --gguf--model-path was removed, I think it was only used when that flag was used? We never really pointed it to local safetensors I don't think.

src/instructlab/lab.py
cdoern: This isn't my TODO,it just looks like it is since I moved around the flags. I think this is out of scope for the PR, I'd hate to break input-dir.

src/instructlab/lab.py
cdoern: nope, thanks. I will add this back

src/instructlab/lab.py
cdoern: fixed this up a bit, thanks for calling it out

src/instructlab/train/pytorch_train.py
cdoern: removed

src/instructlab/lab.py
cdoern: If the user knows what they are doing enogh to set these, they probably have good reason for it :) 

I think the fallback will error out as it can only be 1 or 0.

The ratio will work as long as its set, but will be worse the further you get from 0.0

requirements.txt
cdoern: oh this is great. Thanks!

src/instructlab/lab.py
cdoern: Not sure, @alimaredia any specific reason?

src/instructlab/lab.py
cdoern: good point. I'll remove one

src/instructlab/lab.py
nathan-weinberg: All good!

src/instructlab/lab.py
nathan-weinberg: I would def be in favor of using the logger for this, but reasonable to do this as a followup and not as part of this PR IMHO

src/instructlab/lab.py
nathan-weinberg: @jaideepr97 you good for removing it or prefer keeping it in?

src/instructlab/lab.py
leseb: Thanks for the clarification!

cdoern:
holding because of https://github.com/instructlab/instructlab/actions/runs/9114567829/job/25058839214

nathan-weinberg:
@cdoern please add a log line so we know which model we are downloading for training - currently there is no indication as to whether it's merlinite or granite

src/instructlab/lab.py
```diff
@@ -972,33 +955,52 @@ def convert(self, value, param, ctx) -> "torch.device":
...
+)
+@click.option(
+    "--backend",
+    type=click.Choice(["pytorch", "mlx"]),
+    show_default=True,
+    default="pytorch",
+    help=("Backend to use for training. Options are 'mlx' or 'pytorch'."),
+)
+@click.option(
+    "--train-style",
```
chudegao, line 972: '--eval' is the most common used parameter for this purpose, set default value to True.

russellb (changes requested):
There is a design doc aiming to gather consensus on how to move training forward to support multiple modes of operation with varying levels of cost, complexity, and quality of output. This PR does some of the same. I would prefer we reach agreement on where we are headed before considering merging this.

https://github.com/instructlab/dev-docs/pull/52

tests/test_lab_train.py
```diff
@@ -248,152 +269,67 @@ def test_skip_preprocessing(
...
-            assert result.exit_code == 0
-            convert_llama_to_gguf_mock.assert_called_once()
+            # TODO: Why would we call convert without a model? this seems broken
+            # convert_llama_to_gguf_mock.assert_called_once()
+            # assert (
+            #     convert_llama_to_gguf_mock.call_args[1]["model"]
+            #     == "./training_results/final"
+            # )
+            # assert convert_llama_to_gguf_mock.call_args[1]["pad_vocab"] is True
+            # assert len(convert_llama_to_gguf_mock.call_args[1]) == 2
```
russellb, line 312: I'd like to see all the commented-out code removed throughout

russellb (commented):
My previous request for changes was about the UX related changes and that the alignment with the proposed training commands design was not clear. The new options have been hidden now, making it easier to continue to change them to get to the proposed design once it's agreed on.  I'm not blocking on that point anymore.

src/instructlab/train/pytorch_train.py
```diff
@@ -50,6 +52,7 @@
     htcore = None
     hpu = None
     hpu_backends = None
+# 'fork' incompatible with some hardware accelerator
```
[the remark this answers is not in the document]
cdoern: apologies, will remove this

src/instructlab/train/pytorch_train.py
```diff
@@ -291,8 +325,32 @@ def model_generate(user, **kwargs):
     tokenizer.padding_side = "right"
     output_dir = "./training_results"
     per_device_train_batch_size = 1
+
+    # torch compile only for MPS
```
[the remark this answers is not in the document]
cdoern: will add another comment as to why this is the case.

src/instructlab/train/pytorch_train.py
```diff
@@ -254,15 +275,28 @@ def model_generate(user, **kwargs):
         )
         return tokenizer.batch_decode([o[:-1] for o in outputs])[0]
 
-    assistant_old_lst = [
-        model_generate(d["user"]).split(response_template.strip())[-1].strip()
-        for d in tqdm(test_dataset)
-    ]
+    # TODO: INFERENCING IS BROKEN ON MACOS, CANNOT INF ON CPU AND TRAIN ON GPU
+    # MPS seems to be slow on this part. But, if the user is doing CPU or CUDA, this should still happen.
```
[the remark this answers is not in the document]
cdoern: IMO, yes. This is only the case when using MPS. If using cpu on mac this works.

This is just a "sanity check" that is currently not even acted on, just for the user's benefit.

src/instructlab/train/pytorch_train.py
```diff
@@ -291,8 +325,32 @@ def model_generate(user, **kwargs):
     tokenizer.padding_side = "right"
     output_dir = "./training_results"
     per_device_train_batch_size = 1
+
+    # torch compile only for MPS
+    if device.type == "mps":
+        torch_compile = True
+        torch_backend = "aot_eager"
```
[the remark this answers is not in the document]
cdoern: added another comment but the backend needs to be rebuilt in order to get mps compiled I believe. i can retry it without this to get the exact error if you want!

src/instructlab/train/pytorch_train.py
```diff
@@ -291,8 +325,32 @@ def model_generate(user, **kwargs):
...
+        evaluation_strategy = "epoch"
+        load_best_model_at_end = True
+    else:
+        raise ValueError(train_style)
+    # TODO: This might break MacOS
+    fp16 = dtype == "fp16"
+    bf16 = not fp16 and device.type != "mps"
     max_seq_length = 300
 
+    # TODO: This to @cdoern smells like it needs to be a different function, file, etc. Too different from the rest of train. Or at least, add flags for a bunch of this.
```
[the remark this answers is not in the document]
cdoern: why? imo leaving a TODO is better practice than forgetting about this entirely.

tests/test_lab_train.py
```diff
@@ -248,152 +269,67 @@ def test_skip_preprocessing(
...
-            assert len(linux_train_mock.call_args[1]) == 7
-            is_macos_with_m_chip_mock.assert_called_once()
-            assert not os.path.isfile(LINUX_GGUF_FILE)
+            assert train_mock.call_args[1]["test_file"] == "test_generated/test_1.jsonl"
+            assert train_mock.call_args[1]["num_epochs"] == 1
+            assert train_mock.call_args[1]["device"] is not None
+            assert not train_mock.call_args[1]["four_bit_quant"]
+            assert len(train_mock.call_args[1]) == 8
+            # TODO: why did this ever work??? linux train will create the gguf file unless we exit 0 earlier
+            # assert not os.path.isfile(LINUX_GGUF_FILE)
```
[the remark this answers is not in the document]
cdoern: This and the above one, I think, were tests that were written with the wrong functionality in mind. They passed by luck I think. This one seems to be saying " assert the gguf final file is not created on a successful train " which makes no sense?

I could totally be misinterpreting here. But it is possible the testing suite set this up differently and I somehow broke it?

src/instructlab/train/pytorch_train.py
```diff
@@ -193,40 +201,53 @@ def linux_train(
...
 
+    torchDtype = None
+    if dtype == "fp16":
+        torchDtype = torch.float16
+    if dtype == "bf16":
+        torchDtype = torch.bfloat16
+    if dtype == "fp32":
+        torchDtype = torch.float32
+    else:
+        torchDtype = "auto"
```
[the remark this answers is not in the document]
cdoern: This PR cleans up linux train as well, in that dtype was hardcoded depending on GPU type. Adding another gpu type would lead to another hardcoded situation for the dtype. Bubbling this up as an option seems to be the right move as we add support for more gpu backends.

src/instructlab/lab.py
```diff
@@ -1293,9 +1302,18 @@ def test(data_dir, model_dir, adapter_file):
...
     show_default=True,
 )
-@utils.macos_requirement(echo_func=click.secho, exit_exception=click.exceptions.Exit)
+@click.option(
+    "-su",
+    "--skip-unknown",
+    is_flag=True,
+    default=True,
+    help="Whether to skip unknown tensors while converting to GGUF.",
+)
```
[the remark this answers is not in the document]
cdoern: I can delete this. I think this is honestly unrelated to the train code.

The issue here was conversion after training with mlx would  fail due to unknown tensors but if you skip them the result is honestly a pretty bad model. We should revisit this thought probably

leseb:
@cdoern Is this work tracked in a design document? What's the status? Thanks!

JamesKunstle:
@cdoern do you still want to keep this open?

courtneypacheco:
Hi @cdoern! I'm looking to do some housekeeping and close out stale PRs, including drafts. 

This PR has been marked as stale and I haven't seen any recent activity, so I'm going to close it for now, but you're more than welcome to reopen it if you need to complete it.

PR #1160, opened 2024-05-14
import confluence documents

### LINKED ISSUE #1152: import Confluence documents
https://atlassian-python-api.readthedocs.io/ can be used to import Confluence documents.

makelinux:
yaml example:

```
document:
  confluence:
    host: "https://spaces.redhat.com"
    spaces:
      - name: telco5gocp
        patterns:
          - Debugging an OCP host
```
Closes #1152
Related changes: https://github.com/instructlab/schema/pull/17


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "src/instructlab/schema"]
 	path = src/instructlab/schema
-	url = https://github.com/instructlab/schema.git
+	url = git@github.com:RedHat-Israel/instructlab-schema.git
```
tiran, line 3: This looks like a mistake

requirements.txt
```diff
@@ -41,3 +41,5 @@ langchain-text-splitters
 # the below library should NOT be imported into any python files
 # it is for CLI usage ONLY
 yamllint>=1.35.1,<1.36.0
+markdownify
+atlassian-python-api
```
tiran, line 45: Please keep the requirements file sorted.

src/instructlab/utils.py
```diff
@@ -16,13 +16,18 @@
 import tempfile
 
 # Third Party
+from atlassian import Confluence
 from git import Repo, exc
 from langchain_text_splitters import RecursiveCharacterTextSplitter
+from markdownify import markdownify as md
```
tiran, line 22:
```suggestion
from markdownify import markdownify
```

src/instructlab/utils.py
```diff
@@ -239,6 +244,57 @@ def git_clone_checkout(
...
+    cfg = config.read_config()
+    C = Confluence(
+        url=cf["host"], username=cfg.confluence.user, token=cfg.confluence.token
+    )
+    docs = []
+    for s in cf["spaces"]:
+        for t in s["titles"]:
+            p = C.get_page_id(s["name"], t)
+            page_html = C.get_page_by_id(p, "space,body.view,version,container")
+            m = md(page_html["body"]["view"]["value"])
```
tiran, line 270:
```suggestion
            m = markdownify(page_html["body"]["view"]["value"])
```

tiran (commented):
Thank you!

The CI failures are going to be fixed by #1162.

src/instructlab/config.py
```diff
@@ -137,6 +147,7 @@ class Config(BaseModel):
 
     # additional fields with defaults
     general: _general = _general()
+    confluence: Optional[_confluence]
```
tiran, line 150: You also have to assign a default value to make the field an optional argument.
```suggestion
    confluence: Optional[_confluence] = None
```

src/instructlab/utils.py
```diff
@@ -239,6 +244,57 @@ def git_clone_checkout(
...
+    Retrieve the content of Confluence documents
+
+    Args:
+        source (dict): Source info containing references
+
+    Returns:
+         List[str]: List of document contents.
+    """
+    cf = source["confluence"]
+    cfg = config.read_config()
```
tiran, line 261: I don't think this is the correct approach. You have to pass the config from the `ctx` in `instructlab/lab.py` into the function, then access it with `ctx.obj.config`.

tiran (changes requested):

src/instructlab/utils.py
```diff
@@ -239,6 +244,58 @@ def git_clone_checkout(
...
+    """
+    Retrieve the content of Confluence documents
+
+    Args:
+        source (dict): Source info containing references
+
+    Returns:
+         List[str]: List of document contents.
+    """
+    cfg = ctx.obj.config.confluence
```
tiran, line 260: You have declared `confluence` as an optional value. That means `cfg` can be `None`. You need to handle the case here to avoid `AttributeError`.

russellb (changes requested):
I placed a `hold` since this should depend on backend support, as well.

As for these changes, I think it would be nice to consider making this an optional dependency, since it would not be part of the most common flow.

bjhargrave:
There is also the issue of knowledge stability. In the current git-repo-with-knowledge, a commit SHA is provided so the knowledge can be reliably collected at SDG/training time which matches the contribution time. This is important. The example here of using confluence do not seem to have a way to pin the knowledge to a specific revision like git does with a commit SHA. So I think this is a concern for repeatability and safety (since the knowledge could be poisoned after the contribution is made but before the SDG and training are done).

russellb:
Adding a new knowledge source seems like a good candidate for a design doc in the `dev-docs` repo to answer some of these key design questions. The addition spans many components, so a project-wide design doc there makes sense to me.

makelinux:
I've added Confluence versioning.

hickeyma (changes requested):
Thanks @makelinux for pushing this capability. As @russellb mentioned in https://github.com/instructlab/instructlab/pull/1160#issuecomment-2113899187, do you mins pushing a design doc for this capability?

makelinux:
Hi, @russellb , @hickeyma, here is the requested document [Confluence document source design notes](https://github.com/instructlab/dev-docs/pull/64)

makelinux:
> I placed a `hold` since this should depend on backend support, as well.
@russellb , Confluence pages usually are private, unlike Wikipedia.  What do you mean by **backend support**?

tests/test_utils.py
```diff
@@ -5,6 +5,7 @@
 import logging
 
 # Third Party
+# from atlassian import Confluence
```
leseb, line 8: Why is this commented?

makelinux: By mistake. Fixed.

src/instructlab/utils.py
```diff
@@ -239,6 +242,71 @@ def git_clone_checkout(
...
+    """
+    Retrieve the content of Confluence documents
+
+    Args:
+        source (dict): Source info containing references
+
+    Returns:
+         List[str]: List of document contents.
+    """
+    # pylint: disable=C0415
```
leseb, line 258: Use human-readable `import-outside-toplevel` flag instead of the code. I know parts of the code are using the code but I believe we trying move away from this. Thanks!

leseb (commented):
Small nits otherwise this feels ready to go.

russellb:
> > I placed a `hold` since this should depend on backend support, as well.
> > @russellb , Confluence pages usually are private, unlike Wikipedia.  What do you mean by **backend support**?

I mean if there's an expectation that we would accept any PRs into `instructlab/taxonomy` with pointers to confluence, then we have some backend code that does weekly model builds out of that repo that would need to understand it. If the intent is that this would only be for private taxonomy forks, then I'd agree only this PR is necessary.

I see a design doc is up now (thank you!) so let's also get that merged first -- https://github.com/instructlab/dev-docs/pull/64

PR #1161, opened 2024-05-14
Update llama_cpp_python version to 0.2.75

alimaredia:
# Changes

Resolves # https://github.com/instructlab/instructlab/security/dependabot/1

Testing needs to be done on M3 Macs to ensure https://github.com/abetlen/llama-cpp-python/issues/1286 doesn't still occur.

### DIFF
--- a/src/instructlab/chat/chat.py
+++ b/src/instructlab/chat/chat.py
@@ -376,6 +376,7 @@ class ConsoleChatBot:  # pylint: disable=too-many-instance-attributes
                         **create_params,
                     )
                 except openai.BadRequestError as e:
+                    logger.debug(f"BadRequestError: {e}")
                     if e.code == "context_length_exceeded":
                         if len(self.info["messages"]) > 1:
                             # Trim the oldest entry in our message history
@@ -390,6 +391,10 @@ class ConsoleChatBot:  # pylint: disable=too-many-instance-attributes
                         )
                         self.info["messages"].pop()
                         raise KeyboardInterrupt from e
+                except openai.InternalServerError as e:
+                    logger.debug(f"InternalServerError: {e}")
+                    self.info["messages"].clear()
+                    raise KeyboardInterrupt from e
                 assert (
                     next(response).choices[0].delta.role == "assistant"
                 ), 'first response should be {"role": "assistant"}'
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,9 +11,7 @@ GitPython>=3.1.42,<4.0.0
 httpx>=0.25.0,<1.0.0
 instructlab-quantize>=0.1.0
 jsonschema>=4.21.1,<5.0.0
-# pin version, lift restriction after testing >=0.2.58
-# see https://github.com/abetlen/llama-cpp-python/issues/1286
-llama_cpp_python[server]==0.2.55
+llama_cpp_python[server]==0.2.75
 mlx>=0.5.1,<0.6.0; sys_platform == 'darwin' and platform_machine == 'arm64'
 # HabanaLabs / Intel Gaudi env comes with Python 3.10 and slightly older
 # versions of some dependencies. Use '3.10' as an indicator.
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -114,41 +114,43 @@ test_bind_port(){
 
 test_ctx_size(){
     # A context size of 55 will allow a small message
-    ilab serve --max-ctx-size 55 &> "$TEST_CTX_SIZE_LAB_SERVE_LOG_FILE" &
+    ilab serve --max-ctx-size 25 &> "$TEST_CTX_SIZE_LAB_SERVE_LOG_FILE" &
     PID_SERVE=$!
 
     # Make sure the server has time to open the port
     # if "ilab chat" tests it before its open then it will run its own server without --max-ctx-size 1
     wait_for_server
 
-    # Should succeed
+    # SHOULD SUCCEED: ilab chat will trim the SYS_PROMPT then take the second message
     ${chat_shot} "Hello"
 
-    # now chat with the server and exceed the context size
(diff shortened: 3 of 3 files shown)


nathan-weinberg:
@alimaredia can we fix the lint error in this as well? Not sure how that got in

alimaredia:
@nathan-weinberg Wouldn't that impact the backport if we just want to backport requirements.txt. If so the linting should be addressed in a separate PR.

russellb:
Ran the e2e job here: https://github.com/instructlab/instructlab/actions/runs/9087518967

russellb:
Ran the e2e job here: https://github.com/instructlab/instructlab/actions/runs/9087518967

alimaredia:
@nathan-weinberg https://github.com/instructlab/instructlab/pull/1162 fixes the linting issues. Once that is merged I can rebase my PR and the linting test should pass.

nathan-weinberg:
@Mergifyio rebase

nathan-weinberg:
@Mergifyio rebase

russellb:
the functional test failure looks like a real problem

tiran:
Could you try with a larger max ctx size? 4096 might be too small.

src/instructlab/chat/chat.py
```diff
@@ -376,6 +376,7 @@ def start_prompt(self, logger, content=None, box=True):
                         **create_params,
                     )
                 except openai.BadRequestError as e:
+                    logger.debug("BadRequestError")
```
nathan-weinberg, line 379:
```suggestion
                    logger.debug(f"BadRequestError: {e}")
```

src/instructlab/chat/chat.py
```diff
@@ -391,6 +392,10 @@ def start_prompt(self, logger, content=None, box=True):
                             )
                             self.info["messages"].pop()
                             raise KeyboardInterrupt
+                except openai.InternalServerError as e:
+                    logger.debug("InternalServerError")
```
nathan-weinberg, line 396:
```suggestion
                    logger.debug(f"InternalServerError: {e}")
```

src/instructlab/chat/chat.py
```diff
@@ -391,6 +392,10 @@ def start_prompt(self, logger, content=None, box=True):
                             )
                             self.info["messages"].pop()
```
markstur, line 393: this is trimming the newest message and also recent llama-cpp versions don't seem to tolerate changing the message list size

alimaredia:
@russellb @markstur @tiran @nathan-weinberg 

What started as just trying to bump the version of llama_cpp_python (because of https://github.com/instructlab/instructlab/security/dependabot/1) turned into realizing that our trimming code in chat doesn't trim the way we'd expect it to anymore. 

When max-ctx-size goes from 25 -> 55 , certain prompts throw a `openai.InternalServerError` even though both max-ctx-sizes get rounded up to 64 in llama_cpp. Below is a breakdown of when the trimming code in chat is executed with different prompts and two small max-ctx-sizes. 

For now I think just handling `openapi.InternalServerError` is a good starting point and a follow up issue should get created. What do you all think?

```
M1 = "Hello"
M2 = "Hello, joe was born in 2000. How old is joe"
M3 = "Hello, I am a ci message that should not finish because I am too long for the context window, tell me about your day please?"
M4 = "Hello, I am a ci message that should not finish because I am too long for the context window, tell me about your day please? How many tokens could you take today. Could you tell me about the time you could only take 55 tokens"
M5 = "I was born in 2000. How old am I?"

25 max context window (ilab serve --max-ctx-size 25):
ilab chat -qq M1 
    - BadRequestError: Trims message
    - Responds back fine
ilab chat -qq M2
    - BadRequestError: Trims message
    - Responds back fine
ilab chat -qq M3
    - BadRequestError: Trims message
    - BadRequestError: Final message too large for context size
ilab chat -qq M4
    - BadRequestError: Trims message
    - BadRequestError: Final message too large for context size
ilab chat -qq M5
    - BadRequestError: Trims message
    - Responds back fine

55 max context window (ilab serve --max-ctx-size 55):
ilab chat -qq M1
    - Responds back fine
ilab chat -qq M2
    - BadRequestError: Trims message
    - Responds back fine
ilab chat -qq M3
    - BadRequestError: Trims message
    - Responds back fine
ilab chat -qq M4
    - BadRequestError: Trims message
    - InternalServerError
ilab chat -qq M5
    - InternalServerError
```

src/instructlab/chat/chat.py
tiran: - linters do not like this approach. Use `logger.debug("BadRequestError: %s", e)` instead.
- You can also pass `exc_info=True` to any logging call in order to print the exception name, value, and traceback.

alimaredia:
> this is trimming the newest message and also recent llama-cpp versions don't seem to tolerate changing the message list size

@markstur could you go into detail about this more or send me a reproducer for what you're talking about? That line should be trimming the last remaining message and then raising a `KeyboardInterupt`.

russellb (approved):
Thank you for your diligent efforts on this!

As discussed, there's more to dig into here to figure out why it's responding with an internal server error. Something unexpected is still happening on the server side. Please file an issue to track down the source of the internal server error at some point.

src/instructlab/chat/chat.py
markstur: Ignore the confusion about "newest" vs "latest" -- I read a comment out-of-place.

The main thing here though is that we still get InternalError happening when we try to shorten messages this way and I think we're going to make that a separate issue.

markstur:
I think this works reasonably well when the context is 512 (maybe even 256).   It's the real short context tests where the internal error is such a problem.  I suspect llama-cpp-python needs to fix something with the batch size vs context size, but haven't been able to figure out why I only reproduce the problem with small contexts (so far).

We probably should merge this soon as the lesser of evils.  Wondering about a few things though:

1.  Can/should we just enforce a reasonable minimum max-ctx-size?
2. Shouldn't we always keep the first message (the prompt) when trimming?   Need reasonable ctx size for that.

nathan-weinberg (approved):

russellb:
I’m going to let this merge. If anyone has follow ups let’s file an issue to make sure it doesn’t get lost.

nathan-weinberg:
@Mergifyio backport release-v0.15

PR #1167, opened 2024-05-15
add doxygen configuration

makelinux:
doxygen for auto generated documentation.

Preview:

image

image

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -23,6 +23,9 @@ dataset
 dev
 ditaa
 DKMS
+Doxyfile
+doxygen
+Doxygen
 dr
 Dropdown
 env
@@ -40,6 +43,7 @@ habana
 Habana
 hipBLAS
 HPU
+html
 ilab
 impactful
 Inferencing
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -1,0 +1,20 @@
+PROJECT_NAME           = "InstructLab CLI"
+EXTRACT_ALL            = yes
+QUIET                  = yes
+INPUT                  = ../src
+RECURSIVE              = yes
+EXCLUDE_PATTERNS       = __init__.py \
+                         __main__.py
+SOURCE_BROWSER         = yes
+REFERENCED_BY_RELATION = yes
+REFERENCES_RELATION    = yes
+HTML_OUTPUT            = doxygen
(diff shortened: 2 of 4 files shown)


russellb (changes requested):
How about a Makefile that makes it easy to run the generation?

Also, how did you decide on Doxygen vs alternatives? My impression had been Sphinx is kind of the standard for Python.

makelinux:
To generate need just run `doxygen`. Let's take a look and compare efforts and results.

makelinux:
Doxygen generates documentation for API and internals, including call graphs.

Sphinx AFAIK generates only API documentation.

Doxyfile
```diff
@@ -0,0 +1,19 @@
+PROJECT_NAME           = "Instruct Lab CLI"
```
nathan-weinberg, line 1:
```suggestion
PROJECT_NAME           = "InstructLab CLI"
```

nathan-weinberg (changes requested):
Where do the generated docs get saved? Is there any docs anywhere as to how to install the utility?

makelinux:
> Where do the generated docs get saved? Is there any docs anywhere as to how to install the utility?

Generated documentation is saved by default to directory html.

About Doxygen: https://www.doxygen.nl/download.html

nathan-weinberg:
If we decide to include this, we should include docs about installing/using it - otherwise this is just an unexplained config file

makelinux:
Yes. it can live in docs. Thank you.

nathan-weinberg:
Putting a hold on this until the docs around it I requested are added

hickeyma (commented):
Preference for sphinx as well

leseb:
What's the latest on this? Are we aiming for Sphinx only? Thanks!

makelinux:
Doxygen has several unique features. Just two of them: graphs, annotated directory tree.

tiran:
Who is the target user for the documentation and how does any of these features benefit the target user?

makelinux:
Our target audience includes not only users of the ilab CLI but also developers contributing to the project. Developers, in particular, benefit significantly from Doxygen features, as it facilitates comfortable and efficient project navigation.

hickeyma (commented):
@makelinux @tiran This PR and #1184 seem to be duplicates in functionality to add infrastructure to build documentation from code. Do you mind working out which implementation we are going to progress with?

PR #1168, opened 2024-05-15
ci: fix typo in actionlint workflow

nathan-weinberg:
Was digging into why https://github.com/instructlab/instructlab/pull/1166 wasn't automerging and it seems this typo is to blame


bjhargrave (approved):
LGTM

russellb (approved):

PR #1169, opened 2024-05-15
mergify: Automate a label for CI failures

russellb:
When I pull up the PR queue to do some reviews, having data to filter on helps me prioritize which PRs to look at first. Sometimes I want to filter out any PRs where CI isn't passing to help narrow down the list to the ones most likely to be ready for approval.

This PR makes mergify automate adding and removing the `ci-failures` label based on if there are any checks currently failing.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -153,3 +153,19 @@ pull_request_rules:
         - release-v0.14
   conditions:
     - label=backport-release-v0.14
+
+- name: Apply ci-failure label if any CI checks have failed
+  conditions:
+      - "#check-failure>0"
+  actions:
+    label:
+      add:
+        - ci-failure
+
+- name: Remove ci-failure label if no failures are present
+  conditions:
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):

leseb (commented):
GitHub already shows if the CI is failing in the PR queue with the ❌ emoji. However, I don't think we can filter out PR based on CI failing so using a label through this Mergify automation makes sense to me. Is that the filtering you are looking for that led to this PR? Thanks!

russellb:
> GitHub already shows if the CI is failing in the PR queue with the ❌ emoji. However, I don't think we can filter out PR based on CI failing so using a label through this Mergify automation makes sense to me. Is that the filtering you are looking for that led to this PR? Thanks!

Exactly. I know it's visible, but I didn't see how to filter on it when looking at a list.

alimaredia (approved):

PR #1174, opened 2024-05-15
Extract model_family from model_path automatically by default.

### LINKED ISSUE #1060: Users incorrectly think that by specifying the `--model-family` flag then the correct model is selected for `ilab generate`
Users think by specifying the `--model-family` flag for the `ilab generate` command that it will use the correct model. They are unaware that you also need to specify the model using the `--model` flag. Otherwise, the default model is selected for them and this is unexpected behavior.

The use of the `--model-family` flag can be confusing. Suggest removing the flag in favor of adding the logic to select the correct template in the code for the model of a particular family. A user should not need to know what family a model is from.

From @laurachiticariu:
"
`ilab generate --model-family mixtral`  still seem to use `merlinite` to generate data. Are there instructions on how to make it use `mixtral` instead? (e.g., perhaps I need to serve `mixtral` using some combo of `ilab download` and `ilab serve`
"

This was also raised by @anik120 in https://github.com/instructlab/instructlab/pull/1002#discussion_r1583318874.

makelinux:
Extract model_family from model_path automatically by default.

  --model-family forces model_family.

Resolves #1060

### DIFF
--- a/src/instructlab/config.py
+++ b/src/instructlab/config.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+from os import path
+from re import match
 from typing import Optional
 
 # Third Party
@@ -190,3 +192,7 @@ def write_config(cfg, config_file=DEFAULT_CONFIG):
 def get_api_base(host_port):
     """Returns server API URL, based on the provided host_port"""
     return f"http://{host_port}/v1"
+
+
+def get_model_family(forced, model_path):
+    return forced if forced else match(r"^\w*", path.basename(model_path)).group(0)
--- a/src/instructlab/generator/generate_data.py
+++ b/src/instructlab/generator/generate_data.py
@@ -20,7 +20,7 @@ import click
 import tqdm
 
 # Local
-from ..config import DEFAULT_MULTIPROCESSING_START_METHOD
+from ..config import DEFAULT_MULTIPROCESSING_START_METHOD, get_model_family
 from ..utils import chunk_document, read_taxonomy
 from . import utils
 from .utils import GenerateException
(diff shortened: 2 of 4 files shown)


cdoern:
I am all for implicitly figuring out the model family, but removing the flag isn't necessary.

I vote for 
1. implicitly deciding if the flag isn't specified and 
2. overriding if the user gives it.

Pulling the wool over the user's eyes IMO never goes well, especially with abstracted things like this.

lhawthorn:
I agree with @cdoern's assessment - it is always better to give more information rather than less.

makelinux:
> I agree with @cdoern's assessment - it is always better to give more information rather than less.

I've already implemented it.

nathan-weinberg:
@cdoern is the current implementation good with you?

src/instructlab/config.py
```diff
@@ -190,3 +192,7 @@ def write_config(cfg, config_file=DEFAULT_CONFIG):
 def get_api_base(host_port):
     """Returns server API URL, based on the provided host_port"""
     return f"http://{host_port}/v1"
+
+
+def get_model_family(forced, model_path):
+    return forced if forced else match(r"^\w*", path.basename(model_path)).group(0)
```
russellb, line 198: It looks like `match()` can return `None`, but that isn't handled here

russellb (changes requested):
one minor request for change

src/instructlab/config.py
makelinux: How can it return `None`?

nathan-weinberg: https://docs.python.org/3/library/re.html#re.match

makelinux: `^\w*` will return at least an empty string. In will not return `None`.

russellb: Ah, right. I didn't read the regex itself closely enough. Thanks.

nathan-weinberg (approved):

PR #1175, opened 2024-05-15
docs: make document structure more consistent

nathan-weinberg:
I noticed some inconsistencies in the doc structure of this repo - this PR makes some proposed changes:
1. rename `governance.md` to `GOVERNANCE.md` to match all other top-level Markdown files
2. move the `containerization.md` file to `containers/README.md` to match scheme in other subdirectories such as `notebooks/`
3. move `CONTRIBUTOR_ROLES.md` to `CONTRIBUTING/` directory with other contribution documents


russellb (changes requested):
I think governance and contributor roles can be removed. If you want, they could be added as links inside the CONTRIBUTING.md doc. I don't think every doc that exists in community needs its own corresponding pointer doc here.

Btw the link in the governance doc is broken anyway.

makelinux:
Once upon a time, someone named readme file in uppercase so that it could be quickly found. Since then, the small readme file gets lost among many documents with names in uppercase.

russellb (approved):

kelbrown20 (approved):

PR #1178, opened 2024-05-16
ci: add 'one-approval' label on PR that has one approval

leseb:
If a PR has been approved already by one of the maintainers, the PR will be labeled with 'one-approval'. This is useful for maintainers to help them determine which PR is getting close to completion and thus prioritize the review.
With the help of filters, one can look for PRs that have one approval already.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -169,3 +169,19 @@ pull_request_rules:
     label:
       remove:
         - ci-failure
+
+- name: Apply 'one-approval' label if one of the maintainer approved the PR
+  conditions:
+      - "#approved-reviews-by=1"
+  actions:
+    label:
+      add:
+        - one-approval
+
+- name: Remove 'one-approval' label if the approval was reset
+  conditions:
+      - "#approved-reviews-by!=1"
+  actions:
+    label:
+      remove:
+        - one-approval


leseb:
@russellb PTAL

leseb:
Side note: the "Approved review" filter on PRs will only filter PRs that have two approvals (GitHub config?) so this will still be useful.

.github/mergify.yml
```diff
@@ -169,3 +169,19 @@ pull_request_rules:
...
+  conditions:
+      - "#approved-reviews-by=1"
+  actions:
+    label:
+      add:
+        - one-approval
+
+- name: Remove 'one-approval' label if the approval was reset
+  conditions:
+      - "#approved-reviews-by=0"
```
russellb, line 183:
```suggestion
      - "#approved-reviews-by!=1"
```

russellb (changes requested):
I think the current config would lave the label on even after there's more than just one approval on it

.github/mergify.yml
leseb: It's a good point, thanks!

nathan-weinberg (approved):
I don't want to get too label-happy here but this does seem like it could be a cool addition - willing to give it a shot!

russellb (approved):

PR #1179, opened 2024-05-16
feat: redirect llama_cpp server stderr/stdout to a logger

leseb:
Our logger now includes the llamap_cpp logs instead of printing to stderr and stdout, making the logs hard to read and parse. It looks like llma_cpp print debug messages to stderr instead of stdout so the startup logs look a little bit confusing when setting our log level to DEBUG.


russellb (approved):
nice improvement! thank you

nathan-weinberg (approved):

PR #1181, opened 2024-05-16
POC: Implement model signing/verification with Sigstore

### LINKED ISSUE #821: Add command to sign model using Sigstore
Would like to see about adding a command to sign a model using Sigstore. Does this sound like a feature/workflow we would want integrated into this CLI tool?

mbestavros:
# Changes

Resolves #821 

This adds a new `ilab` command that allows users to sign their completed `GGUF` models with Sigstore and later verify them.

Signing uses Sigstore's interactive OIDC flow, so a user will be prompted to log into the identity provider of their choice (Github, Google, or Microsoft):

```shell
> ilab sign --model-path models/merlinite-7b-lab-Q4_K_M.gguf
```

The command creates a `.sigstore.json` bundle file within the `models` directory that corresponds to the signed model. (In the above example, it would be `models/merlinite-7b-lab-Q4_K_M.gguf.sigstore.json`.)

Verification expects a bundle and identity provided via flags. This example assumes Google as the identity provider:

```shell
> ilab verify --model-path merlinite-7b-lab-Q4_K_M.gguf --identity user@gmail.com --issuer https://accounts.google.com
```

Marking as a proof-of-concept, since @cdoern mentioned that there's a big command hierarchy rework underway at https://github.com/instructlab/instructlab/pull/1157.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ click-didyoumean>=0.3.0
 datasets>=2.18.0
 gguf>=0.6.0
 GitPython>=3.1.42
+sigstore>=3.0.0
 # Linux: 4-bit quantization with BitsAndBytes is not ready to use, yet.
 # see https://github.com/instructlab/instructlab/issues/579
 # bitsandbytes; sys_platform=='linux' and platform_machine=='x86_64'
--- a/src/instructlab/model/verify.py
+++ b/src/instructlab/model/verify.py
@@ -1,0 +1,65 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+from pathlib import Path
+
+# Third Party
+from sigstore.errors import VerificationError
+import click
+
+# First Party
+from instructlab import configuration as config
+from instructlab import signing, utils
+
+
+@click.command()
+@click.option(
+    "--model-path",
+    type=click.Path(),
+    default=config.DEFAULT_MODEL_PATH,
+    show_default=True,
+    help="Path to the model to be signed.",
+)
+@click.option(
+    "--bundle-path",
+    type=click.Path(),
+    default=None,
+    show_default=True,
+    help="Path to save the Sigstore bundle file after signing.",
+)
+@click.option(
+    "--identity",
+    help="Certificate identity to verify against."
+    "Typically an email provided by the OIDC identity provider."
+    "Example: a model signed using a Github account under hello@instructlab.ai would use hello@instructlab.ai.",
+)
+@click.option(
+    "--issuer",
+    help="Certificate identity's issuing authority.",
+    default="https://github.com/login/oauth",
+    show_default=True,
+)
+@click.option(
+    "--staging",
+    is_flag=True,
+    help="Use Sigstore's staging environment.",
+)
+@click.pass_context
+@utils.display_params
+def verify(ctx, model_path, bundle_path, identity, issuer, staging):
+    """Signs a model with Sigstore"""
+
+    if bundle_path is None:
+        bundle_path = f"{model_path}.sigstore.json"
+
+    try:
+        signing.verify_model(
+            model_path=Path(model_path),
+            bundle_path=Path(bundle_path),
+            identity=identity,
+            issuer=issuer,
+            staging=staging,
+        )
+        print(f"✅ {bundle_path} passed verification")
+    except VerificationError as e:
+        print(f"❌ {bundle_path} failed verification: {str(e)}")
--- a/src/instructlab/model/model.py
+++ b/src/instructlab/model/model.py
@@ -9,8 +9,10 @@ from .chat import chat
 from .convert import convert
 from .download import download
 from .serve import serve
+from .sign import sign
 from .test import test
 from .train import train
+from .verify import verify
 
 
 @click.group(cls=DYMGroup)
@@ -31,3 +33,5 @@ model.add_command(convert)
 model.add_command(chat)
 model.add_command(test)
 model.add_command(download)
+model.add_command(sign)
+model.add_command(verify)
--- a/src/instructlab/model/sign.py
+++ b/src/instructlab/model/sign.py
@@ -1,0 +1,44 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+from pathlib import Path
+
+# Third Party
+import click
+
+# First Party
+from instructlab import configuration as config
+from instructlab import signing, utils
+
+
+@click.command()
+@click.option(
+    "--model-path",
+    type=click.Path(),
+    default=config.DEFAULT_MODEL_PATH,
+    show_default=True,
+    help="Path to the model to be signed.",
+)
+@click.option(
+    "--bundle-path",
+    type=click.Path(),
+    default=None,
+    show_default=True,
+    help="Path to save the Sigstore bundle file after signing.",
+)
+@click.option(
+    "--staging",
+    is_flag=True,
+    help="Use Sigstore's staging environment.",
+)
+@click.pass_context
+@utils.display_params
+def sign(ctx, model_path, bundle_path, staging):
+    """Signs a model with Sigstore"""
+
+    if bundle_path is None:
+        bundle_path = f"{model_path}.sigstore.json"
+
+    signing.sign_model(
+        model_path=Path(model_path), bundle_path=Path(bundle_path), staging=staging
+    )
--- a/src/instructlab/signing.py
+++ b/src/instructlab/signing.py
@@ -1,0 +1,80 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+from pathlib import Path
+import hashlib
+
+# Third Party
+from sigstore.hashes import Hashed
+from sigstore.models import Bundle
+from sigstore.oidc import Issuer
+from sigstore.sign import SigningContext
+from sigstore.verify import Verifier
+from sigstore.verify.policy import Identity
+from sigstore_protobuf_specs.dev.sigstore.common.v1 import HashAlgorithm
+
+
+def sign_model(model_path: Path, bundle_path: Path, staging: bool = False) -> None:
+    """Signs a model with Sigstore"""
+
+    # Sigstore setup
+    if staging:
+        issuer = Issuer.staging()
+        signing_ctx = SigningContext.staging()
+    else:
+        issuer = Issuer.production()
+        signing_ctx = SigningContext.production()
+    identity = issuer.identity_token()
+
+    # Hash the model file
+    model_hash = incremental_hash(model_path=model_path)
+
+    # Sign the model's hash and write Sigstore bundle to disk
+    with signing_ctx.signer(identity, cache=True) as signer:
+        result = signer.sign_artifact(model_hash)
+    with bundle_path.open("w", encoding="utf8") as bundle_file:
+        bundle_file.write(result.to_json())
+
+
+def verify_model(
+    model_path: Path,
+    bundle_path: Path,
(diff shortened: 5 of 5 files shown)


src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,54 @@
...
+from pathlib import Path
+
+# Third-party
+from sigstore.sign import SigningContext
+from sigstore.oidc import Issuer
+from sigstore.models import Bundle
+from sigstore.verify import Verifier
+from sigstore.verify.policy import Identity
+
+def sign(model_path, bundle_path, staging=False):
```
tiran, line 13: I recommend to use pathlib for input and output

```suggestion
def sign_model(model_path: pathlib.Path, bundle_path: pathlib.Path, staging: bool=False) -> None:
```

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,54 @@
...
+    """Signs a model with Sigstore"""
+    # pylint: disable=C0415
+
+    issuer = Issuer.production()
+    if staging:
+        issuer = Issuer.staging()
```
tiran, line 19:
```suggestion
    if staging:
        issuer = Issuer.staging()
    else:
        issuer = Issuer.production()
```

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,54 @@
...
+    identity = issuer.identity_token()
+
+    # The artifact to sign
+    artifact = Path(model_path).read_bytes()
```
tiran, line 23: This approach consumes a lot of memory. Models are very large, several GB to tens of GB. I suggest that you use incremental hashing and `sigstore.hashes.Hashed` here.

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,54 @@
...
+
+    signing_ctx = SigningContext.production()
```
tiran, line 25: I think you need a special case for staging here.

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,54 @@
...
+# Third-party
+from sigstore.sign import SigningContext
+from sigstore.oidc import Issuer
+from sigstore.models import Bundle
+from sigstore.verify import Verifier
+from sigstore.verify.policy import Identity
+
+def sign(model_path, bundle_path, staging=False):
+    """Signs a model with Sigstore"""
+    # pylint: disable=C0415
```
tiran, line 15: Please use long, human-readable names for pylint disable comments:
```suggestion
    # pylint: disable=import-outside-toplevel
```

src/instructlab/provenance/sigstore.py
mbestavros: Done, thanks for the recommendation.

src/instructlab/provenance/sigstore.py
mbestavros: Good call, added.

src/instructlab/provenance/sigstore.py
mbestavros: Good idea, I added an implementation I based off `sigstore-python`'s implementation of incremental hashing. Let me know how it looks.

src/instructlab/provenance/sigstore.py
mbestavros: Aha, I knew I was missing something. Thanks for the catch!

mbestavros:
Latest push updates a bunch of stuff to address @tiran's review (notably switching to hashing instead of complete file reads), structures the `sigstore` library a little better, and improves handling when verification fails. Also fixes a bug when using the `--staging` flag.

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,83 @@
...
+def incremental_hash(model_path: Path) -> Hashed:
+    """Incrementally hash a file and create a sigstore.hashes.Hashed object"""
+
+    sha = hashlib.sha256()
+    view = memoryview(bytearray(128 * 1024))
+    with open(model_path, 'rb') as model_file:
+        file_buffer = model_file.readinto(view)
+        while file_buffer:
+            sha.update(view[:file_buffer])
+            file_buffer = model_file.readinto(view)
```
tiran, line 82: `readinto` does not return a file buffer. It returns the amount of read data in bytes. You also don't need the buffer / memoryview trick. My implementation `hashlib.file_digest` in Python 3.11+ uses the trick as a micro-optimization.

```suggestion
    h = hashlib.sha256()
    with open(model_path, "rb") as f:
        while True:
            buf = f.read(128 * 1024)
            if not buf:
                break
            h.update(buf)
        file_buffer = model_file.readinto(view)
        while file_buffer:
            sha.update(view[:file_buffer])
            file_buffer = model_file.readinto(view)
```

mbestavros: Thanks for the suggestion - added!

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,89 @@
...
+            model_hash,
+            bundle,
+            Identity(
+                identity=identity,
+                issuer=issuer,
+            ),
+        )
+        return f"✅ {bundle_path} passed verification"
+    except VerificationError as e:
+        return f"❌ {bundle_path} failed verification: {str(e)}"
```
tiran, line 77: I would prefer to raise the exception here and let the CLI entry point map the exception to an error message. It's separates internal logic from UI.

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def incremental_hash(model_path: Path) -> Hashed:
+    """Incrementally hash a file and create a sigstore.hashes.Hashed object"""
+    h = hashlib.sha256()
+    with open(model_path, "rb") as f:
```
tiran, line 83:
```suggestion
    with model_path.open("rb") as f:
```

src/instructlab/lab.py
```diff
@@ -1370,6 +1370,90 @@ def convert(ctx, model_dir, adapter_file, skip_de_quantize, skip_quantize, model
...
+@click.option(
+    "--bundle-path",
+    type=click.Path(),
+    default=None,
+    show_default=True,
+    help="Path to save the Sigstore bundle file after signing.",
+)
+@click.option(
+    "--identity",
+    help="Certificate identity to verify against.",
```
tiran, line 1423: Can you include an example here, please? How would a user verify that the model is signed by a specific entity, e.g. a GitHub org or user account?

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,89 @@
...
+        signing_ctx = SigningContext.production()
+    identity = issuer.identity_token()
+
+    # Hash the model file
+    model_hash = incremental_hash(model_path=model_path)
+
+    # Sign the model's hash and write Sigstore bundle to disk
+    with signing_ctx.signer(identity, cache=True) as signer:
+        result = signer.sign_artifact(model_hash)
+    with open(bundle_path, "w", encoding="utf8") as bundle_file:
```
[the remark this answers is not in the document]
mbestavros: Added!

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,89 @@
...
+    bundle_path: Path,
+    identity: str,
+    issuer: str,
+    staging: bool = False,
+) -> str:
+    """Verifies a model and its bundle with Sigstore"""
+
+    # The input to verify
+    # input_ = model_path.read_bytes()
+
```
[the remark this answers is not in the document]
mbestavros: Whoops - removed!

src/instructlab/provenance/sigstore.py
```diff
@@ -0,0 +1,89 @@
...
+    if staging:
+        verifier = Verifier.staging()
+    else:
+        verifier = Verifier.production()
+
+    # Hash the model file
+    model_hash = incremental_hash(model_path=model_path)
+
+    # Read the Sigstore bundle to verify with from disk
+    bundle = Bundle.from_json(Path(bundle_path).read_bytes())
```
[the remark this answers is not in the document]
mbestavros: `click.Path()` (which appears to be the convention for path options elsewhere in the CLI) doesn't give us `pathlib.Path` objects consistently (often times it's just a string), so I'm just casting all the input paths to `pathlib.Path` in the command stubs in my latest push for consistency. @tiran I'm open to suggestions if you have a better solution.

src/instructlab/provenance/sigstore.py
mbestavros: Done!

src/instructlab/lab.py
mbestavros: I added some inline documentation, but if it's too wordy I can rework.

I also just need to write proper documentation for this outside of code.

russellb:
Marked as a Draft since the title and summary say it's a POC. Please update once you feel it's ready for review to be merged. Thanks!

requirements.txt
```diff
@@ -5,6 +5,7 @@ click-didyoumean>=0.3.0
 datasets>=2.18.0
 gguf>=0.6.0
 GitPython>=3.1.42
+sigstore>=3.0.0
```
stefanberger, line 8: Hi! Are you aware of this library here: https://github.com/sigstore/model-transparency/

It may be a better library to use for signing models that are split into multiple tensor files and have possibly other files laying around that may be worth signing or verying. This library would sign all the files and files found in directories under a `model_path`: https://github.com/sigstore/model-transparency/blob/13459e976001d892f254f8e9927a092461d47ed8/src/sign.py#L191-L197

My feeling is that you're getting into an area of what will be the industry-accepted way of signing models and I think the model signing library may be better.

src/instructlab/model/verify.py
```diff
@@ -0,0 +1,65 @@
...
+    help="Certificate identity to verify against."
+    "Typically an email provided by the OIDC identity provider."
+    "Example: a model signed using a Github account under hello@instructlab.ai would use hello@instructlab.ai.",
+)
+@click.option(
+    "--issuer",
+    help="Certificate identity's issuing authority.",
+    default="https://github.com/login/oauth",
+    show_default=True,
+)
```
stefanberger, line 41: Another option could be to pick up a .yaml file here that contains the identity and issuer and is provided by the signer ( ideally this file would also be signed) and its contents are displayed when the signature verification is done so that the user can see whether the signature expectations are met correctly.
```
oidc:
  -identity: user@gmail.com
  -issuer: https://accounts.google.com
```

requirements.txt
mbestavros: @stefanberger Thanks for the review (and hello again, long time no see)!

> Are you aware of this library here:

Yes! I've been involved in the Sigstore model-signing SIG since before I wrote this PR - I left it in draft format with the intention of reworking it using the model-signing library once it hit stable. That has not yet happened, but when it does, I might revisit.

> My feeling is that you're getting into an area of what will be the industry-accepted way of signing models and I think the model signing library may be better.

Totally agreed! I wrote this PR mostly as a very quick and dirty proof of concept of model signing - I wasn't intending to ship it in this state.

stefanberger: > Yes! I've been involved in the Sigstore model-signing SIG since before I wrote this PR - I left it in draft format with the intention of reworking it using the model-signing library once it hit stable.

Do you know what it will take for it to 'hit stable'?

booxter:
I don't think this work here is moving forward. Closing but feel free to reopen.

PR #1184, opened 2024-05-17
Add Sphinx documentation, MyST, and RTD

tiran:
# Changes

Add infrastructure to build documentation and API docs with Sphinx, MyST (for markdown parsing), napoelon autodoc (for automatic API documentation), and furo theme. An untested RDT config is included, too.

`tox -e docs` and `make docs` now builds the documentation

I had to fix several minor issues in our markdown documents. For example Pygments does not recognize `ShellSession` and Sphinx cannot reference files outside the doc tree.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -118,7 +118,7 @@ jobs:
       - name: Test Workflow Complete
         run: echo "Test Workflow Complete"
 
-  man:
+  docs:
     runs-on: ubuntu-latest
     steps:
       - name: "Harden Runner"
@@ -146,8 +146,7 @@ jobs:
           pip cache remove llama_cpp_python
 
       - name: Run tox docs target (expect failure since tox is not present)
-        run: |
-          make man
+        run: make docs
         continue-on-error: true
 
       - name: Run tox docs target
@@ -169,8 +168,7 @@ jobs:
           python -m pip install tox
 
       - name: Run tox docs target
-        run: |
-          make man
+        run: make docs
 
       - name: Check that man pages were generated
         uses: andstor/file-existence-action@076e0072799f4942c8bc574a82233e1e4d13e9d6 # v3.0.0
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,33 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+  jobs:
+    pre_install:
+      # install torch without CUDA bindings to speed up builds
+      - pip install --extra-index-url https://download.pytorch.org/whl/cpu -r requirements.txt
+
+# Build documentation in the "docs/" directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+  # You can configure Sphinx to use a different builder, for instance use the dirhtml builder for simpler URLs
+  # builder: "dirhtml"
+  # Fail on all warnings to avoid broken references
+  # fail_on_warning: true
+
+# See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
+python:
+  install:
+    - requirements: docs/requirements.txt
+    - method: pip
+      path: .
+
+submodules:
+  include: all
--- a/Makefile
+++ b/Makefile
@@ -164,7 +164,9 @@ man: docs
 
 .PHONY: sphinx-apidoc
 sphinx-apidoc: check-tox
-	tox -e docs exec -- sphinx-apidoc --no-toc -o docs/api src/instructlab src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
+	rm -f docs/api/instructlab.*.rst
+	tox -e docs exec -- sphinx-apidoc --no-toc -o docs/api src/instructlab \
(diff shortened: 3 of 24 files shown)


.github/workflows/test.yml
```diff
@@ -150,7 +150,7 @@ jobs:
           python -m pip install tox
 
       - name: Run tox docs target
-        run: make man
+        run: tox -e docs
```
leseb, line 153: why not use a make target for this?

.readthedocs.yaml
```diff
@@ -0,0 +1,40 @@
...
+  configuration: docs/conf.py
+  # You can configure Sphinx to use a different builder, for instance use the dirhtml builder for simpler URLs
+  # builder: "dirhtml"
+  # Fail on all warnings to avoid broken references
+  # fail_on_warning: true
+
+# Optionally build your docs in additional formats such as PDF and ePub
+# formats:
+#   - pdf
+#   - epub
```
leseb, line 28: Can we remove these?

.readthedocs.yaml
```diff
@@ -0,0 +1,40 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
```
leseb, line 9: Can we have Fedora here as well?

Makefile
```diff
@@ -147,12 +147,16 @@ verify: check-tox ## Run tox -e fmt,lint,spellcheck against code
...
 	tox -e docs
 
-.PHONY: docs
-docs: man ## Run tox -e docs against code
+.PHONY: man
+man: docs
+
+.PHONY: sphinx-apidoc
+sphinx-apidoc:
+	sphinx-apidoc --no-toc -o docs/api src/instructlab src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
```
leseb, line 159: This assumes `sphinx-apidoc` is installed. Can we first check it's present?

docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
leseb, line 1:
```suggestion
# SPDX-License-Identifier: Apache-2.0

# Minimal makefile for Sphinx documentation
```

.github/workflows/test.yml
tiran: Why? It's an unnecessary level of indirection. The other test files run tox directly, too.

.readthedocs.yaml
tiran: Only Ubuntu is supported, https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os

.readthedocs.yaml
tiran: Yes, I have removed the comment block.

Makefile
tiran: I have an even better idea: `tox -e docs exec -- sphinx-apidoc ...`

.github/workflows/test.yml
leseb: It is an unnecessary level of indirection but it helps us validate the Make file at the same time.

Makefile
```diff
@@ -147,12 +147,16 @@ verify: check-tox ## Run tox -e fmt,lint,spellcheck against code
...
+docs: check-tox  ## Generate Sphinx docs and man pages
 	tox -e docs
 
-.PHONY: docs
-docs: man ## Run tox -e docs against code
+.PHONY: man
+man: docs
+
+.PHONY: sphinx-apidoc
+sphinx-apidoc:
```
leseb, line 158:
```suggestion
sphinx-apidoc: check-tox
```

tox.ini
```diff
@@ -69,12 +69,15 @@ commands =
...
 [testenv:docs]
-description = docs
+description = sphinx docs and man pages
 deps =
+    -r docs/requirements.txt
     # TODO: switch to a release tag when it's cut
     # We need https://github.com/click-contrib/click-man/pull/62 to handle hidden options
     git+https://github.com/click-contrib/click-man@24ec8377e3c24378417f6fc4f571b4f585d7df6b
 commands =
+    # add "-n" once all errors are fixed
```
leseb, line 79: Which errors?

leseb (approved):
Nice work thanks!

tox.ini
tiran: Nit pick mode was complaining about missing references. I fixed them by adding intersphinx, explicit nickpick ignores, and changing the code a bit. For example I had to make the Pydantic submoduels public, so they show up in the docs and can be referenced.

leseb: Thanks!

leseb (dismissed):
Thanks!

hickeyma (commented):
See https://github.com/instructlab/instructlab/pull/1167#pullrequestreview-2076963104 about duplication of functionality with #1167.

russellb (commented):
This seems to have a mix of docs targeting 2 different groups:

- Python API docs
- `ilab` CLI usage docs

I would prefer we keep these separate. There are some others looking at creating an InstructLab user documentation site that would focus on that type of content.

I also don't think there are any APIs in `instructlab` that we would consider public and intended for use by anyone else. Do you (or anyone) think that these API docs would be useful to anyone working on the code? That would be the primary audience for it right now.

tiran:
Closing this PR for now

If we want internal API docs, then we can come back to this PR later.

PR #1185, opened 2024-05-17
e2e: Make use of `--non-interactive` for `ilab init`

russellb:
The current implementation is a bit fragile. If the interactive flow
changed at all, this code would break. It also results in some
confusing output in a CI log.

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -51,7 +51,7 @@ test_smoke() {
 
 test_init() {
     task Initializing ilab
-    [ -f config.yaml ] || printf "taxonomy\ny" | ilab init
+    [ -f config.yaml ] || ilab init --non-interactive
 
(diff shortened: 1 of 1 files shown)


russellb:
e2e test run to exercise this change: https://github.com/instructlab/instructlab/actions/runs/9131290650

nathan-weinberg (approved):

Gregory-Pereira:
For reference the CI failure in question can be seen [here](https://github.com/containers/ai-lab-recipes/actions/runs/9127918557/job/25099222405#step:12:32). Our E2E testrun using Ilab in bootc images fails due to assuming our ENV is interactive, and waiting for the `y` entry to be applied to taxonomy cloning, even when running in CI mode.

Gregory-Pereira (approved):

leseb (approved):

cdoern (approved):
lgtm

PR #1193, opened 2024-05-21
use actual converted file name

makelinux:
don't try to guess it or use hard-coded

from #1109 
for #1006

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -1117,14 +1117,14 @@ def train(
             )
             return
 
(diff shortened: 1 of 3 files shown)


russellb (approved):

leseb (approved):

nathan-weinberg:
@makelinux does this serve as a resolution for https://github.com/instructlab/instructlab/issues/1006? Want to know if we should close that one out or not

makelinux:
> @makelinux does this serve as a resolution for #1006? Want to know if we should close that one out or not
No, it is just one change.

PR #1195, opened 2024-05-21
fix: do not fail when sending 'exit' to the chat

leseb:
This is a partial revert of https://github.com/instructlab/instructlab/commit/b8c9e63c550d07369f5381eb92b5285440cfa24b
which removes the unused 'content' argument of the `_handle_quit()`
function signature. Most of the time the function is not called with two
arguments unless the user types 'exit' in the chat, then `_handle_quit()`
is called with two arguments and we get:

```
TypeError: ConsoleChatBot._handle_quit() takes 1 positional argument but
2 were given
```

To avoid linters complaining about an argument not being accessed we use
the  `_` keyword.

Fixes: https://github.com/instructlab/instructlab/issues/1196

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -40,10 +40,10 @@ cleanup() {
     rm -rf test_taxonomy
     # revert port change from test_bind_port()
     sed -i.bak 's/9999/8000/g' config.yaml
-    rm -f config.yaml.bak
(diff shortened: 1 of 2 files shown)


nathan-weinberg (approved):
Good catch @leseb - I missed this as the tests didn't cover that case - thanks for the fix!

russellb (approved):

PR #1207, opened 2024-05-23
docs: update release documentation

nathan-weinberg:
Cleans up and adds some additional detail to our excellent release docs written by @alimaredia and @markstur


docs/release-strategy.md
```diff
@@ -47,19 +47,26 @@ The Release Manager for each release is identified in the Description of the Mil
...
+    - #frontend
+    - #users
+1. Create a milestone on GitHub for the next release without a milestone.
+
+### Z-Stream
+
+1. Backport all relevant commits from `main` to the `release-X.Y` branch - this can be done automatically with Mergify or manually if preferred.
+1. Validate the release branch with an E2E test if desired.
+1. Create a new release on GitHub targeting the release branch and using the previous Z-Stream tag as the previous release (e.g. `0.15.0` preceeds `0.15.1`).
+1. Move the `stable` tag to the new release - see [STABLE.md](STABLE.md) for in-depth directions.
```
markstur, line 68: This stable comment would be wrong in case of releasing 15.2 when an 16.0 already exists.  Probably just add a "When appropriate, " prefix in both places.

markstur (commented):
Looks good, but can you tweak the not-quite-right "stable" line if it's convenient enough to catch that nit now.  See inline.

docs/release-strategy.md
nathan-weinberg: Hmmm, will we ever do that in our current scheme? Since we're only maintaining the most recent Y-Stream version?

markstur: > Hmmm, will we ever do that in our current scheme? Since we're only maintaining the most recent Y-Stream version?

I'm sure we will support one-back with security updates eventually, but perhaps instead of tweaking this text right now we should just plan on getting rid of the "stable" tag stuff now that we have PyPI.  It's fine.

markstur (approved):
lgtm

docs/release-strategy.md
hickeyma: I agree, all references to "stable" should be removed.

hickeyma (changes requested):
Looks good, thanks @nathan-weinberg. 

Can we remove stable documentation and all references in `release-strategy` doc, as its no longer needed. 

I have pushed #1214 to update the install instructions to use the package.

docs/release-strategy.md
nathan-weinberg: Posted in `#frontend`

nathan-weinberg:
@russellb just to be clear, are you `-1` on just removing the tag and these docs references or just the former?

russellb:
> @russellb just to be clear, are you `-1` on just removing the tag and these docs references or just the former?

`docs/STABLE.md` seems like it could go.

Assuming we keep the `stable` tag, I think the release process should still include references to updating it.

I think keeping `stable` but not updating it would be even worse than just deleting it, since people will think they're getting the latest but won't. It'd be better to just break them completely so they know they need to make a change to get the latest release now.

in preference order:

1. keep `stable` tag and keep updating it for "a while" (a few months?). Document that it's deprecated and that instructions should be changed.
2. drop the `stable` tag as proposed in this PR
3. keep the `stable` tag but don't move it anymore

russellb:
Thanks, @nathan-weinberg.  Let's go ahead and pick a date and document it.  How about "the stable tag will be removed on or after September 1st, 2024."

and we broadcast a message in slack and the mailing lists about changing scripts and instructions everywhere

hickeyma (commented):
> Thanks, @nathan-weinberg. Let's go ahead and pick a date and document it. How about "the stable tag will be removed on or after September 1st, 2024."
and we broadcast a message in slack and the mailing lists about changing scripts and instructions everywhere

My initial opinion is to remove it as it was always a "stop gap" approach due to private repo and no pypi package. However, this looks like a good compromise for now.

hickeyma (approved):

docs/release-strategy.md
```diff
@@ -47,19 +47,26 @@ The Release Manager for each release is identified in the Description of the Mil
...
-6. Move the `stable` tag to the newest release.
-7. Announce release on the following channels:
-    - InstructLab Slack
-8. Create a milestone on GitHub for the next release without a milestone.
-    - For a Z-Stream release skip this step.
+### Y-Stream
+
+1. Determine a commit on the main branch that will serve as the basis for the next release - most of the time this should be the latest commit.
+1. Create a new release branch in the format `release-vX.Y` - this should match `main` when created.
+1. Validate the release branch with an E2E test if desired.
```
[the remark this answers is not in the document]
nathan-weinberg: I think that's a fair point! And relatively easy/cheap to do. I'm going to remove `if desired` here.

nathan-weinberg:
Thanks for the reviews all - have squashed commits and incorporated everyone's feedback. Kept `STABLE.md` doc as removed but we can continue to maintain the tag until Sept 1st of this year as @russellb suggested - noted this in the release doc. Once this is merged I can send out a notification to Slack/Email lists about that.

docs/release-strategy.md
```diff
@@ -47,19 +47,26 @@ The Release Manager for each release is identified in the Description of the Mil
...
-5. Validate the release (manual testing).
-6. Move the `stable` tag to the newest release.
-7. Announce release on the following channels:
-    - InstructLab Slack
-8. Create a milestone on GitHub for the next release without a milestone.
-    - For a Z-Stream release skip this step.
+### Y-Stream
+
+1. Determine a commit on the main branch that will serve as the basis for the next release - most of the time this should be the latest commit.
+1. Create a new release branch in the format `release-vX.Y` - this should match `main` when created.
```
russellb, line 53: > this should match `main` when created.

This is not always true based on out point 1 is worded

docs/release-strategy.md
```diff
@@ -47,19 +47,26 @@ The Release Manager for each release is identified in the Description of the Mil
...
+1. Validate the release branch with an E2E test.
```
russellb, line 54: A link to the doc with instructions on how to run the `e2e` job would be helpful, I think

```suggestion
1. Validate the release branch with an [E2E test](ci.md).
```

docs/release-strategy.md
```diff
@@ -47,19 +47,26 @@ The Release Manager for each release is identified in the Description of the Mil
...
+1. Move the `stable` tag to the new release (note this tag is set to be deprecated on September 1st, 2024 - users should use PyPi to install the latest "stable" release)
+1. Announce release on the following channels on Slack:
+    - #announce
+    - #frontend
+    - #users
+1. Create a milestone on GitHub for the next release without a milestone.
+
+### Z-Stream
+
+1. Backport all relevant commits from `main` to the `release-X.Y` branch - this can be done automatically with Mergify or manually if preferred.
```
russellb, line 65:
```suggestion
1. Backport all relevant commits from `main` to the `release-X.Y` branch - this can be done automatically with Mergify or manually if preferred. A backport using Mergify is done by adding a comment to the PR with the change merged to `main` with the contents `@mergifyio backport <branch-name>`.
```

docs/release-strategy.md
```diff
@@ -47,19 +47,26 @@ The Release Manager for each release is identified in the Description of the Mil
...
+1. Validate the release branch with an E2E test.
```
russellb, line 66:
```suggestion
1. Validate the release branch with an [E2E test](ci.md).
```

docs/release-strategy.md
```diff
@@ -47,19 +47,26 @@ The Release Manager for each release is identified in the Description of the Mil
...
+1. Create a new release on GitHub targeting the release branch and using the previous Z-Stream tag as the previous release (e.g. `0.15.0` preceeds `0.15.1`).
+1. Move the `stable` tag to the new release (note this tag is set to be deprecated on September 1st, 2024 - users should use PyPi to install the latest "stable" release)
+1. Announce release on the following channels on Slack:
+    - #announce
+    - #frontend
+    - #users
```
russellb, line 72:
```suggestion
    - #announce
```

docs/release-strategy.md
```diff
@@ -47,19 +47,26 @@ The Release Manager for each release is identified in the Description of the Mil
...
+
+1. Determine a commit on the main branch that will serve as the basis for the next release - most of the time this should be the latest commit.
+1. Create a new release branch in the format `release-vX.Y` - this should match `main` when created.
+1. Validate the release branch with an E2E test.
+1. Create a new release on GitHub targeting the release branch and using the latest Y-Stream tag as the previous release (e.g. `0.15.1` preceeds `0.16.0`).
+1. Move the `stable` tag to the new release (note this tag is set to be deprecated on September 1st, 2024 - users should use PyPi to install the latest "stable" release)
+1. Announce release on the following channels on Slack:
+    - #announce
+    - #frontend
+    - #users
```
russellb, line 60: Just `#announce` seems like enough to me ...

```suggestion
    - #announce
```

russellb (approved):
I gave various suggestions, but none of them are blockers

docs/release-strategy.md
nathan-weinberg: Agreed!

docs/release-strategy.md
nathan-weinberg: Good catch - pushing some updated text to clarify.

Issue #77, opened 2024-02-27
Add Python coding style


hickeyma:
It is imperative to follow a coding style like python [pep8](https://peps.python.org/pep-0008/) to ensure good coding practises. 

The work can be broken down into the following parts:

- [x] Add linting tool (suggestion [pylint](https://pylint.readthedocs.io/en/stable/)) (#78)
- [x] Enable linting in CI/CD (PR will fail until the style has been applied correctly)(#85)
- [x] Add formatters (tools that automatically format your code according to a set of rules or style guides) (suggestion:  [black](https://github.com/psf/black) and [isort](https://pycqa.github.io/isort/)(#107)
- [x] Enable formatters in CI/CD (PR will fail until the style has been applied correctly)(#120)
- [x] Add [tox](https://tox.wiki/en/4.12.1/) which is a generic virtual environment management and test command line tool that can also manage performing linting and formatting. It can be used for the unit testing also. (#78)
- [x] Add contributor doc (add when enabling CI/CD) (#85)(#121)

PR #1211, opened 2024-05-27
fix: make all bash script pass shellcheck

leseb:
All our bash scripts are now passing the shellcheck linter.

### DIFF
--- a/.github/workflows/shellcheck.yaml
+++ b/.github/workflows/shellcheck.yaml
@@ -1,0 +1,47 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Shellcheck
+
+on:
+  push:
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - '**.sh'
+      - '.github/workflows/shellcheck.yaml' # This workflow file
+  pull_request:
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - '**.sh'
+      - '.github/workflows/shellcheck.yaml' # This workflow file
+
+env:
+  LC_ALL: en_US.UTF-8
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # https://github.com/actions/checkout/issues/249
+          fetch-depth: 0
+          submodules: true
+
+      - name: Run ShellCheck
+        uses: ludeeus/action-shellcheck@master
+        with:
+          check_together: 'yes'
+          format: gcc
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -58,6 +58,14 @@ pull_request_rules:
       - and:
         - -files~='notebooks/.*\.ipynb$'
 
+    - or:
+      - and:
+        - or:
+          - files~='.*\.sh$'
+        - check-success=shellcheck
+      - and:
+        - -files~='.*\.sh$'
+
   actions:
     merge:
       method: merge
--- a/containers/rocm/remove-gfx.sh
+++ b/containers/rocm/remove-gfx.sh
@@ -13,9 +13,9 @@ case "$AMDGPU_ARCH" in
 		rm -f /usr/lib*/rocblas/library/*gfx8*
 		rm -f /usr/lib*/rocblas/library/*gfx10*
 		rm -f /usr/lib*/rocblas/library/*gfx11*
-		rm -f ${TORCH}/lib/rocblas/library/*gfx8*
-		rm -f ${TORCH}/lib/rocblas/library/*gfx10*
-		rm -f ${TORCH}/lib/rocblas/library/*gfx11*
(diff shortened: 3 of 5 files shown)


hickeyma (commented):
@leseb Is this PR related to #933?

leseb:
> @leseb Is this PR related to #933?

Somehow but no direct connection. This PR cleans up all the warnings. The other one was partial and I think #933 could stick with the CI job only. Thanks!

hickeyma (commented):
> Somehow but no direct connection. This PR cleans up all the warnings. The other one was partial and I think https://github.com/instructlab/instructlab/pull/933 could stick with the CI job only. Thanks!

My thinking is that on adding checker/linter is a follows:

1. Add checker to code base
2. Enable checker and fix all suggestions

When you merge 1 you then merge 2. This way you have a base where going forward the checker is in place, so any PR afterwards knows it is introducing the checker errors.

Will merging this PR just fix the shell checker errors for this moment in time? If so, what about the next person who makes changes/adds scripts and adds checker errors?

leseb:
> > Somehow but no direct connection. This PR cleans up all the warnings. The other one was partial and I think #933 could stick with the CI job only. Thanks!
> 
> My thinking is that on adding checker/linter is a follows:
> 
> 1. Add checker to code base
> 2. Enable checker and fix all suggestions
> 
> When you merge 1 you then merge 2. This way you have a base where going forward the checker is in place, so any PR afterwards knows it is introducing the checker errors.
> 
> Will merging this PR just fix the shell checker errors for this moment in time? If so, what about the next person who makes changes/adds scripts?

Sorry for the confusion, my thinking is that this PR should be merged since it fixes all the shellcheck issues, then #933 should follow so that any changes to bash scripts will be lint.
Does that make sense? Thanks!

I commented on https://github.com/instructlab/instructlab/pull/933#issuecomment-2134972953 let's wait a few, if no response I'll just cherry-pick https://github.com/instructlab/instructlab/pull/933/commits/f1148a0edf169f9d12637869392fb1f89fadc618 in this PR.

hickeyma:
> Sorry for the confusion, my thinking is that this PR should be merged since it fixes all the shellcheck issues, then https://github.com/instructlab/instructlab/pull/933 should follow so that any changes to bash scripts will be lint.

If you do it this way, you are waiting on the tool to be added and enabled. My suggestions is to follow the approach in https://github.com/instructlab/instructlab/issues/77 where tools are added first and then PR follows with tool enabled in CI/CD containing all changes in relevant files as suggested by the tool. This way you are enabled and ready to go.

leseb:
> @leseb I see that #933 is closed and not being continued. Does this mean we run shellcheck on the fly and update changes like this PR? Or are we going to add it to CI/CD?

I'm writing the CI job at the moment and will push shortly an update to this PR.

leseb:
Simulated failure can be seen [here](https://github.com/instructlab/instructlab/actions/runs/9285708013/job/25550833963?pr=1211)

Screenshot 2024-05-29 at 14 04 58

.github/workflows/shellcheck.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - "main"
+      - "release-**"
+    paths:
+      - '**.sh'
+  pull_request:
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - '**.sh'
```
russellb, line 17: How about when this file is changed as well?

russellb (approved):
one minor suggestion, but lgtm!

.github/workflows/shellcheck.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+name: Shellcheck
+
+on:
+  push:
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - '**.sh'
```
bjhargrave, line 11:
```suggestion
      - '**.sh'
      - '.github/workflows/shellcheck.yaml' # This workflow
```

.github/workflows/shellcheck.yaml
bjhargrave, line 17:
```suggestion
      - '**.sh'
      - '.github/workflows/shellcheck.yaml' # This workflow
```

.github/workflows/shellcheck.yaml
```diff
@@ -0,0 +1,38 @@
...
+  pull_request:
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - '**.sh'
+
+permissions:
```
bjhargrave, line 19:
```suggestion
env:
  LC_ALL: en_US.UTF-8

defaults:
  run:
    shell: bash


permissions:
```

bjhargrave (changes requested):

.github/workflows/shellcheck.yaml
leseb: Why do we need?

```
defaults:
	  run:
	    shell: bash
```

leseb: Actually, why do we need all the suggestions? Thanks!

leseb:
Need https://github.com/instructlab/instructlab/pull/1246 for CI to be green

russellb:
It looks like @hickeyma 's comments were addressed, so I'm going to drop that pending review and let this merge. Let me know if I missed anything, though!

PR #1222, opened 2024-05-29
CPU training optimization

russellb:
This is a rebased copy of the PR originally in #1109 and is related to issue
#1006.

I wanted to test it out and see if we should revisit the changes, but needed to
rebase first. Since I did the rebase, I figured I'd go ahead and post it as a
draft while playing with it.

I'll update the title, description, and series of commits prior to a request for
review if we get that far.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -40,3 +40,4 @@ langchain-text-splitters
 # the below library should NOT be imported into any python files
 # it is for CLI usage ONLY
 yamllint>=1.35.1,<1.36.0
+psutil>=5.9.8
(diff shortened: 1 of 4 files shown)


tiran:
My PR #993 contains most bits of #1006, too.

russellb:
> My PR #993 contains most bits of #1006, too.

Oh, thanks! I'll focus on that one, then.

russellb:
Actually, I'll keep this draft open for the moment since I'm in the middle of testing these CPU focused changes on their own ... but obviously if we can get #993 in this is not necessary

russellb:
Similar to the comment on #1225, ilab gets OOM killed using this PR on my laptop with 32 GB RAM.

russellb:
I'm working through some follow-up changes necessary on #993 now

PR #1231, opened 2024-05-30
introduce a release note for release

leseb:
From now on, let's get into the habit of editing the
`CHANGELOG.md` file on PR adding notable features and/or
breaking changes that users need to be aware of.

### DIFF
--- a/docs/release-strategy.md
+++ b/docs/release-strategy.md
@@ -68,3 +68,23 @@ The following are the steps for how Y-stream and Z-stream releases gets cut.
 1. Announce release via the following:
     - The `#announce` channel on Slack
     - The `announce` mailing list
+
+## Release Notes
+
+The project maintains a single `CHANGELOG.md` file that documents all releases. To ensure our users
+are well-informed about new features, improvements, and breaking changes, we maintain a
+`CHANGELOG.md` file. This file serves as a centralized place to document changes that will be
+included in the next (X) or (Y) release. Given that the project is in its early stages, we are
+currently focused on documenting changes for the next (Y) release.
+
+### Editing Release Notes
+
+When submitting a Pull Request (PR) that introduces notable features or breaking changes, committers
+need to update the `CHANGELOG.md` file. Clearly describe the changes, their impact, and
+any actions users might need to take. We want clear, concise, and user-friendly notes.
+
+### Branching for a New Release
+
+Each time we prepare for a new (X) or (Y) release, we branch out from the main codebase.
+As part of this branching process, the contents of `CHANGELOG.md` are reviewed and
+finalized.
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -84,6 +84,7 @@ backends
 backport
 backported
 cli
+codebase
 compositional
 coreutils
 cpp
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+## v0.17
+
+### Features
+
+### Breaking Changes


nathan-weinberg:
Maybe we should put this in the `docs/` driectory - wdyt @leseb?

alimaredia:
@leseb In the scenario where someone opens up a PR with a breaking change. Should the author edit this file in that PR or will a release manager/maintainer be responsible for updating this doc before the release.

leseb:
> @leseb In the scenario where someone opens up a PR with a breaking change. Should the author edit this file in that PR or will a release manager/maintainer be responsible for updating this doc before the release.

From my perspective, whoever introduces feature X with a PR has to update the PendingReleaseNotes.md in that PR. Most of the time, maintainers will have to ask for it but we could do better and add a note in the PR template.

leseb:
> Maybe we should put this in the `docs/` directory - wdyt @leseb?

In most of the projects I've worked with, we had this file at the root of the repo. It's simple, easy to find and won't cause any trouble if `docs` disappears or something else.

alimaredia:
> > @leseb In the scenario where someone opens up a PR with a breaking change. Should the author edit this file in that PR or will a release manager/maintainer be responsible for updating this doc before the release.
> 
> From my perspective, whoever introduces feature X with a PR has to update the PendingReleaseNotes.md in that PR. Most of the time, maintainers will have to ask for it but we could do better and add a note in the PR template.

That sounds good to me. What do you think about adding automation around that to this PR? Say a maintainer adds a "Release Notes" label, this then the PR can't get merged unless the new file changed?

leseb:
> I support this in general. It's probably worth an update to the release process doc for how to handle this, as well.

Thanks for the suggestion, I added a new section in `docs/release-strategy.md`. PTAL. Thanks!

nathan-weinberg (approved):

docs/release-strategy.md
```diff
@@ -68,3 +68,29 @@ The following are the steps for how Y-stream and Z-stream releases gets cut.
...
+need to update the `PendingReleaseNotes.md` file. Clearly describe the changes, their impact, and
+any actions users might need to take. We want clear, concise, and user-friendly notes.
+
+### Branching for a New Release
+
+Each time we prepare for a new major (X) release, we branch out from the main codebase.
+As part of this branching process, the contents of `PendingReleaseNotes.md` are reviewed and
+finalized.
+
+### Resetting the Release Notes
```
alimaredia, line 90: @leseb why not just get rid of the overhead of clearing `PendingReleaseNotes.md` and have the file being a running list of breaking changes with those for the upcoming release at the top. This was how it was done in Ceph (https://github.com/ceph/ceph/blob/main/PendingReleaseNotes). 

Though this file can get long I think there's value to having all of the breaking changes between versions in one place.

PendingReleaseNotes.md
```diff
@@ -0,0 +1,5 @@
+# v0.17 Pending Release Notes
+
+## Breaking Changes
+
+## Features
```
russellb, line 5: This format will be necessary to allow for multiple release's worth of notes in the file.

```suggestion
# Release Notes

## v0.17 Pending Release Notes

### Breaking Changes

### Features
```

docs/release-strategy.md
russellb: +1

and in that case, I'd rename the file to `ReleaseNotes.md`, though `RELEASE_NOTES.md` would be more consistent with other files in the same directory.

leseb: After some more digging, it looks like most of the notable open-source projects out there use a `CHANGELOG.md` file. The projects maintain a single `CHANGELOG.md` file that documents all releases. I like the suggestion @alimaredia @russellb, I think this approach provides a great balance between consistency, ease of management, and centralized information. Thanks!

docs/release-strategy.md
```diff
@@ -68,3 +68,22 @@ The following are the steps for how Y-stream and Z-stream releases gets cut.
 1. Announce release via the following:
     - The `#announce` channel on Slack
     - The `announce` mailing list
+
+## Release notes
```
nathan-weinberg, line 72:
```suggestion
## Release Notes
```

docs/release-strategy.md
```diff
@@ -68,3 +68,22 @@ The following are the steps for how Y-stream and Z-stream releases gets cut.
...
+
+The project maintains a single `CHANGELOG.md` file that documents all releases.
+To ensure our users are well-informed about new features, improvements, and breaking changes, we
+maintain a `CHANGELOG.md` file. This file serves as a centralized place to document changes
+that will be included in the next major (X) release.
```
nathan-weinberg, line 77:
```suggestion
that will be included in the next major (X) release.
```
Y-streams as well?

docs/release-strategy.md
```diff
@@ -68,3 +68,22 @@ The following are the steps for how Y-stream and Z-stream releases gets cut.
...
+
+### Editing Release Notes
+
+When submitting a Pull Request (PR) that introduces notable features or breaking changes, committers
+need to update the `CHANGELOG.md` file. Clearly describe the changes, their impact, and
+any actions users might need to take. We want clear, concise, and user-friendly notes.
+
+### Branching for a New Release
+
+Each time we prepare for a new major (X) release, we branch out from the main codebase.
```
nathan-weinberg, line 87: Same question here about Y-Streams

docs/release-strategy.md
leseb: The Y-streams are backports only right? So they will only include critical bug fixes, no features nor breaking changes. From my perspective, it's just Major release updates.

docs/release-strategy.md
leseb: https://github.com/instructlab/instructlab/pull/1231#discussion_r1626088030

docs/release-strategy.md
nathan-weinberg: Z-Streams are backports only

docs/release-strategy.md
nathan-weinberg: Responded there!

docs/release-strategy.md
leseb: Rah sorry, actually I mixed things up. I was talking about (Y) releases all along and not (X). In the end, both X and Y will get the same treatment. Thanks for pointing this out!

nathan-weinberg: Np at all!

nathan-weinberg (approved):

russellb (approved):

PR #1234, opened 2024-05-31
chat: Fix ssl when only cert file is passed

booxter:
httpx.Client expects cert argument to be one of: a 2+ tuple, or a string, or None.

Before the change, we were passing a 1-tuple when only the cert file was configured through cli, which is violating the typing requirements of the library; and it also made the library not initialize SSL context at all. See:

https://github.com/encode/httpx/blob/37593c1952f4972040f6163da67e3777fd3d2e94/httpx/_config.py#L183

### DIFF
--- a/src/instructlab/generator/utils.py
+++ b/src/instructlab/generator/utils.py
@@ -17,7 +17,7 @@ import httpx
 
 # Local
 from ..config import DEFAULT_API_KEY, DEFAULT_MODEL_OLD
-from ..utils import get_sysprompt
+from ..utils import get_ssl_cert_config, get_sysprompt
 
 StrOrOpenAIObject = Union[str, object]
 
@@ -124,8 +124,7 @@ def openai_completion(
 
         # do not pass a lower timeout to this client since generating a dataset takes some time
         # pylint: disable=R0801
-        orig_cert = (tls_client_cert, tls_client_key, tls_client_passwd)
-        cert = tuple(item for item in orig_cert if item)
+        cert = get_ssl_cert_config(tls_client_cert, tls_client_key, tls_client_passwd)
         verify = not tls_insecure
         client = OpenAI(
             base_url=api_base,
--- a/src/instructlab/chat/chat.py
+++ b/src/instructlab/chat/chat.py
@@ -23,7 +23,7 @@ import openai
 
 # Local
 from ..config import DEFAULT_CONNECTION_TIMEOUT, DEFAULT_MODEL_OLD
-from ..utils import get_sysprompt
+from ..utils import get_ssl_cert_config, get_sysprompt
 
 HELP_MD = """
 Help / TL;DR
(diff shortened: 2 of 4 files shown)


booxter:
Yes I noticed it's not just here. This should be caught by type checking, e.g. mypy.

russellb (dismissed):
fair point that there are other places that could be fixed, but I'm fine with a subset going in if an issue is filed at least

booxter:
Thanks Russell. I think the request to fix everywhere is fair. I'd like to hold it off until I consolidate all ssl config code into one place and fix it once and for all. I will set it to draft for now.

booxter:
Consolidated ssl cert code in a single place now. I wonder if we should have some tests for these options, I can't find any in the functional test suite...

src/instructlab/generator/utils.py
```diff
@@ -124,8 +124,7 @@ def openai_completion(
 
         # do not pass a lower timeout to this client since generating a dataset takes some time
         # pylint: disable=R0801
-        orig_cert = (tls_client_cert, tls_client_key, tls_client_passwd)
-        cert = tuple(item for item in orig_cert if item)
+        cert = get_ssl_cert_config(tls_client_cert, tls_client_key, tls_client_passwd)
         verify = not tls_insecure
         client = OpenAI(
```
booxter, line 129: the whole code to set up OpenAI() is duplicate and should probably be consolidated. But there's a slight different atm - in one of these places, timeout is not passed, while two others pass default timeout defined in instructlab tree. Perhaps it's an oversight too and should be fixed (and then consolidated into a single function), but I leave it out of this PR.

russellb:
> Consolidated ssl cert code in a single place now. I wonder if we should have some tests for these options, I can't find any in the functional test suite...

Tests right now are ...

- `scripts/functional-tests.sh` - runs in all the Test workflows
- `tests/*.py` -- a mix of what I would call unit tests and functional tests, runs in all Test workflows
- `scripts/basic-workflow-tests.sh` -- runs in the on-demand e2e workflow (runs on a GPU enabled worker)

russellb (approved):
I'll have to pull the generator/ directory change into the new `sdg` repo. We're in the middle of moving that code. The change is simple enough that I don't mind doing it, though.

booxter:
@russellb I don't see where the options are passed though at the moment... I suspect this change is not fully tested because of it (automatically, I mean).

booxter:
> I'll have to pull the generator/ directory change into the new sdg repo

If sdg is supposed to be a shared library, then maybe the new function should go there too.

russellb:
> > I'll have to pull the generator/ directory change into the new sdg repo
> 
> If sdg is supposed to be a shared library, then maybe the new function should go there too.

Yeah, that's what I was thinking.

leseb (approved):

Issue #1236, opened 2024-05-31
help command in chat console hit error


chudegao:
```
(dlipy3) [root@llm opt]# ilab chat -gm
DEBUG 2024-05-31 05:02:43,254 server.py:63 Trying to connect to http://127.0.0.1:8000/v1...
DEBUG 2024-05-31 05:02:45,737 server.py:81 Trying port 58166...
DEBUG 2024-05-31 05:02:45,737 server.py:101 Port 58166 is available.
DEBUG 2024-05-31 05:02:45,737 server.py:105 Connection to http://127.0.0.1:8000/v1 failed. Starting a temporary server at http://127.0.0.1:58166/v1 with model_path models/merlinite-7b-lab-Q4_K_M.gguf...
=========greedy_mode: True
╭──────────────────────────────────────────────────── system ────────────────────────────────────────────────────╮
│ Welcome to InstructLab Chat w/ MERLINITE-7B-LAB-Q4_K_M (type /h for help)                                      │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
>>> INFO 2024-05-31 05:02:47,732 server.py:206 Starting server process, press CTRL+C to shutdown server...default]
INFO 2024-05-31 05:02:47,732 server.py:207 After application startup complete see http://127.0.0.1:58166/docs for API.
>>>                                                                                                   [S][default]
>>> /s save.log                                                                                       [S][default]
>>> /h                                                                                                [S][default]
Traceback (most recent call last):
  File "/opt/anaconda3/envs/dlipy3/bin/ilab", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/opt/anaconda3/envs/dlipy3/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/dlipy3/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/dlipy3/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1237, opened 2024-05-31
fix ilab chat run /h command failed issue

chudegao:
# Changes

Resolves #https://github.com/instructlab/instructlab/issues/1236

chat.py: def _handle_help(slef) missed one parameter even it does not use it. Update to def _handle_help(self, _)


leseb (changes requested):
Looks good, please update your commit message body with the same content as the PR. (`chat.py: def _handle_help(slef) missed one parameter even it does not use it. Update to def _handle_help(self, _)`)

Thanks!

leseb (approved):

nathan-weinberg (approved):

PR #1239, opened 2024-05-31
ci: try to fix flaky test_generate() functional test

leseb:
The test seems to fail intermittently, so let's try to push the timeout to 20m as well as printing a clear indication of why the functional test is going to fail.

Failure example: https://github.com/leseb/rook/blob/bcd2511a7269ceb72c5c4a255fa760de85dea1e1/pkg/operator/ceph/cluster/osd/spec.go#L169-L182

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -231,8 +231,11 @@ EOF
 
     sed -i.bak -e 's/num_instructions:.*/num_instructions: 1/g' config.yaml
 
-    # This should be finished in a minut or so but time it out incase it goes wrong
-    timeout 10m ilab generate --taxonomy-path test_taxonomy/compositional_skills/simple_math.yaml
+    # This should be finished in a minute or so but time it out incase it goes wrong
+    if ! timeout 20m ilab generate --taxonomy-path test_taxonomy/compositional_skills/simple_math.yaml; then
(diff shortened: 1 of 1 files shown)


scripts/functional-tests.sh
```diff
@@ -232,7 +232,10 @@ EOF
     sed -i.bak -e 's/num_instructions:.*/num_instructions: 1/g' config.yaml
 
     # This should be finished in a minut or so but time it out incase it goes wrong
-    timeout 10m ilab generate --taxonomy-path test_taxonomy/compositional_skills/simple_math.yaml
+    if ! timeout 20m ilab generate --taxonomy-path test_taxonomy/compositional_skills/simple_math.yaml; then
```
nathan-weinberg, line 235: nit: could you fix the text in the above comment while you're in here :smile:

leseb: Actually, the comment is still valid, most of the time the process takes a few mins.

nathan-weinberg: there's a typo - `minut` should be `minute`

leseb: Ahhhhh the typo, ok ok :)

nathan-weinberg (approved):
Thank you @leseb!

russellb (approved):

PR #1241, opened 2024-06-02
refactor TrainingArguments

makelinux:
Current code contains some duplicated code and variables, which don't smell good:
```
output_dir=output_dir,
num_train_epochs=num_epochs,
per_device_train_batch_size=per_device_train_batch_size,
```

Introduce common train_args for GaudiSFTTrainer/GaudiTrainingArguments and FTTrainer/STrainingArguments to avoid duplication and improve maintainability.

### DIFF
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -239,7 +239,6 @@ def linux_train(
 
     if four_bit_quant:
         print("LINUX_TRAIN.PY: USING 4-bit quantization with BitsAndBytes")
-        use_fp16 = True
         bnb_config = BitsAndBytesConfig(
             load_in_4bit=True,
             bnb_4bit_quant_type="nf4",
@@ -248,7 +247,6 @@ def linux_train(
         )
     else:
         print("LINUX_TRAIN.PY: NOT USING 4-bit quantization")
-        use_fp16 = False
         bnb_config = None
 
     # Loading the model
@@ -335,7 +333,14 @@ def linux_train(
     )
 
     tokenizer.padding_side = "right"
-    per_device_train_batch_size = 1
+    train_args = {
+        "output_dir": "./training_results",
+        "fp16": four_bit_quant,
+        "num_train_epochs": num_epochs,
+        "per_device_train_batch_size": 1,
+        "save_strategy": "epoch",
+        "report_to": "none",
+    }
     max_seq_length = 300
 
     if device.type == "hpu":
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -16,6 +16,7 @@ mlx>=0.5.1,<0.6.0; sys_platform == 'darwin' and platform_machine == 'arm64'
 # HabanaLabs / Intel Gaudi env comes with Python 3.10 and slightly older
 # versions of some dependencies. Use '3.10' as an indicator.
 # Habana installer has NumPy 1.23.5
+munch
```
leseb, line 19: can we pin this dep?

makelinux: I've pinned. But why it is important?

leseb: The repo is following this practice.
One example is future-proofing, It allows us to gradually adopt new features and improvements from the dependency, making it easier to stay current and reducing the effort required for major upgrades. This was we can avoid pulling a recent version that would break backward compatibility.

src/instructlab/train/linux_train.py
```diff
@@ -289,31 +288,30 @@ def model_generate(user, **kwargs):
     )
 
     tokenizer.padding_side = "right"
-    output_dir = "./training_results"
-    per_device_train_batch_size = 1
+    train_args = Munch(
```
tiran, line 291: Why did you introduce a new dependency? Wouldn't `dict()` work here, too?

src/instructlab/train/linux_train.py
```diff
@@ -289,31 +288,30 @@ def model_generate(user, **kwargs):
...
-            # gaudi_config_name=gaudi_config_name,
-        )
+        gta = GaudiTrainingArguments(**train_args, use_habana=True)
+        gta.bf16 = True
+        gta.use_lazy_mode = True
+        # create checkpoint directories
+        gta.save_on_each_node = True
+        # gaudi_config_name=gaudi_config_name,
+        if gta.per_device_train_batch_size == 1:
+            gta.per_device_train_batch_size = 8
```
tiran, line 313: Have you verified that this approach works with Intel Gaudi software 1.15.1 and 1.16.0, and Intel Gaudi 2 cluster?

tiran (changes requested):

src/instructlab/train/linux_train.py
makelinux: No.

I've just assured than result of `GaudiTrainingArguments` is the same as `before`:

```
import dictdiffer

for diff in list(dictdiffer.diff(gta.to_dict(), training_arguments.to_dict()):
    print(diff)
```

src/instructlab/train/linux_train.py
makelinux: Munch is more readable than dict for accessing individual members.

leseb (dismissed):
@tiran PTAL :)

markstur (commented):
lgtm, but let's have tiran try it out

hickeyma (dismissed):
LGTM, thanks @makelinux. 

@tiran do you mind checking it also? Added hold until then.

tiran:
I still haven't had time to test this change with Intel Gaudi hardware. Could you please leave the Gaudi section as is?

makelinux:
> I still haven't had time to test this change with Intel Gaudi hardware. Could you please leave the Gaudi section as is?

OK. Funny enough that Habana Labs is located 15 minutes from me, but I have no folks there.

PR #1244, opened 2024-06-03
Make use of new library from `instructlab/sdg`

### LINKED ISSUE #1245: Make use of new library from `instructlab/sdg`
The SDG code from this repo is being moved into a new library. Once the first release of the library is made, this repo needs to be updated to use it.

new repo: https://github.com/instructlab/sdg

russellb:
This PR drops the sdg code from this repo and replaces it with the new
library housed in `instructlab/sdg`. The first release of the library
will maintain an identical interface as used today to allow for a
quick transition.

Closes #1245 


requirements.txt
```diff
@@ -40,3 +40,4 @@ langchain-text-splitters
 # the below library should NOT be imported into any python files
 # it is for CLI usage ONLY
 yamllint>=1.35.1,<1.36.0
+instructlab.sdg==0.0.1rc1
```
bjhargrave, line 43:
```suggestion
instructlab-sdg==0.0.1rc1
```

Isn't the package name with a dash?

https://github.com/instructlab/sdg/blob/c46d5067fa2ee19dccb0530026b2d8109147464d/pyproject.toml#L8

russellb: good catch, thanks!

nathan-weinberg (approved):
LGTM - added folks from CLI and SDG squads to look as well

src/instructlab/data/generate.py
```diff
@@ -146,13 +146,13 @@ def generate(
 ):
     """Generates synthetic data to enhance your example data"""
     # pylint: disable=C0415
+    # Third Party
+    from instructlab.sdg.generate_data import generate_data
```
cdoern, line 150: is this the final way this will work? How will the user know which sdg version they are running? is that all internal?

For now, I feel like that is fine if we just do the work for them but eventually I feel like this should be surfaced

russellb: the code will certainly evolve. The goal at this iteration was moving the CLI code as close to as-is as possible

leseb (approved):

leseb:
@Mergifyio rebase

russellb:
@mergifyio rebase

russellb:
I tried disabling the pip cache to see if a bad cache was the cause of the failure related to the schema package, but that's not it -- same failure

russellb:
I fixed the problem that was breaking CI. There's a bit more explanation in the fix here: https://github.com/instructlab/sdg/pull/28

I believe this is ready now.

leseb (approved):

leseb (dismissed):

tiran (approved):

requirements.txt
```diff
@@ -12,6 +12,7 @@ httpx>=0.25.0
 instructlab-quantize>=0.1.0
 instructlab-schema>=0.2.0
 jsonschema>=4.21.1
+instructlab-sdg>=0.0.2,<0.1.0
```
bjhargrave, line 15:
```suggestion
instructlab-sdg>=0.0.2
jsonschema>=4.21.1
```

Moved up a line to sort properly and also removed the upper range in line with recent changes to most dependencies.

bjhargrave (changes requested):
Otherwise LGTM

leseb (approved):

bjhargrave (approved):

PR #1250, opened 2024-06-03
ci: install gh cli on e2e runner

leseb:
The current flow was assuming the 'gh' CLI was present, we now install and configure it. Also, we remove the git usage since using gh is shorter.
Failing run error:

```
/home/runner/work/_temp/09f60100-6344-4e59-add5-004d08de3070.sh: line 1: gh: command not found
```

Link: https://github.com/instructlab/instructlab/actions/runs/9353362921/job/25743665032

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -31,21 +31,59 @@ jobs:
             echo "is_pr=false" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Check if gh cli is installed
+        id: gh_cli
+        run: |
+          if command -v gh &> /dev/null ; then
+            echo "gh_cli_installed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "gh_cli_installed=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Configure gh CLI APT repository
+        if: steps.gh_cli.outputs.gh_cli_installed == 'false'
+        run: |
+          (type -p wget >/dev/null || (sudo apt update && sudo apt-get install wget -y))
+          sudo mkdir -p -m 755 /etc/apt/keyrings
+          wget -qO- https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo tee /etc/apt/keyrings/githubcli-archive-keyring.gpg > /dev/null
+          sudo chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -31,17 +31,38 @@ jobs:
             echo "is_pr=false" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Install gh CLI
+        run: |
+          (type -p wget >/dev/null || (sudo apt update && sudo apt-get install wget -y)) \
+          && sudo mkdir -p -m 755 /etc/apt/keyrings \
```
bjhargrave, line 37: Do we need all the `&&`s in here to have a single command? I think we can use multiple lines since the step will stop when any command fails.

This should also probably be 2 steps: Setup apt, install gh cli.

.github/workflows/e2e.yml
```diff
@@ -31,17 +31,38 @@ jobs:
...
+          && wget -qO- https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo tee /etc/apt/keyrings/githubcli-archive-keyring.gpg > /dev/null \
+          && sudo chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg \
+          && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+          && sudo apt update \
+          && sudo apt install gh -y
+
+      - name: test gh CLI
+        run: gh --version
```
bjhargrave, line 45:
```suggestion
        run: |
          gh --version
```

to be consistent with the other steps.

.github/workflows/e2e.yml
```diff
@@ -31,17 +31,38 @@ jobs:
...
+
+      - name: set default repo
+        run: gh repo set-default ${{ github.server_url }}/${{ github.repository }}
```
bjhargrave, line 48:
```suggestion
        run: |
          gh repo set-default ${{ github.server_url }}/${{ github.repository }}
```

russellb:
An alternative to both #1250 and #1253 is to just got back to the github script solution to this and then we don't have to worry about checking for and installing `gh`.

bjhargrave:
I still like using gh instead of typescript which no one will understand to maintain. Yes, it may be work to install gh cli, but our future selves will appreciate it when we have to improve this workflow some day :-)

bjhargrave (approved):

russellb (approved):

russellb:
follow-up here #1261

PR #1253, opened 2024-06-03
e2e: Install gh command if it's missing

russellb:
Commenting on PRs is failing because `gh` is not present. This should
install it.


bjhargrave:
This seems like a shorter version of #1250.

russellb:
> This seems like a shorter version of #1250.

ah I missed that other one.  At this point the github script one was less lines overall ...

russellb:
addressed by #1250

PR #1255, opened 2024-06-04
linux_train returns training_results_dir

makelinux:
Refactoring to avoid procedural programming and
rigid hard-coded key values in different files
to use functional programming and flexible style.

The "training_results" string can also be easily parameterized when it is necessary.

### DIFF
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -151,13 +151,15 @@ def linux_train(
     num_epochs: Optional[int] = None,
     device: torch.device = torch.device("cpu"),
     four_bit_quant: bool = False,
-):
+    output_dir: Path = Path("training_results"),
+) -> Path:
     """Lab Train for Linux!"""
     print("LINUX_TRAIN.PY: NUM EPOCHS IS: ", num_epochs)
     print("LINUX_TRAIN.PY: TRAIN FILE IS: ", train_file)
     print("LINUX_TRAIN.PY: TEST FILE IS: ", test_file)
 
     print(f"LINUX_TRAIN.PY: Using device '{device}'")
+    output_dir = Path(output_dir)
     if device.type == "cuda":
         # estimated by watching nvtop / radeontop during training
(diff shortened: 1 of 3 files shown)


leseb (commented):
Not sure why DCO does not see your sign-off, can you amend and push again? Thanks!

src/instructlab/train/linux_train.py
```diff
@@ -396,3 +397,4 @@ def model_generate(user, **kwargs):
     model.save_pretrained("./training_results/merged_model")
 
     print("LINUX_TRAIN.PY: FINISHED")
+    return "./training_results"
```
leseb, line 400: Can we use a const for `training_results` since it is used and re-used in a few places?

makelinux: I've made argument with default value.

leseb (approved):

src/instructlab/train/linux_train.py
```diff
@@ -393,6 +394,7 @@ def model_generate(user, **kwargs):
 
     print("LINUX_TRAIN.PY: MERGING ADAPTERS")
     model = trainer.model.merge_and_unload()
-    model.save_pretrained("./training_results/merged_model")
+    model.save_pretrained(output_dir + "/merged_model")
```
russellb, line 397: I don't think it matters in this case, but this is slightly preferred:

```suggestion
    model.save_pretrained(os.path.join(output_dir, "/merged_model"))
```

russellb: I would have applied this, but looks like maintainer access to the PR branch was turned off

leseb (approved):

russellb (approved):

leseb:
@Mergifyio rebase

leseb:
The branch needs to be rebased for Mergify to merge it, this is due to https://github.com/instructlab/instructlab/commit/8be639723828884dc88945b5c9572c49863ffae8. However, the lint workflow passes so I'm going ahead with this change.

PR #1258, opened 2024-06-04
sort train_files and test_files by real time

### LINKED ISSUE #1227: ilab train may use unexpected generated data
ilab generate can generate different data in same directory. ilab train are trying to use the latest one. However, if user generate data with different models, current implementation(sort by name) can not guarantee it use the latest one. 
Some proposal:
1. ilab generate check output data dir, make sure it's empty. Or
2. ilab generate  make sure only same model data are in the output data dir. Or
3. ilab train make sort files under input-dir with time only.

makelinux:
Sorting by name is error prone. One model name
will shadow another, when it goes alphabetically
after.

Fixes #1227

### DIFF
--- a/src/instructlab/model/train.py
+++ b/src/instructlab/model/train.py
@@ -3,6 +3,7 @@
 # Standard
 from glob import glob
 from pathlib import Path
+import logging
 import os
 import shutil
 
@@ -12,6 +13,8 @@ import torch
 
 # First Party
 from instructlab import utils
+
+logger = logging.getLogger(__name__)
 
 
 @click.command()
@@ -142,9 +145,11 @@ def train(
             )
             raise click.exceptions.Exit(1)
 
-        # generated input files reverse sorted by name (contains timestamp)
+        # generated input files reverse sorted by modification time
         def get_files(pattern):
-            return sorted(Path(input_dir).glob(pattern), reverse=True)
+            return sorted(
+                Path(input_dir).glob(pattern), key=os.path.getmtime, reverse=True
+            )
 
         train_files = get_files("train_*")
         test_files = get_files("test_*")
@@ -160,7 +165,9 @@ def train(
                 "Found multiple files from `ilab generate`. Using the most recent generation.",
                 fg="yellow",
             )
-        # First file is latest (by above reverse sort and timestamped names)
+        # The first file is latest
+        logger.debug("train_file=%s", train_files[0])
+        logger.debug("test_file=%s", test_files[0])
         shutil.copy(train_files[0], train_file)
         shutil.copy(test_files[0], test_file)
 


markstur (commented):
How does this improve the model shadowing mentioned in the comment?  Isn't it the same problem switching between models with a mod time?

Also curious how the timestamp used in the file name when the output is created by a run of generate would be wrong.  Seems pretty clear as-is.  Can you clarify what this really fixes.

Otherwise mod-time change seem pretty straightforward and makes sense.  I'm just unclear about what it really fixes and whether a file mod-time is really better than a naming convention.  Mod-time is a bit sneakier (e.g. view/save can update) where the file name isn't perfect but is fairly obvious.

makelinux:
@markstur , 
InstructLab support different model families: merlinite and mixtral for now.
Generated file names are concatenated model name and timestamp, for example:
`test_merlinite-7b-lab-Q4_K_M_2024-06-06T09_04_57.jsonl`

After reverse sorting "mixtral" will **always** go before "merlinite", no matter what timestamp is.

leseb (changes requested):
I have a personal preference for using file attributes instead of sorting on file names that include a timestamp. Using `mtime` for tracking recent changes and updates makes sense to me, that way we are sure to get the last modified file.

Please include a sorting example before and after. Once done; this is a LGTM. Thanks!

chudegao:
I think this pr can address this issue https://github.com/instructlab/instructlab/issues/1227

nathan-weinberg:
> I think this pr can address this issue #1227

They are already linked

makelinux:
> > I think this pr can address this issue #1227
> 
> They are already linked

I've just linked thanks to this comment.

makelinux:
@leseb 

Run generate first time with mixtral and then with merlinite:                         
 ```
ilab generate --model models/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf         
                                                                            
ilab generate --model models/merlinite-7b-lab-Q4_K_M.gguf                   
```
Run different versions of train:                                           
Without sorting:                                                             
```
ilab model train                                                            
generated/train_mixtral-8x7b-instruct-v0.1.Q4_K_M_2024-06-13T14_32_42.jsonl 
generated/test_mixtral-8x7b-instruct-v0.1.Q4_K_M_2024-06-13T14_32_42.jsonl  
                                                                            
```
With sorting:                                                                
```
ilab model train                                                            
generated/train_merlinite-7b-lab-Q4_K_M_2024-06-13T15_36_00.jsonl           
generated/test_merlinite-7b-lab-Q4_K_M_2024-06-13T15_36_00.jsonl  
```

**Review by leseb — DISMISSED** [review]
Ideally, we would have had one commit for the sorting and another for the added log debug. However, I don't want to block the progress for this. You can still split the commits if you prefer :), thanks!

File: src/instructlab/model/train.py [hunk 1]
```diff
@@ -222,7 +224,9 @@ def get_files(pattern):
                 "Found multiple files from `ilab generate`. Using the most recent generation.",
                 fg="yellow",
             )
-        # First file is latest (by above reverse sort and timestamped names)
+        # The first file is latest
+        ctx.obj.logger.debug("train_file=%s", train_files[0])
+        ctx.obj.logger.debug("test_file=%s", test_files[0])
```
bjhargrave, line 229: This will need to be changed to use the new logging style where each module uses it's own logger.

```suggestion
        # The first file is latest
        logger.debug("train_file=%s", train_files[0])
        logger.debug("test_file=%s", test_files[0])
```

Near the top of the file

```python
import logging

bjhargrave (changes requested):

leseb (approved):

bjhargrave (approved):

PR #1233, opened 2024-05-30
Misc code cleanup in lab.train endpoint

booxter:
This is code cleanup (see each commit message in the series for details of what is done and why). It also fixes a `NameError` issue when `data-dir` is passed on linux (found the issue during code cleanup).

PR #1259, opened 2024-06-04
refactor copying of training_results

makelinux:
Use clean loop with table-like arguments
instead of flat interleaved repetitive code.


leseb:
Addressed in https://github.com/instructlab/instructlab/pull/1233 it seems.

PR #1263, opened 2024-06-04
Support namespace packages under `instructlab`

russellb:
The new `sdg` repository includes a new Python library. The current
consensus on the following issue is that it should be published as a
Python namespace package, `instructlab.sdg`.

This magic snippet is required to allow this. See the discussion on
the issue for further discussion:

https://github.com/instructlab/sdg/issues/2

### DIFF
--- a/src/instructlab/__init__.py
+++ b/src/instructlab/__init__.py
@@ -1,0 +1,1 @@
+__path__ = __import__("pkgutil").extend_path(__path__, __name__)


bjhargrave:
Timing is perfect! I am just working on making a python package for the schema and will exploit this to use the `instructlab.schema` package as we currently do using submodules.

bjhargrave (approved):

tiran:
> One suggests this is not needed, am I missing something?

Both options are valid. PEP 420 namespace packages are the future, but some tooling and our test setup is incompatible with them. For example pylint needs a different configuration to handle them. You can easily reproduce the problem locally by removing `src/instructlab/__init__.py` and then running `tox -e fastlint`.

Legacy namespace packages with `pkgutil` work out of the box.

My recommendation is to use legacy namespace package for now, then refactor and improve after $DEADLINE.

russellb:
@tiran are the changes included here sufficient to merge?

tiran (approved):

PR #1269, opened 2024-06-05
train: fix type violation when passing test_file and train_file

booxter:
The linux_train function defines these arguments as `str`, but we are now passing lists of strings. The `load_dataset` function down the stream accepts either a `str` or a sequence of strings.

This patch changes the type of the arguments to `Path` since it's a stronger type. It then converts to string before passing it further to `load_dataset`.

This also reverts a test case change that made the check ineffective because it no longer validated what the code under test did (instead checking that pathlib library behavior is valid.)

This is a follow-up to e1d59105.

### DIFF
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+from pathlib import Path
 from typing import Optional
 import logging
 import os
@@ -144,8 +145,8 @@ def report_hpu_device(args_device: torch.device) -> None:
 
 def linux_train(
     ctx: click.Context,
-    train_file: str,
-    test_file: str,
+    train_file: Path,
+    test_file: Path,
     model_name: str,
     num_epochs: Optional[int] = None,
     device: torch.device = torch.device("cpu"),
@@ -175,9 +176,11 @@ def linux_train(
 
     print("LINUX_TRAIN.PY: LOADING DATASETS")
     # Get the file name
-    train_dataset = load_dataset("json", data_files=train_file, split="train")
+    train_dataset = load_dataset(
+        "json", data_files=os.fspath(train_file), split="train"
+    )
 
-    test_dataset = load_dataset("json", data_files=test_file, split="train")
+    test_dataset = load_dataset("json", data_files=os.fspath(test_file), split="train")
     train_dataset.to_pandas().head()
 
     tokenizer = AutoTokenizer.from_pretrained(model_name, trust_remote_code=True)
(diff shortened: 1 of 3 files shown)


booxter:
@leseb @russellb

booxter:
We should probably have better test coverage for this flow. Evidently I was able to break training flow with no test catching it...

src/instructlab/train/linux_train.py
```diff
@@ -144,8 +145,8 @@ def report_hpu_device(args_device: torch.device) -> None:
 
 def linux_train(
     ctx: click.Context,
-    train_file: str,
-    test_file: str,
+    train_file: Path,
```
leseb, line 148: Would having a str simplify it? Why do we need it to be a Path since the `load_dataset()` expects a str and we don't use `train_file` anywhere else?

booxter: I deem it better to keep type more tight. Path is a **semantically stronger** type (it represents a particular thing - a path to a file - and not just a string - which may be any string.)

leseb: Agreed `load_dataset` could potentially benefit from accepting `Path`. The semantically stronger type argument is appealing.

leseb (dismissed):

src/instructlab/train/linux_train.py
```diff
@@ -175,9 +176,9 @@ def linux_train(
 
     print("LINUX_TRAIN.PY: LOADING DATASETS")
     # Get the file name
-    train_dataset = load_dataset("json", data_files=train_file, split="train")
+    train_dataset = load_dataset("json", data_files=str(train_file), split="train")
```
tiran, line 179: `str(path_like)` is okay for human-readable representation, but not for converting something path-like to a filesystem representation. `os.fspath` does the right thing.

```suggestion
    train_dataset = load_dataset("json", data_files=os.fspath(train_file), split="train")
```

src/instructlab/train/linux_train.py
```diff
@@ -175,9 +176,9 @@ def linux_train(
...
 
-    test_dataset = load_dataset("json", data_files=test_file, split="train")
+    test_dataset = load_dataset("json", data_files=str(test_file), split="train")
```
tiran, line 181:
```suggestion
    test_dataset = load_dataset("json", data_files=os.fspath(test_file), split="train")
```

tiran (changes requested):

tiran (approved):

russellb (approved):

booxter:
@russellb FYI after mypy check merged, CI is broken; but this PR fixes it.

nathan-weinberg (approved):

PR #1271, opened 2024-06-05
ci: run e2e job when 2 approvals are given on a PR

leseb:
The follow-up process occurs in two stages.  First, Mergify triggers the end-to-end (e2e) workflow on a pull request (PR) after it receives two approvals.  Once the e2e workflow runs on the PR, if it succeeds, the workflow adds an 'e2e-success' label to the PR.  Mergify then has a new condition for auto-merging, which relies on the presence of the 'e2e-success' label on the PR.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -12,6 +12,7 @@ pull_request_rules:
     - label!=do-not-merge
     - label!=needs-rebase
     - check-success=DCO
+    - label=e2e-success # this will wait for the e2e workflow to pass, this label will be set by the workflow
 
     # If files are changed in .github/, the actionlint check must pass
     - or:
@@ -215,3 +216,18 @@ pull_request_rules:
     label:
       remove:
         - e2e-trigger
+
+- name: Trigger e2e workflow when two approvals are given
+  conditions:
+    - "#approved-reviews-by>=2"
+    - or:
+      - files~=.*\.py$
+      - files=scripts/basic-workflow-tests.sh
+      - files=.github/workflows/e2e.yml
+  actions:
+    github_actions:
+      workflow:
+        dispatch:
+          - workflow: e2e.yml
+            inputs:
+              pr_or_branch: "{{ number }}"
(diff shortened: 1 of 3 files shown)


leseb:
Adding 'hold' label to allow discussion on this new workflow.

russellb (changes requested):
This is a nice and simple solution, thanks!

I know it'll make the configuration more complicated, but I think we should limit which PRs this fork of the workflow applies to. Maybe it could be any PR that touches a Python source file, the e2e workflow file, or the e2e test script?

leseb:
> This is a nice and simple solution, thanks!
> 
> I know it'll make the configuration more complicated, but I think we should limit which PRs this fork of the workflow applies to. Maybe it could be any PR that touches a Python source file, the e2e workflow file, or the e2e test script?

Done, PTAL!

.github/mergify.yml
```diff
@@ -215,3 +216,22 @@ pull_request_rules:
...
+  conditions:
+    - "#approved-reviews-by>=2"
+    - -files~=tests/.*$ # do not trigger on .py test files
+    - and:
+      - or:
+        - files~=.*\.py$
+      - or:
+        - files=scripts/basic-workflow-tests.sh
+      - or:
+        - files=.github/workflows/e2e.yml
```
russellb, line 230:
```suggestion
    - or:
      - files~=.*\.py$
      - files=scripts/basic-workflow-tests.sh
      - files=.github/workflows/e2e.yml
```

.github/mergify.yml
```diff
@@ -215,3 +216,22 @@ pull_request_rules:
     label:
       remove:
         - e2e-trigger
+
+- name: Trigger e2e workflow when two approvals are given
+  conditions:
+    - "#approved-reviews-by>=2"
+    - -files~=tests/.*$ # do not trigger on .py test files
```
russellb, line 223: but this means if there's large functional change that includes tests, it will not run.  That doesn't seem quite right, either.

it would have to be something like changing the simple `*.py` check to be checking for python changes outside of that directory.  It'd be a more complicated regex, but doable I think.

.github/mergify.yml
```diff
@@ -12,6 +12,7 @@ pull_request_rules:
     - label!=do-not-merge
     - label!=needs-rebase
     - check-success=DCO
+    - label=e2e-success # this will wait for the e2e workflow to pass, this label will be set by the workflow
```
russellb, line 15: This will have to now be conditional as well using the same conditions as what triggers the e2e workflow

russellb (changes requested):

markstur:
I approve the idea, but currently has changes requested/needed and also fails Lint...

leseb:
Closing in favor of https://github.com/instructlab/instructlab/pull/1340

PR #1274, opened 2024-06-05
Add mypy github action matcher

booxter:


booxter:
@bjhargrave added the matcher. You can see the result here: https://github.com/instructlab/instructlab/pull/1274/files is it what you'd like to see? Let me know; I will drop the top patch then and then we can merge. Thanks!

.github/workflows/lint.yml
```diff
@@ -68,11 +68,15 @@ jobs:
         run: |
           echo "::add-matcher::.github/workflows/matchers/pylint.json"
           tox -e lint
+          echo "::remove-matcher owner=pylint-error::"
```
booxter, line 71: I'm not sure if we have to remove them across separate steps. Maybe someone could confirm.

bjhargrave: I think removing the matcher is safest. It can avoid the previous matcher also matching the errors.

bjhargrave:
> is it what you'd like to see?

Yes, it looks as expected! Thanks for adding.

booxter:
@bjhargrave ack. I removed the breaking patch, this should be ready now.

.github/workflows/lint.yml
booxter: Ack. I am not sure the state of matchers survive across steps; but I guess it won't make it worse.

leseb (commented):
I know you don't mind separating commits so how about one commit for removing the pylint matcher and one for adding mypy matcher? Thanks!

.github/workflows/lint.yml
bjhargrave, line 71: We want mypy to run even when lint fails. But this only removes the pylint matcher when lint succeeds. We probably need to move the remove-matcher to a separate step that runs even when lint fails or into the mypy step.

booxter: They do survive as proven here: https://github.com/instructlab/instructlab/pull/1286/files

Will update as you suggested.

bjhargrave (approved):

russellb (approved):

leseb (approved):

PR #1275, opened 2024-06-05
mypy: tackle all violations in excluded files

booxter:
Next steps will be: adding more type annotations to existing functions; enabling stricter validations (--strict), revisiting existing type annotations and maybe tightening where possible (str -> Path, fewer Unions etc.)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -45,3 +45,4 @@ wandb>=0.16.4
 # the below library should NOT be imported into any python files
 # it is for CLI usage ONLY
 yamllint>=1.35.1
+referencing>=0.35.0
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -5,7 +5,7 @@ from functools import cache, wraps
 from importlib import resources
 from importlib.abc import Traversable
 from pathlib import Path
-from typing import Any, Dict, List, Mapping, Optional, Union
+from typing import Any, List, Mapping, Optional, TypedDict
 from urllib.parse import urlparse
 import copy
 import glob
@@ -16,10 +16,10 @@ import platform
 import re
 import subprocess
 import tempfile
-import typing
 
 # Third Party
 from git import Repo, exc
+from referencing import Resource
 import click
 import git
 import gitdb
@@ -28,10 +28,6 @@ import yaml
 
 # Local
 from . import common
-
-if typing.TYPE_CHECKING:
-    # Third Party
-    import referencing
 
 logger = logging.getLogger(__name__)
 
@@ -189,8 +185,14 @@ def get_taxonomy_diff(repo="taxonomy", base="origin/main"):
     return updated_taxonomy_files
 
 
+class SourceDict(TypedDict):
+    repo: str
+    commit: str
+    patterns: List[str]
+
+
 def get_documents(
-    source: Dict[str, Union[str, List[str]]],
+    source: SourceDict,
     skip_checkout: bool = False,
 ) -> List[str]:
     """
@@ -202,9 +204,9 @@ def get_documents(
     Returns:
          List[str]: List of document contents.
     """ ""
-    repo_url = source.get("repo")
-    commit_hash = source.get("commit")
-    file_patterns = source.get("patterns")
+    repo_url = source.get("repo", "")
+    commit_hash = source.get("commit", "")
+    file_patterns = source.get("patterns", [])
     with tempfile.TemporaryDirectory() as temp_dir:
         try:
             repo = git_clone_checkout(
@@ -251,7 +253,7 @@ def get_sysprompt(model=None):
 
 
 @cache
-def _load_schema(path: Traversable) -> "referencing.Resource":
+def _load_schema(path: Traversable) -> Resource:
     """Load the schema from the path into a Resource object.
 
     Args:
@@ -265,7 +267,6 @@ def _load_schema(path: Traversable) -> "referencing.Resource":
     """
     # pylint: disable=import-outside-toplevel
     # Third Party
-    from referencing import Resource
     from referencing.exceptions import NoSuchResource
     from referencing.jsonschema import DRAFT202012
 
@@ -275,7 +276,7 @@ def _load_schema(path: Traversable) -> "referencing.Resource":
             contents=contents, default_specification=DRAFT202012
         )
     except Exception as e:
-        raise NoSuchResource(ref=str(path)) from e
+        raise NoSuchResource(str(path)) from e
     return resource
 
 
@@ -295,7 +296,7 @@ def validate_yaml(contents: Mapping[str, Any], taxonomy_path: Path) -> int:
     # Third Party
     from jsonschema.protocols import Validator
     from jsonschema.validators import validator_for
-    from referencing import Registry, Resource
+    from referencing import Registry
     from referencing.exceptions import NoSuchResource
 
     errors = 0
@@ -304,7 +305,9 @@ def validate_yaml(contents: Mapping[str, Any], taxonomy_path: Path) -> int:
 
     def retrieve(uri: str) -> Resource:
         path = schemas_path.joinpath(uri)
-        return _load_schema(path)
+        # This mypy violation will be fixed in:
+        # https://github.com/instructlab/schema/pull/33
+        return _load_schema(path)  # type: ignore[arg-type]
 
     schema_name = taxonomy_path.parts[0]
     if schema_name not in TAXONOMY_FOLDERS:
@@ -318,7 +321,10 @@ def validate_yaml(contents: Mapping[str, Any], taxonomy_path: Path) -> int:
         schema = schema_resource.contents
         validator_cls = validator_for(schema)
         validator: Validator = validator_cls(
-            schema, registry=Registry(retrieve=retrieve)
+            # mypy doesn't understand attrs classes fields, see:
+            # https://github.com/python/mypy/issues/5406
+            schema,
+            registry=Registry(retrieve=retrieve),  # type: ignore[call-arg]
         )
 
         for validation_error in validator.iter_errors(contents):
@@ -345,14 +351,10 @@ def validate_yaml(contents: Mapping[str, Any], taxonomy_path: Path) -> int:
 
 
 def get_version(contents: Mapping) -> int:
-    version = contents.get("version", 1)
-    if not isinstance(version, int):
-        # schema validation will complain about the type
-        try:
-            version = int(version)
-        except ValueError:
-            version = 1  # fallback to version 1
-    return version
+    try:
+        return int(contents.get("version", 1))
+    except (ValueError, TypeError):
+        return 1  # fallback to version 1
 
 
 # pylint: disable=broad-exception-caught
@@ -360,31 +362,31 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
     seed_instruction_data = []
     warnings = 0
     errors = 0
-    file_path = Path(file_path).resolve()
+    file_path_p = Path(file_path).resolve()
     # file should end with ".yaml" explicitly
-    if file_path.suffix != ".yaml":
+    if file_path_p.suffix != ".yaml":
         logger.warning(
-            f"Skipping {file_path}! Use lowercase '.yaml' extension instead."
+            f"Skipping {file_path_p}! Use lowercase '.yaml' extension instead."
         )
         warnings += 1
         return None, warnings, errors
-    for i in range(len(file_path.parts) - 1, -1, -1):
-        if file_path.parts[i] in TAXONOMY_FOLDERS:
-            taxonomy_path = Path(*file_path.parts[i:])
+    for i in range(len(file_path_p.parts) - 1, -1, -1):
+        if file_path_p.parts[i] in TAXONOMY_FOLDERS:
+            taxonomy_path = Path(*file_path_p.parts[i:])
             break
     else:
-        taxonomy_path = file_path
+        taxonomy_path = file_path_p
     # read file if extension is correct
     try:
-        with open(file_path, "r", encoding="utf-8") as file:
+        with open(file_path_p, "r", encoding="utf-8") as file:
             contents = yaml.safe_load(file)
         if not contents:
-            logger.warning(f"Skipping {file_path} because it is empty!")
+            logger.warning(f"Skipping {file_path_p} because it is empty!")
             warnings += 1
             return None, warnings, errors
         if not isinstance(contents, Mapping):
             logger.error(
-                f"{file_path} is not valid. The top-level element is not an object with key-value pairs."
+                f"{file_path_p} is not valid. The top-level element is not an object with key-value pairs."
             )
             errors += 1
             return None, warnings, errors
@@ -401,7 +403,7 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
                         "parsable",
                         "-c",
                         yaml_rules,
-                        file_path,
+                        str(file_path_p),
                         "-s",
                     ]
                 else:
@@ -412,7 +414,7 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
                         "parsable",
                         "-d",
                         DEFAULT_YAML_RULES,
-                        file_path,
+                        str(file_path_p),
                         "-s",
                     ]
             else:
@@ -422,17 +424,17 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
                     "parsable",
                     "-d",
                     DEFAULT_YAML_RULES,
-                    file_path,
+                    str(file_path_p),
                     "-s",
                 ]
             try:
                 subprocess.check_output(yamllint_cmd, text=True)
-            except subprocess.SubprocessError as e:
-                lint_messages = [f"Problems found in file {file_path}"]
+            except subprocess.CalledProcessError as e:
+                lint_messages = [f"Problems found in file {file_path_p}"]
                 parsed_output = e.output.splitlines()
                 for p in parsed_output:
                     errors += 1
-                    delim = str(file_path) + ":"
+                    delim = str(file_path_p) + ":"
                     parsed_p = p.split(delim)[1]
                     lint_messages.append(parsed_p)
                 logger.error("\n".join(lint_messages))
@@ -451,7 +453,7 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
             documents = get_documents(source=documents)
             logger.debug("Content from git repo fetched")
 
-        for seed_example in contents.get("seed_examples"):
+        for seed_example in contents.get("seed_examples", []):
             question = seed_example.get("question")
             answer = seed_example.get("answer")
             context = seed_example.get("context", "")
@@ -467,7 +469,7 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
             )
     except Exception as e:
         errors += 1
-        raise TaxonomyReadingException(f"Exception {e} raised in {file_path}") from e
+        raise TaxonomyReadingException(f"Exception {e} raised in {file_path_p}") from e
 
     return seed_instruction_data, warnings, errors
 
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -82,7 +82,7 @@ def is_macos_with_m_chip():
         result = subprocess.check_output(["sysctl", "-a"], text=True)
         is_m_chip = "machdep.cpu.brand_string: Apple" in result
         return is_m_chip
-    except subprocess.SubprocessError:
+    except subprocess.CalledProcessError:
         return False
 
 
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -393,8 +393,7 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
         version = get_version(contents)
         if version > 1:  # no linting for version 1 yaml
             if yaml_rules is not None:
-                is_file = os.path.isfile(yaml_rules)
-                if is_file:
+                if os.path.isfile(yaml_rules):
                     logger.debug(f"Using YAML rules from {yaml_rules}")
                     yamllint_cmd = [
                         "yamllint",
@@ -476,8 +475,7 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
 # TODO: remove `_logger` parameter after instructlab.sdg is fixed.
 def read_taxonomy(_logger, taxonomy, taxonomy_base, yaml_rules):
     seed_instruction_data = []
-    is_file = os.path.isfile(taxonomy)
-    if is_file:  # taxonomy is file
+    if os.path.isfile(taxonomy):
         seed_instruction_data, warnings, errors = read_taxonomy_file(
             taxonomy, yaml_rules
         )
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -378,7 +378,7 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
         taxonomy_path = file_path_p
     # read file if extension is correct
     try:
-        with open(file_path_p, "r", encoding="utf-8") as file:
+        with file_path_p.open(encoding="utf-8") as file:
             contents = yaml.safe_load(file)
         if not contents:
             logger.warning(f"Skipping {file_path_p} because it is empty!")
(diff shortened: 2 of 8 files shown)


booxter:
The remaining violations in the file are tackled in https://github.com/instructlab/instructlab/pull/1287 Both PRs combined should allow us to enable mypy for this module.

leseb (dismissed):

booxter:
I'm collecting all relevant fixes for mypy here now. Seems counterproductive to split these in tiny PRs.

booxter:
Converted to draft so that reviewers are aware this is ongoing work.

booxter:
This is ready for review.

leseb (dismissed):
Very nice, meticulous work! Congrats on clearing that list of violations!

src/instructlab/data/generator/generate_data.py
```diff
@@ -4,7 +4,7 @@
 from datetime import datetime
 from functools import partial
 from pathlib import Path
-from typing import Optional
+from typing import List, Optional, TypedDict
```
bjhargrave, line 7: I am not sure I would bother changing files in this package since it is moving to the sdg repo. For now, I would leave the package in the exclude list.

bjhargrave (commented):
There seems a lot of new code and deleted code in this PR which seems beyond linting fixes. I am unsure if those changes are safe.

booxter:
> There seems a lot of new code and deleted code in this PR which seems beyond linting fixes. I am unsure if those changes are safe.

which specific cases do you mean? there was some dead code removed (that violates mypy). Beyond that, I don't think I touched the code.

src/instructlab/data/generator/generate_data.py
booxter: We can do that too. I don't know the timeline of the migration of the code. We can move these changes to sdg (as part of mypy enablement there.) If it's the only concern, I can update the PR. But I'd like to see if any other specific issues are of concern.

src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
```diff
@@ -1613,15 +1577,15 @@ def do_dump_model(model_plus: ModelPlus) -> None:
 
 
 def convert_llama_to_gguf(
-    model: Path,
```
bjhargrave, line 1616: > which specific cases do you mean?

A param name change can break callers.

src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
```diff
@@ -1151,12 +1121,6 @@ def add_meta_arch(self, params: Params) -> None:
         self.gguf.add_head_count(params.n_head)
         self.gguf.add_head_count_kv(params.n_head_kv)
 
-        if params.n_experts:
```
bjhargrave, line 1154: > which specific cases do you mean?

This code is deleted. And above params.n_experts was deleted.

booxter:
btw I split the PR into small commits, so we can discuss each of them individually. If you think a particular change is invalid, please point to the commit with a link so we discuss the same thing.

src/instructlab/mlx_explore/gguf_convert_to_mlx.py
```diff
@@ -75,11 +64,7 @@ def __init__(self, args: ModelArgs):
         self.k_proj = nn.Linear(dim, n_kv_heads * head_dim, bias=False)
         self.v_proj = nn.Linear(dim, n_kv_heads * head_dim, bias=False)
         self.o_proj = nn.Linear(n_heads * head_dim, dim, bias=False)
-        rope_scale = (
```
bjhargrave, line 78: > which specific cases do you mean?

This is now just 1 since rope_scale was removed from ModelArgs.

src/instructlab/mlx_explore/gguf_convert_to_mlx.py
```diff
@@ -64,7 +51,9 @@ def __init__(self, args: ModelArgs):
 
         dim = args.hidden_size
         self.n_heads = n_heads = args.num_attention_heads
-        self.n_kv_heads = n_kv_heads = args.num_key_value_heads
+        self.n_kv_heads = n_kv_heads = (
+            args.num_key_value_heads or self.num_attention_heads
```
bjhargrave, line 55: > which specific cases do you mean?
 Not sure why this `or` was added or if the proper value is on the right hand side?

bjhargrave (commented):
A couple of the code changes I was referring to.

src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
booxter: Let me double check. I *think* that this code was dead because of how params are loaded, but I will confirm.

src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
booxter: That's why the patch changes the callers where needed to adjust.

The reason for the change in variable name is because the function later reuses the same name to store a different type, which is a warning in mypy. (You are not advised to mix types like this.)

booxter: An alternative here can be a rename of the local variable instead. I don't have huge preference.

src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
booxter: OK, the explanation as to why these two options are removed is here: 76410ae05cd81aeee305a48ed83da373d0ec28bf To quote,

```
    mypy: Remove use of undefined GGUFWriter symbols

    Neither add_expert_count nor add_expert_used_count are defined in the
    version of gguf library we use.  Actually, these symbols are not yet
    released on pypi, and are available in git repo of the library only:

    https://github.com/ggerganov/llama.cpp/commit/799a1cb13b0b1b560ab0ceff485caed68faa8f1f

    This makes mypy complain about it.

    This patch removes code related to these parameters. If and when we
    upgrade to gguf version that supports them, we may revert this patch.
```

Since GGUFWriter doesn't have the functions, passing these settings through config file would crash the code anyway. In the future, gguf may be updated to include these functions, then this may be revisited.

Let me know if this line of thought is ok for you.

src/instructlab/data/generator/generate_data.py
booxter: It's unclear if you insist, or it's an opinion. For now, I will leave these changes included. But please let me know. (If so, could you please list the exact modules that are going to be migrated - or is the list captured somewhere?)

booxter:
The latest is just a merge from `main`. No significant changes to the patch series itself.

src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
bjhargrave: I guess I am worried were are changing 3rd party files just to make mypy happy. Perhaps we should exclude these 3rd party source files (e.g. this and the mlx_explore folder) from mypy checks.

booxter: Depends on how you look at it. It's not just to make `mypy` happy: this patch removes a bug that results in a crash depending on what's in the config file. With this patch included, the unknown options will be silently ignored.

I was wondering about the nature of this file. The way these files are managed in this tree is ambiguous. On one hand, they clearly come from `llamacpp` repo. On the other hand, it's not pulled in through a library (which makes sense - since this code is NOT part of their package); and we don't track which particular version of the file we pulled here. (I actually tried to figure this out - and I suspect what we have here in-tree is a **modified** version of one of older versions of the `llamacpp` example file.)

We already fixed some bugs in this file, removed `click` code, and `blackened` / `ruffed` the file. (Just as examples. There are more.) If the decision is to not touch the file (and maybe some other files - which are these?), then we should document this (there's a README uder this path, we could update it.) Ideally, we would also document where this code comes from then (the exact commit message), or maybe even try to pull it through git submodules. (Not sure it's possible at this point though - the code diverged.)

(My personal opinion is that this would not be a good decision to make, because this code already diverged from wherever we took it from. But if the team feels otherwise, we could try to freeze new changes to the module, somehow, at least as a review policy.)

src/instructlab/data/generator/generate_data.py
booxter: These are gone now in the latest tree.

src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
booxter: @bjhargrave all patches for the modules you identified (plus one more) that are copy-pasted are dropped, series flattened. This should be less controversial, I think.

src/instructlab/utils.py
```diff
@@ -354,7 +361,9 @@ def _load_schema(path: "importlib.resources.abc.Traversable") -> "referencing.Re
             contents=contents, default_specification=DRAFT202012
         )
     except Exception as e:
-        raise NoSuchResource(ref=str(path)) from e
+        # mypy doesn't understand attrs classes fields, see:
+        # https://github.com/python/mypy/issues/5406
+        raise NoSuchResource(ref=str(path)) from e  # type: ignore[call-arg]
```
bjhargrave, line 366:
```suggestion
        raise NoSuchResource(str(path)) from e
```

This will make mypy happy.

src/instructlab/utils.py
```diff
@@ -330,11 +338,11 @@ def get_sysprompt(model=None):
 
 
 @cache
-def _load_schema(path: "importlib.resources.abc.Traversable") -> "referencing.Resource":
+def _load_schema(path: Path) -> Resource:
```
bjhargrave, line 341: This really is a Traversable. It just moved in Python 3.11.

```python
try:
    from importlib.resources.abc import Traversable  # type: ignore[import-not-found]
except ImportError:  # python<3.11
    from importlib.abc import Traversable
```

We should use the proper type rather than coerce it to Path.

requirements.txt
```diff
@@ -46,3 +46,4 @@ tenacity>=8.3.0,!=8.4.0
 # the below library should NOT be imported into any python files
 # it is for CLI usage ONLY
 yamllint>=1.35.1
+referencing>=0.35.0,<0.36.0
```
bjhargrave, line 49: This should be brought in by `jsonschema`. Not sure why it should be explicit here.

src/instructlab/utils.py
```diff
@@ -388,7 +394,8 @@ def validate_yaml(contents: Mapping[str, Any], taxonomy_path: Path) -> int:
 
     def retrieve(uri: URI) -> Resource:
         path = schemas_path.joinpath(uri)
-        return _load_schema(path)
+        with resources.as_file(path) as file:
```
bjhargrave, line 397: Why do this? It is not necessary except to satisfy the change to use Path from Traversable above.

src/instructlab/utils.py
```diff
@@ -536,7 +545,7 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
             documents = get_documents(source=documents)
             logger.debug("Content from git repo fetched")
 
-        for seed_example in contents.get("seed_examples"):
+        for seed_example in contents.get("seed_examples", []):
```
bjhargrave, line 548: Not sure why this is necessary? The schema validation has already made sure the key is present.

bjhargrave (changes requested):
Much simpler with the removal of the copied files!

requirements.txt
booxter: I assume that if it's directly imported, then it's a direct dependency.

src/instructlab/utils.py
booxter: You can see all reasons for each change documented in commit messages, e.g. here for this change: https://github.com/instructlab/instructlab/pull/1275/commits/344f697b739f12771b9e1a6de4b846b775a04805

The reason why this has to change to Path is because we apply LRU cache decorator that expects hashable arguments, and Traversable is not hashable.

src/instructlab/utils.py
booxter: Yes, it's to satisfy the other change, which is needed to fix type violation where `@cache` decorator is applied to a function that receives (potentially) non-hashable arguments (`Traversable`). Both changes belong to the same commit.

src/instructlab/utils.py
booxter: Yes, though this fact is not known to type checking layer. `mypy` reasons that the field **may** be missing (e.g. a bug in schema validation module may expose it.)

leseb (approved):
Thanks a lot for taking care of this.

tiran:
I forgot about this PR...

A few suggestions:
- We dropped support for older versions of Python and require at least 3.10. You can now use `str | None` instead of `Optional[str]`. I'm appending a Ruff configuration change that lets you auto-fix the code.
- IMHO the PR is fixing too much at once. I really, really want proper type information in InstructLab. But a big patch is harder to land and disruptive for others. Do you have an idea how to split the PR into smaller chunks? Maybe address one module at a time?
- InstructLab mixes and matches str and pathlib.Path for paths. We should use `pathlib.Path` consistently for all paths and start with CLI args. Clicks `click.Path` type does not return a `pathlib.Path`. I created a custom `ClickPath` object for Fromager. Please take a look at https://github.com/python-wheel-build/fromager/pull/112
- Personal opinion: I don't like `from pathlib import Path` or `from typing import Any`. I prefer `import pathlib` and `pathlib.Path` for readability. Can I persuade you to avoid from imports? :) 

<details>
<summary>Auto-fixer for type annotations</summary>

```
diff --git a/pyproject.toml b/pyproject.toml
index 3484a10..fe441bf 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ include = [
 ignore = ["W002"]
 
 [tool.ruff]
-target-version = "py39"
+target-version = "py310"
 # same as black's default line length
 line-length = 88
 
@@ -76,16 +76,21 @@ line-length = 88
 # Allow fix for all enabled rules (when `--fix`) is provided.
 fixable = ["ALL"]
 unfixable = []
+# modern type annotations
+extend-safe-fixes = ["UP006", "UP007"]
 
 # Fixers will be enabled gradually.
 select = [
   # "B",  # flake8-bugbear
   # "E",  # pycodestyle
   # "F",  # Pyflakes
+  "F401", # unused imports
   "Q", # flake8-quotes
   # Ruff does not support isort's import_headings feature, yet.
   # "I",  # isort
   # "UP",  # pyupgrade
+  "UP006",
+  "UP007",
   # "SIM",  # flake8-simplify
   "TID", # flake8-tidy-imports
 ]
```

</details>

booxter:
@tiran I appreciate the feedback.

The series started as separate PR per module patched, but these never went anywhere, so I figured that at least once I get some attention from reviewers, I will have a chance to land it before yet another git conflict. Technically, it's simple to split these, since I already split each change into a separate commit. But I don't see how this will get us anywhere, except into months of struggling for reviewer attention.

Re: Optional vs None, from xxx import yyy, and path types - we can follow up. I want to follow up with path type switches, but I am holding it off for when we enable the checks for the whole tree and stop adding more excludes to the list instead of fixing legit failures. (Which seems to be a habit here.) I don't see how squashing more changes like this here will get us quicker to where we could at least start enforcing "no new mypy violations" for new patches.

requirements.txt
tiran: Don't pin upper version unless a higher version introduces a known breaking change.

src/instructlab/utils.py
```diff
@@ -304,7 +308,8 @@ def validate_yaml(contents: Mapping[str, Any], taxonomy_path: Path) -> int:
 
     def retrieve(uri: URI) -> Resource:
```
tiran, line 309: I think this should be a normal `str`.

src/instructlab/utils.py
```diff
@@ -360,31 +368,31 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
...
+            f"Skipping {file_path_p}! Use lowercase '.yaml' extension instead."
         )
         warnings += 1
         return None, warnings, errors
-    for i in range(len(file_path.parts) - 1, -1, -1):
-        if file_path.parts[i] in TAXONOMY_FOLDERS:
-            taxonomy_path = Path(*file_path.parts[i:])
+    for i in range(len(file_path_p.parts) - 1, -1, -1):
+        if file_path_p.parts[i] in TAXONOMY_FOLDERS:
+            taxonomy_path = Path(*file_path_p.parts[i:])
```
tiran, line 381: Not your fault, but oh boy, these three lines are unreadable.

src/instructlab/utils.py
```diff
@@ -360,31 +368,31 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
...
             break
     else:
-        taxonomy_path = file_path
+        taxonomy_path = file_path_p
     # read file if extension is correct
     try:
-        with open(file_path, "r", encoding="utf-8") as file:
+        with open(file_path_p, "r", encoding="utf-8") as file:
```
tiran, line 387:
```suggestion
        with file_path_p.open(encoding="utf-8") as file:
```

src/instructlab/utils.py
tiran: This code can break the cache under some conditions. `resource.as_file` can point to a random temporary file instead of a stable file name.
I recommend to leave the code as it for now. Schema loading and validation should be moved to the `instructlab.schema` package. The package should use an explicit cache with key cache `(version, basename)`.

leseb (approved):
@booxter thanks for your patience while working on this. I'd like to move forward with this, it has been around for quite some time and the constant rebase/conflict makes it painful I'm sure. I think @tiran has small nits but we should be able to sort it out rapidly. Thanks again.

tiran:
Another nit pick is: This PR has 35 commits. InstructLab project does not use squash commits. I don't think it's a good idea to add 35 commits + 1 merge commit to the git history of main branch. It's way too noisy.

src/instructlab/utils.py
```diff
@@ -260,7 +268,6 @@ def _load_schema(path: "importlib.resources.abc.Traversable") -> "referencing.Re
     """
     # pylint: disable=import-outside-toplevel
     # Third Party
-    from referencing import Resource
     from referencing.exceptions import NoSuchResource
     from referencing.jsonschema import DRAFT202012
```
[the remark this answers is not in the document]
booxter: I am happy to do it. But this would have nothing to do with mypy enablement. I will make a note to follow up with cleaning up these pylint exceptions separately.

src/instructlab/utils.py
bjhargrave: I have fixed this in https://github.com/instructlab/schema/pull/33.

That PR is to centralize the taxonomy parsing code into the instruct lab-schema package which can be used by other package which need to parse qna.yaml files.

src/instructlab/utils.py
booxter: I'll replace with line-specific ignore comment for now.

booxter:
> I don't think it's a good idea to add 35 commits + 1 merge commit to the git history of main branch.

Elsewhere (@russellb ?) it was suggested that splitting in smaller pieces if they are logically separate (and they are here) is the right way of doing it. (And that's how I build my series in other projects over years.) I doubt you want to squash all of them into a one, but please confirm.

At the end of the day, it's not hard to squash the series and force-push. I just find it counterproductive to lose the context of the changes when dropping separate commit messages.

src/instructlab/utils.py
```diff
@@ -287,14 +294,11 @@ def validate_yaml(contents: Mapping[str, Any], taxonomy_path: Path) -> int:
...
-    # pylint: disable=import-outside-toplevel
-    # Standard
-    from importlib import resources
-
     # Third Party
+    # pylint: disable=C0415
     from jsonschema.protocols import Validator
     from jsonschema.validators import validator_for
-    from referencing import Registry, Resource
+    from referencing import Registry
```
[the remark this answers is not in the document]
booxter: I will send a separate pylint cleanup PR later to handle these.

leseb:
> > I don't think it's a good idea to add 35 commits + 1 merge commit to the git history of main branch.
> 
> Elsewhere (@russellb ?) it was suggested that splitting in smaller pieces if they are logically separate (and they are here) is the right way of doing it. (And that's how I build my series in other projects over years.) I doubt you want to squash all of them into a one, but please confirm.
> 
> At the end of the day, it's not hard to squash the series and force-push. I just find it counterproductive to lose the context of the changes when dropping separate commit messages.

💯

src/instructlab/utils.py
booxter: Strictly speaking, it is not related to mypy, but I will add this to the series to speed up review.

booxter:
Changes since the previous version:

- dropped the patch that attempted to handle non-hashable object used as key to LRU cache (replaced with ignore comment);
- added a patch to switch from open(Path) to Path.open();
- removed upper-cap for referencing version in requirements.txt.

tiran:
> At the end of the day, it's not hard to squash the series and force-push. I just find it counterproductive to lose the context of the changes when dropping separate commit messages.

Personally I prefer to keep separate commits in the PR but land each PR as one commit in the git commit history. This approach ensures a clean git history, but keeps the history of a change in the PR. With merge commits, every commit works, has passed the tests, and allows to perform git bisect + test.

Our approach with several commits + merge commit for each PR can create a broken git history, because any of the 3, 5, or 30 commits in a PR can cause a breakage. For example commit 2 breaks a test, commit 4 fixes it, and commits 1-5 land in the git history. This makes `git bisect` for bug hunting impossible, because commit 2 and 3 introduce a false positive error. The article https://danielrotter.at/2023/10/28/git-bisect-and-the-importance-of-a-clean-history.html explains the concept of git bisect and clean history in more words.

InstructLab decided to go for merge commits with several sub commits. I have to accept it, but I don't have to like it. I pity the maintainer that has to bisect a complex bug in the future...

booxter:
I can only agree that, ideally, each commit would be CI'd. But that's something that the review / CI platform the project chose supports. (E.g. gerrit doesn't have this problem since each patch is validated.)

As I said, I can oblige to whatever the decision of the core team is. I hold off because I get inconsistent signals.

---

BTW since we don't gate pre-merge, the repo will have breaking patches merged regardless (though fewer). One would have to build a serial validation/merge pipeline to avoid this.

leseb (dismissed):

tiran (commented):
The changes are fine.

I still think it is too noisy to land these changes as a 38 separate commits. It's way too noisy in the git history. IMO we don't need fine-grained history for code refactoring. After all these changes should not change behavior.

leseb (approved):
@booxter has been very patient with this large change, and the constant rebasing must have been painful. The PR has no conflicts, so I think we should free him now :slightly_smiling_face: It’s been a while already. Honestly, when I look at all the questionable commits that get merged these days, I cringe every time I check the commit history. @booxter did a great job with well-formatted commit messages. I don’t think we should be blocking him on the number of commits. Thanks!

tiran:
I like to add another data point. I haven't mentioned this before, but I did not want to insinuate foul play. 

As an open source project, we welcome contributions from everybody. In return, we recognize and celebrate the work from our contributors. For some contributors, metrics such as contribution and activity statistics are important. I fear that merging 38 separate commits will skew the statistics and set bad precedence. To put the number in relation: Russel is currently the top contributor with 89 commits in total (main branch, excluding merge commits). Places 9 has 38 commits and place 10 has 33 commits.

To be clear, I don't think that the 38 commits in this PR were created out of malicious intent. You are doing a great job. I really want to have your fixes in, so we can improve our typing and code quality even further.

I hope my explanation shows that there are not only technical issues (bisect, bloated history, hard to revert), but also a human issue.

With this new information, would you be willing to compromise and squash your PR into a smaller number of commits? I'm no longer insisting on one commit, but would prefer something along the line of half a dozen to a dozen commits.

Given your technical skills and experience with typing, I would also like to nominate you to become a triager and eventually a maintainer -- if you are interested. I'll have to talk to some people and figure out how to initiate the process.

booxter:
@tiran don't get me wrong, I am not frustrated about any particular rule or policy (though of course I have my opinions), I just want to see some clarity as to what is actually considered "the right way" in this project. Signals so far are diametrically different. :) This is a bit frustrated indeed. (Honestly, I didn't think this PR will get so contentious.)

I understand people watch stats. Ideally, people don't use these numbers to celebrate or judge others' impact. (Which is e.g. minuscule here.) But if this keeps people happy, so be it. I merged most of mypy patches into one; left unrelated changes outside; only one mypy specific patch is left separated because it's not run-of-the-mill one.

Let me know if this is enough.

bjhargrave (approved):

tiran (approved):

src/instructlab/utils.py
nathan-weinberg: Looking at this now and boy do I concur

PR #1276, opened 2024-06-06
[DRAFT] Add vllm backend for ilab serve

### LINKED ISSUE #1106: ilab should support multiple serving backends
ilab currently uses llama as the main backend to serve the models. llama doesn't have as much extensive GPU and multi GPU support. 

in developing some new tooling, custom vllm support needed to be quickly developed. ilab should natively be able to support vllm as a backend and other backends that exist. 

something like `--backend` when running `ilab serve` `chat` or `generate` would make sense since all of these commands have the ability to start a server.

This might need to be an enhancement. I am open to writing one if needed.

cc @jaideepr97

alimaredia:
# Changes

Resolves #1106

Initial support serving models with vllm as a backend. Further discussion on design is being tracked at https://github.com/instructlab/instructlab/issues/1106.

### DIFF
--- a/requirements-vllm.txt
+++ b/requirements-vllm.txt
@@ -1,0 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
+
+-r requirements.txt
+
+# newer vllm does not serve Merlinite
+vllm<0.4.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,13 +27,13 @@ PyYAML>=6.0.0,<7.0.0
 rich>=13.3.1,<14.0.0
 rouge-score>=0.1.2,<0.2.0
 sentencepiece>=0.2.0,<0.3.0
-tokenizers>=0.15.2,<0.16.0
+tokenizers>=0.19.1,<0.20.0
 toml>=0.10.2,<0.11.0
 # Habana installer has 2.2.0a0+git8964477 with oneMKL
 torch>=2.2.0a0,<3.0.0 ; python_version == '3.10'
 torch>=2.2.1,<3.0.0 ; python_version != '3.10'
 tqdm>=4.66.2,<5.0.0
-transformers>=4.30.0,<=4.38.2
+transformers>=4.41.2,<5.0.0
 trl>=0.7.11,<0.8.0
 wandb>=0.16.4,<0.17.0
 langchain-text-splitters
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -13,6 +13,7 @@ import os
 import shutil
 import sys
 import typing
+import subprocess
 
 # Third Party
 from click_didyoumean import DYMGroup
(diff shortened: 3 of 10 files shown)


requirements-vllm.txt
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
+
+-r requirements.txt
+
+# newer vllm does not serve Merlinite
+vllm<0.4.3
```
tiran, line 6: Instead of another requirements file, how about you use an optional dependency?

Add this to `pyproject.toml`
```toml
[project.optional-dependencies]
vllm = ["vllm<0.4.3"]
```
then you can install vllm support with `pip install instructlab[vllm]`

requirements.txt
```diff
@@ -27,13 +27,13 @@ PyYAML>=6.0.0,<7.0.0
 rich>=13.3.1,<14.0.0
 rouge-score>=0.1.2,<0.2.0
 sentencepiece>=0.2.0,<0.3.0
-tokenizers>=0.15.2,<0.16.0
+tokenizers>=0.19.1,<0.20.0
```
tiran, line 30: This break Intel Gaudi support. Is the version bump really required? Could you just add the higher version to the optional dependencies instead of main dependencies?

requirements.txt
```diff
@@ -27,13 +27,13 @@ PyYAML>=6.0.0,<7.0.0
...
 toml>=0.10.2,<0.11.0
 # Habana installer has 2.2.0a0+git8964477 with oneMKL
 torch>=2.2.0a0,<3.0.0 ; python_version == '3.10'
 torch>=2.2.1,<3.0.0 ; python_version != '3.10'
 tqdm>=4.66.2,<5.0.0
-transformers>=4.30.0,<=4.38.2
+transformers>=4.41.2,<5.0.0
```
tiran, line 36: Same problem, this version bump also breaks Intel Gaudi support.

src/instructlab/lab.py
```diff
@@ -310,61 +311,108 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
...
 @click.pass_context
-def serve(
-    ctx, model_path, gpu_layers, num_threads, max_ctx_size, model_family, log_file
-):
+def serve(ctx, model_path, log_file, llama_cpp_gpu_layers, llama_cpp_num_threads, llama_cpp_max_ctx_size, llama_cpp_model_family, use_vllm, vllm_chat_template, vllm_args):
     """Start a local server"""
-    # pylint: disable=C0415
-    # Local
-    from .server import ServerException, server
+    from pathlib import PurePath
```
tiran, line 357: Please move this to top-level import. Also why are you using `PurePath` instead of `Path` here? There is rarely a need to use `PurePath` except for special circumstances. See top level module docs of https://docs.python.org/3/library/pathlib.html

src/instructlab/lab.py
```diff
@@ -310,61 +311,108 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
...
-            ctx.obj.logger,
-            model_path,
-            gpu_layers,
-            max_ctx_size,
-            model_family,
-            num_threads,
-            host,
-            port,
+    if use_vllm or serve_safetensors:
+        import importlib.util
```
tiran, line 370: Move this to top-level import

src/instructlab/lab.py
```diff
@@ -310,61 +311,108 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
...
+
+        ctx.obj.logger.info(f"Serving backend is vllm")
+
+        vllm_found = importlib.util.find_spec("vllm")
+        if vllm_found is None:
+            ctx.obj.logger.info(f"ERROR: vllm not installed")
+            raise click.exceptions.Exit(1)
+
+        vllm_cmd = ["python", "-m",  "vllm.entrypoints.openai.api_server", "--model", model_path]
```
tiran, line 379: `python` may point to a different Python interpreter. `sys.executable` is the path to the current Python interpreter.
```suggestion
        vllm_cmd = [sys.executable, "-m",  "vllm.entrypoints.openai.api_server", "--model", model_path]
```

src/instructlab/lab.py
```diff
@@ -310,61 +311,108 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
...
+
+        if vllm_chat_template is not None:
+            vllm_cmd = vllm_cmd + vllm_chat_template.split()
```
tiran, line 382:
```suggestion
            vllm_cmd.extend(vllm_chat_template.split())
```

src/instructlab/lab.py
```diff
@@ -310,61 +311,108 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
...
+
+        if vllm_args is not None:
+            vllm_cmd = vllm_cmd + vllm_args.split()
```
tiran, line 385:
```suggestion
            vllm_cmd.extend(vllm_args.split())
```

src/instructlab/lab.py
```diff
@@ -310,61 +311,108 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
...
+
+        vllm_cmd_str = " ".join(vllm_cmd)
+        ctx.obj.logger.info(f"vllm serving command is: {vllm_cmd_str}")
+        subprocess.run(args=vllm_cmd)
```
tiran, line 389: You either need to check the status of the return object or use some function like `subprocess.check_call` to check whether the call fails.

src/instructlab/lab.py
```diff
@@ -310,61 +311,108 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
...
+    else:
+        # pylint: disable=C0415
```
tiran, line 391: Please use human-readable short names instead of number constants. They are easier to recognize.
```suggestion
        # pylint: disable=import-outside-toplevel
```

src/instructlab/lab.py
```diff
@@ -310,61 +311,108 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
...
+        # Local
+        from .server import ServerException, server
+
+        ctx.obj.logger.info(f"Serving backend is llama-cpp")
+        ctx.obj.logger.info(
+            f"Using model '{model_path}' with {llama_cpp_gpu_layers} gpu-layers and {llama_cpp_max_ctx_size} max context size."
```
tiran, line 397: It's considered bad practice to use f-strings with logging. You should use logging's string interpolation instead.

tiran (changes requested):

requirements.txt
tiran: I have filed https://github.com/huggingface/optimum/issues/1895 with optimum and asked to lift the upper version bound of transformers. This will also allow us to bump the tokenizers version.

leseb (changes requested):
Thanks for the initial shot, how about we freeze this for a moment and validate the design first? I'd like to avoid back and forth between the design change and the current code.

requirements.txt
tiran: I got feedback on my feature request. The next versions of optimum and optimum-habana will support `transformers >= 4.40.0, < 4.41.0`. In the mean time, we have to support `>=4.38.2`.

alimaredia:
closed for https://github.com/instructlab/instructlab/pull/1386

PR #1277, opened 2024-06-06
Log funcName instead of asctime.

makelinux:
Logging timetstamp is very noisy and is not very useful.

Logging function name is much more valuable.
To compensate lack of timestamp dump it on initialization.

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -3,6 +3,7 @@
 # pylint: disable=too-many-lines
 
 # Standard
+from datetime import datetime
 from glob import glob
 from os.path import basename, dirname, exists
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


tiran:
I find time stamp verify useful to understand the timing of events or to correlate events in other logs.

leseb (changes requested):
I think timestamps in logs are crucial during debugging sessions. I'd like to keep it.

makelinux:
Timestamps can be added precisely just before network operations.

leseb:
> Timestamps can be added precisely just before network operations.

It's not only about network operations.

tiran:
We could make the log format customizable. Or detect systemd service and use journald's time stamps instead.

PR #1278, opened 2024-06-06
Log funcName

makelinux:
Add a colon after `lineno` because it is a widely used convention since `grep`.


leseb (approved):
I'm fine with the change, I don't think the impact is huge given we don't have many systems parsing the logs. Also the `--log-file` option is quite recent. Thanks!

nathan-weinberg (approved):

PR #1287, opened 2024-06-06
mypy: Remove use of undefined GGUFWriter symbols

booxter:
Neither add_expert_count nor add_expert_used_count are defined in the version of gguf library we use.  Actually, these symbols are not yet released on pypi, and are available in git repo of the library only:

https://github.com/ggerganov/llama.cpp/commit/799a1cb13b0b1b560ab0ceff485caed68faa8f1f

This makes mypy complain about it.

This patch removes code related to these parameters. If and when we upgrade to gguf version that supports them, we may revert this patch.

### DIFF
--- a/src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
+++ b/src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
@@ -190,8 +190,6 @@ class Params:
     n_ff: int
     n_head: int
     n_head_kv: int
-    n_experts: int | None = None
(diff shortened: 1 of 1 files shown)


booxter:
Overall, this module seems to be a copy-pasted version of `convert.py` from gguf library repo. Shouldn't we reuse the code from the original library?.. @xukai92 you added this file here. What's the story behind it?

booxter:
Squashed this into https://github.com/instructlab/instructlab/pull/1275

PR #1297, opened 2024-06-07
Use correct model family for granite

### LINKED ISSUE #1281: [v0.16.0] Granite/Mistral models respond with nonesense/hallucinate
**Describe the bug**
Models like granite or mistral when served with the v0.16.0 hallucinate pretty badly when even prompted with a simple 'hello'. Both for the base and trained models.

**The exact same models work fine with v0.15.1**

**To Reproduce**
1. Serve https://huggingface.co/instructlab/granite-7b-lab-GGUF/blob/main/granite-7b-lab-Q4_K_M.gguf with the v.0.16.0
2. Chat and send a "hello"

**Expected behavior**
Model shouldn't hallucinate that badly

**Screenshots**

Screenshot 2024-06-06 at 14 40 23

 - Hardware Specs: M1, 32GB
 - OS Version: Sonoma 14.5
 - Python Version: Python 3.12.3
 - InstructLab Version: [output of `ilab --version`]

**Additional context**

russellb:
This change fixes a regression introduced in commit
bf332e84a3ddeaa4150bbeb6fb05648d6d0e8537 from PR #1174.

Prior to this commit, ilab would correctly use the `merlinte` model
family when using a granite model as that was the default. The change
in #1174 added some additional logic trying to guess the model family
based on filename to avoid having to specify `--model-family` in some
cases.

This change made the code automatically set the model family to
`granite`, which the code did not understand, which resulted in broken
model behavior.

The code has been updated to do the following:

- Understand that `granite` maps to `merlinite` for the model family

- When guessing a model family based on filename, don't assume the
  result is valid. Validate it against the set of known model families
  to `ilab` first.

- Fall back to the previous default of `merlinite` if the guess
  doesn't result in something `ilab` understands.

Fixes #1281

### DIFF
--- a/src/instructlab/config.py
+++ b/src/instructlab/config.py
@@ -38,6 +38,17 @@ DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=30.0)
 # use spawn start method, fork is not thread-safe
 DEFAULT_MULTIPROCESSING_START_METHOD = "spawn"
 
+# When otherwise unknown, ilab uses this as the default family
+DEFAULT_MODEL_FAMILY = "merlinite"
+
+# Model families understood by ilab
+MODEL_FAMILIES = set(("merlinite", "mixtral"))
+
+# Map model names to their family
+MODEL_FAMILY_MAPPINGS = {
+    "granite": "merlinite",
+}
+
 
 class ConfigException(Exception):
     """An exception that a configuration file has an error."""
@@ -197,4 +208,11 @@ def get_api_base(host_port):
 
 
 def get_model_family(forced, model_path):
-    return forced if forced else match(r"^\w*", path.basename(model_path)).group(0)
+    if forced and forced.lower() not in MODEL_FAMILIES:
+        raise ConfigException("Unknown model family: %s" % forced)
+
+    # Try to guess the model family based on the model's filename
+    guess = match(r"^\w*", path.basename(model_path)).group(0).lower()
+    guess = MODEL_FAMILY_MAPPINGS.get(guess, guess)
+
+    return guess if guess in MODEL_FAMILIES else DEFAULT_MODEL_FAMILY


russellb:
I validated the fix manually.

It's worth a follow-up to consider what test coverage should be added. If nothing else, a small unit test on `get_model_family()` is a start.

src/instructlab/config.py
```diff
@@ -38,6 +38,17 @@
 # use spawn start method, fork is not thread-safe
 DEFAULT_MULTIPROCESSING_START_METHOD = "spawn"
 
+# When otherwise unknown, ilab uses this as the default family
+DEFAULT_MODEL_FAMILY = "merlinte"
```
leseb, line 42:
```suggestion
DEFAULT_MODEL_FAMILY = "merlinite"
```

Did you mean `merlinite`?

src/instructlab/config.py
```diff
@@ -38,6 +38,17 @@
...
+
+# Model families understood by ilab
+MODEL_FAMILIES = set(("merlinite", "mixtral"))
+
+# Map model names to their family
+MODEL_FAMILY_MAPPINGS = {
```
leseb, line 48: I believe a tuple is more appropriate given the static nature of the mapping. How about:

```
MODEL_FAMILY_MAPPINGS = (
    ("granite", "merlinite"),
)
```

src/instructlab/config.py
```diff
@@ -197,4 +208,11 @@ def get_api_base(host_port):
...
 def get_model_family(forced, model_path):
-    return forced if forced else match(r"^\w*", path.basename(model_path)).group(0)
+    if forced and forced.lower() not in MODEL_FAMILIES:
+        raise ConfigException("Unknown model family: %s" % forced)
+
+    # Try to guess the model family based on the model's filename
+    guess = match(r"^\w*", path.basename(model_path)).group(0)
+    guess = MODEL_FAMILY_MAPPINGS.get(guess, guess)
+
+    return guess if guess in MODEL_FAMILIES else DEFAULT_MODEL_FAMILY
```
leseb, line 218: Update accordingly if we agree to go with a tuple

leseb (changes requested):
Can we turn your "manual" validation into a functional test? Thanks!

russellb:
> Can we turn your "manual" validation into a functional test? Thanks!

Not easily ... I was doing `ilab chat` with a granite model and saying "hello!". The response was either something sensible like "Hello, how can I help you?" or absolute garbage, so the failure was obvious enough to my eyes.

Automatically determining the quality of the output of the model is much more complicated. :-)

src/instructlab/config.py
russellb: Yes - that explains the test failure, too. Thank you!

leseb:
> > Can we turn your "manual" validation into a functional test? Thanks!
> 
> Not easily ... I was doing `ilab chat` with a granite model and saying "hello!". The response was either something sensible like "Hello, how can I help you?" or absolute garbage, so the failure was obvious enough to my eyes.
> 
> Automatically determining the quality of the output of the model is much more complicated. :-)

Ah right, makes sense :)

src/instructlab/config.py
russellb: That seems more complicated.  Instead of a direct lookup on the key, we'd have to iterate over all items checking for a match.

src/instructlab/config.py
```diff
@@ -197,4 +208,11 @@ def get_api_base(host_port):
 
 
 def get_model_family(forced, model_path):
-    return forced if forced else match(r"^\w*", path.basename(model_path)).group(0)
+    if forced and forced.lower() not in MODEL_FAMILIES:
+        raise ConfigException("Unknown model family: %s" % forced)
+
+    # Try to guess the model family based on the model's filename
+    guess = match(r"^\w*", path.basename(model_path)).group(0)
```
russellb, line 215:
```suggestion
    guess = match(r"^\w*", path.basename(model_path)).group(0).lower()
```

russellb: added

hickeyma (commented):
Fix looks good eyeballing it. Just testing it manually now. Thanks for pushing this fix so quickly @russellb

nathan-weinberg:
@Mergifyio backport release-v0.16

src/instructlab/config.py
hickeyma: I'm ok with the way it is.

hickeyma (approved):
Tested manually and looks to have fixed the bug.

As for functional tests, lets do this in a separate PR if its going to hold up the fix. @russellb If not doing it here, can you raise an issue to remember it needs to be done?

src/instructlab/config.py
jjasghar: How much over head would be required to and more model families going forward? We will have others right?

hickeyma:
@leseb Can you do a review again pls?

leseb (approved):
I don't feel strongly enough about the tuple to keep advocating for it, so let's move forward!

xukai92 (approved):

xukai92:
The fix looks good to me.
I think the real confusion comes from the mix of model family and chat template: multiple model (merlinite and granite) can use the same template (the current one, which we usually refer as ibm-generic).

leseb:
PR received enough attention so dropped the requested reviews to move forward.

src/instructlab/config.py
nathan-weinberg: Seems like we would just need to flesh out this check, and ensure every supported model family has a proper mapping

xukai92: i expand some of my thoughts here: https://github.com/instructlab/instructlab/issues/1300

russellb:
> As for functional tests, lets do this in a separate PR if its going to hold up the fix. @russellb If not doing it here, can you raise an issue to remember it needs to be done?

Filed a follow-up about test coverage here https://github.com/instructlab/instructlab/issues/1303

gshipley:
The merlinite model was trained on a 2023 Mac Studio with 64gb RAM and Sonoma 14.5

ilab --version
ilab, version 0.15.0

PR #1301, opened 2024-06-07
docs: Improve LoRA doc by adding var and file name

rascasoft:
This commit extends the data upload section of the Collab instructions by adding references to the corresponding variable names for the uploaded files. This should save time to people that ask themselves "where do I put this path that I've copied?".


rascasoft:
Done it here https://github.com/instructlab/instructlab/pull/1397.

PR #1311, opened 2024-06-08
fix: use all prompts in a batch to generate data

### LINKED ISSUE #1292: ilab generate batch issue
**Describe the bug**
In function openai_completion()(https://github.com/instructlab/instructlab/blob/main/src/instructlab/generator/utils.py), it should call llm with all prompts(a batch with 5 prompts), however ,the implementation seems only use the first prompt.

image
image

in line 105: prompt_batches will be a list of list(1*5): [[prompt1, prompt2, prompt3, prompt4, prompt5]]
in line 111: it will enumerate prompt_batches, only once as it only has one element, so batch_id is 0.
in line 152, it only use prompt_batch[batch_id](actually is prompt_batches[0][0]), not enumerate all 5 prompt in prompt_batch(prompt_batches[0]) to use.

openai chat api client.chat.completions.create() does not support batch, so for a batch prompts in prompt_batch need an extra loop to call the api. 

**To Reproduce**
I found the issue by reading the code.

**Expected behavior**
all seed prompts in a batch should be used to call llm to generate data

chudegao:
Cnhance the code logic to use all batch prompts when call LLM api.

Resolves #1292

**Checklist:**

### DIFF
--- a/src/instructlab/data/generator/utils.py
+++ b/src/instructlab/data/generator/utils.py
@@ -146,24 +146,23 @@ def openai_completion(
             raise GenerateException(
                 f"Model {model_name} is not served by the server. These are the served models {model_ids}"
             )
+        for prompt in prompt_batch:
+            messages = [
+                {"role": "system", "content": get_sysprompt()},
+                {"role": "user", "content": prompt},
+            ]
+            # Inference the model
+            try:
+                response = client.chat.completions.create(
+                    messages=messages,
+                    **shared_kwargs,
+                )
+            except OpenAIError as exc:
+                raise GenerateException(
+                    f"There was a problem connecting to the server {exc}"
+                ) from exc
 
-        messages = [
-            {"role": "system", "content": get_sysprompt()},
-            {"role": "user", "content": prompt_batch[batch_id]},
-        ]
-
-        # Inference the model
-        try:
-            response = client.chat.completions.create(
-                messages=messages,
-                **shared_kwargs,
-            )
-        except OpenAIError as exc:
-            raise GenerateException(
-                f"There was a problem connecting to the server {exc}"
-            ) from exc
-
-        completions.extend(response.choices)
+            completions.extend(response.choices)
 
     if return_text:
         completions = [completion.text for completion in completions]


leseb (commented):
Could you please clarify (in the commit message), what issue this is solving or what enhancement it is providing?
The code seems ok to me, what I'm missing is the context.

Thanks!

chudegao:
> Could you please clarify (in the commit message), what issue this is solving or what enhancement it is providing? The code seems ok to me, what I'm missing is the context.
> 
> Thanks!

Updated the commit message.
The issue and details are in https://github.com/instructlab/instructlab/issues/1292  . The original code in openai_completion() only use the first prompt and missed last 4 of a batch(5) prompts. The fix is adding an extra loop to use all 5 prompts to call llm(openai chat completion api) to generate data. 
```
for batch_id, prompt_batch in enumerate(prompt_batches):
    ....
    messages = [
                {"role": "system", "content": get_sysprompt()},
                {"role": "user", "content": prompt_batch[batch_id]},   =====>> here's the issue as prompt_batch is a list of 5 prompts
            ]
```

leseb (commented):
Thanks for clarifying! I’m curious if the decision to use a single prompt was intentional, as I feel I might be missing some context here. @chudegao, could you explain the benefits of using multiple prompts compared to the current approach? Thanks! Just to be clear the code LGTM, it's just the impact of the change that I'm still missing.

chudegao:
> Thanks for clarifying! I’m curious if the decision to use a single prompt was intentional, as I feel I might be missing some context here. @chudegao, could you explain the benefits of using multiple prompts compared to the current approach? Thanks! Just to be clear the code LGTM, it's just the impact of the change that I'm still missing.

@markstur can you help confirm?

russellb (changes requested):
This code was moved over to the `instructlab/sdg` repo. Can you propose a change there?  Thanks!

chudegao:
> This code was moved over to the `instructlab/sdg` repo. Can you propose a change there? Thanks!

Ok

leseb:
Closing based on https://github.com/instructlab/instructlab/pull/1311#issuecomment-2177955894. Looking forward to your PR in `instructlab/sdg`. Thanks!

PR #1317, opened 2024-06-10
Harden the GitHub Action workflows

### LINKED ISSUE #1218: ci: convert repo's github actions to the newly proposed guidelines
GitHub actions need to be converted to follow these principles: https://github.com/instructlab/dev-docs/blob/main/docs/github-actions-use-policy.md

liranr23:
We use SHAs instead of tag names to refer to action versions. Dependabot
will help us manage the SHAs.

Update permissions to minimum necessary.

Add harden-runner to monitor egress of action. After some time, we can
tighten the egress to limit hosts/ports.

We also parallelize the lint workflow's linting work.

Resolves #1218 

**Checklist:**

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -7,29 +7,46 @@ on:
       - "main"
       - "release-**"
     paths:
-      - '.github/**'
+      - '.github/workflows/*.ya?ml'
+      - '.github/workflows/actionlint.*' # This workflow
   pull_request:
     branches:
       - "main"
       - "release-**"
     paths:
-      - '.github/**'
+      - '.github/workflows/*.ya?ml'
+      - '.github/workflows/actionlint.*' # This workflow
+
+env:
+  LC_ALL: en_US.UTF-8
 
 defaults:
   run:
     shell: bash
 
+permissions:
+  contents: read
+
 jobs:
   actionlint:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
+      - name: "Harden Runner"
+        uses: step-security/harden-runner@f086349bfa2bd1361f7909c78558e816508cdc10 # v2.8.0
+        with:
+          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
+
+      - name: "Checkout"
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
         with:
           fetch-depth: 0
           submodules: true
-      - name: Download actionlint
-        id: get_actionlint
-        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash) 1.6.27
-      - name: Check workflow files
-        run: PATH=".:$PATH" actionlint -color
+
+      - name: "Download actionlint"
+        run: |
+          docker build --tag actionlint - < .github/workflows/actionlint.dockerfile
+
+      - name: "Check workflow files"
+        run: |
+          echo "::add-matcher::.github/workflows/matchers/actionlint.json"
+          docker run --volume="${PWD}:/repo" --workdir=/repo actionlint -color
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,4 +1,6 @@
-name: docs
+# SPDX-License-Identifier: Apache-2.0
+
+name: Lint Markdown documents
 
 on:
   push:
@@ -8,6 +10,7 @@ on:
     paths:
       - '**/*.md'
       - '.markdownlint-cli2.yaml'
+      - '.github/workflows/docs.yml' # This workflow
   pull_request:
     branches:
       - "main"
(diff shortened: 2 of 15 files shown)


liranr23:
> Thanks for working on this, small question though, @bjhargrave was originally assigned to #1218. Did you both coordinate on this? Thanks!

I didn't coordinate on this. @bjhargrave sorry about that. I saw it's fresh and decided to do it as a good first issue. @bjhargrave if you wish me to abandon it because it's already being worked on let me know.

leseb:
> > Thanks for working on this, small question though, @bjhargrave was originally assigned to #1218. Did you both coordinate on this? Thanks!
> 
> I didn't coordinate on this. @bjhargrave sorry about that. I saw it's fresh and decided to do it as a good first issue. @bjhargrave if you wish me to abandon it because it's already being worked on let me know.

I'm not suggesting abandoning anything, just sync with @bjhargrave on his progress. Maybe he has not started :). I just want to make sure we don't duplicate efforts. Thanks!

leseb (changes requested):
The "Harden Runner" and the "Checkout" steps are repeated over all the workflows. Can we factor this? It is described [here](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow)

How about creating a new file that has the common steps?

```
# .github/workflows/common-steps.yml
name: Common Steps

on:
  workflow_call:

jobs:
  common_steps:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v4
        with:
          fetch-depth: 0
          submodules: 'true'
```

And call it on all the files? Like:

```yaml
uses: ./.github/workflows/common-steps.yml
```

Do you mind trying it out? Thanks!

.github/workflows/actionlint.yml
```diff
@@ -23,8 +23,12 @@ jobs:
   actionlint:
     runs-on: ubuntu-latest
     steps:
+      - name: "Harden Runner"
```
bjhargrave, line 26: There is a much better solution for actionlint. See https://github.com/instructlab/community/pull/302 for example.

bjhargrave (changes requested):

bjhargrave:
> Do you mind trying it out? Thanks!

This is doesn't help since these reusable workflow files are not processed by dependabot. And using SHAs really means we need dependabot to help manage dependency versions.

As for reducing repetition, see https://github.com/bjhargrave/instructlab_schema/blob/python-package/.github/workflows/lint.yml for an example of using a matrix to supply the variable parts of otherwise identical jobs.



> if you wish me to abandon it because it's already being worked on let me know.

I had not started to this start is fine!

nathan-weinberg:
Given the above convo I am going to reassign the issue to you @liranr23 - thanks for taking this on!

liranr23:
I will try to reach the actionlint soon, but it might take me a bit more time now as I have some more urgent things. Hopefully next week it will be done.

leseb (commented):
Dismissing my previous review based on @bjhargrave comments. Thanks!

bjhargrave:
> it might take me a bit more time now as I have some more urgent things

I can take a look at getting this across the finish line today.

liranr23:
> > it might take me a bit more time now as I have some more urgent things
> 
> I can take a look at getting this across the finish line today.

Sure, I don't mind it :)

bjhargrave (dismissed):

.github/workflows/lint.yml
```diff
@@ -23,25 +23,53 @@ on:
...
 jobs:
   lint:
     runs-on: ubuntu-latest
+    name: "${{ matrix.lint.name }}"
+    strategy:
+      fail-fast: false
+      matrix:
+        lint:
+          - name: "ruff"
+            commands: |
```
leseb, line 48: Why not use `run` here? I couldn't find any reference on `commands` in https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions.

.github/workflows/pypi.yaml
```diff
@@ -1,4 +1,5 @@
----
+# SPDX-License-Identifier: Apache-2.0
```
leseb, line 1: Too bad for consistency that this file uses 4 spaces indent instead of 2 like other files...

leseb (commented):
Just one question otherwise LGTM, thanks!

.github/workflows/pypi.yaml
bjhargrave: I guess this could be "fixed" in a followup PR.

leseb: Definitely.

.github/workflows/lint.yml
leseb: There is no need to change the key name, I was looking for clarification, thanks :D

leseb (approved):

.github/workflows/docs.yml
```diff
@@ -8,20 +10,39 @@ on:
...
   markdown-lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - name: markdownlint-cli2-action
-        uses: DavidAnson/markdownlint-cli2-action@v15
+      - name: "Harden Runner"
+        uses: step-security/harden-runner@17d0e2bd7d51742c71671bd19fa12bdc9d40a3d6 # v2.8.1
+        with:
+          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
```
nathan-weinberg, line 40: @bjhargrave do you think we could change this since this has run many times?

.github/workflows/actionlint.yml
```diff
@@ -7,29 +7,46 @@ on:
...
 jobs:
   actionlint:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
+      - name: "Harden Runner"
+        uses: step-security/harden-runner@f086349bfa2bd1361f7909c78558e816508cdc10 # v2.8.0
+        with:
+          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
```
nathan-weinberg, line 37: @bjhargrave do you think we could change this since this has run many times?

.github/workflows/docs.yml
bjhargrave: We are just introducing hardening to this repo and the specific egress needs to be seen before we can lock it down. Yes, this workflow is similar to the docs workflow in other repos, but I would like to run it a few times in this repo before we change from `audit`.

.github/workflows/actionlint.yml
bjhargrave: As above, I would like to merge using audit and then follow up with replacing audit in a later PR.

.github/workflows/docs.yml
nathan-weinberg: SGTM

nathan-weinberg (approved):

Issue #1310, opened 2024-06-07
instructlab shouldn't stack trace when it can't connect to the server


paulczar:
**Describe the bug**

If you try and connect to a server that doesn't exist (or isn't listening) you'll lose the error in the stack trace.

**To Reproduce**

1. Run

```
$ ilab chat -m models/ggml-model-f16.gguf
```

output

```
Traceback (most recent call last):
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/httpx/_transports/default.py", line 69, in map_httpcore_exceptions
    yield
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/httpx/_transports/default.py", line 233, in handle_request
    resp = self._pool.handle_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/httpcore/_sync/connection_pool.py", line 216, in handle_request
    raise exc from None
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/httpcore/_sync/connection_pool.py", line 196, in handle_request
    response = connection.handle_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/httpcore/_sync/connection.py", line 99, in handle_request
    raise exc
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/httpcore/_sync/connection.py", line 76, in handle_request
    stream = self._connect(request)
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/httpcore/_sync/connection.py", line 122, in _connect
    stream = self._network_backend.connect_tcp(**kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/httpcore/_backends/sync.py", line 205, in connect_tcp
    with map_exceptions(exc_map):
  File "/usr/lib64/python3.11/contextlib.py", line 155, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/httpcore/_exceptions.py", line 14, in map_exceptions
    raise to_exc(exc) from exc
[...]

Issue #1293, opened 2024-06-07
ilab generate failed sometimes


chudegao:
**Describe the bug**
ilab genereate failed sometimes without run 'ilab serve' first
```
(dlipy3) [root@churh1 lab]# ilab generate --num-instructions 1
llama_cpp_python is built without hardware acceleration. ilab generate will be very slow.
DEBUG 2024-06-06 19:11:48,634 server.py:63 Trying to connect to http://127.0.0.1:8000/v1...
DEBUG 2024-06-06 19:11:51,191 server.py:81 Trying port 35228...
DEBUG 2024-06-06 19:11:51,191 server.py:101 Port 35228 is available.
DEBUG 2024-06-06 19:11:51,192 server.py:105 Connection to http://127.0.0.1:8000/v1 failed. Starting a temporary server at http://127.0.0.1:35228/v1...
Generating synthetic data using 'models/merlinite-7b-lab-Q4_K_M.gguf' model, taxonomy:'taxonomy' against http://127.0.0.1:35228/v1 server
DEBUG 2024-06-06 19:11:51,281 utils.py:580 Found new taxonomy files:
DEBUG 2024-06-06 19:11:51,282 utils.py:582 * compositional_skills/writing/freeform/foo-lang/foo-lang.yaml
Cannot find prompt.txt. Using default prompt depending on model-family.
DEBUG 2024-06-06 19:11:51,457 generate_data.py:420 Loaded 5 human-written seed instructions from taxonomy
DEBUG 2024-06-06 19:11:51,458 generate_data.py:464 Generating to: generated/generated_merlinite-7b-lab-Q4_K_M_2024-06-06T19_11_51.json
  0%|                                                                                       | 0/1 [00:00<?, ?it/s]Synthesizing new instructions. If you aren't satisfied with the generated instructions, interrupt training (Ctrl-C) and try adjusting your YAML files. Adding more examples may help.
INFO 2024-06-06 19:11:51,461 generate_data.py:506 Selected taxonomy path compositional_skills->writing->freeform->foo-lang

Traceback (most recent call last):
  File "/opt/anaconda3/envs/dlipy3/lib/python3.11/site-packages/httpx/_transports/default.py", line 69, in map_httpcore_exceptions
    yield
  File "/opt/anaconda3/envs/dlipy3/lib/python3.11/site-packages/httpx/_transports/default.py", line 233, in handle_request
    resp = self._pool.handle_request(req)
[...]

PR #1294, opened 2024-06-07
update ensure_server(): wait longer time to make sure server is started

chudegao:
**Issue resolved by this Pull Request:**
Resolves #1293

**Checklist:**

- [ ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #1322, opened 2024-06-11
make ensure_server to wait readiness

makelinux:
Even when `is_alive`, the server fails to answer to `list_models`
for several seconds.
    
Retry list_models till it succeeds.
    
Fixes: #1310,  #1293

### DIFF
--- a/src/instructlab/server.py
+++ b/src/instructlab/server.py
@@ -2,7 +2,6 @@
 
 # Standard
 from contextlib import redirect_stderr, redirect_stdout
-from time import sleep
 import logging
 import multiprocessing
 import os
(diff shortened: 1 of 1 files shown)


leseb (changes requested):
Similar approach than https://github.com/instructlab/instructlab/pull/1294?

leseb:
> @leseb , similar, but simpler and cleaner: no need to check `is_alive` at all.


I disagree. is_alive informs us if the process is running, while list_models indicates whether the server is responding. They serve two different but useful purposes. If either of these checks fails, we can log and return the appropriate exception, allowing us to identify the issue more precisely.

The current PR only indicates whether the server is responsive, making it difficult to pinpoint the underlying problem. The child process might have died, the llama_cpp internal server could have crashed, or there might be a bug in the OpenAI server.

Additionally, in the future, it would be better to check if someone is assigned to an issue before pushing code. If there is an assignee, go ahead and review their work. If you have an alternative approach in mind, you can include it in your review.

Thanks!

makelinux:
@leseb , why original implementation has `while server_process.is_alive():` , why **whlie is_alive**?
Why not `while not server_process.is_alive():'?
Or I am missing something?

leseb:
> @leseb , why original implementation has `while server_process.is_alive():` , why **whlie is_alive**? Why not `while not server_process.is_alive():'? Or I am missing something?

I think the idea was to wait for a server that might fail, on failure the queue will be incremented, then later we check for the empty queue or not and fail if needed. Does that help? Thanks!

PR #1339, opened 2024-06-12
get abspath for each served model and the path provided

### LINKED ISSUE #1338: After 0.16.0, ilab serve and ilab chat are not handling directory names correctly
**Describe the bug**
After ilab 0.16.0, when passing in --model-path on serve and -m, it is not handling the difference between directories with the ./models/modelname and models/modename correctly

**To Reproduce**
on ilab v0.16.1

ilab serve --model-path ./models/merlinite.gguf
ilab chat -m models/merlinite.gguf

Gives an error even though ./models and models/ references the same file.

Also,
ilab serve --model-path ./models/merlinite.gguf
ilab generate --num-instructions 100

fails that the model is not being served. For generate to work now you have to not have ./ in the serve command as follows:
ilab serve --model-path modes/merlinite.gguf
^^ will work.

**Expected behavior**
The CLI should revert back to understanding that ./models and models/ are the same thing

**Screenshots**

 ilab sysinfo
instructlab.version: 0.16.1
sys.version: 3.11.9 (main, Apr  2 2024, 08:25:04) [Clang 15.0.0 (clang-1500.3.9.4)]
sys.platform: darwin
os.name: posix
platform.release: 23.5.0
platform.machine: arm64
torch.version: 2.3.1
torch.version.cuda: None
torch.version.hip: None
torch.cuda.available: False
llama_cpp_python.version: 0.2.75

cdoern:
if either the path in the config, the path being served, or the path specified by the user on runtime is an absolute/relative path to the same gguf, we fail since currenty we do not expand relative paths. get the abspath when checking if a model is served.

resolves #1338

**Checklist:**

### DIFF
--- a/src/instructlab/model/chat.py
+++ b/src/instructlab/model/chat.py
@@ -693,7 +693,14 @@ def chat_cli(
     model_ids = []
     for m in model_list:
         model_ids.append(m.id)
-    if not any(model == m for m in model_ids):
+
+    model_found = False
+    for m in model_ids:
+        if os.path.abspath(m) == os.path.abspath(model):
+            model_found = True
+            break
+
+    if not model_found:
         if model == cfg.DEFAULTS.MODEL_NAME_OLD:
             logger.info(
                 f"Model {model} is not a full path. Try running ilab config init or edit your config to have the full model path for serving, chatting, and generation."


cdoern:
unsure if there is a more pythonic way to do this, but the `any()` I had before didn't let me compare the two as abspaths so I had to expand it out this way :shrug:

src/instructlab/data/generator/utils.py
```diff
@@ -137,14 +137,21 @@ def openai_completion(
         model_ids = []
         for model in model_list:
             model_ids.append(model.id)
-        if not any(model_name == m for m in model_ids):
-            if model_name == DEFAULT_MODEL_OLD:
+
+        model_found = False
+        for m in model_ids:
```
leseb, line 142: Would a map do the trick?

```python
if os.path.abspath(model_name) not in map(os.path.abspath, model_ids):
    if model_name == DEFAULT_MODEL_OLD:
        logging.info(
            "Model %s is not a full path. Try running ilab init or edit your config to have the full model path for serving, chatting, and generation.",
            model_name,
        )
    raise GenerateException(
        f"Model {model_name} is not served by the server. These are the served models: {model_ids}"
    )
```

nathan-weinberg:
@gshipley can you try this out locally and see if it fixes your issue?

RobotSail:
Should we close this?

src/instructlab/data/generator/utils.py
```diff
@@ -0,0 +1,228 @@
+# SPDX-License-Identifier: Apache-2.0
```
leseb, line 1: ?

cdoern: sorry, accepted the wrong changes when rebasing :)

src/instructlab/model/chat.py
```diff
@@ -693,7 +693,14 @@ def chat_cli(
     model_ids = []
     for m in model_list:
         model_ids.append(m.id)
-    if not any(model == m for m in model_ids):
```
booxter, line 696:
```
    try:
        model_ids = [m.id for m in client.models.list().data]
    except openai.OpenAIError as exc:
        raise ChatException(f"Is the server running? {exc}") from exc
    if os.path.abspath(model) not in (os.path.abspath(m) for m in model_ids):
        ...
```

?

booxter (changes requested):
The new version is not very python. See the alternative.

In addition to this,

- commit message lines should be wrapped to standard;
- could you please add a test for this?

Thank you.

nathan-weinberg:
@cdoern is this still relevant?

ktam3:
Closing in favor of #1871

PR #1340, opened 2024-06-12
e2e: Run on PRs and pushes

russellb:
The e2e workflow seems stable and I've cleared the budget side of
running this a lot more often. Turn on running it on all code changes.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -21,6 +21,21 @@ pull_request_rules:
         - check-success=actionlint
       - -files~=.github/.*$
 
+    # e2e workflow
+    - or:
+      - and:
+        - check-success=e2e
+        - or:
+          - files~='.*\.py$'
+          - files~=pyproject.toml$
+          - files~=requirements.*\.txt$
+          - files~=.github/workflows/e2e.yml$
+      - and:
+        - -files~='.*\.py$'
+        - -files~=pyproject.toml$
+        - -files~=requirements.*\.txt$
+        - -files~=.github/workflows/e2e.yml$
+
     # lint and test must pass if files change that would trigger this job
     - or:
       - and:
@@ -201,17 +216,3 @@ pull_request_rules:
     label:
       remove:
         - one-approval
-
-- name: Trigger e2e workflow based on the 'e2e-trigger' label
-  conditions:
-    - label=e2e-trigger
-  actions:
-    github_actions:
-      workflow:
-        dispatch:
-          - workflow: e2e.yml
-            inputs:
-              pr_or_branch: "{{ number }}"
-    label:
-      remove:
-        - e2e-trigger
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -9,6 +9,24 @@ on:
         description: 'pull request number or branch name'
         required: true
         default: 'main'
+  push:
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - '**.py'
+      - 'pyproject.toml'
+      - 'requirements*.txt'
+      - '.github/workflows/e2e.yml'
+  pull_request:
+    branches:
+      - "main"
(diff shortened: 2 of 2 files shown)


leseb (commented):
Should we remove the code that triggers on labels too? It does not seem needed anymore.

russellb:
> Should we remove the code that triggers on labels too? It does not seem needed anymore.

good point ...

russellb:
> > Should we remove the code that triggers on labels too? It does not seem needed anymore.
> 
> good point ...

done!

nathan-weinberg (approved):

leseb (changes requested):
A few more things need to go away from `.github/workflows/e2e.yml`:

* logic about detecting whether this is a PR or not: it's always a PR
* All the logic about commenting on the PRs on success/failure
* `workflow_dispatch`
* Checkout branch step

Thanks!

nathan-weinberg:
> A few more things need to go away from `.github/workflows/e2e.yml`:
> 
>     * logic about detecting whether this is a PR or not: it's always a PR
> 
>     * All the logic about commenting on the PRs on success/failure
> 
>     * `workflow_dispatch`
> 
>     * Checkout branch step
> 
> 
> Thanks!
> 
> EDIT: I think keeping `gh` CLI does not hurt even if not used anymore.

If we get rid of `workflow_dispatch` we should also update the release documentation as that calls for a manual E2E run to be done on the release branch before drafting the release.

@russellb it may also be good to update the `ci.md` doc with some of the changes, but I won't block based on that

russellb:
> A few more things need to go away from `.github/workflows/e2e.yml`:
> 
> * logic about detecting whether this is a PR or not: it's always a PR

This only applies when manually triggered. I think it doesn't hurt to keep.

> * All the logic about commenting on the PRs on success/failure

I think this will only run if someone manually triggers the job against a PR.

> * `workflow_dispatch`

This is necessary for triggering it against a specified branch at a minimum. I suppose triggering on a PR isn't that interesting since that will already happen automatically.

I think I'd rather keep it though.  We may want to allow manually triggering on a PR with some additional options. Maybe we can let manually triggering it run it with a different model type, options to run more extensive training, ... probably a lot of potenntial.

> * Checkout branch step

I'd like to keep it to allow manually triggering the workflow against a branch. It's part of the release process at the moment.

I suppose it's going to run automatically every time a release branch changes, anyway, so maybe it's not needed in practice ... but similar to the discussion above about PRs, I think we might have use for this to allow triggering the workflow with parameters to do different variations.

Can we keep it for now to allow exploring more inputs for custom job runs?

leseb:
> > A few more things need to go away from `.github/workflows/e2e.yml`:
> > 
> > * logic about detecting whether this is a PR or not: it's always a PR
> 
> This only applies when manually triggered. I think it doesn't hurt to keep.
> 
> > * All the logic about commenting on the PRs on success/failure
> 
> I think this will only run if someone manually triggers the job against a PR.
> 
> > * `workflow_dispatch`
> 
> This is necessary for triggering it against a specified branch at a minimum. I suppose triggering on a PR isn't that interesting since that will already happen automatically.
> 
> I think I'd rather keep it though. We may want to allow manually triggering on a PR with some additional options. Maybe we can let manually triggering it run it with a different model type, options to run more extensive training, ... probably a lot of potenntial.
> 
> > * Checkout branch step
> 
> I'd like to keep it to allow manually triggering the workflow against a branch. It's part of the release process at the moment.
> 
> I suppose it's going to run automatically every time a release branch changes, anyway, so maybe it's not needed in practice ... but similar to the discussion above about PRs, I think we might have use for this to allow triggering the workflow with parameters to do different variations.
> 
> Can we keep it for now to allow exploring more inputs for custom job runs?

Alright alright :). I don't want to block that PR for that, we can revisit the config later. Thanks!

leseb (approved):

PR #1341, opened 2024-06-12
reduce arguments of chat_cli()

makelinux:
chat_cli() has too many arguments, which is anti-pattern. Many of arguments are parameters of the context.

Pass the context instead of individual parameters.

Introduce reusable http_client() to replace tls arguments.

### DIFF
--- a/src/instructlab/model/chat.py
+++ b/src/instructlab/model/chat.py
@@ -30,7 +30,7 @@ from instructlab import configuration as cfg
 from instructlab.server import is_temp_server_running
 
 # Local
-from ..utils import get_ssl_cert_config, get_sysprompt
+from ..utils import get_sysprompt, http_client
 
 HELP_MD = """
 Help / TL;DR
@@ -238,9 +238,8 @@ def chat(
 
     try:
         chat_cli(
-            logger=ctx.obj.logger,
+            ctx,
             api_base=api_base,
-            api_key=api_key,
             config=ctx.obj.config.chat,
             question=question,
             model=model,
@@ -249,10 +248,6 @@ def chat(
             qq=quick_question,
             greedy_mode=greedy_mode,
             max_tokens=max_tokens,
-            tls_insecure=tls_insecure,
-            tls_client_cert=tls_client_cert,
-            tls_client_key=tls_client_key,
-            tls_client_passwd=tls_client_passwd,
         )
     except ChatException as exc:
         click.secho(f"Executing chat failed with: {exc}", fg="red")
@@ -666,9 +661,8 @@ class ConsoleChatBot:  # pylint: disable=too-many-instance-attributes
 
 
 def chat_cli(
-    logger,
+    ctx,
     api_base,
-    api_key,
     config,
     question,
     model,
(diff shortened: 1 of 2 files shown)


src/instructlab/model/chat.py
```diff
@@ -677,19 +671,14 @@ def chat_cli(
...
-    tls_client_passwd: Optional[str] = None,
 ):
     """Starts a CLI-based chat with the server"""
-    cert = get_ssl_cert_config(tls_client_cert, tls_client_key, tls_client_passwd)
-    verify = not tls_insecure
+    logger = ctx.obj.logger
     client = OpenAI(
         base_url=api_base,
-        api_key=api_key,
+        api_key=ctx.params["api_key"],
```
leseb, line 679: Where `ctx.params` coming from? Where is it initialized?

makelinux: It is coming from Click module.

makelinux: I suppose it is initialized [here](https://github.com/pallets/click/blob/923d197b56caa9ffea21edeef5baf1816585b099/src/click/core.py#L920)

src/instructlab/utils.py
```diff
@@ -601,3 +602,14 @@ def read_taxonomy(logger, taxonomy, taxonomy_base, yaml_rules):
 def get_ssl_cert_config(tls_client_cert, tls_client_key, tls_client_passwd):
     if tls_client_cert:
         return tls_client_cert, tls_client_key, tls_client_passwd
+
+
+def http_client(ctx):
```
leseb, line 607: Passing the whole context does not seem very efficient.

src/instructlab/model/chat.py
```diff
@@ -677,19 +671,14 @@ def chat_cli(
     qq,
     greedy_mode,
     max_tokens,
-    tls_insecure,
-    tls_client_cert: Optional[str] = None,
-    tls_client_key: Optional[str] = None,
-    tls_client_passwd: Optional[str] = None,
 ):
     """Starts a CLI-based chat with the server"""
-    cert = get_ssl_cert_config(tls_client_cert, tls_client_key, tls_client_passwd)
```
leseb, line 686: Why not just do:

```python
cart = get_ssl_cert_config(ctx.params["tls_client_cert"], ctx.params["tls_client_key"], ctx.params["tls_client_passwd"])
```

leseb:
@makelinux thanks for the submitting, I like this new approach.

src/instructlab/model/chat.py
makelinux: The context as all objects in python is passed by reference. So it is more efficient and the code is more clean.

leseb: Ok glad to know it's a pointer under the hood. What do you mean by non-relevant details? Even though it's passed by reference, I don't like the idea of simply passing `ctx` everywhere will all the details, at least for entrypoint function but not for specific functions like `http_client`.

makelinux: Non-relevant details is what is not relevant for application logic: chat with LLM.
You can treat the `ctx.params` as runtime configuration. Each application module receives access to all configuration and selects what it needs. We can pass only `ctx.params` to `http_client` instead of `ctx` to reduce the scope.

leseb (approved):
Thanks!

nathan-weinberg (approved):

PR #1346, opened 2024-06-13
mergify: allow auto-merge for specific script without requiring lint/…

leseb:
…test checks

Updated Mergify configuration to support automatic merging when shell files
in `scripts/infra/` are modified. This change modifies the current
linter's regex to ignore all .sh files of not at the root of `scripts`.

This change aims to improve the merge workflow and reduce unnecessary
blocking for specific file changes.

Alternative approach to https://github.com/instructlab/instructlab/pull/1330

**Checklist:**

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -32,13 +32,13 @@ pull_request_rules:
           - files~=pyproject.toml$
           - files~=requirements.*\.txt$
           - files~=tox.ini$
-          - files~=scripts/.*\.sh$
+          - files~=scripts/[^/]+\.sh$
       - and:
         - -files~=.*\.py$
         - -files~=pyproject.toml$
         - -files~=requirements.*\.txt$
         - -files~=tox.ini$
-        - -files~=scripts/.*\.sh$
+        - -files~=scripts/[^/]+\.sh$
 
     - or:
       - and:


.github/mergify.yml
```diff
@@ -21,24 +21,46 @@ pull_request_rules:
...
         - or:
           # see .github/workflows/lint.yml and test.yml
           - files~=.*\.py$
           - files~=pyproject.toml$
           - files~=requirements.*\.txt$
           - files~=tox.ini$
-          - files~=scripts/.*\.sh$
+        # the following line only parses the shell scripts in the scripts directory but not the
+        # infra scripts - it's doing a depth 1 search
+          - files~=scripts/*.sh$
```
bjhargrave, line 55: I don't think this is a good regex.

I think the simple solution is just to just change the lint's regex for .sh to

```
- files~=scripts/[^/]+\.sh$
```

This would properly exclude subfolders of scripts.

It would be nice if mergify config has a way of using globs instead of regex. Then is would be much easier to keep in sync with the workflow's globs.

bjhargrave (changes requested):

.github/mergify.yml
leseb: Why isn't it a good regex? `scripts/*.sh$` excludes subfolders, it's a simple shell glob.

Other than I like your proposal, this is a simpler change. Thanks!

bjhargrave: `scripts/*.sh$` means `scripts` followed by zero-or-more `/` (`/*`) followed by any character (`.`) followed by `sh` followed by end-of-string. So `scriptsxsh` would match as would `scripts/xsh`. I don't think this is the intention.

Shell globs are not regex. The `~=` operated is followed by a regex not a glob.

leseb: > `scripts/*.sh$` means `scripts` followed by zero-or-more `/` (`/*`) followed by any character (`.`) followed by `sh` followed by end-of-string. So `scriptsxsh` would match as would `scripts/xsh`. `scripts/ruff.sh` would not match. I don't think this is the intention.

That's correct, thanks!

> 
> Shell globs are not regex. The [`~=` operator](https://docs.mergify.com/configuration/conditions/#operators-list) is followed by a regex not a glob.

Thanks for the pointer, I did not realize that!

bjhargrave (approved):

nathan-weinberg (approved):

PR #1369, opened 2024-06-14
`ilab model evaluate` command and eval library usage

cdoern:
1) added support for `ilab model evaluate` which allows users to run MMLU Bench MT Bench, MMLU Branch, and MT Branch Benchmarks
2) Add an `_evaluate` class to `config.yaml` so that users can get sane evaluation defaults that they can see and modify. These funnel directly into the evaluation flags as training now does. 
a sample evaluation class looks like:
```
evaluate:
  branch: null
  mmlu:
    batch_size: 5
    few_shots: 2
  mmlu_branch:
    sdg_path: generated
  model_name: null
  mt:
    judge_model: prometheus-eval/prometheus-8x7b-v2.0
    max_workers: 40
    output_dir: eval_data
  mt_branch:
    taxonomy_path: taxonomy
```


cdoern:
this is a placeholder for now, lmk when the library is somewhere I can access and import

nathan-weinberg:
@cdoern you can install directory from test.pypy.org for testing if you wish: https://test.pypi.org/project/instructlab-eval

once we have a `0.0.1` release we'll publish that to production PyPI

src/instructlab/model/evaluate.py
```diff
@@ -1,22 +1,113 @@
...
-    help="benchmarks to run during evaluation",
+    "--benchmark",
+    type=click.Choice(list(benchmark_names_to_classes.keys())),
+    case_sensitive=False,
+    help="Benchmarks to run during evaluation",
+)
+@click.option(
+    "--server-url",
+    type=click.STRING,
+    help="vLLM server endpoint. Needed for running mt_bench or pr_bench.",
```
leseb, line 29: is it only for vLLM? How about llama-ccp?

src/instructlab/model/evaluate.py
```diff
@@ -1,22 +1,113 @@
...
+    type=click.INT,
+    default=5,
+    show_default=True,
+    help="Number of GPUs. Needed for running mmlu or pr_mmlu.",
+)
+@click.option(
+    "--tasks",
+    type=click.STRING,
+    multiple=True,
+    help="List of tasks for MMLU to test the model with. Needed for running mmlu.",
```
leseb, line 55:
```suggestion
    help="List of tasks for mmlu to test the model with. Needed for running mmlu.",
```

src/instructlab/model/evaluate.py
```diff
@@ -1,22 +1,113 @@
...
+@click.option(
+    "--sdg-path",
+    type=click.Path(),
+    multiple=True,
+    help="Path where all the PR MMLU tasks are stored. Needed for running pr_mmlu.",
 )
 @click.pass_context
-def evaluate(ctx, benchmarks):
-    # do eval here
+def evaluate(
```
leseb, line 70:
```suggestion
@utils.display_params
def evaluate(
```

src/instructlab/model/evaluate.py
```diff
@@ -1,22 +1,113 @@
...
+                f"Benchmark {benchmark} requires the following args to be set: {required_args}",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
+        else:
+            evaluator_class = benchmark_names_to_classes[benchmark]
+            if benchmark == "mmlu":
+                evaluator = evaluator_class(tasks, few_shots, batch_size)
+            else:
+                evaluator = evaluator_class(sdg_path, task, few_shots, batch_size)
```
leseb, line 109: Can we have this in a dedicated function so that it can be unit tested accordingly, that's a lot of conditions.

leseb (changes requested):

src/instructlab/model/evaluate.py
cdoern: is this new? What does this do?

nathan-weinberg: Curious as well

src/instructlab/model/evaluate.py
nathan-weinberg: Yeah think we could do something like `selectEvaluator()` where we pass the benchmark and return the proper class if all the needed args are present, @cdoern I can look into doing this if you want

nathan-weinberg: @leseb did something like this in my latest commit

src/instructlab/model/evaluate.py
nathan-weinberg: MMLU is an acronym, so it should be capitalized

src/instructlab/model/evaluate.py
nathan-weinberg: Arg has been removed in subsequent configs, but also I believe vLLM is required - @alimaredia can you confirm?

src/instructlab/model/evaluate.py
```diff
@@ -1,52 +1,145 @@
...
+                    judge_model_name,
+                    taxonomy_path,
+                )
+
+    # ensure knowledge benchmarks have proper arguments if selected
+    if benchmark in ["mmlu", "mmlu_branch"]:
+        required_args = [few_shots, batch_size]
+        if benchmark == "mmlu":
+            required_args.append(tasks)
+        if benchmark == "mmlu_branch":
```
danmcp, line 75: elif

src/instructlab/model/evaluate.py
```diff
@@ -1,52 +1,145 @@
...
+                return evaluator_class(tasks, few_shots, batch_size)
+            else:
+                return evaluator_class(sdg_path, task, few_shots, batch_size)
 
 
 @click.command(cls=DYMGroup)
+@click.option(
+    "--model-name",
+    type=click.Path(),
+    help="Path to model to be evaluated",
```
danmcp, line 95: Name of model to be evaluated?

src/instructlab/model/evaluate.py
```diff
@@ -1,52 +1,145 @@
...
     case_sensitive=False,
     help="Benchmarks to run during evaluation",
 )
 @click.option(
-    "--server-url",
+    "--judge-model-name",
+    type=click.Path(),
+    help="Path to model to be used as a judge for running mt_bench or mt_bench_branch",
+)
+@click.option(
```
danmcp, line 108: Should this have a default?

src/instructlab/model/evaluate.py
```diff
@@ -1,52 +1,145 @@
...
+            if benchmark == "mmlu":
+                return evaluator_class(tasks, few_shots, batch_size)
+            else:
+                return evaluator_class(sdg_path, task, few_shots, batch_size)
 
 
 @click.command(cls=DYMGroup)
+@click.option(
+    "--model-name",
+    type=click.Path(),
```
danmcp, line 94: No longer a path

src/instructlab/model/evaluate.py
```diff
@@ -1,52 +1,145 @@
...
     "--benchmark",
-    type=click.Choice(list(benchmark_names_to_classes.keys())),
+    type=click.Choice(list(BENCHMARK_TO_CLASS_MAP.keys())),
     case_sensitive=False,
     help="Benchmarks to run during evaluation",
 )
 @click.option(
-    "--server-url",
+    "--judge-model-name",
+    type=click.Path(),
```
danmcp, line 105: No longer a path

src/instructlab/model/evaluate.py
```diff
@@ -1,52 +1,145 @@
...
+    help="Path to model to be used as a judge for running mt_bench or mt_bench_branch",
```
danmcp, line 106: Name of the model....

danmcp (changes requested):

src/instructlab/model/evaluate.py
nathan-weinberg: Any particular reason? I don't see a functional difference

src/instructlab/model/evaluate.py
nathan-weinberg: Where will it look for the model once it has the name?

src/instructlab/model/evaluate.py
nathan-weinberg: Same question as above

src/instructlab/model/evaluate.py
nathan-weinberg: Same question as above...

src/instructlab/model/evaluate.py
nathan-weinberg: Same question as above

src/instructlab/model/evaluate.py
danmcp: For mtbench, it looks for the model being served on server_url.  The mmlu work hasn't switched to looking up the hosted model yet.  My comment was just meant to point out the description didn't match the option name.

src/instructlab/model/evaluate.py
danmcp: I would suggest it makes the code more clear (that it will never hit the second if if the first one succeeds) and it avoid the extra unneeded string compare in those cases.

src/instructlab/model/evaluate.py
danmcp: I don't know.  Consistency with other cli dirs would be good.  The other consideration is whether we want to advertise the output to users or not.  Currently there isn't a reason to point users to it so I would suggest a . directory.

src/instructlab/model/evaluate.py
nathan-weinberg: @alinaryan ^^ FYI

src/instructlab/model/evaluate.py
nathan-weinberg: Spoke with @cdoern, adding a default output dir called `eval_results` :+1:

src/instructlab/model/evaluate.py
danmcp: > `elif` would still require the string compare no?

the elif wouldn't compare if the first if was a match

src/instructlab/configuration.py
```diff
@@ -29,6 +29,7 @@
 DEFAULT_MODEL_OLD = "merlinite-7b-lab-Q4_K_M"
 DEFAULT_MODEL = "models/merlinite-7b-lab-Q4_K_M.gguf"
 DEFAULT_MODEL_PATH = "models/merlinite-7b-lab-Q4_K_M.gguf"
+DEFAULT_EVAL_PATH = "eval_results"
```
danmcp, line 32: results may not be the best term here.  The folder will contain a lot of the intermediate states.  The actual results should probably be returned through the command.

nathan-weinberg: Going to change to `eval_data` - @RobotSail this aligns with what's in your proposal, though the pathing will need to be changed later

RobotSail: @nathan-weinberg Yes this is perfect! @danmcp We will move this into a more reasonable location later.

src/instructlab/model/evaluate.py
```diff
@@ -86,77 +84,80 @@ def get_evaluator(
...
+                    "mmlu_professional_psychology",
+                    "mmlu_public_relations",
+                    "mmlu_security_studies",
+                    "mmlu_social_sciences",
+                    "mmlu_sociology",
+                    "mmlu_stem",
+                    "mmlu_us_foreign_policy",
+                    "mmlu_virology",
+                    "mmlu_world_religions",
+                ]
```
nathan-weinberg, line 154: Indentation is off

src/instructlab/model/evaluate.py
```diff
@@ -86,77 +84,80 @@ def get_evaluator(
...
                 return evaluator_class(
                     model_path, tasks, "float16", few_shots, batch_size
                 )
             else:
                 return evaluator_class(
-                    model_path, sdg_path, tasks, "float16", few_shots, batch_size
+                    model_path, sdg_path, ["mmlu_pr"], "float16", few_shots, batch_size
```
nathan-weinberg, line 160: why is this a single-item list?

nathan-weinberg (changes requested):

src/instructlab/model/evaluate.py
danmcp: It can be multiple items.  But we are only going to use it for the whole group.

src/instructlab/model/evaluate.py
danmcp: Apologies, I probably missed a format before pushing.  Giving we are sharing the pull, I'll wait till my next commit to fix.

src/instructlab/model/evaluate.py
nathan-weinberg: Gotcha!

src/instructlab/model/evaluate.py
nathan-weinberg: No worries!

cdoern:
I think CI should just pass on this one :crossed_fingers:

nathan-weinberg (dismissed):

nathan-weinberg (approved):

src/instructlab/model/evaluate.py
```diff
@@ -0,0 +1,504 @@
...
+        taxonomy_path,
+        branch,
+        base_branch,
+        few_shots,
+        batch_size,
+        sdg_path,
+    )
+
+    if benchmark == "mt_bench":
+        # TODO: Replace temp Popen hack with serving library calls. Current library doesn't support server-model-name.
```
danmcp, line 289: Noting this is a WIP here: https://github.com/instructlab/instructlab/compare/main...danmcp:instructlab:evalandvllm?expand=1

danmcp, line 289: Noting this is a WIP here: https://github.com/instructlab/instructlab/compare/main...danmcp:instructlab:evalandvllm?expand=1

src/instructlab/model/evaluate.py
```diff
@@ -0,0 +1,504 @@
...
+                        m_name,
+                    ]
+                )
+                time.sleep(60)
+                evaluator.gen_answers("http://localhost:8000/v1")
+            finally:
+                proc.terminate()
+
+            print(f"Evaluating answers for branch {branch}...")
+            try:
```
danmcp, line 395: This server launch could be shared between the two models.  Addressed here: https://github.com/instructlab/instructlab/compare/main...danmcp:instructlab:evalandvllm?expand=1

danmcp (approved):
I am approving with a few notes:
- The error handling still needs a decent bit of work both here and the eval library.  -> Replace stack traces with nice messages.
- This version is very chatty in the cli.  Will largely be addressed with https://github.com/instructlab/instructlab/compare/main...danmcp:instructlab:evalandvllm?expand=1.  Those commit(s) also address the vllm serving hack still in place in this PR.
- More test cases are needed
- Config defaulting probably still needs some work with the models selected

src/instructlab/model/evaluate.py
```diff
@@ -0,0 +1,504 @@
...
+            required_arg_names.append("base-model")
+        if None in required_args:
+            click.secho(
+                f"Benchmark {benchmark} requires the following args to be set: {required_arg_names}",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
+        evaluator_class = BENCHMARK_TO_CLASS_MAP[benchmark]
+        if benchmark == "mt_bench":
+            return evaluator_class("test_model", "judge_model", output_dir, max_workers)
```
alinaryan, line 71: Should `test_model` be `model`? Applies here and other places too

src/instructlab/model/evaluate.py
```diff
@@ -0,0 +1,504 @@
...
+            required_arg_names.append("base-model")
+        if None in required_args:
+            click.secho(
+                f"Benchmark {benchmark} requires the following args to be set: {required_arg_names}",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
+        evaluator_class = BENCHMARK_TO_CLASS_MAP[benchmark]
+        if benchmark == "mmlu":
+            min_tasks = os.environ.get("INSTRUCTLAB_EVAL_MMLU_MIN_TASKS")
```
alinaryan, line 98: Are you planning to reference the [default 57 tasks](https://github.com/instructlab/eval/pull/24/files#diff-88873e8f609db17f8729f4725117acdfa0db7930066436bd6b0d5c7301511101R14) in this PR?

src/instructlab/model/evaluate.py
```diff
@@ -0,0 +1,504 @@
...
+            if min_tasks is not None:
+                tasks = ["mmlu_abstract_algebra", "mmlu_anatomy", "mmlu_astronomy"]
```
alinaryan, line 100: Is this needed as some reinforcement? If so, should you hard code the full list of 57 tasks here?

src/instructlab/model/evaluate.py
```diff
@@ -0,0 +1,504 @@
...
+)
+@click.option(
+    "--judge-model",
+    type=click.STRING,
+    help="Model to be used as a judge for running mt_bench or mt_bench_branch - can be a local path or the name of a Hugging Face repository",
+)
+@click.option(
+    "--output-dir",
+    type=click.Path(),
+    help="The directory to use for evaluation output from mt_bench or mt_bench_branch",
```
alinaryan, line 219: don't forget to add the default `eval_results`

alinaryan (commented):
Thanks for the strong work on this!! Just have a few comments/q's:

src/instructlab/model/evaluate.py
danmcp: INSTRUCTLAB_EVAL_MMLU_MIN_TASKS is in place for testing purposes to allow for a smaller list of tasks to validate the flow without running a full test.

src/instructlab/model/evaluate.py
danmcp: The default is set to eval_data in configuration.py

src/instructlab/model/evaluate.py
danmcp: In this case test_model is simply the name referenced by mt_bench to call the model being served.  It could be anything so we just want to pick something recognizable.  And calling the model you are serving "model" seems a little under descriptive.

alinaryan: Sounds good to me! Wanted to ensure it wasn't missed with the other renames to 'model'

src/instructlab/model/evaluate.py
alinaryan: Ah. Thought it had to be added in this click arg

alinaryan (approved):

src/instructlab/model/evaluate.py
alinaryan: Spoke with @cdoern - this will be addressed in a follow-up PR. See issue https://github.com/instructlab/instructlab/issues/1534

cdoern:
made follow up issues!

src/instructlab/configuration.py
```diff
@@ -358,13 +410,26 @@ def init(ctx, config_file):
...
         )
-        ctx.default_map = training_dict
+        config_dict["evaluate"] = (
+            config_dict["evaluate"]
+            | config_dict["evaluate"]["mmlu"]
+            | config_dict["evaluate"]["mmlu_branch"]
+            | config_dict["evaluate"]["mt_bench"]
+            | config_dict["evaluate"]["mt_bench_branch"]
+        )
+        # need to delete the individual sub-classes from the map
```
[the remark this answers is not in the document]
cdoern: since `_evaluate` has "top level" options that arent in sub classes, the `|` operator can't pull out primitive values (str, float, etc). So, what I do here is create a map with the entire old eval class, AND I add the flattened sub classes. so it looks something like 

` eval
    mt:
    mmlu
    mmlubranch
    mtbranch
    ALL_OPTIONS_FROM_THOSE_CLASSES_DUPLICATED_IN_TOP_LEVEL
`
I then delete mt, mmlu, and the branch benches so I am left with just the "unfolded" options in a "top level" eval key that can be directly mapped to the flags.

This is super confusing. I chatted with @nathan-weinberg about it, I can whiteboard it for folks on monday too

src/instructlab/configuration.py
```diff
@@ -358,13 +410,26 @@ def init(ctx, config_file):
...
-        training_dict = get_dict(ctx.obj.config)
+        config_dict = get_dict(ctx.obj.config)
         # since torch and train args are sep, they need to be combined into a single `train` entity for the default map
         # this is because the flags for `ilab model train` will only be populated if the default map has a single `train` entry, not two.
-        training_dict["train"] = (
-            training_dict["train"]["train_args"]
-            | training_dict["train"]["torch_args"]
-            | training_dict["train"]["train_args"]["lora"]
-            | training_dict["train"]["train_args"]["deepspeed_options"]
+        config_dict["train"] = (
```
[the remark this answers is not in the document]
cdoern: yep, for training yeah. I need to split out each cmd that we want to auto map

src/instructlab/configuration.py
```diff
@@ -177,6 +179,39 @@ def api_base(self):
         return get_api_base(self.host_port)
 
 
+class _mmlu(BaseModel):
```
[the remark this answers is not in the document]
cdoern: Making a follow up issue for this since I don't want to lose the two approvals we currently have with a small push

src/instructlab/configuration.py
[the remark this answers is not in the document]
alinaryan: That would be helpful, thank you :pray:

cdoern:
Given that tomorrow morning (7/1) is a deadline. This PR needs to be merged before then to add some form of evaluation support.

That being said, there is a stale change request on this PR, some pending reviews that haven't come in yet, etc.

We will be dismissing those in favor of deferring to follow up issues but if any of the reviewers have immediate follow up concerns please feel free to reach out!!!!

russellb:
I filed #1540 as a follow-up to get this tested in `e2e` CI

src/instructlab/model/evaluate.py
nathan-weinberg: full list is also now a default of the MMLU evaluator class as well, so we can initialize it w/o that arg to run full-scale MMLU: https://github.com/instructlab/eval/pull/24

PR #1370, opened 2024-06-14
Training Library Usage

cdoern:
This is the implementation of high fidelity training using the new backend library

1.   adds all flags outlined in new training library to ilab model train
2.   The defaults for these flags come from a new _train class in the config. The train class is composed of the exact classes introduced in the training library, I imported them.
3. I added a --train-profile to ilab init allowing a user to specify a yaml file matching the exact format of the training library args. The two sections in the yaml are train_args and torch_args.
4. I cleaned up mismatched flags and outdated flags necessary for this new library.

the way to run this is:

   * if you want to run linux train just run `ilab config init` and then `ilab train --legacy=true`
   * if you want to use the backend code, ilab config init now has a `--train-profile` flag which takes a valid yaml of the following format:

```
train_args:
  data_path: sample-data/train_all_pruned_SDG.jsonl
  ckpt_output_dir: checkpoints
  data_output_dir: output
  max_seq_len: 4096
  max_batch_len: 10000
  num_epochs: 10
  effective_batch_size: 3840
  save_samples: 25000
  learning_rate: 2e-6
  warmup_steps: 800
  is_padding_free: False
  random_seed: 42
torch_args:
  node_rank: 0
  nnodes: 1
  nproc_per_node: 1
  rdzv_id: 123
  rdzv_endpoint: '127.0.0.1:12345'
```
These are the exact TrainingArgs and TorchrunArgs from the training repo.

These options then funnel into the flags for ilab train which can be used as overrides but the hope is @RobotSail and team can provide pre-baked training yaml for specific hardware so users will not need to specify ANY flags (this works in testing so far which is really cool)

 *   macos train is detected based off the chip type, and users will auto run that if they are on a mac.


src/instructlab/configuration.py
```diff
@@ -148,6 +149,38 @@ def api_base(self):
         """Returns server API URL, based on the configured host and port"""
         return get_api_base(self.host_port)
 
+class _deepspeed(BaseModel):
```
RobotSail, line 152: For local dev, I recommend cloning the current interface PR https://github.com/instructlab/training/pull/12, installing locally via `pip install -e .` and then importing these directly.

src/instructlab/model/train.py
```diff
@@ -226,78 +318,9 @@ def get_files(pattern):
         shutil.copy(train_files[0], train_file)
         shutil.copy(test_files[0], test_file)
 
-    # pylint: disable=import-outside-toplevel
```
RobotSail, line 229: I love to see it

src/instructlab/model/train.py
```diff
@@ -226,78 +318,9 @@ def get_files(pattern):
...
-    if not utils.is_macos_with_m_chip():
-        # Local
```
RobotSail, line 231: I love to see how much code gets deleted with this

src/instructlab/configuration.py
cdoern: yep, I am removing these and opting to use the ones you introduced. Still valuable to have a top level `_train` class in here though

cdoern:
some status here. This PR does the following:

1. adds all flags outlined in new training library to `ilab model train`
2. The defaults for these flags come from a new `_train` class in the config. The train class is composed of the exact classes introduced in the training library, I imported them.
3. I added a `--train-profile` to ilab init allowing a user to specify a yaml file matching the exact format of the training library args. The two sections in the yaml are `train_args` and `torch_args`.
4. I cleaned up mismatched flags and outdated flags necessary for this new library. 


the commits might be a bit messy rn. I will clean these up in the morning. Functionally, this is the structure we should go for. Eventually, all of the top level config stuff should live in a profile. The amount of flags being added to training is... quite a lot.

cdoern:
usage:

1. `ilab config init --train-profile /path/to/trainargs.yaml`
2. `ilab train` (no flags unless you want to override the yaml. use `--legacy` to use old linux train code.


a train profile takes the following format: 

```
train_args:
  data_path: sample-data/train_all_pruned_SDG.jsonl
  ckpt_output_dir: checkpoints
  data_output_dir: output
  max_seq_len: 4096
  max_batch_len: 10000
  num_epochs: 10
  effective_batch_size: 3840
  save_samples: 25000
  learning_rate: 2e-6
  warmup_steps: 800
  is_padding_free: False
  random_seed: 42
torch_args:
  node_rank: 0
  nnodes: 1
  nproc_per_node: 1
  rdzv_id: 123
  rdzv_endpoint: 127.0.0.1:12345
```

thats it :)

cdoern:
@jeremyeder this is the training PR. It should work but the instrutlab-train package isn't published to pypi yet so I have been installing and testing locally.

@RobotSail will be taking over from here weds, thurs, fri, and I will pick up when I am back. 

the way to run this is:

* if you want to run linux train just run `ilab config init` and then `ilab train --legacy=true`
* if you want to use the backend, clone the instructlab/training repo and `pip install -e .` in the same venv then pip install -e . from the instructlab cloned repo. `ilab config init` `ilab train`. `ilab config init` now has a `--train-profile` which takes a valid yaml of the following format:

```
train_args:
  data_path: sample-data/train_all_pruned_SDG.jsonl
  ckpt_output_dir: checkpoints
  data_output_dir: output
  max_seq_len: 4096
  max_batch_len: 10000
  num_epochs: 10
  effective_batch_size: 3840
  save_samples: 25000
  learning_rate: 2e-6
  warmup_steps: 800
  is_padding_free: False
  random_seed: 42
torch_args:
  node_rank: 0
  nnodes: 1
  nproc_per_node: 1
  rdzv_id: 123
  rdzv_endpoint: '127.0.0.1:12345'
```

These are the exact TrainingArgs and TorchrunArgs from the training repo.

These options then funnel into the flags for ilab train which can be used as overrides but the hope is @RobotSail and team can provide pre-baked training yaml for specific hardware so users will not need to specify ANY flags (this works in testing so far which is really cool)

* macos train is detected based off the chip type, and users will auto run that if they are on a mac.

n1hility:
> 3. I added a --train-profile to ilab init allowing a user to specify a yaml file matching the exact format of the training library args. The two sections in the yaml are train_args and torch_args.

Is this just a temporary development capability as a stand in until we get to a config ux? Friday we all talked about how as a step towards profiles we would first map in everything as an option which allow us to have a consistent experience (e.g. everything has a config file setting, a command line argument, and a default (which is provided by the profile)) And then we would have categories etc to hide some of the advance CLI overrides.

The big thing I would be worried about is ending up in a position where it's becomes unclear how you specify something or what the source of truth.(e.g. user has to ask Is it in a config file or an option, or an option to a config file? If it's in a config file, which one holds the setting? If I am a laptop user do I use the config.yaml or the train profile, they both sound like what I need).

n1hility (changes requested):
Sorry somehow forgot to mark this one request changes as part of my comment here: https://github.com/instructlab/instructlab/pull/1370#issuecomment-2177257333

cdoern:
@n1hility 
> The big thing I would be worried about is ending up in a position where it's becomes unclear how you specify something or what the source of truth.(e.g. user has to ask Is it in a config file or an option, or an option to a config file? If it's in a config file, which one holds the setting? If I am a laptop user do I use the config.yaml or the train profile, they both sound like what I need).


`--train-profile` is an argument to `ilab config init`. This flag will populate a new portion of the `config.yaml` called `train` with the options specified in the user's yaml. I thought this made more sense than forcing them to edit the config.yaml directly OR creating a full config.yaml with the train settings pre-specified. 

In my mind, this is a step towards global profiles for all commands that can either live in the config.yaml or as a separate entity. We didn't get enough consensus on where profiles would be represented because folks viewed them as something we should focus on later. However, I was trying to convey that PRs like this will necessitate at least some direction for profiles to be made.

every option in `--train-profile` can be overridden by a flag in the CLI. And if a user wants to run the old linux train they can specify `--legacy`, there is no functionality loss here over all.

cdoern:
The big thing in this PR is: The new train library introduces a lot of new code. Much of it is similar theory to what we run, but I don't feel comfortable hot swapping it in for the current linux_train flow within the week.

the `--legacy` flag enables the old linux_train code and should require minimal additional flags since `ilab init` sets sane defaults for training now. 

The "non-legacy" new workflow is going to have bumps, but overall the design of taking a top level cfg for *just training for now* and *eventually evolving into a profile* is the direction I believe we should go in, especially in a time crunch.

cdoern:
after discussing with the team, we are going to be timeboxing the merging of this pr to EOD today EDT or the morning of 6/25 EDT. This, of course, is assuming I get CI passing! I will not be merging a broken PR.

src/instructlab/configuration.py
```diff
@@ -173,9 +186,53 @@ def get_default_config():
...
+            train_args=TrainingArgs(
+                model_path=DEFAULT_MODEL_REPO,
+                data_path=DEFAULT_GENERATED_FILES_OUTPUT_DIR,
+                ckpt_output_dir=DEFAULT_CKPT_DIR,
+                data_output_dir=DEFAULT_OUT_DIR,
+                max_seq_len=4096,
+                max_batch_len=10000,
+                num_epochs=10,
+                effective_batch_size=3840,
+                save_samples=25000,
```
RobotSail, line 199: This is way too high, set it to `250_000`

src/instructlab/model/train.py
```diff
@@ -87,29 +159,72 @@
         "(reduces GPU VRAM usage and may slow down training)"
     ),
 )
+@click.option("--max-seq-len", type=int)
+@click.option("--max-batch-len", type=int)
```
nathan-weinberg, line 163: Can we add help text to these args?

src/instructlab/model/train.py
```diff
@@ -87,29 +159,72 @@
...
+)
+@click.option("--lora-dropout", type=float, help="dropout for LoRA layers")
+@click.option(
+    "--target-modules",
+    type=str,
+)
+@click.option(
+    "--is-padding-free",
+    type=bool,
+)
```
nathan-weinberg, line 200: Can we add help text to these args?

src/instructlab/model/train.py
```diff
@@ -87,29 +159,72 @@
...
+    "--gpus",
+    "nproc_per_node",
+    type=int,
+    help="this is the number of GPUs to use. This is a torch specific arg and must be called nproc-per-node",
 )
+@click.option("--nnodes", type=int)
+@click.option("--node-rank", type=int)
+@click.option("--rdzv-id", type=int)
+@click.option("--rdzv-endpoint", type=str)
+@click.option("--legacy", hidden=True, type=bool, default=False)
```
nathan-weinberg, line 211: What is `--legacy` for?

src/instructlab/model/train.py
```diff
@@ -119,13 +234,16 @@ def train(
         # By default, generate output-dir is used as train input-dir
         input_dir = ctx.obj.config.generate.output_dir
 
-    effective_data_dir = Path(data_dir or "./taxonomy_data")
+    if four_bit_quant and device != "cuda":
+        ctx.fail("--4-bit-quant option requires --device=cuda")
```
nathan-weinberg, line 238:
```suggestion
        ctx.fail("'--4-bit-quant' option requires '--device=cuda'")
```

src/instructlab/model/train.py
```diff
@@ -164,8 +282,57 @@ def get_files(pattern):
...
+
+        load_and_train(
+            model=dest_model_dir,
+            train=True,
+            data=data_path,
+            adapter_file=adapter_file_path,
+            iters=iters,
+            save_every=10,
+            steps_per_eval=10,
+        )
```
nathan-weinberg, line 334:
```suggestion

        # train the model with LoRA
        load_and_train(
            model=dest_model_dir,
            train=True,
            data=data_path,
            adapter_file=adapter_file_path,
            iters=iters,
            save_every=10,
            steps_per_eval=10,
        )
```

src/instructlab/configuration.py
RobotSail: It'll be `13.48G` per checkpoint, we save 2 checkpoints but we get rid of one set.

src/instructlab/configuration.py
```diff
@@ -173,9 +186,53 @@ def get_default_config():
...
         ),
         serve=_serve(model_path=DEFAULT_MODEL_PATH),
+        train=_train(
+            train_args=TrainingArgs(
+                model_path=DEFAULT_MODEL_REPO,
+                data_path=DEFAULT_GENERATED_FILES_OUTPUT_DIR,
+                ckpt_output_dir=DEFAULT_CKPT_DIR,
+                data_output_dir=DEFAULT_OUT_DIR,
+                max_seq_len=4096,
+                max_batch_len=10000,
```
RobotSail, line 196: This is the one that gets affected by num of GPUs or memory of GPUs.

src/instructlab/configuration.py
```diff
@@ -173,9 +186,53 @@ def get_default_config():
             taxonomy_base=DEFAULT_TAXONOMY_BASE,
         ),
         serve=_serve(model_path=DEFAULT_MODEL_PATH),
+        train=_train(
+            train_args=TrainingArgs(
+                model_path=DEFAULT_MODEL_REPO,
+                data_path=DEFAULT_GENERATED_FILES_OUTPUT_DIR,
```
RobotSail, line 192: Needs to be a specific `.jsonl` file containing the training data.

src/instructlab/configuration.py
```diff
@@ -173,9 +186,53 @@ def get_default_config():
             taxonomy_base=DEFAULT_TAXONOMY_BASE,
         ),
         serve=_serve(model_path=DEFAULT_MODEL_PATH),
+        train=_train(
+            train_args=TrainingArgs(
+                model_path=DEFAULT_MODEL_REPO,
```
RobotSail, line 191: We should change this to use [granite-7b-base](https://huggingface.co/ibm-granite/granite-7b-base/tree/main) and `is_padding_free` should be set to true.

src/instructlab/model/train.py
```diff
@@ -233,72 +400,9 @@ def get_files(pattern):
...
-            save_every=10,
-            steps_per_eval=10,
-        )
-
-        # TODO copy some downloaded files from the PyTorch model folder
-        # Seems to be not a problem if working with a remote download with convert.py
+        # pull the training and torch args from the flags
+        # the flags are populated from the config as a base.
+        params = ctx.params
+        train_args = TrainingArgs(**params)
```
RobotSail, line 406: Just need to make sure that the datapath being received here is valid. We can either:
- check the type of it e.g. file/dir and then give them a descriptive error
- OR we can just figure it out on our own

nathan-weinberg:
@RobotSail could you do a single batch review instead of separate reviews for each comment? I can show ya how

src/instructlab/model/train.py
RobotSail:
```py
if is_dir(data_path):
  raise ValueError("It looks like the datapath you passed is either a directory or does not exist, please doublecheck the value of --data-path and try again")
```

RobotSail, line 406: just make sure that lora and deepspeed is being read correctly

src/instructlab/configuration.py
cdoern: we discussed leaving this as a dir for the linux_train usecase, high fidelity users will be overriding these defaults 100% of the time anyway. So tailoring this to the base usecase makes sense

src/instructlab/model/train.py
RobotSail: @nathan-weinberg `--legacy` is so that we can bring in the new training interface without breaking functionality for existing users who want to use either MacOS or Linux train.

src/instructlab/configuration.py
```diff
@@ -148,6 +153,11 @@ def api_base(self):
         return get_api_base(self.host_port)
 
 
+class _train(BaseModel):
```
JamesKunstle, line 156: IMO since this is a typing name we should use the typical "TrainArgs" or something. a type called "_train" is kinda weird stylistically

src/instructlab/configuration.py
```diff
@@ -173,9 +186,53 @@ def get_default_config():
...
+                warmup_steps=800,
+                is_padding_free=False,
+                random_seed=42,
+            ),
+            torch_args=TorchrunArgs(
+                node_rank=0,
+                nnodes=1,
+                nproc_per_node=1,
+                rdzv_id=123,
+                rdzv_endpoint="127.0.0.1:8000",
```
JamesKunstle, line 210: let's pick a low-contention port as the default

src/instructlab/model/train.py
```diff
@@ -7,15 +7,93 @@
...
+
+    Returns a torch.device object:
+    - type is one of 'cpu', 'cuda', 'hpu'
+    - index is None or device index (e.g. 0 for first GPU)
+    """
+
+    name = "deviceinfo"
+    supported_devices = {"cuda", "cpu", "hpu"}
+
+    def convert(self, value, param, ctx) -> "torch.device":
```
JamesKunstle, line 29: I don't think this output annotation works, I haven't seen it before.

src/instructlab/model/train.py
```diff
@@ -87,29 +159,72 @@
...
-    "--model-name",
-    default="instructlab/merlinite-7b-lab",
-    show_default=True,
-    help="model name to use in training",
+    "--effective-batch-size", type=int, help="total batch size across all GPUs"
+)
+@click.option(
+    "--save-samples",
+    type=int,
+    help="how many amples to process before saving a checkpoint",
```
JamesKunstle, line 170:
```suggestion
    help="The number of samples processed in between checkpoints.",
```

src/instructlab/model/train.py
```diff
@@ -87,29 +159,72 @@
...
+@click.option("--ds-quantize-dtype", type=click.Choice(["nf4", "fp8"]), default=None)
+# below flags are invalid if lora == false
+@click.option("--lora-rank", type=int, help="rank of update matricies")
+@click.option(
+    "--lora-alpha", type=float, help="how influential/strong lors tune will be"
+)
+@click.option("--lora-dropout", type=float, help="dropout for LoRA layers")
+@click.option(
+    "--target-modules",
+    type=str,
```
JamesKunstle, line 195: maybe this should be hidden since it's pretty implementation-specific?

JamesKunstle (changes requested):
couple of minor nit-picks but overall I think this is a really good reorganization. If there are things missed in review we can grab them later.

src/instructlab/configuration.py
RobotSail: Yeah it looks like `12222` is just [being used](https://www.speedguide.net/port.php?port=12222) by "Act of War: Direct Action (Atari)" and the Light Weight Access Point Protocol (LWAPP) so it would probably be fine to use.

RobotSail (approved):
LGTM once everything works and is in a good state

src/instructlab/configuration.py
cdoern: for some odd reason (idk why) the names of the classes need to match the name of the commands for auto populating the flags.

src/instructlab/configuration.py
cdoern: any suggestions?

src/instructlab/model/train.py
```diff
@@ -7,15 +7,99 @@
...
+
+TORCH_DEVICE = TorchDeviceParam()
+
+
 @click.command()
-@click.option("--data-dir", help="Base directory where data is stored.", default=None)
+@click.option(
+    "--data-path",
+    type=click.Path(file_okay=True),
+    help="Base directory where data is stored.",
```
RobotSail, line 88: In the new training, this should be a path to the dataset file we're training on. For Linux/MacOS train this is still correct.

src/instructlab/model/train.py
```diff
@@ -87,29 +165,67 @@
         "(reduces GPU VRAM usage and may slow down training)"
     ),
 )
+@click.option("--max-seq-len", type=int)
```
RobotSail, line 168: This setting determines the maximum length (in tokens) of a SINGLE sample. Think of this as a filter that drops all samples (doesn't train on them) from the dataset exceeding this length.

src/instructlab/model/train.py
RobotSail: This is the maximum overall length of samples that we process in a given batch.

src/instructlab/model/train.py
```diff
@@ -87,29 +165,67 @@
...
+)
+# below flags are invalid if lora == false
+@click.option("--rank", type=int, help="rank of update matricies")
+@click.option("--alpha", type=float, help="how influential/strong lors tune will be")
+@click.option("--dropout", type=float, help="dropout for LoRA layers")
+@click.option(
+    "--target-modules",
+    type=str,
 )
+@click.option("--quantize-data-type", type=str, default=None)
```
RobotSail, line 197: The quantization data type to use when training a LoRA. Valid options are "nf4" and None.

src/instructlab/model/train.py
```diff
@@ -87,29 +165,67 @@
...
+@click.option(
+    "--is-padding-free",
```
RobotSail, line 199: Whether or not we're training a padding-free transformer. For instance, llama architectures are not padding free, but GPTMegatron and GPTDolomite (granite) architectures are padding-free. 

In general when training granite, this should be set to true, but is not necessarily true.

src/instructlab/model/train.py
```diff
@@ -87,29 +165,67 @@
...
+    type=bool,
+)
+@click.option(
+    "--gpus",
+    "nproc_per_node",
+    type=int,
+    help="this is the number of GPUs to use. This is a torch specific arg and must be called nproc-per-node",
+)
+@click.option("--nnodes", type=int)
```
RobotSail, line 208: Number of machines in the training pool. So for the following configuration:
```
== machine 1 ==
GPU 0
GPU 1

== machine 2 ==
GPU 0
GPU 1
GPU 2
```
Both machines would run `ilab model train` with `--nnodes=2`. The machine 1 would provide `--nproc_per_node=2` and machine 2 would specify `--nproc_per_node=3`.

src/instructlab/model/train.py
```diff
@@ -87,29 +165,67 @@
...
+@click.option("--node-rank", type=int)
```
RobotSail, line 209: The rank of this machine in the training group. In the above example, machine 1 would have a rank of 0 and machine 2 would have a rank of 1.

src/instructlab/model/train.py
```diff
@@ -87,29 +165,67 @@
...
+@click.option("--rdzv-id", type=int)
```
RobotSail, line 210: This is the training group ID. So if there are multiple matching endpoints, we can have only the machines that we need for a specific job connect with each other.

src/instructlab/model/train.py
```diff
@@ -87,29 +165,67 @@
...
+@click.option("--rdzv-endpoint", type=str)
```
RobotSail, line 211: This is the rendezvous endpoint which other torchrun jobs will join on. Usually in the format of `url`, can be `<ip>:<port>`, `<dns-name>`, etc.

src/instructlab/configuration.py
russellb: if you're controlling both sides (server and clients), then you can test if a port number is in use before using it.

Roughly

- open a TCP socket
- bind it to localhost + a port number and see if it generates an error
- close the socket
- if you got an error binding, generate a new random high port number and try again

src/instructlab/config/init.py
```diff
@@ -115,8 +117,12 @@ def init(
         model_path = utils.expand_path(
             click.prompt("Path to your model", default=model_path)
         )
+    if train_profile != None:
+        train = config.read_train_profile(train_profile)
     click.echo(f"Generating `{config.DEFAULT_CONFIG}` in the current directory...")
     cfg = config.get_default_config()
+    if train_profile != None:
+        cfg.train = train
```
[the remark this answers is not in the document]
cdoern: thanks!

src/instructlab/configuration.py
```diff
@@ -30,6 +31,7 @@
 DEFAULT_MODEL_OLD = "merlinite-7b-lab-Q4_K_M"
 DEFAULT_MODEL = "models/merlinite-7b-lab-Q4_K_M.gguf"
 DEFAULT_MODEL_PATH = "models/merlinite-7b-lab-Q4_K_M.gguf"
+DEFAULT_MODEL_REPO = "instructlab/granite-7b-lab"
```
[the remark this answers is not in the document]
cdoern: I don't really want to change things not directly related to training, I believe all of the others are generation or serving?

src/instructlab/model/train.py
cdoern: you are right, should probably not be hidden

src/instructlab/model/train.py
cdoern: nit

src/instructlab/model/train.py
cdoern: nit

src/instructlab/configuration.py
cdoern: can't use 7b-base, not public :) switching to 7b-lab

src/instructlab/model/train.py
cdoern: yep, added handling

src/instructlab/model/train.py
cdoern: it already exists, should we leave it for now?

src/instructlab/model/train.py
cdoern: whoops! thanks

src/instructlab/model/train.py
cdoern: I think a first pass where all flags are visible is best, then we can widdle away

scripts/basic-workflow-tests.sh
```diff
@@ -133,7 +133,7 @@ test_train() {
     task Train the model
 
     # TODO Only cuda for now
-    TRAIN_ARGS=("--device=cuda" "--4-bit-quant")
+    TRAIN_ARGS=("--legacy=True" "--device=cuda" "--4-bit-quant")
```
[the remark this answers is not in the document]
cdoern: In a follow up I will be expanding the suite. Unsure of how to test given hardware access. I could add a test to run the lora part of the library. Though, should a comprehensive suite live in the actual library not here?

src/instructlab/configuration.py
RobotSail: Is 7b-lab a base model though? The input model needs to be one that hasn't been trained yet.

src/instructlab/configuration.py
JamesKunstle: huh that is weird, okay w/e not a merge-blocking problem.

src/instructlab/model/train.py
JamesKunstle: yeah if it's not breaking. we can refactor later.

src/instructlab/model/train.py
```diff
@@ -17,14 +24,97 @@
...
+                "are supported for now.",
+                param,
+                ctx,
+            )
+
+        # Detect CUDA/ROCm device
+        if device.type == "cuda":
+            if not torch.cuda.is_available():
+                self.fail(
+                    f"{value}: Torch has no CUDA/ROCm support or could not detect "
```
JamesKunstle, line 64: This is actually something that should be reported differently for merge.

We ought to say that CUDA isn't *available* rather than saying it's *unsupported*, the former is the canonical thing to report if the user's drivers or something are busted, the latter implies that we're using CUDA-less Torch, which we definitely aren't 😁

src/instructlab/model/train.py
```diff
@@ -91,28 +175,87 @@
...
+    type=bool,
+    help="whether or not we are training a padding free transformer.",
+)
+@click.option(
+    "--gpus",
+    "nproc_per_node",
+    type=int,
+    help="this is the number of GPUs to use. This is a torch specific arg and must be called nproc-per-node",
+)
+@click.option("--nnodes", type=int, help="number of machines in the training pool.")
```
JamesKunstle, line 221: for the CLI I think we should remove `--nnodes`, `--node-rank`, `--rdzv-id`, and sensibly default `---rdzv-endpoint` because we aren't supporting multi-node torchrun. Those configuration are purely for multi-proc + multi-node.

JamesKunstle (dismissed):
I'm approving because I'll leave the "blocking-ness" of my requests up to Charlie's discretion. However, I do think that removing the interface for distributed torchrun would be nice because I'd like to hide that training detail from the user, especially because the multi-machine situation is 100% not supported by the CLI.

cdoern:
whoops sorry @JamesKunstle I forgot to pull @russellb changes down :sweat:

src/instructlab/model/train.py
cdoern: I honestly think providing all flags and letting the library decided what they are going to allow in terms of options is the right move but I don't feel too strongly here

src/instructlab/configuration.py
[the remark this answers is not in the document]
nathan-weinberg: It's just really confusing in it's current form - would you be amendable to just renaming the var to something like `DEFAULT_TRAINING_MODEL_REPO`?

src/instructlab/model/train.py
nathan-weinberg: Great - let's put that in the help text :smile:

src/instructlab/model/train.py
nathan-weinberg: Readability is important for users :smile:

src/instructlab/model/train.py
nathan-weinberg: Easy to implement, code looks less sloppy

nathan-weinberg:
@cdoern I have some outstanding comments that were never addressed - I can look at approving once they are

src/instructlab/model/train.py
cdoern: done.

src/instructlab/model/train.py
cdoern: added

src/instructlab/model/train.py
cdoern: done.

src/instructlab/model/train.py
cdoern: alright, done

RobotSail (dismissed):
LGTM

nathan-weinberg (dismissed):

src/instructlab/configuration.py
```diff
@@ -147,6 +157,11 @@ def api_base(self):
         return get_api_base(self.host_port)
 
 
+class _train(BaseModel):
+    train_args: TrainingArgs
+    torch_args: TorchrunArgs
```
n1hility, line 162: I don't consider this blocking but note that there is an API encapsulation problem here. The ilab CLI project governs the config file format, yet it's using a library dep to know how to parse the configuration and parameters. Any sort of change to the library API class will silently fail in the CLI code, and alter the CLI contract (issue new values etc).  Although this would be fine if it was part of the same code base. A common strategy to deal with this is to have a mapping class that copies the values across.

cdoern: I actually just had this convo with Nathan. I am going to investigate having our own types that mirror the training ones but need to be manually updated in a future PR.

n1hility:
> `--train-profile` is an argument to `ilab config init`. This flag will populate a new portion of the `config.yaml` called `train` with the options specified in the user's yaml. I thought this made more sense than forcing them to edit the config.yaml directly OR creating a full config.yaml with the train settings pre-specified.
> 
> In my mind, this is a step towards global profiles for all commands that can either live in the config.yaml or as a separate entity. We didn't get enough consensus on where profiles would be represented because folks viewed them as something we should focus on later. However, I was trying to convey that PRs like this will necessitate at least some direction for profiles to be made.

Thanks for clarifying the override aspect, I previously missed that, and that addresses a lot of my concern! I still feel introducing a secondary config mechanism is a step sidewise from where we want to head. When we introduce profiles I assume we would retire this right away. Would we still support its usage?

Another issue is will we find ourselves applying in several areas. Training profile, generation profile etc, and then finally introduce profiles. IMO the eventual direction we want (or I thought we were heading to in discussions) is that a profile would cross all these domains.

An alternative could have been a generic config merge, where you do ilab config init --include blah.yml. That could have been closer to profiles but it would still have a problem in that profiles would invalidate it. 

All that said, I don't want to hold you up, but would be great to chat about where to go from here.

n1hility (dismissed):

Maxusmusti (approved):
LGTM

nathan-weinberg (approved):

RobotSail (approved):
LGTM

n1hility (dismissed):

JamesKunstle (dismissed):

cdoern:
I rebased, and added one more timeout bump.

JamesKunstle:
We're failing one MacOS test. I manually ran the test to see if it's a test problem or a real problem. The manual testing passed.

i.e. if the server is running and we are in a chat window, and we kill the server while chat is emitting, the chat window will exit and let the user know that it's ceasing to emit a response.

I used:
`watch -n 1 sudo lsof -iTCP -sTCP:LISTEN -nP`

to validate that the serving process that we killed was the one that the chat instance was communicating on.

JamesKunstle (approved):

RobotSail (approved):
LGTM

russellb:
For future reference, this commit history could have used a bit of squashing cleanup prior to merging.

PR #1374, opened 2024-06-16
fix: Remove DEFAULT_YAML_RULES from config

drpaneas:
This variable is never used, as it gets shadowed by the `utils.py`:

```python
DEFAULT_YAML_RULES = """\
extends: relaxed

rules:
  line-length:
    max: 120
"""
```


leseb (approved):
Thanks!

leseb (changes requested):
One small, could you update your commit message? The body is empty, it only has a title and a sign-off. It'd great if you could include:

"""
This variable is never used, as it gets shadowed by the `utils.py`:

```python
DEFAULT_YAML_RULES = """\
extends: relaxed

rules:
  line-length:
    max: 120
"""
```
"""

Thanks!

drpaneas:
@leseb done ;)

leseb (approved):
Perfect, thanks!

russellb (approved):
thanks!

PR #1382, opened 2024-06-17
enable local document sources to remove dependency on remote git fetching

relyt0925:
Currently in the generation process: if any knowledge document or grounded skill is being processed: there is a hard dependency on needing to fetch the documents from the remote git source. For potential private taxonomy sources: this requires the ilab cli git processes to be configured to talk to the private git sources storing the documents. This introduces the ability to add a local document directory in the knowledge or grounded skill directory that can hold all document contents. The SDG process can then read all content from this directory meaning if the content has already been prepulled there is no need to dynmaically reach out to the remote git repo

**Checklist:**

### DIFF
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -534,10 +534,15 @@ def read_taxonomy_file(
         tax_path = "->".join(taxonomy_path.parent.parts)
         task_description = contents.get("task_description")
         documents = contents.get("document")
-        if documents:
+        local_document_directory = os.path.join(file_path.parent, "documents")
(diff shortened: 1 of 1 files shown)


bjhargrave:
I am not sure this is a good idea. 

See some discussion around knowledge sources in:

https://github.com/instructlab/instructlab/pull/1160#issuecomment-2113667565

And any change in knowledge sources would need to be coordinated across the frontend, backend, schema, and taxonomy repos. So such changes should start with a design document in dev-docs to gain consensus on a design which can then be properly documented for users.

Issue #1357, opened 2024-06-14
[doc] Issue of generating data use larger model in TROUBLESHOOTING.md 


chudegao:
In [TROUBLESHOOTING.md](https://github.com/instructlab/instructlab/blob/main/TROUBLESHOOTING.md#how-to-improve-the-quality-of-generated-data) , it said user can use ' --model'  to use better models. But this parameter is not used to specify llm used to generate(see #1230 ). Currently, the only way is first serve a model by 'ilab serve ...' , then run 'ilab generate --endpoint-url ...'.

```
3. Using a better model via --model Larger models can lead to better data generation. This option requires users to be familiar with various existing models, and which specific models would suit their needs. This could mean either using a model with more nodes than the default InstructLab merlinite-7b-lab model, such as the Mixtral-8x7B-Instruct-v0.1 model, or using an unquantized version of the InstructLab merlinite-7b-lab model. It can be used as follows: ilab generate --model Mixtral-8x7B-Instruct-v0.1'

```

PR #1387, opened 2024-06-17
fix: support chat/generate command use custom model path(via '--model')

### LINKED ISSUE #1358: `ilab chat/generate` support use model path to serve llm
Command 'ilab chat/generate' does not support use custom model path to serve llm for chat/generate in command options.  Current parameter '--model' is not used for this purpose(very confused also). 

Hope 'ilab chat/generate' can support '--model-path' parameter like 'ilab serve'.  If reuse '--model' parameter, make the help message more clear.

May also consider support other parameters as serve like: --gpu-layers, --max-ctx-size, --num-threads

chudegao:
Resolves #1358 #1357 

**Checklist:**

- [* ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
- [* ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [* ] Documentation has been updated, if necessary.
- [* ] Unit tests have been added, if necessary.
- [* ] Integration tests have been added, if necessary.

Before the fix, can not use mixtral-8x7b model for chat/generate:
```
(dlipy3) [root@llm opt]# ilab chat --model chu/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf
You are using an aliased command, this will be deprecated in a future release. Please consider using `ilab model chat` instead
Executing chat failed with: Model chu/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf is not served by the server. These are the served models: ['models/merlinite-7b-lab-Q4_K_M.gguf']

(dlipy3) [root@llm opt]# ilab generate --model chu/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf --quiet
You are using an aliased command, this will be deprecated in a future release. Please consider using `ilab data generate` instead
Generating synthetic data using 'chu/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf' model, taxonomy:'taxonomy' against http://127.0.0.1:36043/v1 server
Cannot find prompt.txt. Using default prompt depending on model-family.
  0%|                                                                                     | 0/100 [00:00<?, ?it/s]INFO 2024-06-17 07:41:53,444 generate_data.py:505: generate_data Selected taxonomy path knowledge->textbook->movies->awards->oscar
Generating dataset failed with the following error: Model chu/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf is not served by the server. These are the served models ['models/merlinite-7b-lab-Q4_K_M.gguf']
  0%|                                                                                     | 0/100 [00:00<?, ?it/s]
(dlipy3) [root@llm opt]# 
```

After the fix:
```
(dlipy3) [root@llm opt]# ilab chat --model chu/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf
You are using an aliased command, this will be deprecated in a future release. Please consider using `ilab model chat` instead
╭──────────────────────────────────────────────────── system ────────────────────────────────────────────────────╮
│ Welcome to InstructLab Chat w/ CHU/MIXTRAL-8X7B-INSTRUCT-V0.1.Q4_K_M.GGUF (type /h for help)                   │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
>>> hello 

(dlipy3) [root@llm opt]# ilab generate --model chu/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf --num-instructions 10
You are using an aliased command, this will be deprecated in a future release. Please consider using `ilab data generate` instead
Generating synthetic data using 'chu/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf' model, taxonomy:'taxonomy' against http://127.0.0.1:45245/v1 server
Cannot find prompt.txt. Using default prompt depending on model-family.
  0%|                                                                                      | 0/10 [00:00<?, ?it/s]Synthesizing new instructions. If you aren't satisfied with the generated instructions, interrupt training (Ctrl-C) and try adjusting your YAML files. Adding more examples may help.
INFO 2024-06-17 07:51:00,450 generate_data.py:505: generate_data Selected taxonomy path knowledge->textbook->movies->awards->oscar
Q> What was the inspiration behind the stage design for the 2024 Oscars?
I> 
A> The stage was inspired by contemporary spaces where people can meet, exchange, and create, similar to a modern-day plaza.

 10%|███████▊ 


```

### DIFF
--- a/src/instructlab/client.py
+++ b/src/instructlab/client.py
@@ -18,6 +18,10 @@ class ClientException(Exception):
     """An exception raised when invoking client operations."""
 
 
+class ModelCheckException(Exception):
+    """An exception raised when checking model."""
+
+
 def list_models(
     api_base,
     api_key=DEFAULTS.API_KEY,
@@ -43,3 +47,23 @@ def check_api_base(api_base: str, http_client: httpx.Client | None = None) -> bo
         return True
     except ClientException:
         return False
+
+
+def check_model(model, api_base: str, endpoint_url=None, http_client=None):
+    """Check if right model is selected or served."""
+    if endpoint_url:
+        api_base = endpoint_url
+    model_list = list_models(
+        api_base=api_base,
+        http_client=http_client,
+    )
+    if len(model_list.data) == 0:
+        raise ModelCheckException("No model is served")
+
+    model_ids = [m.id for m in model_list.data]
+    if model not in model_ids:
+        if endpoint_url:
+            msg = f"Model {model} is not served by the endpoint url {endpoint_url}. These are the served models: {model_ids}. Use '--model' to select right model to use."
+        else:
+            msg = f"Model {model} is not served by the server. These are the served models: {model_ids}. Use '--model' to select right model to use. Serve the model {model} by running 'ilab serve --model-path {model}' or update the configuration file"
+        raise ModelCheckException(msg)
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -9,6 +9,7 @@ import click
 
 # First Party
 from instructlab import clickext
+from instructlab import client as ilabclient
 from instructlab.configuration import DEFAULTS
 
 # Local
@@ -22,7 +23,7 @@ logger = logging.getLogger(__name__)
     "--model",
     default=lambda: DEFAULTS.DEFAULT_MODEL,
     show_default="The default model used by the instructlab system, located in the data directory.",
-    help="Name of the model used during generation.",
+    help="Name of the model used during generation. Note '--model' is used to select model rather than serve model.",
 )
 @click.option(
     "--num-cpus",
@@ -196,6 +197,7 @@ def generate(
             click.secho(f"Failed to start server: {exc}", fg="red")
             raise click.exceptions.Exit(1)
     try:
+        ilabclient.check_model(model, api_base, endpoint_url, http_client(ctx.params))
         click.echo(
             f"Generating synthetic data using '{model}' model, taxonomy:'{taxonomy_path}' against {api_base} server"
         )
@@ -222,7 +224,7 @@ def generate(
             tls_client_passwd=tls_client_passwd,
             pipeline=pipeline,
         )
-    except GenerateException as exc:
+    except (GenerateException, ilabclient.ModelCheckException) as exc:
         click.secho(
             f"Generating dataset failed with the following error: {exc}",
             fg="red",
--- a/src/instructlab/model/chat.py
+++ b/src/instructlab/model/chat.py
@@ -73,7 +73,7 @@ PROMPT_PREFIX = ">>> "
     "--model",
     cls=clickext.ConfigOption,
     required=True,  # default from config
-    help="Model name to print in chat process",
+    help="Name of the model used during chat. Note '--model' is used to select model rather than serve model.",
 )
 @click.option(
     "-c",
@@ -241,6 +241,7 @@ def chat(
                 raise click.exceptions.Exit(1) from exc
 
     try:
+        ilabclient.check_model(model, api_base, endpoint_url, http_client(ctx.params))
         chat_cli(
             ctx,
             api_base=api_base,
@@ -253,7 +254,7 @@ def chat(
             greedy_mode=greedy_mode,
             max_tokens=max_tokens,
         )
-    except ChatException as exc:
+    except (ChatException, ilabclient.ModelCheckException) as exc:
         click.secho(f"Executing chat failed with: {exc}", fg="red")
         raise click.exceptions.Exit(1)
     finally:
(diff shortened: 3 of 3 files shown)


russellb (changes requested):
I have mixed feelings on this one.

This parameter is not actually a path. It goes into OpenAI API calls as a model name. It just so happens that our defaults have it in the form of a path.

This change overloads the option to be two things:

1. the model name for the OpenAI API calls
2. overrides the model path for the built-in model server.

I do appreciate the intent behind the change and that it does technically work. I'm more worried that the overloading will bite us at some point later.

Another option I'd feel better about would be to add a new `--model-path` option that only applies when we're launching the server automatically.

... but then we could say we could duplicate ALL options to `ilab serve` under both `ilab chat` and `ilab generate` and that doesn't really seem ideal.

With all of that said, I'm inclined to maybe improve the error messages you get to make it VERY clear what you need to do if it is not clear already.  Your options in this case would be:

- update your config.yaml
- run `ilab serve` manually with the model of choice

chudegao:
@russellb thanks for the comments.  I'm fine for the baseline not support the options ilab serve supported considering further impact and maintainance effort. 

I think there are two major pain points here:

1. Really confused parameter name and help message, so many people think it will serve the model with this model. 
   Solution is updating the help message to make it clear enough.

2. The parameter is actually not needed in most case. If the non-builtin model is served by ilab(ilab serve or temporary server by chat/generate),  we can list the model and get model name then call openai api. It's not reasonable enough to ask user must specify the model in this case. If user use '--endpoint-url' and there are multiple models, the parameter is a must to choose one of models to chat/generate.  My proposal is only verify this parameter in case 2 and ignore it in other cases. The help message will be more clear also.

Use ilab serve to support non built-in model is clear enough  if 1 and 2 are solved.

BTW, for config.yaml, there's no clear doc to guide user and it will impact globally. It's can be treated as an advanced feature for some user. Maybe it can be enhanced(e.g. support command level configuration) in the future.

chudegao:
@russellb , let me know which option you prefer:

1. '--model' parameter only used when '--endpoint-url' is used, otherwise it value is set to empty. model name will be get by list_models().  This allow user no need to specify this parameter when use a non-default model.  The parameter is required only when use a third party model service via  '--endpoint-url' and has multiple models by list_models().  This option is more user friendly.

2. Keep '--model'  as before, can check model is matched with served model or will be served model and provide more clear message earlier. I'v updated readme adding some examples based on this option.

src/instructlab/data/generate.py
```diff
@@ -194,6 +194,9 @@ def generate(
         if not api_base:
             api_base = ctx.obj.config.serve.api_base()
     try:
+        backends.check_model(
+            model, api_base, endpoint_url, http_client(ctx.params), ctx
+        )
```
russellb, line 199: This isn't really a function of the model serving backend. At this point in the code, we may or may not be using `ilab` to run the model server.

russellb (changes requested):
Thanks - I like the direction.  2 things

- I don't think `backends` is the right package for this, since it's more of a client-side thing than server-side
- would prefer a new exception type than just using `Exception`

thanks!

src/instructlab/client.py
```diff
@@ -28,3 +32,31 @@ def list_models(
...
+    )
+    if len(model_list.data) == 0:
+        raise ModelCheckException("No model is served")
+
+    model_ids = [m.id for m in model_list.data]
+    if model not in model_ids:
+        if endpoint_url:
+            # user not specify --model parameter
+            if model in (DEFAULT_MODEL, ctx.obj.config.chat.model):
+                msg = f"Model {model} is not served by the endpoint url {endpoint_url}. These are the served models: {model_ids}. Use '--model' to select right model to use."
```
russellb, line 53: How about we just always use this message? It'll simplify the code and the extra instructions never hurt.

src/instructlab/client.py
```diff
@@ -28,3 +32,31 @@ def list_models(
...
+            else:
+                msg = f"Model {model} is not served by the endpoint url {endpoint_url}. These are the served models: {model_ids}"
+        else:
+            # user not specify --model parameter
+            if model in (DEFAULT_MODEL, ctx.obj.config.chat.model):
+                msg = f"Model {model} is not served by the server. These are the served models: {model_ids}. Use '--model' to select right model to use."
+            else:
+                msg = f"Model {model} is not served by the server. These are the served models: {model_ids}. Serve the model {model} by running 'ilab serve --model-path {model}' or update the configuration file"
```
russellb, line 61: Same thing here. I think we could just always emit the extra instructions.

russellb (commented):
Thank you for the continued follow-up on this! I'm sorry that I keep coming up with comments every time I read it again. I think it's really close ...

chudegao:
> Thank you for the continued follow-up on this! I'm sorry that I keep coming up with comments every time I read it again. I think it's really close ...

Glad to get your comments. It's good idea to provide a more general instructions considering we already provide details in help message and readme. Let me update.

PR #1397, opened 2024-06-17
docs: Improve LoRA doc by adding var and file name

rascasoft:
This commit extends the data upload section of the Collab instructions by adding references to the corresponding variable names for the uploaded files. This should save time to people that ask themselves "where do I put this path that I've copied?"


russellb (approved):
thanks!

nathan-weinberg (approved):

PR #1399, opened 2024-06-17
WIP: Profiles for `ilab`

cdoern:
this is built on top of the training PR


russellb (commented):
I know this is a draft and you're working some things out, but just as some early feedback just in case -- it would be very helpful to not include any changes to the arguments to any comments in the same PR as introducing the profiles feature

PR #1408, opened 2024-06-18
schema: Remove references to schema folder

bjhargrave:
The schema is now provided by the instructlab-schema python package and we no longer submodule the code. So we should remove all references to the removed src/instructlab/schema folder in this repo.

**Checklist:**

### DIFF
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -16,7 +16,6 @@ ignores:
   - ".tox/**"
   - ".venv/**"
   - "build/**"
-  - "src/instructlab/schema/**"
   - "tests/**"
   - "venv/**"
   - ".github/ISSUE_TEMPLATE/**"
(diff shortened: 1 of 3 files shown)


nathan-weinberg (approved):

bjhargrave:
> Thanks, one question, should test_diff_failing_schema_yaml() test move to instructlab-schema? Or does it still make sense to have it here? Thanks!

The schema repo currently does not have any real code. It is a delivery vehicle for the schema json files. 

If we think we want to move code for reading taxonomy files and schema validation into the schema repo, then it would make sense to also move such testing for valid/invalid taxonomy yaml files.

Since both `ilab taxonomy diff` and the sdg code both need to read the taxonomy yaml files, there may be some value in collecting that code into a central package like instructlab-schema. Currently sdg needs to reach back into the cli package to read the taxonomy which is a circular dependency...

russellb:
I like the idea for using that lib for both reading and validating taxonomy files

PR #1415, opened 2024-06-18
Add backend configuration and auto-detection support

leseb:
This commit introduces the ability to configure and automatically detect
the backend for serving models in InstructLab. One new field, `backend`
has been added to the configuration and the CLI to allow users to
specify the backend. This field is optional.

Additionally, the commit includes the implementation of backend validation and
auto-detection based on the model file properties, specifically targeting GGUF
file format for now. The `serve` command has been updated to utilize these new
configurations and perform backend validation or auto-detection as required.

Unit tests for backend validation and changes to the configuration tests to
include the new fields are also part of this commit.

Co-authored-by: Ali Maredia <amaredia@redhat.com>

**Checklist:**


src/instructlab/model/backends/backends.py
```diff
@@ -4,6 +4,9 @@
 import abc
 import logging
 import pathlib
+import sys
+
+supported_backends = ["llama"]
```
tiran, line 9:
```suggestion
SUPPORTED_BACKENDS = frozenset({"llama"})
```

src/instructlab/model/backends/backends.py
```diff
@@ -29,3 +32,58 @@ def run(self):
...
+    Args:
+        file_path (Path): The path to the file.
+    Returns:
+        bool: True if the file is a GGUF file, False otherwise.
+    """
+    # Third Party
+    import gguf
+
+    try:
+        gguf.GGUFReader(model_path, "r")
```
tiran, line 49: That approach is very slow. It takes almost 2 seconds on my laptop. I guess it loads the entire file into memory. IMO it's sufficient to check that the file is present and has the right magic header:

```
>>> from gguf.constants import GGUF_MAGIC
>>> import struct
>>> with open("models/merlinite-7b-lab-Q4_K_M.gguf", "rb") as f:
...     buf = f.read(4)
>>> header = struct.unpack("<I", buf)
>>> header[0] == GGUF_MAGIC
```

russellb (dismissed):

russellb (approved):

russellb (approved):

russellb (approved):

russellb:
I missed that the corresponding design doc has unresolved discussion on whether `--backend foo` should be added or not.

https://github.com/instructlab/dev-docs/pull/81

I think that needs to be clearly resolved before it'd be appropriate to merge this. I care more about ensuring there's consensus than which way it goes on this particular issue.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## v0.18
+
+### Features
+
+1. Add '--backend' flag to `ilab model serve` command to allow for specifying the backend to use
+   when serving a model. This is useful when you have multiple backends installed and want to
+   specify which one to use. Currently, the only supported backend is `llama`.
```
chudegao, line 7: llama is well known as Meta's LLM, maybe use llama.cpp/llama-cpp/llama-cpp-python is better?

leseb: Good point thanks, I'll rename it to "llama-cpp" instead.

chudegao (approved):
LGTM

src/instructlab/model/backends/backends.py
```diff
@@ -29,3 +32,107 @@ def run(self):
...
+        # Read the first 4 bytes
+        first_four_bytes = mmapped_file.read(4)
+
+        # Convert the first four bytes to an integer
+        first_four_bytes_int = struct.unpack("<I", first_four_bytes)[0]
+
+        # Close the memory-mapped file
+        mmapped_file.close()
+
+        return first_four_bytes_int == GGUF_MAGIC
```
tiran, line 65: Why do you use mmap here? I kinda get it. `mmap` is a great way to deal with large files and too few people know about the power of mmap. It seems a bit overpowered here, though. A simple `read()` would do the trick, too. Also it's a good idea to close the file as early as possible and do the rest of the work outside the context manager.

```suggestion
    with open(model_path, "rb") as f:
        first_four_bytes = f.read(4)

    # Unpack first four bytes as little endian uint32
    first_four_bytes_int = struct.unpack("<I", first_four_bytes)[0]

    return first_four_bytes_int == GGUF_MAGIC
```

pyproject.toml
```diff
@@ -136,7 +136,7 @@ exclude = [
   "^src/instructlab/train/lora_mlx/models/mixtral\\.py$",
   "^src/instructlab/train/lora_mlx/models/models\\.py$",
   "^src/instructlab/utils\\.py$",
-  "^src/instructlab/model/backends/llama\\.py$",
+  "^src/instructlab/model/backends/llama_cpp\\.py$",
```
tiran, line 139: Are there still typing violations in the file?

src/instructlab/model/backends/backends.py
leseb: Absolutely, I'm using it for efficiency. It might be a bit overpowered, but at least it helps more people become aware of it 😅. Anyway, I'll just proceed with a simple read.

pyproject.toml
leseb: Yes it does, this file used to be `server.py`. That PR should fix it https://github.com/instructlab/instructlab/pull/1275. Thanks!

src/instructlab/model/backends/backends.py
```diff
@@ -29,3 +32,99 @@ def run(self):
...
+    """
+    # Check if the model is a GGUF file
+    logger.debug(f"Auto-detecting backend for model {model_path}")
+    auto_detected_backend = determine_backend(model_path)
+
+    logger.debug(f"Auto-detected backend: {auto_detected_backend}")
+    # When the backend is not set using the --backend flag, determine the backend automatically
+    # 'backend' is optional so we still check for None or empty string in case 'config.yaml' hasn't
+    # been updated via 'ilab config init'
+    if backend is None or backend == "":
```
tiran, line 115: This should never be `None` because the function requires a `str` type input.
Or maybe it would be better to change the backend variable to `Optional[str]` and only support `None` or a valid name.

leseb: I went with supporting None and added `Optional[str]`. Thanks for the recommendation!

leseb:
https://github.com/instructlab/instructlab/pull/1442 contains all of that now.

PR #1426, opened 2024-06-19
fix: clean training result final directory to make sure training can work after a failure training

### LINKED ISSUE #1416: ilab train fails with `file exists` if its already been [partially?] trained.
**Describe the bug**

I had a failure running train, and when I ran it a second time I got an error to say that the destination path already exists.  Likely it should ask if I want to overwrite the previous training before it gets too far.

```
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/instruct/instructlab/venv/lib64/python3.11/site-packages/instructlab/model/train.py", line 273, in train
    shutil.move(file, final_results_dir)
  File "/usr/lib64/python3.11/shutil.py", line 823, in move
    raise Error("Destination path '%s' already exists" % real_dst)
shutil.Error: Destination path 'training_results/final/model-00001-of-00003.safetensors' already exists
```

chudegao:
Resolves #1416

**Checklist:**

### DIFF
--- a/src/instructlab/model/train.py
+++ b/src/instructlab/model/train.py
@@ -244,7 +244,9 @@ def train(
         )
 
         final_results_dir = training_results_dir / "final"
-        final_results_dir.mkdir(exist_ok=True)
+        if final_results_dir.exists():
+            shutil.rmtree(final_results_dir)
+        final_results_dir.mkdir()
 
         gguf_models_dir = Path("./models")
         gguf_models_dir.mkdir(exist_ok=True)


src/instructlab/model/train.py
```diff
@@ -244,7 +244,9 @@ def get_files(pattern):
         )
 
         final_results_dir = training_results_dir / "final"
-        final_results_dir.mkdir(exist_ok=True)
+        if final_results_dir.exists():
```
leseb, line 247: This just checks the existence of the directory right? Shouldn't we verify the directory as files instead? How about always cleaning the dir instead of the condition on `final_results_dir.exists()`?

chudegao: final_results_dir.exists() is a must as shutil.rmtree() will hit error if the directory does not exist

final_results_dir is created by ilab train, it will always be a directory. 

Only if user manually create final_results_dir as a file, we need to check it's a file and remove the file. If this is a valid case, I can add more check. Let me know your further comments.

leseb: Thanks for the clarification.

leseb (approved):

tiran (approved):

PR #1434, opened 2024-06-20
Improve support for instructlab NS packages

tiran:
- print version of `instructlab.sdg` in `ilab sysinfo`
- fix `tox -e fastlint` / `make verify`
- configure `isort` to make new `instructlab` namespace packages to a a new import section.

**Checklist:**

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,10 +1,13 @@
 [settings]
 profile=black
 from_first=true
+sections=FUTURE,STDLIB,THIRDPARTY,INSTRUCTLABNS,FIRSTPARTY,LOCALFOLDER
 import_heading_future=Future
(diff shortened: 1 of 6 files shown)


leseb (approved):
Dividing into individual commits would have made sense but I'm not going to block on this. Thanks!

markstur:
why not treat them as third-party?  I'm a big fan of sort and the grouping of std, local, 3rd... but those isort comments and adding more groups doesn't make the code look better to me.  Just longer.  So... just asking.

tiran:
The fastlint fix, verify fix, and sysinfo improvement have landed with other PRs. The new isort section for 2nd party instructlab packages got some opposition. Let's not land that change.

PR #1436, opened 2024-06-20
Update README.md

laurachiticariu:
Clarify system requirements upfront.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,9 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 
 ## 📋 Requirements
 
-- **🍎 Apple M1/M2/M3 Mac or 🐧 Linux system** (tested on Fedora). We anticipate support for more operating systems in the future.
+- **🍎 Apple M1/M2/M3 Mac or 🐧 Linux system** (tested on Fedora). Note Linux
+ is not fully supported (testing a trained model does not currently work on Linux).
+  We anticipate support for more operating systems in the future.
 - C++ compiler
 - Python 3.9+ (<3.12 for PyTorch JIT)
 - Approximately 60GB disk space (entire process)


README.md
```diff
@@ -77,7 +77,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 
 ## 📋 Requirements
 
-- **🍎 Apple M1/M2/M3 Mac or 🐧 Linux system** (tested on Fedora). We anticipate support for more operating systems in the future.
+- **🍎 Apple M1/M2/M3 Mac or 🐧 Linux system** (tested on Fedora). Note the Linux is only partially supported (serving and testing a trained model does not work on Linux). We anticipate support for more operating systems in the future.
```
joesepi, line 80:
```suggestion
- **🍎 Apple M1/M2/M3 Mac or 🐧 Linux system** (tested on Fedora). Note Linux
 is only partially supported (serving and testing a trained model does not work on Linux).
  We anticipate support for more operating systems in the future.
```

joesepi (approved):
LGTM. Thanks!

README.md
```diff
@@ -77,7 +77,9 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 
 ## 📋 Requirements
 
-- **🍎 Apple M1/M2/M3 Mac or 🐧 Linux system** (tested on Fedora). We anticipate support for more operating systems in the future.
+- **🍎 Apple M1/M2/M3 Mac or 🐧 Linux system** (tested on Fedora). Note Linux
+ is only partially supported (serving and testing a trained model does not work on Linux).
```
chudegao, line 81: I think serving is supported already,  test and convert is not supported on linux so far.  There's a [PR](https://github.com/instructlab/instructlab/pull/1337) under review for test support. Convert is not needed for linux after train as trained model is a converted model already.

leseb: Indeed serving is supported.

leseb (changes requested):
Please rephrase based on comment reviews. Thanks

README.md
laurachiticariu: I have learned from Akash Srivastava that while test and serve is supported for the base models, it is not supported for models trained through `ilab train` today. You are saying there is a PR in the works to support this. Once the [PR you mention](https://github.com/instructlab/instructlab/pull/1337) is merged, it should change this readme accordingly. Until then, I'd appreciate if you merge this to avoid confusing users who are testing this now.

bjhargrave: After reading through all the discussion here, I think the proposed text is valid and correct:

> Note Linux is only partially supported (serving and testing a trained model does not work on Linux).

So I think we can merge this for now and update later (along with https://github.com/instructlab/instructlab/blob/main/README.md#-test-the-newly-trained-model) when support is added to allow linux to serve and test an ilab trained model.

chudegao: serve a trained model on linux is supported. Here's an example:
```
(dlipy3) [root@llm chu]# ilab serve --model-path models/ggml-model-f16.gguf
You are using an aliased command, this will be deprecated in a future release. Please consider using `ilab model serve` instead
INFO 2024-06-25 04:57:20,095 serve.py:56: serve Using model 'models/ggml-model-f16.gguf' with -1 gpu-layers and 4096 max context size.
INFO 2024-06-25 04:57:22,738 llama.py:268: server Starting server process, press CTRL+C to shutdown server...
INFO 2024-06-25 04:57:22,738 llama.py:269: server After application startup complete see http://10.244.128.4:4001/docs for API.
```
models/ggml-model-f16.gguf is trained model on linux.

ilab model convert is only need for mac. Maybe some updates are needed in 'Serve the newly trained model' part to make mac and linux serve more clear.

laurachiticariu: Ok, so training a model on Linux is not supported on Linux. But serving a trained model is supported on Linux? Did I get this right? So the full life cycle is still not supported on Linux. Since I don't have a Linux machine to verify this statement, how do you all propose to modify this PR to ensure the statements are accurate? Can someone with knowledge of the code actually refresh this PR with the accurate wording so users don't get confused? Since 5 days ago when I proposed this change, I got another 10 users who complained the top level statement on Linux support is misleading.

markstur (approved):
This is at least an improvement.

README.md
chudegao: training on linux is supported also.

chudegao: But serving a trained model is supported on Linux?   -- Yes. serveing is supported.  

So the full life cycle is still not supported on Linux.   --- I don't think so. Actually, testing is part of training on linux, you can see the test result when do training.(just can not do test separately).

chudegao: But serving a trained model is supported on Linux?   -- Yes. serving is supported.  

So the full life cycle is still not supported on Linux.   --- I don't think so. Actually, testing is part of training on linux, you can see the test result when do training.(just can not do test separately).

joesepi, line 81: Is this accurate? If so, can we get the block lifted @leseb @chudegao 🙏 

```suggestion
- **🍎 Apple M1/M2/M3 Mac or 🐧 Linux system** (tested on Fedora). Note Linux
 is not fully supported (testing a trained model does not currently work on Linux).
```

bjhargrave: This will need to be added to the PR before approval.

joesepi:
@leseb -- I've committed my suggested change in the hopes that it accurately reflects the current state of the project and can be merged. Please and thank you! :)

joesepi (approved):

leseb:
@joesepi for next time, it'd be nice to squash commits, thanks!

nathan-weinberg:
I can squash real quick

laurachiticariu:
Thank you all!

PR #1441, opened 2024-06-21
change instructlab logger instead of root

makelinux:
Setting the root log level to DEBUG or INFO generates noisy logs from all modules, not just from openai and httpx.

Changing only the instructlab log level produces more desirable results and avoids the need for workarounds like explicitly setting the log level of external modules.

### DIFF
--- a/src/instructlab/log.py
+++ b/src/instructlab/log.py
@@ -68,7 +68,7 @@ def configure_logging(
     stream = logging.StreamHandler()
     stream.setFormatter(CustomFormatter(fmt=fmt))
     root.addHandler(stream)
-    root.setLevel(log_level)
+    logging.getLogger("instructlab").setLevel(log_level)
 
     # Set logging level of OpenAI client and httpx library to ERROR to suppress INFO messages
     logging.getLogger("openai").setLevel(logging.ERROR)


tiran (changes requested):
Please keep the code that modifies the default logger for openai and httpx.

src/instructlab/log.py
```diff
@@ -68,7 +68,7 @@ def configure_logging(
     stream = logging.StreamHandler()
     stream.setFormatter(CustomFormatter(fmt=fmt))
     root.addHandler(stream)
-    root.setLevel(log_level)
+    logging.getLogger("instructlab").setLevel(log_level)
```
bjhargrave, line 71:
```suggestion
    logging.getLogger(__name__.split(".", 1)[0]).setLevel(log_level)
```

I think something like this would be preferable to hard coding "instructLab".

tiran:
I have been thinking about the problem a bit. I would prefer to keep the current behavior and have debug level set debugging for **ALL** loggers. Debug mode is a developer feature to diagnose a problem. What if the problem is in another library?

How about we add a debug level option? Debug level 1 (or `ilab -v`) enables debug logging just for `instructlab` loggers. Debug level 2 (or `ilab -vv`) enables debug logging for all loggers. I can make a PR for the feature.

tiran:
I have created #1443, will continue working on it on Monday.

makelinux:
@tiran, I recommend merging this PR because your solution is not yet ready. Please note that this PR restores the original behavior, where only InstructLab's debug messages are printed.

BTW, you can emit a lot traces with `python -m trace --trace -C .`.

PR #1444, opened 2024-06-21
Modify E2E script for MacOS Compatibility

RobotSail:
Due to Apple distributing an older version of BASH (3.2), the e2e script
does not currently support the array operations being utilized by the current
workflow.
This PR simplifies the logic in the e2e script for compatibility with MacOS.

**Checklist:**

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euf
+set -euf 
 
 # This is a basic workflow test of the ilab CLI.
 #
@@ -39,7 +39,13 @@ set_defaults() {
     fi
 
     NUM_INSTRUCTIONS=1
-    GENERATE_ARGS+=("--num-cpus" "$(nproc)")
+	# nproc doesn't exist on MacOS, so we have to use its equivalent:
+	# `sysctl -n hw.logicalcpu`: https://github.com/memkind/memkind/issues/33#issuecomment-540614162
+	if [[ "$(uname)" == "Darwin" ]]; then
+		GENERATE_ARGS+=("--num-cpus" "$(sysctl -n hw.logicalcpu)")
+	else
+		GENERATE_ARGS+=("--num-cpus" "$(nproc)")
+	fi
     TRAIN_ARGS+=("--num-epochs" "1")
 }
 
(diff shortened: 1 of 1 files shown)


bjhargrave:
Why not just install a better version of bash?

https://github.com/instructlab/instructlab/blob/184582a0c86af46606d2e256093e2c243d6ee902/.github/workflows/test.yml#L77-L80

RobotSail:
Although this is definitely a solution, the problem is that it requires going through a third-party packaging system to do so.

Additionally, we're only using arrays in this script to pass conditional parameters to ilab.
So it seems like overkill to prescribe installing a 3rd party packaging system to upgrade bash just to run this script.


> [instructlab/.github/workflows/test.yml](https://github.com/instructlab/instructlab/blob/184582a0c86af46606d2e256093e2c243d6ee902/.github/workflows/test.yml#L77-L80)

bjhargrave:
> So it seems like overkill to prescribe installing a 3rd party packaging system

Well, we use apt to install things on the ubuntu runners. And brew is already available on the macos runners to make this straight forward.

It would be easier on future developers to use a modern bash so that they also don't have to struggle with coding to an older bash. Ideally we would use the same bash version on both macos and ubuntu runners.

RobotSail:
> > So it seems like overkill to prescribe installing a 3rd party packaging system
> 
> Well, we use apt to install things on the ubuntu runners. And brew is already available on the macos runners to make this straight forward.
> 
> It would be easier on future developers to use a modern bash so that they also don't have to struggle with coding to an older bash. Ideally we would use the same bash version on both macos and ubuntu runners.

Although I don't disagree that assuming an updated bash makes things easier, the reality is that people running this script locally will run into this problem again, and the solution (updating bash) is not obvious and will serve to frustrate future developers. 

Since our usage of arrays here is simple, updating the script to avoid using them seems like the most straightforward option.

bjhargrave:
> the reality is that people running this script locally will run into this problem again

I would certainly hope that no "real" developer would be using the system version of bash on their Mac! :-)

RobotSail:
> I would certainly hope that no "real" developer would be using the system version of bash on their Mac! :-)

I am.

scripts/basic-workflow-tests.sh
```diff
@@ -39,7 +39,13 @@ set_defaults() {
     fi
 
     NUM_INSTRUCTIONS=1
-    GENERATE_ARGS+=("--num-cpus" "$(nproc)")
+	# nproc doesn't exist on MacOS, so we have to use its equivalent:
+	# `sysctl -n hw.logicalcpu`: https://github.com/memkind/memkind/issues/33#issuecomment-540614162
+	if [[ "$(uname)" == "Darwin" ]]; then
```
markstur, line 44: the indenting is off.  You are introducing tabs into a file that uses spaces to indent.  Please keep it consistent with how it was.

markstur (changes requested):
introduces unwanted tabs

scripts/basic-workflow-tests.sh
RobotSail: Oh great catch, my editor displays them identically. Thanks!

leseb (changes requested):
Bash 3.2 is almost 20 years old. I don't think we should support it, especially if there are issues with arrays. Moreover our [doc](https://github.com/instructlab/instructlab?tab=readme-ov-file#bash-version-44-or-newer) clearly indicates we support Bash >= 4.4

Agu1l4r821 (approved):

RobotSail:
Not stale, will need to update though

courtneypacheco:
Hi @RobotSail. Are you still working on this PR? We're looking to do some housekeeping and close out stale PRs.

If we don't hear back within 7 days, we will close this PR, but please know that you are more than welcome to reopen it if you'd like! Thank you!

PR #1446, opened 2024-06-22
add host port option

reid41:
**Checklist:**


leseb (commented):
Please rebase, thanks!

nathan-weinberg:
@reid41 can you sign off your commit please?

tiran (changes requested):
The new code will always override the settings from `config.yaml`. You need a different approach that will use the default from the configuration and only overrides the settings when host and port are passed via CLI args.

You are also missing new arguments `host: str | None` and `port: int | None` and test cases to verify the code.

leseb:
@reid41 please look at the CI issues, thanks

src/instructlab/model/serve.py
```diff
@@ -64,6 +64,20 @@
...
+    show_default=True,
+    help="The host address to bind to.",
+)
+@click.option(
+    "--port",
+    type=click.INT,
+    default=None,
+    show_default=True,
+    help="The port number to bind to.",
+)
```
tiran, line 80: I recommend that you add an option `--host-port` instead. It will automatically get its default value from `config.serve.host_port`.

```suggestion
@click.option(
    "--host-port",
    type=str,
    required=True,  # default from config
    help="Listen host:port for the server (example: 'localhost:8080').",
)
```

src/instructlab/model/serve.py
```diff
@@ -76,12 +90,18 @@ def serve(
     log_file,
     backend,
     vllm_args: typing.Iterable[str],
+    host: str | None,
+    port: int | None,
```
tiran, line 94:
```suggestion
    host_port: str,
```

src/instructlab/model/serve.py
```diff
@@ -76,12 +90,18 @@ def serve(
...
     # First Party
     from instructlab.model.backends import llama_cpp, vllm
-
-    host, port = utils.split_hostport(ctx.obj.config.serve.host_port)
+    
+    config_host, config_port = utils.split_hostport(ctx.obj.config.serve.host_port)
+    if host is None:
+        host = config_host
+    if port is None:
+        port = config_port
```
tiran, line 104:
```suggestion
    host, port = utils.split_hostport(host_port)
```

tests/test_host_port.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-License-Identifier: Apache-2.0
+# pylint: disable=duplicate-code
+
+from click.testing import CliRunner
+from instructlab import lab
+
+
+class TestServe:
```
tiran, line 8: pytest does not need a test class. Please use functions here.

tests/test_host_port.py
```diff
@@ -0,0 +1,39 @@
...
+    def test_serve_with_defaults(self):
```
tiran, line 9: Use [pyest.mark.parametrize](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html) instead of repeat the same test code over and over again.

tiran (changes requested):

reid41:
@leseb hi, updated

leseb:
@reid41 please add a sign-off to your commit message and run `make fix` locally, this should address the CI issues. Thanks

reid41:
@leseb help to take a look?

PR #1453, opened 2024-06-24
docs: Auto generate cli_reference.md

### LINKED ISSUE #1449: Auto generate CLI commands for reference
This PR #1352  adds a manual CLI reference to the InstructLab documentation. @russellb proposed auto generating this list.

russellb:
This file serves as a reference for the CLI commands. Add a script to
generate a reference that includes the help text for every command.

Closes #1449


nathan-weinberg:
Should we also update the README to point to this file instead of having the static example?

russellb:
> Should we also update the README to point to this file instead of having the static example?

sure, that seems like a good idea.

leseb (changes requested):
Please add a CI test that runs `make cli-ref` and then makes sure there is no diff, so whoever adds a new command/flag knows the cli reference needs to be updated. Thanks!

leseb:
How does this cope with https://github.com/instructlab/instructlab/pull/1578? Aren't they both achieving the same thing?

russellb:
> How does this cope with #1578? Aren't they both achieving the same thing?

Yeah, we don't need both. That one looks like it's nicer, but I haven't really read through it properly. We can definitely close this if/when we merge that one, though.

ktam3:
@russellb - can we close this?

PR #1454, opened 2024-06-24
e2e: Add skills to e2e testing

russellb:
The e2e workflow included a single sample knonwledge qna. Expand this
to included both a grounded and freeform skill, as well. This will
ensure the test covers more of the data generation code paths.


scripts/basic-workflow-tests.sh
```diff
@@ -110,12 +110,16 @@ test_taxonomy() {
...
-    cp "$SCRIPTDIR"/test-data/basic-workflow-fixture-qna.yaml taxonomy/knowledge/sports/overview/softball/qna.yaml
-    head taxonomy/knowledge/sports/overview/softball/qna.yaml | grep --color '1st base'
+    step Update taxonomy with sample qna additions
+    mkdir -p taxonomy/knowledge/sports/overview/e2e-softball
+    mkdir -p taxonomy/compositional_skills/extraction/answerability/e2e-yes_or_no
+    mkdir -p taxonomy/compositional_skills/extraction/inference/qualitative/e2e-siblings
+
+    step Put new qna files into place
+    cp "$SCRIPTDIR"/test-data/e2e-qna-knowledge.yaml taxonomy/knowledge/sports/overview/e2e-softball/qna.yaml
+    head taxonomy/knowledge/sports/overview/e2e-softball/qna.yaml | grep --color '1st base'
```
leseb, line 120: Why `head`? Could we get multiple `'1st base'` strings in the file? Also I think the error handling for this case could improve. How about we put this through a `if ! command` and log an error message?

russellb: This is not new code

leseb: Oh.. right. Nevermind then.

russellb:
I'm going to make some more changes here to ensure each taxonomy type is exercised

russellb:
This is ready for review again. It changes the test workflow to put one sample qna.yaml in place at a time and generating 1 sample with them.

nathan-weinberg (approved):

aakankshaduggal (approved):

scripts/basic-workflow-tests.sh
```diff
@@ -108,14 +108,27 @@ test_chat() {
...
-    test -d taxonomy || git clone https://github.com/instructlab/taxonomy || true
+    TESTNUM=$1
+    if [ "$TESTNUM" -ne 1 ] && [ "$TESTNUM" -ne 2 ] && [ "$TESTNUM" -ne 3 ]; then
+        echo "Invalid test number: $TESTNUM"
+        exit 1
+    fi
 
-    step Make new taxonomy
-    mkdir -p taxonomy/knowledge/sports/overview/softball
+    test -d taxonomy || git clone https://github.com/instructlab/taxonomy || true
```
leseb, line 117: Why force success on this? What if the cloning fail for a good reason? Network down or GitHub unavailable? It's probably better to catch it here.

scripts/basic-workflow-tests.sh
```diff
@@ -157,7 +170,12 @@ test_exec() {
     PID=$!
 
     test_chat
-    test_taxonomy
+
+    test_taxonomy 1
```
leseb, line 174: `for i in $(seq 1 3)`?

scripts/basic-workflow-tests.sh
russellb: not new code, but i agree that `|| true` doesn't make sense

scripts/basic-workflow-tests.sh
russellb: sure, that would work, too.

russellb (commented):
not sure if these 2 items are meant to be blocking?

leseb:
> not sure if these 2 items are meant to be blocking?

Not blocking since I just left a comment :)

russellb:
this won't auto merge because it changed mergify config

PR #1455, opened 2024-06-25
python version bump

cdoern:
dolomite does not have python 3.9 support. Remove 3.9 tests and edit docs, pyproject, etc.

**Checklist:**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,16 +44,11 @@ jobs:
     strategy:
       matrix:
         python:
-          - "3.9"
           - "3.10"
           - "3.11"
           - "3.12"
         platform:
           - "ubuntu-latest"
-        exclude:
-          # GHA runner has no ARM64 3.9
-          - python: "3.9"
-            platform: "macos-latest"
         include:
           - python: "3.11"
             platform: "macos-latest"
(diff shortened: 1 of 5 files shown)


russellb (approved):
lgtm -- 3.10 is the minimum I am aware of that we need to support.

I would suggest expanding your commit message and PR description a bit more to clarify where this is coming from.  You can link to this repo, which is where we are publishing `instructlab-dolomite` from -- https://github.com/instructlab/GPTDolomite

RobotSail (approved):
LGTM - it would be nice if we could get dolomite to support 3.9 eventually but it's not a high priority.

bjhargrave:
Shouldn't this be mentioned in CHANGELOG as it raises the floor on python versions supported? People using python 3.9 will be surprised when they update ilab and find it may not work.

It seems like this change should be widely communicated.

nathan-weinberg:
> Shouldn't this be mentioned in CHANGELOG as it raises the floor on python versions supported? People using python 3.9 will be surprised when they update ilab and find it may not work.
> 
> It seems like this change should be widely communicated.

+1, I raised this exact point in Slack...

russellb:
good point re: ChangeLog. Can someone suggest an addition in a PR?

PR #1458, opened 2024-06-25
Drop some more sdg-only code

russellb:
This code was only used by the sdg library. It has been moved there
and is no longer needed here as of instructlab-sdg 0.0.3.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ GitPython>=3.1.42
 httpx>=0.25.0
 instructlab-quantize>=0.1.0
 instructlab-schema>=0.2.0
-instructlab-sdg>=0.0.2
+instructlab-sdg>=0.0.4
 jsonschema>=4.21.1
 llama_cpp_python[server]==0.2.75
 mlx>=0.5.1,<0.6.0; sys_platform == 'darwin' and platform_machine == 'arm64'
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -39,7 +39,6 @@ tqdm>=4.66.2
 transformers>=4.30.0
 trl>=0.7.11
 wandb>=0.16.4
-langchain-text-splitters
 # do not use 8.4.0 due to a bug in the library
 # https://github.com/instructlab/instructlab/issues/1389
 tenacity>=8.3.0,!=8.4.0
```
leseb, line 44: This can go away if `langchain-text-splitters` goes away.

leseb: It's being pulled by `langchain-text-splitters`.

russellb: OK, that will need to wait for a version of `instructlab-sdg` with this requirement in place.  langchain moved there, but not this part because I didn't realize it was related.

russellb: https://github.com/instructlab/sdg/pull/38

russellb: done

nathan-weinberg (approved):

JamesKunstle (approved):
love seeing -152 on a PR

PR #1462, opened 2024-06-25
com.redhat.component label

lmilbaum:
**Checklist:**


Gregory-Pereira (approved):
/lgtm
Testing if this unblocks Konflux builds please merge

nathan-weinberg (approved):

russellb (approved):
This looks like it's redhat product specific, which would not typically be appropriate for the upstream image.  It's also harmless, so I'm ok with it for now, but at some point we should be prepared to have a product specific version of these with redhat product customizations. I know for sure that we are going to need some customizations in the very near future that will not go here.

lmilbaum:
> This looks like it's redhat product specific, which would not typically be appropriate for the upstream image. It's also harmless, so I'm ok with it for now, but at some point we should be prepared to have a product specific version of these with redhat product customizations. I know for sure that we are going to need some customizations in the very near future that will not go here.

Thank you for your feedback. This label is set by the parent image. This PR is intended to override it.

PR #1466, opened 2024-06-26
[WIP] Add end-to-end training orchestration

JamesKunstle:
The full-scale training loop implemented in the `LAB: Large-Scale Alignment for ChatBots` method requires multiple phases, illustrated as: `TRAIN-->EVAL-->TRAIN-->EVAL`. We've built the atomic components of this training flow (training, model evaluation), but we haven't wrapped those steps in a resilient training interface.

This PR implements a concept of `end-to-end` training, abbreviated as `e2e`. 

> more to come as PR develops! Pushing for visibility.

**Checklist:**

### DIFF
--- a/src/instructlab/model/train.py
+++ b/src/instructlab/model/train.py
@@ -443,8 +443,25 @@ def train(
             # end-to-end training
             _validate_e2e_paths(e2e_data_dir, e2e_checkpoints_dir)
             final_checkpoint = run_e2e_training(ctx.params, train_args, torch_args)
+            print(f"E2E output best checkpoint: {final_checkpoint}")
         else:
             run_training(train_args=train_args, torch_args=torch_args)
+
+
+"""
+expect storage like:
+~/.local/var/instructlab/e2e
+    | phase5
+        | checkpoints
+        | eval
+        | data ?
+    | phase10
+        | checkpoints
+        | eval
+        | data ?
+        
+`ilab model train --e2e --model ibm-granite/granite-7b-lab --phase5_data ./phase5/train_data.jsonl --phase10_data ./phase10/train_data.jsonl --model-out ./new_model`
+"""
 
 
 def _validate_e2e_data_dir(data_path: str) -> None:
--- a/src/instructlab/model/train.py
+++ b/src/instructlab/model/train.py
@@ -15,6 +15,9 @@ from instructlab.training import (
     TrainingArgs,
     run_training,
 )
+
+from instructlab.train.run_e2e_training import run_e2e_training
+
 import click
 
 # First Party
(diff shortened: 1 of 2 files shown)


src/instructlab/model/train.py
```diff
@@ -15,6 +15,9 @@
     TrainingArgs,
     run_training,
 )
+
+from instructlab.train.run_e2e_training import run_e2e_training
```
JamesKunstle, line 19: implementing this in parallel, just waiting on `instructlab.eval` to publish their recent merges.

src/instructlab/model/train.py
```diff
@@ -467,10 +490,91 @@ def get_files(pattern):
...
+            run_e2e_training(ctx.params, train_args, torch_args)
+        else:
+            run_training(train_args=train_args, torch_args=torch_args)
+
+
+def _validate_e2e_data_dir(data_path: str) -> None:
+    """Checks for existence of both data files for e2e run.
+
+    Args:
+        data_path: Top-level directory of e2e data directory.
```
cdoern, line 513: there are so many flags called data_path or data_dir in training, can we make this more specific?

src/instructlab/model/train.py
```diff
@@ -467,10 +490,91 @@ def get_files(pattern):
...
+        click.secho(
+            f"Error finding end-to-end checkpoint cache directory. Please make sure you're passing the TOP LEVEL directory of \
+                your e2e_checkpoints directory, and that all sub-directories are in their proper locations. \
+                \nReceived: {ckpt_abs_path}",
+            fg="red",
+        )
+        raise click.exceptions.Exit(1)
+
+
+def _validate_e2e_paths(data_path: str, checkpoint_path: str) -> None:
```
cdoern, line 565: data_path ditto

src/instructlab/model/train.py
cdoern: so, this will in the training lib call the eval lib? or is the idea that we will call eval in the cli and call training over and over?

IMO, I am pro keeping all the orchestration in the CLI not the libraries

cdoern (commented):
I like the structure, the flags look good.

src/instructlab/model/train.py
JamesKunstle: 10000%, what about `e2e_data_dir`

src/instructlab/model/train.py
JamesKunstle: I'd like to build orchestration below the level of the CLI since the training/eval libraries are sharing their interfaces like any other library.... and because I think it's janky to subprocess like 25 different things during training. But there's also the argument to be made that calling the CLI recursively also self-tests, which is useful. idk, how about a prototype and we see if it's reasonable?

JamesKunstle:
@RobotSail This is the PR I'm working on that will use your design for consistent user-space storage. Just wanted to bring it to your attention in case the context was useful

JamesKunstle:
@nathan-weinberg @alimaredia @alinaryan Just requesting since this is consuming your work. Not done yet.

PR #1471, opened 2024-06-26
Implementation of Dedicated Storage

RobotSail:
This PR implements the dev-doc outlined for a dedicated config & storage directory:

https://github.com/instructlab/dev-docs/pull/104

This PR also introduces 2 new commands, only one of which is intended as an addition to instructlab:
- `ilab config show`: displays the current active config
- `ilab system reset`: erases all of the data in the `~/.local/share/instructlab` and `~/.config/instructlab` directories. This was intended as a debug command for myself during dev, but included it in here in case folks feel that it's useful.

This PR also moves the `ilab sysinfo` command to now be under the new `ilab system` command group, so the following change is happening:
- `ilab sysinfo` is now becoming `ilab system sysinfo`.


src/instructlab/debug/debug.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@click.command()
+@click.pass_context
+@utils.display_params
+def reset(
+    ctx,
+):
+    """Start a local server"""
+
```
JamesKunstle, line 31: Since this is _really_ destructive maybe the user should have to type an exact random string or something into a shell query.

```
WARNING: `ilab debug reset` will purge all of your local `ilab` data. To proceed, please confirm by repeating the following string:

> FR42BH
>
```

JamesKunstle (commented):
This is such an upgrade! Awesome work

JamesKunstle:
Will definitely want this to be merged prior to #1466 because I want this feature for e2e training.

src/instructlab/debug/debug.py
RobotSail: I was originally going to drop this from the PR once it's ready, but I really like this approach!

RobotSail: I will implement this

JamesKunstle:
@RobotSail Just one more thought- I know we're not targetting multi-tenancy in any real way, but is there any risk to using user-profile-centric data directories like `~/.local/share` if multiple uses might be logging into the same machine to use this? I can see it as a feature rather than a risk, especially if multiple users want their own version of the ilab flow, but since this is pretty resource intensive it might be inconvenient for a user to think that the data isn't available, when it is, just under another users' home directory.

RobotSail:
@JamesKunstle That's a good point. I think for the v1 it should just work and be inefficient wrt not having multi-tenancy. My thinking is that once it gets to the point of multiple people using this on the same device and complaining about the storage size, that's when it would be appropriate to incorporate this as a design.

src/instructlab/configuration.py
```diff
@@ -300,42 +352,62 @@ def get_model_family(forced, model_path):
...
+
+    for storage_dir in [
+        STORAGE_DIR_NAMES.CHECKPOINTS,
+        STORAGE_DIR_NAMES.MODELS,
+        STORAGE_DIR_NAMES.DATASETS,
+        STORAGE_DIR_NAMES.TAXONOMY,
+        STORAGE_DIR_NAMES.INTERNAL,
+    ]:
+        os.makedirs(
+            os.path.join(ILAB_DATA_HOME, storage_dir), exist_ok=True, mode=0o777
```
n1hility, line 369: probably was 755 here

src/instructlab/debug/debug.py
```diff
@@ -0,0 +1,39 @@
...
+def debug(ctx):
+    """Command Group for debug-related ilab calls"""
+    ctx.obj = ctx.parent.obj
+    ctx.default_map = ctx.parent.default_map
+
+
+@click.command()
+@click.pass_context
+@utils.display_params
+def reset(
```
n1hility, line 27: Two issues with debug

1. Command structure is noun-verb (this is verb verb)
2. Reseting your state is not really related to debugging

I'd go with "system" (e.g. podman system reset is similar) or @cdoern 's suggestion of "storage"

src/instructlab/debug/debug.py
```diff
@@ -0,0 +1,39 @@
...
+    ctx,
+):
+    """Start a local server"""
+
+    # recursively removes the data at the config and share directories
+    print(f"removing {ILAB_CONFIG_HOME}")
+    os.system(f"rm -rf {ILAB_CONFIG_HOME}")
```
n1hility, line 34: I'd recommend switching this to shutil.rmtrree  ( or a glob os.remove() ) to fix two problems:

1. Portability (E.g. support windows)
2. Fixes string escaping issues. If someone uses a space in the path to their home dir or in XDG_CONFIG_HOME for example this would potentially rm -rf the wrong folder.

src/instructlab/configuration.py
```diff
@@ -29,27 +30,81 @@
...
+            "The HOME directory does not exist. Please make sure it's set and run this again."
+        )
+    ILAB_DATA_HOME = os.path.expandvars(
+        os.path.join("${HOME}", ".local", "share", STORAGE_DIR_NAMES.ILAB)
+    )
+else:
+    ILAB_DATA_HOME = os.path.expandvars(
+        os.path.join("${XDG_DATA_HOME}", STORAGE_DIR_NAMES.ILAB)
+    )
+
```
n1hility, line 77: Minor nit (non-blocking) Some of the duplication and double lookups could be replaced with something like (need to scroll):

```suggestion
def getHomeDir(*subdir: str) -> str: 
    home = os.getenv("HOME")
    if not home:
        raise RuntimeError(
           "The HOME directory does not exist. Please make sure it's set and run this again."
        )
    return os.path.join(home, *subdir)  

def getHomeEnvPath(env: str, suffix: str, *homePaths: str) -> str:
    path = os.getenv(env) or getHomeDir(*homePaths)
    return os.path.join(path, suffix)

ILAB_CONFIG_HOME = getHomeEnvPath("XDG_CONFIG_HOME",  STORAGE_DIR_NAMES.ILAB, ".config")
ILAB_DATA_HOME =  getHomeEnvPath("XDG_DATA_HOME", STORAGE_DIR_NAMES.ILAB, ".local", "share")

```

n1hility (changes requested):
Looks good, just a few minor things. Namely debug command.

src/instructlab/model/debug.py
```diff
@@ -0,0 +1,3 @@
+# Third Party
+from click_didyoumean import DYMGroup
```
cdoern, line 2: why is this file here? is this meant to be in its own group?

src/instructlab/configuration.py
```diff
@@ -29,27 +30,81 @@
 # Local
 from . import log
 
+
+class STORAGE_DIR_NAMES:
```
cdoern, line 34: this is nice, I really like this setup!

src/instructlab/configuration.py
```diff
@@ -29,27 +30,81 @@
...
+    ILAB_DATA_HOME, STORAGE_DIR_NAMES.INTERNAL, "prompt.txt"
+)
+DEFAULT_GENERATED_FILES_OUTPUT_DIR = os.path.join(
+    ILAB_DATA_HOME, STORAGE_DIR_NAMES.DATASETS
+)
 DEFAULT_CONNECTION_TIMEOUT = httpx.Timeout(timeout=30.0)
 # use spawn start method, fork is not thread-safe
 DEFAULT_MULTIPROCESSING_START_METHOD = "spawn"
 
+
```
cdoern, line 107: random line

src/instructlab/configuration.py
```diff
@@ -300,42 +352,62 @@ def get_model_family(forced, model_path):
...
+    # since torch and train args are sep, they need to be combined into a single `train` entity for the default map
+    # this is because the flags for `ilab model train` will only be populated if the default map has a single `train` entry, not two.
+    training_dict["train"] = (
+        training_dict["train"]["train_args"]
+        | training_dict["train"]["torch_args"]
+        | training_dict["train"]["train_args"]["lora"]
+        | training_dict["train"]["train_args"]["deepspeed_options"]
+    )
+    # default_map holds a dictionary with default values for each command parameters
+    ctx.default_map = get_dict(ctx.obj.config)
```
cdoern, line 413: this needs to be `training_dict` not `get_dict`, this is how the flags are auto populated from the cfg

cdoern (commented):
I like the idea here, a few comments on the naming, the structure, and some breakages to config initialization I found.

src/instructlab/configuration.py
RobotSail: Good point

src/instructlab/debug/debug.py
RobotSail: Done!

src/instructlab/debug/debug.py
RobotSail: Very good catch, changing this!

src/instructlab/configuration.py
RobotSail: This is definitely cleaner than my original approach. I mainly wanted to avoid prematurely abstracting, but this describes it well. I'll include this in the PR.

src/instructlab/lab.py
```diff
@@ -15,6 +15,7 @@
 # NOTE: Subcommands are using local imports to speed up startup time.
 from .config import config as config_group
 from .data import data as data_group
+from .debug import debug
```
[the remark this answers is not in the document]
RobotSail: That's fair, I changed it to `system` per @n1hility's recommendation, since `ilab system reset` implies resetting the config. `ilab storage` makes a lot of sense though, since we'll also probably need a way to view the total storage usage with a command like `ilab storage view` and `ilab storage clear`. 

We can make a dev-doc for this down the road though.

src/instructlab/model/debug.py
RobotSail: Deleted

src/instructlab/configuration.py
RobotSail: TY

src/instructlab/configuration.py
RobotSail: Fixed, thanks for catching this!

src/instructlab/configuration.py
RobotSail: Thank you very much for this suggestion, I included it!

src/instructlab/configuration.py
```diff
@@ -29,23 +30,65 @@
...
+
+
+def get_home_env_path(env: str, suffix: str, *home_paths: str) -> str:
+    """Neatly formats the environment paths for the user home & storage directories"""
+    path = os.getenv(env) or get_home_dir(*home_paths)
+    return os.path.join(path, suffix)
+
+
+ILAB_CONFIG_HOME = get_home_env_path(
+    "XDG_CONFIG_HOME", STORAGE_DIR_NAMES.ILAB, ".config"
```
fabianlim, line 61: `XDG_CONFIG_HOME` is for specifying where the python packages live?

RobotSail: No, `XDG_CONFIG_HOME` specifies where we should store config files on the user's system. From the [FreeDesktop XDG Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

> $XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

RobotSail: The python packages will be installed in the `site-packages` of whichever python is used on the user's system, and `ilab` will be placed in its binaries folder.

src/instructlab/debug/debug.py
RobotSail: Done

cdoern:
I won't block on it but I think we should split commit https://github.com/instructlab/instructlab/pull/1471/commits/e0b2ee1164f777acb3bc771f06ab5e0b1c6d0775 into its own PR since it isn't related directly and is a large convo we are all having at the moment.

I agree with all of these changes but adding something that changes the flag parsing in is a bit misleading IMO, if others disagree we can keep it!

cdoern:
I am saying ^ bc I think we can merge this ASAP, but not with the flag parsing

RobotSail:
> I won't block on it but I think we should split commit [e0b2ee1](https://github.com/instructlab/instructlab/commit/e0b2ee1164f777acb3bc771f06ab5e0b1c6d0775) into its own PR since it isn't related directly and is a large convo we are all having at the moment.

This would need to go in with the same PR, because it'll be broken otherwise.

cdoern:
after rebasing I am good to merge

cdoern:
(and after another review :) )

src/instructlab/config/show.py
```diff
@@ -0,0 +1,17 @@
...
+
+@click.command()
+@click.pass_context
+@utils.display_params
+def show(
+    ctx,
+):
+    """Displays the current config as YAML"""
+    # TODO: make this use pretty colors like jq/yq
+    config_yaml = yaml.load(ctx.obj.config.model_dump_json(), Loader=yaml.FullLoader)
```
leseb, line 16: What if `yaml.load` fail? How is the error handled?

src/instructlab/config/show.py
```diff
@@ -0,0 +1,17 @@
...
+    print(yaml.dump(config_yaml, indent=4))
```
leseb, line 17: Use `click.echo`?

src/instructlab/configuration.py
```diff
@@ -29,25 +30,67 @@
...
+    CHECKPOINTS = "checkpoints"
+    MODELS = "models"
+    TAXONOMY = "taxonomy"
+    INTERNAL = (
+        "internal"  # for storing all ilab-internal files the user doesn't need to see
+    )
+
+
+def get_home_dir(*subdir: str) -> str:
+    home = os.getenv("HOME")
```
leseb, line 46:
```suggestion
    home = str(Path.home())
```

How about this instead of the entire function?

src/instructlab/system/reset.py
```diff
@@ -0,0 +1,46 @@
+import click
```
leseb, line 1: This needs the license header

src/instructlab/system/reset.py
```diff
@@ -0,0 +1,46 @@
...
+
+from instructlab import utils
+from instructlab.configuration import ILAB_CONFIG_HOME, ILAB_DATA_HOME
+
+
+def delete_storage_dirs():
+    """
+    Deletes all of the data in the instructlab storage & config directories.
+    """
+    print(f"removing {ILAB_CONFIG_HOME}...")
```
leseb, line 13: Use click.echo

leseb (changes requested):
The description mentions  `ilab debug reset` but I think it's `ilab system reset` right? Also, given that this is not part of the storage design, I would drop it and concentrate on the proposal only. This can be added later if everyone agrees. This will reduce the scope of reviews too. Thanks.

A few more things:

* We need to update the CHANGELOG.md to reflect the addition of the new command.
* Add integration test for the new `config show` command

src/instructlab/config/show.py
RobotSail: Added an exception handle

src/instructlab/config/show.py
RobotSail: Fixed

src/instructlab/configuration.py
RobotSail: Added it to check for both `Path.home()` and `os.getenv("HOME")`

cdoern:
spoke to @RobotSail , lets keep everything in this PR in the name of getting this all done on time :100:

russellb:
> spoke to @RobotSail , lets keep everything in this PR in the name of getting this all done on time 💯

you need @leseb's buy-in for that. It is not appropriate to bypass requested changes from a maintainer.

RobotSail:
> The description mentions `ilab debug reset` but I think it's `ilab system reset` right? Also, given that this is not part of the storage design, I would drop it and concentrate on the proposal only. This can be added later if everyone agrees. This will reduce the scope of reviews too. Thanks.
> 
> A few more things:
> 
> * We need to update the CHANGELOG.md to reflect the addition of the new command.
> * Add integration test for the new `config show` command

Is this really something we want to block on? This PR is already blocking #1466 and others which we need ASAP. 

While I agree that we should keep PRs atomic, everything introduced in this PR is related and would need to go around the same time anyway.

leseb:
> > The description mentions `ilab debug reset` but I think it's `ilab system reset` right? Also, given that this is not part of the storage design, I would drop it and concentrate on the proposal only. This can be added later if everyone agrees. This will reduce the scope of reviews too. Thanks.
> > A few more things:
> > 
> > * We need to update the CHANGELOG.md to reflect the addition of the new command.
> > * Add integration test for the new `config show` command
> 
> Is this really something we want to block on? This PR is already blocking #1466 and others which we need ASAP.
> 
> While I agree that we should keep PRs atomic, everything introduced in this PR is related and would need to go around the same time anyway.

I would say yes, we cannot "wildly" add new flags to the CLI without achieving consensus among maintainers. It's essential to avoid the "while we are at it" approach in critical PRs like this one. This PR's sole objective is to add `config show`, and adding more content to it would be counterproductive given its significance. That's the primary issue at hand.

cdoern:
I want to be clear, I wasn't suggesting completely bypassing people's opinions on this one, I was just stating mine in that I think keeping this all in one PR is ok

cdoern:
What we have been doing for the last week or so due to upcoming deadlines is 

1. merge a PR that "works", fulfills the requirements, passes existing CI
2. add a follow up PR working on tests, and in the same process smoothing over whatever functional issues are uncovered
3. merge that PR 

We have been doing this in order to "unblock work" which I get is a weird process but I just want to be candid that this is what has been going on.

scripts/basic-workflow-tests.sh
```diff
@@ -23,13 +23,21 @@ BOLD='\033[1m'
...
+export XDG_DATA_HOME="${E2E_TEST_DIR}/.local/share"
+
+# ensure that our mock e2e dirs exist
+for dir in "${XDG_CONFIG_HOME}" "${XDG_DATA_HOME}"; do
+	mkdir -p "${dir}"
+done
 
 step() {
-    echo -e "$BOLD$* - $(date)$NC"
+    echo -e "${BOLD}${*} - $(date)${NC}"
```
leseb, line 36: These changes are not needed, let's not complicate the work and change something that is already working. This is valid for the cosmetic bash changes in all .sh files. Thanks.

RobotSail: I reset it locally, will push up in a bit.

RobotSail: Resolved

scripts/functional-tests.sh
```diff
@@ -6,7 +6,24 @@ set -ex
 # build a prompt string that includes the time, source file, line number, and function name
 export PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
-pip install .
+pip install -e .
```
leseb, line 9: Why changing this line?

scripts/functional-tests.sh
```diff
@@ -64,10 +85,10 @@ ilab --version
 ilab sysinfo
 
 # pipe 3 carriage returns to ilab config init to get past the prompts
-echo -e "\n\n\n" | ilab config init
+echo -e "\n\n\n" | ilab init
```
leseb, line 88: Why reverting this?

scripts/functional-tests.sh
```diff
@@ -333,23 +358,54 @@ wait_for_server(){
...
+# Returns:
+#  The uppercase string.
+########################################
+function to_upper() {
+    local to_convert="${1}"
+    local major_version
+    major_version=$(echo "${BASH_VERSION}" | cut -d'.' -f1)
+
+    # compatibility script to use built-in operator on bash 4+
+    # and default to tr on bash 3 or older.
```
leseb, line 375: The project only [supports](https://github.com/instructlab/instructlab?tab=readme-ov-file#bash-version-44-or-newer) Bash 4.4. So this is not needed.

src/instructlab/configuration.py
```diff
@@ -29,25 +31,68 @@
...
+    CHECKPOINTS = "checkpoints"
+    MODELS = "models"
+    TAXONOMY = "taxonomy"
+    INTERNAL = (
+        "internal"  # for storing all ilab-internal files the user doesn't need to see
+    )
+
+
+def get_home_dir(*subdir: str) -> str:
+    home = str(Path.home()) if Path.home() else os.getenv("HOME")
```
leseb, line 47: I think `Path.home()` is sufficient why do you need to do `os.getenv("HOME")`?

scripts/functional-tests.sh
RobotSail: @leseb MacOS still ships with BASH version 3, so it needs to be something that we can support out-of-the-box.

scripts/functional-tests.sh
RobotSail: `ilab config init` was previously protected by a config-guard which ensured that the `config` command group doesn't actually have a config object sent to it. Since this is something we need to change to view the config with `ilab config show`, I removed the `config` group from the no-config guard. So we need to call `ilab init` directly instead of having it be under config.

scripts/functional-tests.sh
RobotSail: In the scope of this test, there's no real reason to change it as it will `pip install` the version of the source at the time of running.

However; this becomes a problem when developing locally as the version of `ilab` now being used in the CLI will be a static version that's not reflective of any changes we make.

src/instructlab/configuration.py
leseb: Give me one system that the project supports where `~` is not present and we will revert to using `os.getenv("HOME")`. Thanks.

RobotSail: You're right, `Path.home()` is simpler, fixing this.

scripts/functional-tests.sh
RobotSail: We discussed this offline, the resolution is we will not support bash 3 and this will be dropped from the PR once the rest of this file works.

src/instructlab/configuration.py
```diff
@@ -29,25 +31,63 @@
...
+
+def get_home_dir(*subdir: str) -> str:
+    return os.path.join(str(Path.home()), *subdir)
+
+
+def get_home_env_path(env: str, suffix: str, *home_paths: str) -> str:
+    """Neatly formats the environment paths for the user home & storage directories"""
+    home_path = os.getenv(env)
+    full_path = home_path if home_path else get_home_dir(*home_paths)
+    return os.path.join(full_path, suffix)
```
tiran, line 54: Please don't roll your own platform directory code, use [platformdirs](https://pypi.org/project/platformdirs/). You are not pulling a new package. InstructLab has an indirect dependency on platformdirs already.

src/instructlab/configuration.py
```diff
@@ -29,25 +31,63 @@
...
 DEFAULT_MODEL_REPO = "instructlab/granite-7b-lab"
 DEFAULT_JUDGE_MODEL_MT = "prometheus-eval/prometheus-8x7b-v2.0"
 DEFAULT_EVAL_PATH = "eval_data"
 DEFAULT_TAXONOMY_REPO = "https://github.com/instructlab/taxonomy.git"
-DEFAULT_TAXONOMY_PATH = "taxonomy"
+DEFAULT_TAXONOMY_PATH = os.path.join(ILAB_DATA_HOME, STORAGE_DIR_NAMES.TAXONOMY)
 DEFAULT_TAXONOMY_BASE = "origin/main"
+DEFAULT_CHECKPOINTS_DIR = os.path.join(ILAB_DATA_HOME, STORAGE_DIR_NAMES.CHECKPOINTS)
+DEFAULT_DATASET_DIR = os.path.join(ILAB_DATA_HOME, STORAGE_DIR_NAMES.DATASETS)
+DEFAULT_INTERNAL_DIR = os.path.join(ILAB_DATA_HOME, STORAGE_DIR_NAMES.INTERNAL)
```
tiran, line 79: Are these really user data files? At least the checkpoints directory probably belongs into user cache dir.

src/instructlab/configuration.py
```diff
@@ -381,55 +419,97 @@ def get_model_family(forced, model_path):
...
+
+    for storage_dir in [
+        STORAGE_DIR_NAMES.CHECKPOINTS,
+        STORAGE_DIR_NAMES.MODELS,
+        STORAGE_DIR_NAMES.DATASETS,
+        STORAGE_DIR_NAMES.TAXONOMY,
+        STORAGE_DIR_NAMES.INTERNAL,
+    ]:
+        os.makedirs(
+            os.path.join(ILAB_DATA_HOME, storage_dir), exist_ok=True, mode=0o755
```
tiran, line 436: Is there any particular reason why you are overriding the user's umask?

scripts/functional-tests.sh
tiran: Strong -1

Don't use `pip install -e` in tests. The option makes pip use a different approach inject a package into a virtual env. It does not actually install the package. Instead it drops a `.pth` file into site-packages and modifies `sys.path`. With `-e`, we are not testing what our users are deploying.

scripts/functional-tests.sh
```diff
@@ -6,7 +6,24 @@ set -ex
...
 export PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
-pip install .
+pip install -e .
+
+export TEST_DIR='__functional-test-data'
+export XDG_DATA_HOME="${TEST_DIR}/.local/share"
+export XDG_CONFIG_HOME="${TEST_DIR}/.config"
+export PACKAGE_NAME='instructlab'  # name we use of the top-level XDG directories for CLI data
+export ILAB_DATA_DIR="${XDG_DATA_HOME}/${PACKAGE_NAME}"
```
tiran, line 15: This is incompatible with macOS. XDG is a Linux standard. Apple has a different layout for user data.

tiran (changes requested):
Don't use `pip install -e`

Please use `platformdirs` instead re-inventing the wheel. The XDG standard is for Linux. We are also supporting macOS. platformdirs will give you the correct path for every platform:

```python
>>> platformdirs.user_data_dir("instructlab")
'/home/heimes/.local/share/instructlab'
>>> platformdirs.user_cache_dir("instructlab")
'/home/heimes/.cache/instructlab'
>>> platformdirs.user_config_dir("instructlab")
'/home/heimes/.config/instructlab'
```

I also recommend any new code should adopt `pathlib,Path` instead of `str` for paths.

src/instructlab/configuration.py
RobotSail: @tiran I didn't know this already existed, thank you!

src/instructlab/configuration.py
RobotSail: What would you recommend instead? I was under the impression that `0o755` would be fine for this.

src/instructlab/configuration.py
RobotSail: Good idea, I'll replace all of this to use `platformdirs` instead.

src/instructlab/configuration.py
RobotSail: Based on the [documentation of mkdir](https://docs.python.org/3/library/os.html#os.mkdir) (what gets used under the hood of `os.makedirs`), it seems like the umask is subtracted away from the mode we set.
My understanding is that if I'm setting `0o755` and the umask is `0077`, then the resulting permissions will still be `0700`. 

Maybe I'm not understanding something, but I don't believe we are overriding the user's umask here.
Please correct me if I'm wrong here.

RobotSail:
The functional tests work locally but require a new release of the SDG library.

scripts/functional-tests.sh
RobotSail: @tiran I've updated everything to rely on `platformdirs` instead, so we aren't explicitly referencing XDG anymore.

RobotSail:
@tiran 

> I also see another potential problem with this change. XDG directories are only suitable for (human) user environments. They don't work well in other environments such as automation, system accounts, or in system-level services under systemd.
> 
> FreeIPA also uses XDG. It took me a long time and several patches to make FreeIPA's client libraries properly work in system accounts and systemd services. Please take this into account, too.

Would this be something beyond just switching to using `platformdirs` vs. explicitly referencing XDG?

scripts/functional-tests.sh
RobotSail: @tiran That's a fair point, but how will we clean up the `pip install .`-ed package? I don't see anything in the script which actually cleans this up from the python environment after the tests finish. 

When users are developing locally and running these tests, it will be very confusing if their `ilab` CLI is suddenly fixed to a static version from the time of the last test build.

tiran: The `pip install .` line should not be there in the first place. It's the job of tox to install and re-install the package. Tox monitors its virtual envs and takes care of installing dependencies or re-creating an outdated env. It will also rebuild and reinstall the current package every time you run a tox env.

If users want to re-use a virtual env as a development env, then they should run `pip install -e .`. Other options are `tox devenv` or changing `package = wheel` to `package = editable` in `tox.ini`.

.gitignore
```diff
@@ -10,6 +10,9 @@ generated/
 training_results/
 train-output/
 
+# We use this directory for storing the data from functional tests.
+__functional-test-data/
```
tiran, line 14: Tests should not write to the local directory. Please assume that the local directory is read-only for tests and that they can only write to a dedicated area.

requirements.txt
```diff
@@ -46,3 +47,4 @@ wandb>=0.16.4
 # it is for CLI usage ONLY
 yamllint>=1.35.1
 referencing>=0.35.0
+instructlab-training>=0.0.4
```
tiran, line 50: Could you please order the dependencies in alphanumeric order? It makes it easier to find dependencies.

tiran (commented):
I recommend that we first land my PR #1591 and address #1594. It will make this PR a lot easier to implement. I can implement #1594 shortly after #1591 has landed.

.gitignore
RobotSail: Where should we write instead?

RobotSail: Maybe `/tmp`?

RobotSail:
@tiran Is it necessary to block until those are merged? This PR is currently blocking #1466 and is something we need to go in ASAP.

requirements.txt
RobotSail: Sure.

RobotSail: Fixed.

.gitignore
RobotSail: Fixed.

tiran: In shell scripts, use mktemp and then add a trap that removes the directory on exit.

RobotSail: I'll do that instead, thanks!

cdoern:
@tiran @RobotSail I think this one needs to get in first to unblock some pretty critical downstream work.

This PR is one of the last pieces of work that needs to get in before upcoming deadlines

tiran:
> @tiran @RobotSail I think this one needs to get in first to unblock some pretty critical downstream work.
> 
> This PR is one of the last pieces of work that needs to get in before upcoming deadlines

I'm a bit worried that this design will make it harder to implement a systemd service. XDG standard does not play well with a system service in systemd. It's design for user sessions and hum end users.

Suggestion:
I'll help you push this PR over the finish line to unblock PR #1466 despite the issues. Let's ignore the design problems (XDG vs. systemd) and implementation issues (str instead of pathlib for paths) for now. They are not an immediate blocker and we have $DEADLINES.

After this PR has landed, I'll rebase my `--help` and default lookup PR and we merge that. It should be ready for review when you are awake on Monday. Once that has landed, I can clean up the str/pathlib typing confusion problems and rework the new code from this PR. I have an idea how to make the implementation more elegant and flexible.

The approach will help you to keep your velocity while still having a more or less clean-ish version for the $DEADLINE.

scripts/basic-workflow-tests.sh
```diff
@@ -24,13 +24,21 @@ BOLD='\033[1m'
 NC='\033[0m' # No Color
 
 SCRIPTDIR=$(dirname "$0")
+E2E_TEST_DIR="$(pwd)/__e2e_test"
+export XDG_CONFIG_HOME="${E2E_TEST_DIR}/.config"
+export XDG_DATA_HOME="${E2E_TEST_DIR}/.local/share"
+
```
tiran, line 30: Let's also define the third and last XDG home directory in case we need it later.
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

```suggestion
export XDG_STATE_HOME="${E2E_TEST_DIR}/.local/share"

```

src/instructlab/configuration.py
```diff
@@ -380,7 +525,52 @@ def get_model_family(forced, model_path):
...
+        DEFAULTS.MODELS,
+        DEFAULTS.TAXONOMY,
+        DEFAULTS.INTERNAL,
+        DEFAULTS.CHATLOGS,
+        DEFAULTS.CONFIG_HOME,
+        DEFAULTS.CACHE_HOME,
+    ]
+
+    for dirpath in dirs_to_make:
+        os.makedirs(dirpath, exist_ok=True, mode=0o755)
```
tiran, line 547: I recommend to keep the default mode value. The actual mode is a combination of the mode and the process' umask. The default value allows users to set a more relaxed umask (e.g. group writable) and stricter value (only accessible by the owner).

```suggestion
        os.makedirs(dirpath, exist_ok=True)
```

src/instructlab/configuration.py
```diff
@@ -390,48 +580,46 @@ def __init__(self, config_obj: Config):
...
+    log.configure_logging(log_level=config_obj.general.log_level.upper())
+    ctx.obj = Lab(config_obj)
+    # default_map holds a dictionary with default values for each command parameters
+    config_dict = get_dict(ctx.obj.config)
+    # since torch and train args are sep, they need to be combined into a single `train` entity for the default map
+    # this is because the flags for `ilab model train` will only be populated if the default map has a single `train` entry, not two.
+
+    # XXX(osilkin): the following block of code is to hack around conflicting keys
+    #               being merged and erasing each other
+    #               issue: https://github.com/instructlab/instructlab/issues/1523
```
tiran, line 611: PR #1591 will solve this mess and get rid of config_dict flattening. The new code in the PR allows to specify where a value comes from, e.g. `config_section=train_args.lora` to look up values from `train.train_args.lora` in the `train` callback.

src/instructlab/data/generate.py
```diff
@@ -10,57 +10,55 @@
...
-
 logger = logging.getLogger(__name__)
 
 
 @click.command()
 @click.option(
     "--model",
-    default=config.DEFAULT_MODEL,
-    show_default=True,
+    default=lambda: config.DEFAULTS.DEFAULT_MODEL,
```
tiran, line 19: Most of the `default` in this file should be **not** necessary. They should all get their values from `ctx.default_map`.

src/instructlab/system/reset.py
```diff
@@ -0,0 +1,53 @@
...
+    click.echo(f"removing {DEFAULTS.CACHE_HOME}...")
+    shutil.rmtree(DEFAULTS.CACHE_HOME)
+
+
+@click.command()
+@click.option(
+    "--skip-verify",
+    default=False,
+    help="Whether or not to skip verification of data deletion.",
+)
```
tiran, line 31: Use this approach to implement a prompt: https://click.palletsprojects.com/en/8.1.x/options/#yes-parameters

src/instructlab/system/system.py
```diff
@@ -0,0 +1,22 @@
...
+
+@click.group(cls=DYMGroup)
+@click.pass_context
+def system(ctx):
+    """Command group for all system-related command calls"""
+    ctx.obj = ctx.parent.obj
+    ctx.default_map = ctx.parent.default_map
+
+
+system.add_command(reset)
```
tiran, line 22: Could you please move `sysinfo` here, too?

tests/test_config.py
```diff
@@ -13,34 +13,35 @@
 class TestConfig:
     @pytest.fixture(autouse=True)
     def _init_tmpdir(self, tmpdir):
+        os.environ["HOME"] = str(tmpdir)
```
tiran, line 16: This change will leak to other tests. You have to use a fixture that resets the original value. You also want to drop any `XDG_*` env vars to force default values.

Something like this should work (untested).

```
# add to conftest.py
import os
from unittest import mock

@pytest.fixture(autouse=True)
def tmp_home(tmpdir):
    with mock.patch.dict(os.environ):
        os.environ["HOME"] = str(tmpdir)
        for key in list(os.environ):
        if key.startswith("XDG_"):
            os.environ.pop(key)
       yield
```

Then make it an auto-use fixture.

tests/test_config.py
```diff
@@ -13,34 +13,35 @@
...
         self.tmpdir = tmpdir
 
-    def _assert_defaults(self, cfg):
+    def _assert_defaults(self, cfg: config.Config):
         assert cfg.general is not None
         assert cfg.general.log_level == "INFO"
 
         assert cfg.chat is not None
-        assert cfg.chat.model == "models/merlinite-7b-lab-Q4_K_M.gguf"
+        assert cfg.chat.model == config.DEFAULTS.DEFAULT_MODEL
```
tiran, line 24: I recommend that you leave the hard-coded values here / replace them with correct new hard-coded values. For tests it is sometimes a good idea to violate the DRY dogma. It makes it easier to detect bugs.

tests/test_lab_init.py
```diff
@@ -22,26 +18,54 @@ class TestLabInit:
     @patch("instructlab.config.init.Repo.clone_from")
     def test_init_noninteractive(self, mock_clone_from):
         runner = CliRunner()
-        with runner.isolated_filesystem():
+        with runner.isolated_filesystem() as td:
+            os.environ["HOME"] = td
```
tiran, line 22: Similar problem ... I'll get back to you with a proper fixture.

tiran (changes requested):

src/instructlab/system/system.py
RobotSail: Sure thing!

scripts/basic-workflow-tests.sh
RobotSail: What's the difference between this and `XDG_DATA_HOME`? 
Also - I can replace this from being "XDG_*" to just being DATA, CONFIG, etc., whatever we need here if you'd like.

tiran: The state home is for less important data than data home. Application directory standards have multiple levels of importance and often different directories for local and roaming data.
Files in data home should be backed up and synchronized between machines. Files in state home are less important and not portable (local, not synced).

tests/test_config.py
tiran: https://github.com/instructlab/instructlab/pull/1617 provides infrastructure to solve this problem.

src/instructlab/configuration.py
```diff
@@ -17,55 +18,190 @@
...
+        v: str | typing.Any = self._cache[key]
+        if not isinstance(v, str):
+            raise ValueError(f"Expected {key=} to be a string, got {v=}: {type(v)}")
+        return v
+
+    @property
+    def CONFIG_HOME(self) -> str:
+        return self._set_and_retrieve(
+            "CONFIG_HOME", platformdirs.user_config_dir(ILAB_PACKAGE_NAME)
+        )
```
tiran, line 107: I didn't notice this code before, because GH had collapsed the diff for `configuration.py`.

The way how you wrote this code, will make it slower, not faster. Every propery access will evaluate the default (2nd parameter) and then do extra work to look up the key in a dict. I recommend that you simplify this code and remove all caching. Path lookup is not a performance critical part of InstructLab. Besides it is super faster already. On my system, platformdirs lookup + `os.path.join` takes less than 5 microseconds:

```
$ python3 -m timeit -s 'import os, platformdirs' 'os.path.join(platformdirs.user_data_dir("instructlab"), "checkpoints")'
50000 loops, best of 5: 4.54 usec per loop
```

```suggestion
    @property
    def CONFIG_HOME(self) -> str:
        return platformdirs.user_config_dir(ILAB_PACKAGE_NAME)
```

tiran (changes requested):

.gitignore
n1hility: I disagree on the expectation to do a build in a read-only directory. A common practice for builds is to store output relative to the source (although using an ENV for redirection). Then you can inspect and access the content, and it's easy to know what (and for a build target to know what to clean). CI systems tend to place your working directory on a location that is ephermal and has fast storage, so when you start using non-relative you may be picking less optimal locations and could be making things more complicated.  As an example of the latter in CI build chains you may have a task that slurps up test output for formatting, but if its temp now you need logic to write out the location and find it. An example of the former is that tempfs is often undersized so if you have some significant test output it run out of storage. That said, CI scripts should reassign TMP/TMPDIR anyway.

tests/test_config.py
n1hility: +1 Keeping test constants separate and not linked to impl constants is good advice (although ok for tests to share between themselves)

.gitignore
tiran: Actually we both agree on the same thing. I now see how my original statement was not clear enough and could be easily misunderstood. What I meant to say is that tests should not modify the source directory / local git clone. Tests should not add, remove, or modify files like code, tests, and test data.

Instead tests should isolate themselves from the project data and any other test case. This makes tests reproducible and allows parallel execution of tests. Our pytest-based test cases use fixtures for tmp path and file system isolation for a unique CWD for each test. pytest has [options](https://docs.pytest.org/en/stable/how-to/tmp_path.html#temporary-directory-location-and-retention) to retain temporary files, move its base directory to a different disk, and deal with distributed tests.

Your remarks about tempfs and TMP sound plausible, too. I rarely have to deal with these low level things and usually let the test runner take care of my environment.

n1hility: ah yes sorry for misunderstanding, you are completely right here thanks!

leseb (changes requested):
Summarizing my request change one more time since my previous review was ignored:

* All the bash cosmetic changes like `s/$FOO/${FOO}` should be removed
* All the changes in the 'expect' commands should be removed (from the `scripts/functional-tests.sh` file)
* As agreed, the `to_upper()` compatibility for bash 3.2 should be removed
* I don't think we need to do any changes in `scripts/basic-workflow-tests.sh` unless they are required by the new `config` CLI
* `ilab system reset` was not discussed in any forum and not part of the design doc so it should be removed and proposed as a followup. It also has request changes and is not functionally tested
* we need to keep the focus on what needs to be implemented by the design doc and not get distracted by other items if we want to move faster with this PR.

If it sticks with the design proposal of `ilab config show`, the PR shouldn't be that big in terms of lines of code.
Most of the commits must be squashed, this work does not require 23 separate commits.

Thanks @RobotSail, we are almost there, keep it up!

cdoern:
@RobotSail @leseb I think if we are to address the above changes Seb put forward then we should break out the fixes for `save_samples` into a separate PR we can merge ASAP since main branch is currently broken. 

I was hoping we could avoid doing this but it is necessary probably

RobotSail:
@cdoern I can do that right now.

RobotSail:
The PR which splits out the `save_samples` change can be found here: https://github.com/instructlab/instructlab/pull/1626

scripts/basic-workflow-tests.sh
RobotSail: It seems like I missed them in this file and only did it in the functional-tests script. Changing this one as well.

src/instructlab/configuration.py
RobotSail: @tiran The point of caching this is not to provide a performance speedup but instead to prevent these values from being defined until they are needed.

The reason to cache here is so that after they are loaded, the variable (and all other common dependencies) consistently provides the same answer, even if the underlying envs change further in the script.

As to the point about speed, I wouldn't worry about it as these are O(1) operations. This is Python afterall, so things will only be so fast.

tiran: The only thing that could change behavior are env vars like HOME or XDG_DATA_HOME. If consistency is your concern, then you can cache the value of the three directories from platformdirs early. This allows for consistent values and still allows tests to `mock.patch` values:

```python
class _InstructlabDefaults:
    _cache_dir: str
    _config_dir: str
    _data_dir: str

    def __init__(self):
        self._reset()
    
    def _reset(self):
        pd = platformdirs.PlatformDirs(appname=ILAB_PACKAGE_NAME)
        self._cache_home = pd.user_cache_dir
        self._config_dir = pd.user_config_dir
        self._data_dir = pd.user_data_dir
    
    @property
    def CHECKPOINTS(self) -> str:
        return path.join(self._cache_home, STORAGE_DIR_NAMES.CHECKPOINTS)
```

RobotSail: Thanks @tiran, I'll use this method then.

RobotSail: This is much cleaner!

.gitignore
RobotSail: I removed this directory from the .gitignore, and now both the functional test & e2e tests are using the mktemp + trap approach.

scripts/basic-workflow-tests.sh
RobotSail: Gotcha. I've defined all 3 and have the scripts using them.

src/instructlab/configuration.py
RobotSail: Fixed

src/instructlab/data/generate.py
RobotSail: I've filed an issue for this [here](https://github.com/instructlab/instructlab/issues/1628), let's tackle it in a separate PR.

src/instructlab/system/reset.py
RobotSail: Filed for an issue for it [here](https://github.com/instructlab/instructlab/issues/1629), let's tackle it in a separate PR.

src/instructlab/configuration.py
RobotSail: resolved by #1591

tests/test_config.py
RobotSail: That makes sense, I'll fix this.

scripts/functional-tests.sh
RobotSail: Is there anyway we can automate this? I feel very strongly that our tests should not be overwriting an editable installation of `ilab` into a static one during development.

tests/test_config.py
RobotSail: The current tests work well-enough, we should prioritize this before merging that one.

src/instructlab/system/system.py
RobotSail: Done.

src/instructlab/config/init.py
```diff
@@ -56,19 +56,20 @@
 @click.option("--train-profile", type=click.Path(), default=None)
 def init(
     interactive,
-    model_path,
+    model_path: str,
```
JamesKunstle, line 59: the type in the @click.option is click.Path

src/instructlab/configuration.py
```diff
@@ -18,55 +18,139 @@
...
+    def EVAL_DATA(self) -> str:
+        return path.join(self.INTERNAL, "eval_data")
+
+
+DEFAULTS = _InstructlabDefaults()
+
 
 # Model families understood by ilab
-MODEL_FAMILIES = set(("merlinite", "mixtral"))
+MODEL_FAMILIES = {"merlinite", "mixtral"}
```
JamesKunstle, line 147: granite | mistral | dolomite?

src/instructlab/configuration.py
```diff
@@ -109,16 +193,18 @@ class _chat(BaseModel):
...
 
     # additional fields with defaults
     vi_mode: bool = False
     visible_overflow: bool = True
     context: str = "default"
-    session: Optional[str] = None
-    logs_dir: str = "data/chatlogs"
+    session: typing.Optional[str] = None
+    logs_dir: str = Field(
+        default_factory=lambda: DEFAULTS.CHATLOGS
```
JamesKunstle, line 204: this applies to other default dir names in config like MODELS: I'd recommend renaming to MODELS_DIR or something like that because DEFAULTS.CHATLOGS doesn't immediately make me think that the variable points to the chatlogs dir.

JamesKunstle: non-blocking btw

src/instructlab/model/download.py
```diff
@@ -18,7 +17,7 @@
 @click.command()
 @click.option(
     "--repository",
-    default="instructlab/merlinite-7b-lab-GGUF",  # TODO: add to config.yaml
+    default=config.DEFAULTS.MERLINITE_GGUF_REPO,  # TODO: add to config.yaml
```
JamesKunstle, line 20: elsewhere you import `from instructlab.configuration import DEFAULTS` which I kinda like better.
non-blocking, just a stylistic comment.

src/instructlab/model/test.py
```diff
@@ -71,13 +73,19 @@ def test(data_dir, model_dir, adapter_file):
...
+        load_and_train(
+            model=model_dir,
+            no_adapter=True,
+            max_tokens=100,
+            prompt=prompt,
+        )
 
         if adapter_file_exists:
             print("\n-----model output AFTER training----:\n")
+            assert processed_adapter is not None
```
JamesKunstle, line 85:
```suggestion
            assert processed_adapter is not None, ""
```

src/instructlab/system/reset.py
JamesKunstle, line 1: v smart way to do this

src/instructlab/system/sysinfo.py
```diff
@@ -0,0 +1,146 @@
...
+    info.update(_torch_hpu_info())
+    info.update(_llama_cpp_info())
+    return info
+
+
+@click.command()
+def sysinfo():
+    """Print system information"""
+    for key, value in get_sysinfo().items():
+        print(f"{key}: {value}")
```
JamesKunstle, line 146: perhaps consider click.echo?

JamesKunstle: non-blocking

JamesKunstle (approved):
said it before- this is a major QOL improvement because we have a handle on where we _actually_ want to store data. all looks good to me.

src/instructlab/system/sysinfo.py
RobotSail: Sure I'll switch this to use `click.echo`. I just copied this over from the previous file, so this isn't my code.

src/instructlab/configuration.py
RobotSail: It looks like there's some mapping logic which exists to handle the Granite case: https://github.com/instructlab/instructlab/blob/a94048fd65ddaa5e2a38b8034536a584897af832/src/instructlab/configuration.py#L63 

We can open an issue for this if you'd like, but we'd need to handle this in a separate PR.

src/instructlab/configuration.py
RobotSail: That makes sense, I think we can make that change.

src/instructlab/model/download.py
RobotSail: Okay we'll use this instead.

cdoern (approved):

cdoern:
e2e is failing @RobotSail is that a legitimate failure or a quick fix?

tests/test_lab_init.py
tiran: Please refactor the code to use `cli_runner`. It will automatically mock patch `os.environ` and reset it at the end of the test.

tests/test_lab_init.py
```diff
@@ -22,26 +18,54 @@ class TestLabInit:
...
+            # lazily import this module as a result of the following issue:
+            # https://github.com/instructlab/instructlab/issues/1582
+            # First Party
+            from instructlab import lab
+            from instructlab.configuration import DEFAULTS
```
tiran, line 27: This does not work reliable. You cannot guarantee that no other test module has imported `instructlab.lab` before this test runs.

tests/test_lab_init.py
```diff
@@ -22,26 +18,54 @@ class TestLabInit:
...
+
+            DEFAULTS._reset()
```
tiran, line 29: I suggest to move `DEFAULTS._reset()` to the `cli_runner` fixture.

russellb (changes requested):
that's a real e2e failure

also please clean up the commit history before merging, thanks!

tiran (changes requested):
Please rebase to get the new `cli_runner` fixture

Remove `os.environ["HOME"] = td`. The new fixture already sets `HOME` for you

I'm confused why you need the local, deferred import. Pytest imports all test modules first, then runs the tests. Other modules are importing `instructlab.lab` on module level.

src/instructlab/system/sysinfo.py
tiran: `click.echo` as a replacement for `print` is a bit overkill. I can refactor the code in another PR to use `click.echo` with additional bells and whistles like colors to highlight the key.

tests/test_lab_init.py
mairin: Is this a blocking requested change or can we move it to an issue? There are other PRs that are blocked waiting on this one.

tests/test_lab_init.py
mairin: Is this a blocking requested change or can we move it to an issue? There are other PRs that are blocked waiting on this one.

RobotSail: Fixed this locally.

n1hility:
I'm +1 once commits are squashed to a small set we and CI is passing

leseb:
The remaining list of things to address before this can be merged as agreed by @RobotSail @cdoern (and others):

- [ ] remove of `ilab system reset` (will be submitted in another PR)
- [ ] squash commits
- [ ] All the changes in the ‘expect’ commands should be removed (from the scripts/functional-tests.sh file)
- [ ] As agreed, the to_upper() compatibility for bash 3.2 should be removed

Once it's all done, feel free to dismiss my review, I trust your judgment on this. Thanks!

src/instructlab/config/init.py
tiran: `type=click.Path` creates a function parameter of type `str` by default. The type annotation `str` is correct.

tests/test_config.py
tiran: You are leaking information between tests case. This will break other cases in very hard to detect ways.

tiran: PR #1617 provides the infrastructure to fix the problem:
- accept `tmp_path_home` instead of `tmpdir`
- drop ` os.environ["HOME"] = str(tmpdir)`

tests/test_lab_init.py
```diff
@@ -20,61 +20,94 @@ class TestLabInit:
     # When using `import X`, you should use `X.Y` to patch.
     # https://docs.python.org/3/library/unittest.mock.html#where-to-patch?
     @patch("instructlab.config.init.Repo.clone_from")
-    def test_init_noninteractive(self, mock_clone_from, cli_runner: CliRunner):
-        result = cli_runner.invoke(lab.ilab, ["init", "--non-interactive"])
-        assert result.exit_code == 0
-        assert "config.yaml" in os.listdir()
-        mock_clone_from.assert_called_once()
+    def test_init_noninteractive(self, mock_clone_from):
```
tiran, line 23: Please use the new `cli_runner` fixture. It takes care of filesystem isolation and setting `HOME` for you.

tiran (dismissed):
Your rebase went wrong somehow. You have undone the `cli_runner` fixture for several test cases. The new fixture addresses the problem with leaking `os.environ["HOME"]`.

tests/test_lab_init.py
tiran: This comment is merely a suggestion.

The removal of `cli_runner` fixture and assignment to `os.environ` are blockers. I created https://github.com/instructlab/instructlab/pull/1617 specifically to solve the problem with leaking env vars.

tests/test_lab_init.py
RobotSail: I tried updating the tests to use it, I guess I missed a few. Thanks!

src/instructlab/system/sysinfo.py
RobotSail: Either way is fine with me. Marking as resolved for now as it sounds like this is a separate PR problem.

tests/test_lab_init.py
RobotSail: Fixed this in commit https://github.com/instructlab/instructlab/pull/1471/commits/dca8b1bba3fd3c044cc9f62ef1827a8df4d75517

src/instructlab/config/init.py
RobotSail: Thank you for the clarification @tiran !

src/instructlab/model/download.py
RobotSail: Resolved this.

src/instructlab/configuration.py
RobotSail: Resolved in commit https://github.com/instructlab/instructlab/pull/1471/commits/c8d73df3a538b99e2acd6d362bbbccacd82737eb

src/instructlab/system/reset.py
RobotSail: Yeah I wonder who I got the idea from 😉

src/instructlab/config/init.py
JamesKunstle: ++ the more you know!

src/instructlab/configuration.py
JamesKunstle: ++

scripts/functional-tests.sh
RobotSail: I made an issue for this here: https://github.com/instructlab/instructlab/issues/1641 

Removing the `pip install -e .` line here and closing this.

RobotSail:
@leseb 
>  All the changes in the ‘expect’ commands should be removed (from the scripts/functional-tests.sh file)

The changes I made to the `expect` commands are purely to format the new strings in. Since those tests were written to expect specific strings (which have now changed due to location differences) and config file locations. This will not be something I'm changing in this PR.

n1hility (approved):

PR #1495, opened 2024-06-27
Serve vllm using subprocess instead of multiprocessing

alimaredia:
**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -310,7 +310,7 @@ def get_uvicorn_config(app: uvicorn.Server, host: str, port: int) -> Config:
         timeout_keep_alive=0,  # prevent clients holding connections open (we only have 1)
     )
 
-def select_backend(logger: logging.Logger, serve_config: serve_config, model_family: str) -> BackendServer:
+def select_backend(logger: logging.Logger, serve_config: serve_config) -> BackendServer:
     from .llama_cpp import Server as llama_cpp_server
     from .vllm import Server as vllm_server
     backend_instance = None
@@ -333,7 +333,7 @@ def select_backend(logger: logging.Logger, serve_config: serve_config, model_fam
             gpu_layers=serve_config.gpu_layers,
             max_ctx_size=serve_config.max_ctx_size,
             num_threads=None,  # exists only as a flag not a config
-            model_family=model_family,
+            model_family=serve_config.model_family,
             host=host,
             port=port,
         )
@@ -344,7 +344,7 @@ def select_backend(logger: logging.Logger, serve_config: serve_config, model_fam
             logger=logger,
             api_base=serve_config.api_base(),
             model_path=model_path,
-            model_family=model_family,
+            model_family=serve_config.model_family,
             host=host,
             port=port,
         )
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -21,6 +21,8 @@ import uvicorn
 # Local
 from ...client import ClientException, list_models
 from ...configuration import get_api_base
+from ...configuration import _serve as serve_config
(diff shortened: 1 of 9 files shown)


src/instructlab/model/backends/backends.py
```diff
@@ -202,7 +203,8 @@ def ensure_server(
     host="localhost",
     port=8000,
     queue=None,
-) -> Tuple[multiprocessing.Process, str]:
+    server_process_func=None,
```
alimaredia, line 206: @n1hility @danmcp FYI I couldn't think of a less invasive way to refactor the code than to not set a type annotation for the `server_process_func`. Can you think of anything cleaner?

src/instructlab/model/backends/backends.py
```diff
@@ -202,7 +203,8 @@ def ensure_server(
...
+) -> Tuple[multiprocessing.Process, subprocess.Popen, str]:
```
alimaredia, line 207: @n1hility @danmcp I don't feel great about returning this tuple (one process for each backend) but in line with my comment one line aboe it felt like the least invasive thing I could do.

src/instructlab/model/backends/vllm.py
```diff
@@ -41,154 +17,109 @@ def __init__(
...
             raise exc
 
     def shutdown(self):
         """Shutdown vllm server"""
-        # Needed when the chat is connected to a server and the temporary server is started
+        # Needed when a temporary server is started
         if self.process is not None:
-            self.logger.debug("Terminating temporary server.")
             self.process.terminate()
-            self.process.join(timeout=30)
```
n1hility, line 107: you probably want a process.wait() timeout, and on exception process.kill

src/instructlab/model/backends/backends.py
n1hility: we discussed options on chat, decided this cleanup can wait till later

src/instructlab/model/backends/backends.py
n1hility: this is good practice IMO could always improve the abstraction later

n1hility (approved):
My only lingering concern is that the introduction of vllm and llamacpp args contradicts our dev doc choice. Although after discussing with Ali, following up in a later PR is a good idea to unblock eval.

n1hility (approved):

danmcp (dismissed):

src/instructlab/model/backends/backends.py
```diff
@@ -248,41 +270,71 @@ def ensure_server(
...
         temp_api_base = get_api_base(host_port)
         logger.debug(f"Starting a temporary server at {temp_api_base}...")
+        llama_cpp_server_process = None
+        vllm_server_process = None
+
+        if backend == VLLM:
+            # TODO: resolve how the hostname is getting passed around the class and this function
+            vllm_server_process = server_process_func(port=port)
+            count = 0
+            # TODO should this be configurable?
```
nathan-weinberg, line 280: Agree but this can definitely be a followup

nathan-weinberg (approved):
Just sat with @alimaredia who talked me through all my comments

n1hility (approved):

danmcp (approved):

src/instructlab/model/chat.py
```diff
@@ -176,43 +175,10 @@ def chat(
...
-                model_path=model_path,
-                model_family=model_family,
-                host=host,
-                port=port,
-            )
+        ctx.obj.config.serve.llama_cpp.llm_family = model_family
+        backend_instance = backends.select_backend(logger, ctx.obj.config.serve)
 
         try:
             # Run the llama server
```
danmcp, line 184: I could be reading this wrong and it was already the case in the code, but is this run connected to the finally below?  And shouldn't it be?

PR #1498, opened 2024-06-27
[DNM] Vllm subprocess breakout ci

alimaredia:
This PR is just a breakdown of https://github.com/instructlab/instructlab/pull/1495 to test commits one by one.

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -137,15 +137,20 @@ class _generate(BaseModel):
     prompt_file: StrictStr = DEFAULT_PROMPT_FILE
     seed_file: StrictStr = "seed_tasks.json"
 
+
 class _serve_vllm(BaseModel):
     """Class describing configuration of vllm serving backend."""
+
     # arguments to pass into vllm process
     vllm_args: str = ""
 
+
 class _serve_llama_cpp(BaseModel):
     """Class describing configuration of llama-cpp serving backend."""
+
     gpu_layers: int = -1
     max_ctx_size: PositiveInt = 4096
+
 
 class _serve(BaseModel):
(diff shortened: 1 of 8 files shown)


src/instructlab/configuration.py
```diff
@@ -138,20 +138,39 @@ class _generate(BaseModel):
     seed_file: StrictStr = "seed_tasks.json"
 
 
+class _serve_vllm(BaseModel):
```
cdoern, line 141: you need to mansually splice these into the `default_map`. I do something similar for training below.

src/instructlab/configuration.py
```diff
@@ -190,7 +209,16 @@ def get_default_config():
...
         ),
-        serve=_serve(model_path=DEFAULT_MODEL_PATH),
+        serve=_serve(
+            model_path=DEFAULT_MODEL_PATH,
+            llama_cpp=_serve_llama_cpp(
+                gpu_layers=-1,
+                max_ctx_size=4096,
+            ),
+            vllm=_serve_vllm(
+                vllm_args="",
```
cdoern, line 219: if the default is "". this should be optional

cdoern: maybe?

src/instructlab/model/serve.py
```diff
@@ -95,6 +95,11 @@ def serve(
     backend_instance = None
     if backend == backends.LLAMA_CPP:
         # Instantiate the llama server
+        if gpu_layers == None:
```
cdoern, line 98: this is where the defaults should auto be populated into the flags from the cfg

cdoern (commented):
initial pass, looks good so far

PR #1507, opened 2024-06-28
e2e: Reorder generate test cases

russellb:
The script runs generation against 3 taxonomy types. Reorder so
knowledge is last. Whatever runs last is what training runs again. I
want knowledge last for testing some SDG library output in that repo,
since that's what is working first in the major rework happening at
the moment.

The impact shouldn't be noticable.


russellb:
I'm going to quick merge this -- it's a trivial thing and I want it for testing some SDG library work.  I will keep an eye on this in case I typoed something, but otherwise it should have no impact for most people

Issue #1508, opened 2024-06-28
HTML tags are not stripped off from the knowledge document


vincentywdeng:
**Describe the bug**
When processing knowledge document, html tags are no stripped. As html tags doesn't contain meaningful information when doing `ilab data generate`. It is better to remove them

**To Reproduce**
1. Follow [Oscars knowledge examples](https://github.com/instructlab/taxonomy?tab=readme-ov-file#knowledge-yaml-examples) which trying to get this [knowledge document(https://github.com/juliadenham/oscars2024_knowledge/blob/main/oscars2024_results.md)
2. Set a break point at https://github.com/instructlab/instructlab/blob/214fe45abd8d0d2db28bf62d414c72b8d5858a48/src/instructlab/utils.py#L537 , you will find html tags are not removed
3. Set break point at [encode_prompt method in sdg](https://github.com/instructlab/sdg/blob/6c9d69a8ff14a443655ee14b2eb7bd9cd0159f50/src/instructlab/sdg/generate_data.py#L139),  the document still contain html tags

**Expected behavior**
The tags should be stripped as it didn't contain meaningful information

**Screenshots**

**Device Info (please complete the following information):**
 - Hardware Specs: Apple M1 Max 32 GB Memory
 - OS Version: Sonoma 14.5
 - Python Version: Python 3.10.13
 - InstructLab Version: 0.17.1.dev113

**Additional context**

nathan-weinberg:
Closing in favor of the SDG repo issue

PR #1509, opened 2024-06-28
Stripped off html tags from knowledge document

vincentywdeng:
Issue: #1508

This PR stripped off html tags from knowledge document as they didn't contain meaningful information and removing them save precious context space

**Checklist:**

### DIFF
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -2,6 +2,7 @@
 
 # Standard
 from functools import cache, wraps
+from html.parser import HTMLParser
 from pathlib import Path
 from typing import Any, Dict, List, Mapping, Optional, Union
 from urllib.parse import urlparse
@@ -39,6 +40,25 @@ rules:
 
 class TaxonomyReadingException(Exception):
     """An exception raised during reading of the taxonomy."""
+
+
+class HTMLTagStripper(HTMLParser):
+    def reset(self):
+        super().reset()
+        self.clean_text = ""
+
+    def strip_html_tags(self, text: str) -> str:
+        """Returns text without html tags for given text input.
+        
+        :param text: The original text to be processed
+        """
+        self.reset()
+        self.feed(text)
+        self.close()
+        return self.clean_text
+
+    def handle_data(self, data):
+        self.clean_text += data
 
 
 def macos_requirement(echo_func, exit_exception):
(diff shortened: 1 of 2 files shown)


leseb (changes requested):
Thanks for contributing. I just approved to run the CI. Could you squash your commits? Thanks

src/instructlab/utils.py
```diff
@@ -448,8 +468,13 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
         tax_path = "->".join(taxonomy_path.parent.parts)
         task_description = contents.get("task_description")
         documents = contents.get("document")
+
         if documents:
-            documents = get_documents(source=documents)
+            tag_stripper = HTMLTagStripper()
+            documents = [
+                tag_stripper.strip_html_tags(doc)
+                for doc in get_documents(source=documents)
```
bjhargrave, line 476: I think any stripping should be in the get_documents method. Also note that the sdg repo has a copy of this method which will need similar treatment. In reality, this change should probably only be in the sdg repo. The only user in this repo is now `ilab taxonomy diff` which does not need to actually process the files.

bjhargrave (changes requested):

src/instructlab/utils.py
vincentywdeng: Thanks for the comment, moved to get_document. I'm not understand this 
```
 In reality, this change should probably only be in the sdg repo. The only user in this repo is now ilab taxonomy diff which does not need to actually process the files.
```
Because the command `ilab data generate` seems till go to utils.py in this repo. 
My version is `0.17.1.dev240`

bjhargrave: https://github.com/instructlab/instructlab/blob/b65cfc9d9a3ee4ae041c780c7b46228e35c8e22a/src/instructlab/data/generate.py#L164

The generate_data method is now in the instructlab.sdg package.

vincentywdeng: Thanks @bjhargrave , just notice that my sdg version is not new. I'll create a ticket in that repo.

vincentywdeng: Issue created 
Issue: https://github.com/instructlab/sdg/issues/79
PR: https://github.com/instructlab/sdg/pull/80

vincentywdeng:
close this PR as the code was moved to sdg repo

PR #1522, opened 2024-06-28
generate: Allow pipeline selection

russellb:
3edc0de generate: Allow pipeline selection
3ce1541 tests: drop test_lab_generate

commit 3edc0de760770af9c8da7570062c8e298c900f7b
Date:   Fri Jun 28 16:42:44 2024 -0400

    generate: Allow pipeline selection
    
    Some major enhancements have gone into the `instructlab/sdg` library
    this week, including a more extensive pipeline that produces higher
    quality results. This new pipeline requires a larger teacher model
    than what we default to in the CLI, so it is optional and not used by
    default.
    
    This option provides the ability to turn it on by specifying
    `--pipeline full`. The default value is `simple`, which maps to a
    simple pipeline aimed at producing similar quality results as before,
    though it is a new and different implementation. It's intended to be
    good enough for basic testing and demos on a laptop or equivalent
    small environment.
    
commit 3ce1541812c5a25707f4fa715f2244a8a2861cfe
Date:   Fri Jun 28 17:06:17 2024 -0400

    tests: drop test_lab_generate
    
    These tests try to mock things internal to the SDG library. They're
    not that valuable anymore. Focused unit testing should be in the
    library. This repo does run end-to-end tests running `ilab data
    generate`, so we get coverage that way. We also run generate in
    the functional test suite.
    

russellb:
Since the patch was so tiny, I went straight to the code PR instead of a design doc, though I'm happy to start one of those if we need to get into a more detailed design discussion!

I do plan to propose a design next week for how we would expose a more advanced capability to add custom pipeline extensions that would hang off of the end of one of the built-in pipelines. I'll definitely write that up properly.

The alternative for this PR would be to have full configuration for ALL pipelines. The configuration is pretty complex, though. Here's what would have to go into a config to represent the "full" pipeline selectable by this PR: https://github.com/instructlab/sdg/blob/6c9d69a8ff14a443655ee14b2eb7bd9cd0159f50/src/instructlab/sdg/default_flows.py#L24-L396

requirements.txt
```diff
@@ -11,7 +11,8 @@ GitPython>=3.1.42
 httpx>=0.25.0
 instructlab-quantize>=0.1.0
 instructlab-schema>=0.2.0
-instructlab-sdg>=0.0.4
+#instructlab-sdg>=0.0.4
+instructlab-sdg @ git+https://github.com/russellb/sdg.git@new-cli-integration
```
tiran, line 15: You are probably installing from git for testing.
Just for the record, PyPI does not accept packages with these types of links. Our CI doesn't detect the issue, yet.

russellb: Thanks!  Yeah, this was just for testing. That branch merged into the repo, so it could chain to the instructlab/sdg repo and main branch, but it's still not in a release.

I posted it here for early feedback and testing on the approach

requirements.txt
```diff
@@ -12,7 +12,7 @@ httpx>=0.25.0
 instructlab-eval>=0.0.3
 instructlab-quantize>=0.1.0
 instructlab-schema>=0.2.0
-instructlab-sdg>=0.0.4
+instructlab-sdg>=0.0.4.1
```
RobotSail, line 15: Not blocking this PR, just a quick question: what is the reason for doing this vs. just bumping to 0.0.5?

RobotSail (approved):
LGTM

requirements.txt
russellb: because we've merged a bunch of stuff to `main` that I didn't want to release yet. I made `v0.0.4.1` that is one tiny trivial change on top of `v0.0.4` to allow the `pipeline` parameter to be specified, though it doesn't actually support any other pipeline types yet (in v0.0.4.1 anyway, it does in `main`)

russellb: https://github.com/instructlab/sdg/compare/v0.0.4...v0.0.4.1

russellb: ^^^ the release, vs what's going to be in v0.0.5 -- https://github.com/instructlab/sdg/compare/v0.0.4...main

RobotSail: Ah I see. I didn't realize that you could have version numbers beyond the standard semver ones in PyPi.

src/instructlab/data/generate.py
```diff
@@ -125,6 +125,15 @@
     "--model-family",
     help="Force model family to use when picking a generation template",
 )
+@click.option(
+    "--pipeline",
+    type=click.STRING,
+    default="simple",
```
cdoern, line 131: should this be a click.choice with the only option as simple? otherwise we will need to change flag type in the future.

cdoern, line 131: if not a click.choice then how about constants for each pipeline type in configuration.py :thinking:

cdoern (commented):
just one thought about the flag types.

requirements.txt
russellb: Yeah, I had to check before  I did it.

https://peps.python.org/pep-0440/

> The canonical public version identifiers MUST comply with the following scheme:
> 
> `[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`

or the important piece of it:

> Release segment: `N(.N)*`

so it looks like you can have as many `.N` segments as you want

src/instructlab/data/generate.py
russellb: oh good call. i didn't realize that was an option, thanks

russellb: updated to use click.choice, thanks for the suggestions

cdoern (approved):
lgtm

requirements.txt
nathan-weinberg: X.Y.Z.??? :wink:

russellb: @nathan-weinberg what’s do you mean?

anyway this release will probably last 1-2 days max

nathan-weinberg: I was just wondering what would come after Z (to be clear I have no issue with this)

tiran: That depends on your native language. For Swedish people it's `å`, `ä`, and `ö`. In German, `z` is the last letter because `ä` comes after `a` (`ä` is sometimes transliterated to `ae`).

tiran (approved):

src/instructlab/data/generate.py
```diff
@@ -125,6 +125,15 @@
     "--model-family",
     help="Force model family to use when picking a generation template",
 )
+@click.option(
+    "--pipeline",
+    type=click.Choice(["simple", "full"], case_sensitive=False),
```
leseb, line 130: Why do we add `full` if it's not implemented?

russellb: it's implemented in `instructlab-sdg` main and I want to be able to test it

requirements.txt
nathan-weinberg: That would be fun but hard to write on my US keyboard :joy:

PR #1526, opened 2024-06-29
Fix various issues with tox and tests

tiran:
`tox -e py3-unit` or `make tests` no longer works when the default Python interpreter is 3.12. InstructLab currently requires Python 3.10 or 3.11. All tox environments without a fully qualified Python version prefix now enforce Python 3.11 as default `basepython`. We cannot set `basepython = python3.11` in `[testenv]`, because that breaks `[testenv:py310]` enviroments. The `[testenv:py3-*]` also enforce Python 3.11 as `basepython`.

`tox` no longer uses `requirements-dev.txt`. The requirements file installs additional packages that are not needed for testing, e.g. tox within tox. It also pulls in all dependencies from `requirements.txt`. We want to verify that `pip install instructlab` installs all dependencies automatically.

Other changes and improvements:

- run `mypy` with `make tests` instead of `make verify`. The verify target is for fast local testing without installing Torch stack.
- use a tox label to run all fast verify targets
- update Ruff to 0.5.0
- drop `ruff.sh` and use two separate targest for ruff checking and ruff fixing. Ruff and isort don't need much time and disk space to install.

**Checklist:**


tox.ini
```diff
@@ -7,14 +7,21 @@ envlist = ruff, lint, mypy, spellcheck, py3-{unit, functional}, tomllint
...
 # are huge. This reduces venv from 5.7 GB to 1.5 GB.
 setenv =
     PIP_EXTRA_INDEX_URL=https://download.pytorch.org/whl/cpu
 package = wheel
 wheel_build_env = pkg
-deps = -r requirements-dev.txt
+
+[testenv:py3]
+description = run tests (unit, unitcov, functional)
+basepython = python3.11
```
bjhargrave, line 19: Doesn't this mean that we never test on python 3.10 now? The test.yml worklflow includes python 3.10. But then this overrides that and tests with 3.11?

bjhargrave (commented):
This change does not seem to result in tox actually running the unit tests. 

See https://github.com/instructlab/instructlab/actions/runs/9723803470/job/26839329939#step:12:27 for no output from the test execution.

tox.ini
tiran: tox's rules for inheritance and Python version mapping are not very intuitive. It's one of the reasons I personally prefer nox over tox.

Tox default:
- `[toxenv:py3]`, `[toxenv:py310]`, and `[toxenv:py311]` are independent and inherit settings from `[toxenv]`.
- `[toxenv:py3]` uses `python3` binary as default base Python. This can point to any Python 3 version. On my system, it is currently Python 3.12
- `[toxenv:py310]` uses `python3.10` binary as default base Python
- `[toxenv:py311]` uses `python3.11` binary as default base Python

To my surprise, ``[toxenv:py3-unit]` does not inherit settings from `[toxenv:py3]`.

My changes modifies `[toxenv:py3-unit]` and `[toxenv:py3-function]` to use `python3.11` instead of `python3`, so `tox -e py3-unit` works when `python3` points to `python3.12`.

Our CI already maps GH envs to tox envs with fully qualified Python version, so we properly test with 3.10 and 3.11.

```
[gh]
python =
    3.11 = py311-{unitcov, functional}
    3.10 = py310-{unitcov, functional}
```

bjhargrave: OK, but the unit tests are not running in this configuration. Look at the output from py3-unit. It does not call pytest. It just sets up pytest for execution but does not run it.

tiran: Mpf, I got fooled by tox's non-intuitive inheritance again. Fixing now...

.github/workflows/lint.yml
```diff
@@ -46,10 +46,11 @@ jobs:
         lint:
           - name: "ruff"
             commands: |
-              tox -e ruff -- check
+              tox -e ruff
           - name: "pylint"
             commands: |
               echo "::add-matcher::.github/workflows/matchers/pylint.json"
+              tox -e fastlint
```
bjhargrave, line 53: Is there a reason to run both fastlint and lint? It seems that fastlint is a subset of lint?

tiran: It behaves slightly differently than regular lint and has been broken two times in the last few weeks. Pylint cannot figure out on its own that `instructlab.eval` is a 3rd party package, so we need to teach it every external instructlab package.

bjhargrave (approved):

russellb (approved):

tox.ini
```diff
@@ -8,13 +8,20 @@ minversion = 4.4
...
 description = run tests (unit, unitcov, functional)
+passenv =
+    CMAKE_ARGS
 # Use PyTorch CPU build instead of CUDA build in test envs. CUDA dependencies
 # are huge. This reduces venv from 5.7 GB to 1.5 GB.
 setenv =
     PIP_EXTRA_INDEX_URL=https://download.pytorch.org/whl/cpu
+    CMAKE_ARGS={env:CMAKE_ARGS:-DLLAMA_NATIVE=off}
 package = wheel
 wheel_build_env = pkg
```
booxter, line 19: I've noticed that `pkg` environment is built using `python3.12`. I am not sure it's an actual issue. (For what I observe, tests still pass.)

```
$ .tox/pkg/bin/python --version
Python 3.12.3
```

The following change seems to switch to 3.11:

```
--- a/tox.ini
+++ b/tox.ini
@@ -34,6 +34,9 @@ allowlist_externals =
 [testenv:py3]
 basepython = python3.11

+[testenv:pkg]
+basepython = {[testenv:py3]basepython}
+
 [testenv:py3-unit]

```

Since you are clearly more knowledgeable about packaging / tox, I'd like to ask: do you anticipate any problems with us using py3.12 for wheel build environment? Or it's a non-issue?

PR #1530, opened 2024-06-30
offshoot new_port() from ensure_server()

makelinux:
Current implementation of ensure_server() is
dirty mix of code.

Start to refactor it to make the code clean and manageable.


tiran:
FWIW, the concept and idea of `new_port` is wrong.

We should use port `0` to have the OS select a random port for us, then get the port number from uvicorn. See https://github.com/encode/uvicorn/issues/761

leseb (changes requested):
This should wait for https://github.com/instructlab/instructlab/pull/1533.

PR #1531, opened 2024-06-30
Fix: No str.split or f-string for subprocess args

tiran:
Never use `str.split` to split arguments to subprocess or f-strings to create arguments for subprocess. This breaks as soon as there a non-ASCII characters in any path, e.g. file names with a white space.

You either need `shlex.split` and `shlex.quote` (which does not work on non-Unix shells) or use list of string all the way.

Also fix a typing bug with port. It's now an `int` all the way.

**Checklist:**

### DIFF
--- a/src/instructlab/model/serve.py
+++ b/src/instructlab/model/serve.py
@@ -3,6 +3,7 @@
 # Standard
 import logging
 import pathlib
+import typing
 
 # Third Party
 import click
@@ -56,7 +57,12 @@ logger = logging.getLogger(__name__)
 @click.option(
     "--vllm-args",
     type=str,
-    help="Specific arguments to pass into vllm. The value passed into this flag must be surrounded by double quotes.\n Example: --vllm-args '--dtype auto --enable-lora'",
+    multiple=True,
+    help=(
+        "Specific arguments to pass into vllm. Each arg must be passed "
+        "separately and surrounded by quotes.\n"
+        " Example: --vllm-args='--dtype=auto' --vllm-args='--enable-lora'"
+    ),
 )
 @click.pass_context
 @utils.display_params
@@ -69,7 +75,7 @@ def serve(
     model_family,
     log_file,
     backend,
-    vllm_args,
+    vllm_args: typing.Iterable[str],
 ):
     """Start a local server"""
     # pylint: disable=import-outside-toplevel
@@ -116,9 +122,6 @@ def serve(
 
     if backend == backends.VLLM:
         # Instantiate the vllm server
-        if vllm_args is None:
-            vllm_args = ""
-
         backend_instance = vllm.Server(
             logger=logger,
             api_base=ctx.obj.config.serve.api_base(),
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -144,7 +144,7 @@ class _serve_vllm(BaseModel):
     """Class describing configuration of vllm serving backend."""
 
     # arguments to pass into vllm process
-    vllm_args: str = ""
+    vllm_args: list[str] | None = None
 
 
 class _serve_llama_cpp(BaseModel):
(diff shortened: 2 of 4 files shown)


nathan-weinberg (approved):

bjhargrave (dismissed):

leseb (commented):
How are users expected to call `vllm-args` from the CLI now?

tiran:
> How are users expected to call `vllm-args` from the CLI now?

It's now a list of strings instead of a single string. Let me add a unit test as example.

```
serve:
  vllm:
    vllm_args:
      - "--egg"
      - "--spam"
```

leseb:
> > How are users expected to call `vllm-args` from the CLI now?
> 
> It's now a list of strings instead of a single string. Let me add a unit test as example.
> 
> ```
> serve:
>   vllm:
>     vllm_args:
>       - "--egg"
>       - "--spam"
> ```

Yes I understand the structure for the config but how about the CLI `--vllm-args` though? Thanks

tiran:
> Yes I understand the structure for the config but how about the CLI `--vllm-args` though? Thanks

I didn't know that there was a `--vllm-args` CLI option because there was no unit test or integration test. I now fixed the option to be multi-arg, so you can do `--vllm-args='--dtype auto' --vllm-args='--enable-lora'`.

src/instructlab/model/serve.py
```diff
@@ -56,7 +56,12 @@
 @click.option(
     "--vllm-args",
     type=str,
-    help="Specific arguments to pass into vllm. The value passed into this flag must be surrounded by double quotes.\n Example: --vllm-args '--dtype auto --enable-lora'",
+    multiple=True,
+    help=(
+        "Specific arguments to pass into vllm. Each arg must be passed "
+        "separately and surrounded by quotes.\n"
+        " Example: --vllm-args='--dtype auto' --vllm-args='--enable-lora'"
```
bjhargrave, line 63: Isn't `'--dtype auto'` really 2 args?

```suggestion
        " Example: --vllm-args='--dtype' --vllm-args='auto' --vllm-args='--enable-lora'"
```

or should it be `'--dtype=auto'`?

src/instructlab/model/serve.py
```diff
@@ -116,9 +121,6 @@ def serve(
 
     if backend == backends.VLLM:
         # Instantiate the vllm server
-        if vllm_args is None:
```
alimaredia, line 119: @tiran what was the reason for removing this? `vllm_args` defaults to `None` if the flag is not passed in.

alimaredia: Upon playing around with the code unset `--vllm-args` defaults to an empty tuple.

src/instructlab/model/serve.py
```diff
@@ -69,7 +74,7 @@ def serve(
     model_family,
     log_file,
     backend,
-    vllm_args,
+    vllm_args: list[str],
```
alimaredia, line 77: @tiran I noticed that if the flag is not set, the type of `vllm_args` is a tuple not a list. Do you need to change anything to handle this case?

alimaredia (changes requested):
I've run this PR by hand. Both `ilab model serve` with `--vllm-args` and `ilab model chat` with `vllm_args` set in the config worked as expected. 

The only thing I couldn't follow is if we could be handing the case of `--vllm-args` not being passed into `ilab model serve`. The value of `vllm_args` becomes a tuple, but we expect a list in vllm.py.

src/instructlab/model/serve.py
tiran: I have updated the code to use `typing.Iterable[str]` in the outer layer and `list[str]` internally.

alimaredia (approved):

nathan-weinberg (approved):

alimaredia:
Removing the "hold" label as Sebastians question about the structure of `--vllm-args` for ilab serve has been addressed by Christian.

Nevertheless future design discussion about `--backend-args` instead of `--vllm-args` need to take place in a future PR.

Issue #1467, opened 2024-06-26
ilab startup time increased after training lib merge


russellb:
On my laptop, `ilab` started in about 1.5s prior to the merge and 3s after. This is from `time ilab --help`.

russellb:
This may have the same fix as #1465 since they're both related to what is imported at startup

JamesKunstle:
Yep I think these two issues are directly linked to imports from the `training` library in `configuration.py` and `train.py`.  When I comment `training` repo stuff out of those files, `ilab` boots immediately and those logs aren't emitted.

tiran:
There are also some more places:

- The import of torch in `train` should be guarded by `if typing.TYPE_CHECKING`.
```
if typing.TYPE_CHECKING:
    import torch
```
- `instructlab.model.serve` needlessly imports `llama_cpp`

I also recommend to add a unit test, so we do not regress. Here is an easy way to dump imports from a subprocess:
```
import json
import subprocess
import sys

modules_txt = subprocess.check_output([sys.executable, "-c", "import sys, json, instructlab.lab; json.dump(sorted(sys.modules), sys.stdout)"], text=True)
modules = json.loads(modules_txt)
```

Other ways are `python3 -v -m instructlab` or `python3 -X importtime -m instructlab`.

tiran:
I got imports down from 5,906 modules and 6.5s to 1,387 imports and 0.8s startup time.

tiran:
PR #1716 speed up the CLI even further. `ilab model train --help` now takes 0.4s instead of 1.6s on my machine.

PR #1532, opened 2024-06-30
Fix slow `ilab` CLI startup

tiran:
Avoid importing `llama_cpp` by importing the `ServerException` from its
original module.

Defer import of `run_training` to avoid import of `torch`.

Move imports from `instructlab.eval` into function to avoid import of
`torch`.

Add a test case to test for slow and heavy imports like `torch` and
`llama_cpp`. 

Allow `import-outside-toplevel` globally.

See: #1467
See: https://github.com/instructlab/training/pull/116

**Checklist:**

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -1,28 +1,33 @@
 # Standard
+import enum
 import logging
 import os
+import typing
 
 # Third Party
-from instructlab.eval.evaluator import Evaluator
-from instructlab.eval.mmlu import MMLUBranchEvaluator, MMLUEvaluator
-from instructlab.eval.mt_bench import MTBenchBranchEvaluator, MTBenchEvaluator
 import click
 
 # Local
 from ..utils import http_client
 
+if typing.TYPE_CHECKING:
+    # Third Party
+    from instructlab.eval.evaluator import Evaluator
+
 logger = logging.getLogger(__name__)
 
-BENCHMARK_TO_CLASS_MAP = {
-    "mmlu": MMLUEvaluator,
-    "mmlu_branch": MMLUBranchEvaluator,
-    "mt_bench": MTBenchEvaluator,
-    "mt_bench_branch": MTBenchBranchEvaluator,
-}
 
 JUDGE_MODEL_NAME = "judge_model"
 TEST_MODEL_NAME = "test_model"
 BASE_TEST_MODEL_NAME = "base_test_model"
+
+
+# Python 3.10 does not have StrEnum
+class Benchmark(str, enum.Enum):
+    MMLU = "mmlu"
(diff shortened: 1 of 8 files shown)


JamesKunstle (approved):
lgtm, very helpful changes!

src/instructlab/model/evaluate.py
```diff
@@ -204,8 +208,7 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
 )
 @click.option(
     "--benchmark",
-    type=click.Choice(list(BENCHMARK_TO_CLASS_MAP.keys())),
-    # case_sensitive=False,
+    type=click.Choice(["mmlu", "mmly_branch", "mt_branch", "mt_bench_branch"]),
```
nathan-weinberg, line 211:
```suggestion
    type=click.Choice(["mmlu", "mmlu_branch", "mt_branch", "mt_bench_branch"]),
```

nathan-weinberg (changes requested):
Has evaluation been tested with these changes?

src/instructlab/model/evaluate.py
tiran: Sigh ... And that's why ad-hoc string constants are bad and enums are better.

leseb (commented):
How can we prevent new code additions from slowing down the CLI again? How about adding a CI test that compares the startup time before and after a new code addition? Thanks!

tiran:
I have included a test case that will fail if `instructlab.lab` code starts depending on `torch`, `vllm`, or `llama_cpp` packages again. The test runs Python in a subprocess and blocks these modules before it tries to import `instructlab.lab`.

I don't like time-based tests, because they depend on the performance of the local machine.

leseb:
> I have included a test case that will fail if `instructlab.lab` code starts depending on `torch`, `vllm`, or `llama_cpp` packages again. The test runs Python in a subprocess and blocks these modules before it tries to import `instructlab.lab`.
> 
> I don't like time-based tests, because they depend on the performance of the local machine.

I saw the test and it is a good guard rail but does not help understand the startup time.
I understand your concern about the time-based test. However this test will always run on the CI runner, the test will be local, and so will be the diff so it does not matter how performant it is. The CI does not have to fail, it can simply print the startup time before and after new code is being added.

tiran:
Can you work on a timing check in a subsequent PR?

leseb:
> Can you work on a timing check in a subsequent PR?

Will do!

leseb (approved):

danmcp:
I've tested out the changes with evaluate.  Changes look good.  And much faster!

nathan-weinberg (approved):

danmcp (approved):

PR #1539, opened 2024-06-30
Use new vllm/llama-cpp backends for evaluate

danmcp:
Followup to #1369

This PR is using the new serving backends and support the vllm and llama_cpp path.  It replaces temporary code serving vllm directly.

**Checklist:**

- [ x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
- [x ] Unit tests have been added, if necessary.

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -1,13 +1,17 @@
 # Standard
+import logging
 import os
-import subprocess
-import time
 
 # Third Party
 from instructlab.eval.evaluator import Evaluator
 from instructlab.eval.mmlu import MMLUBranchEvaluator, MMLUEvaluator
 from instructlab.eval.mt_bench import MTBenchBranchEvaluator, MTBenchEvaluator
 import click
+
+# Local
+from ..utils import http_client
+
+logger = logging.getLogger(__name__)
 
 BENCHMARK_TO_CLASS_MAP = {
     "mmlu": MMLUEvaluator,
@@ -16,8 +20,9 @@ BENCHMARK_TO_CLASS_MAP = {
     "mt_bench_branch": MTBenchBranchEvaluator,
 }
 
-# TODO: Remove after proper vLLM integration
-# pylint: disable=consider-using-with
+JUDGE_MODEL_NAME = "judge_model"
+TEST_MODEL_NAME = "test_model"
+BASE_TEST_MODEL_NAME = "base_test_model"
 
 
 def get_evaluator(
@@ -68,10 +73,12 @@ def get_evaluator(
             raise click.exceptions.Exit(1)
         evaluator_class = BENCHMARK_TO_CLASS_MAP[benchmark]
         if benchmark == "mt_bench":
-            return evaluator_class("test_model", "judge_model", output_dir, max_workers)
+            return evaluator_class(
+                TEST_MODEL_NAME, JUDGE_MODEL_NAME, output_dir, max_workers
+            )
         return evaluator_class(
-            "test_model",
-            "judge_model",
+            TEST_MODEL_NAME,
+            JUDGE_MODEL_NAME,
             taxonomy_path,
             branch,
             output_dir,
@@ -191,6 +198,41 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
     return qna_to_avg_scores
 
 
+def launch_server(
+    ctx,
+    model,
+    model_name,
+    max_workers,
+) -> tuple:
+    # pylint: disable=import-outside-toplevel
+    # First Party
+    from instructlab.model.backends import backends
+
+    if not ctx.obj.config.serve.backend:
+        ctx.obj.config.serve.backend = backends.VLLM
+    if ctx.obj.config.serve.backend == backends.VLLM:
+        ctx.obj.config.serve.vllm.vllm_args.extend(["--served-model-name", model_name])
+    elif ctx.obj.config.serve.backend == backends.LLAMA_CPP:
+        # mt_bench requires a larger context size
+        ctx.obj.config.serve.llama_cpp.max_ctx_size = 5120
+        # llama-cpp fails fast on too many incoming requests and returns errors to client
+        ctx.obj.config.evaluate.mt_bench.max_workers = min(max_workers, 16)
+
+    ctx.obj.config.serve.model_path = model
+
+    backend_instance = backends.select_backend(logger, ctx.obj.config.serve)
+    try:
+        # http_client is handling tls params
+        backend_instance.run_detached(http_client(ctx.params))
+        api_base = backend_instance.api_base
+    except Exception as exc:
+        click.secho(f"Failed to start server: {exc}", fg="red")
+        raise click.exceptions.Exit(1)
+    if not api_base:
+        api_base = ctx.obj.config.serve.api_base()
+    return backend_instance, api_base
+
+
 @click.command()
 @click.option(
     "--model",
@@ -253,9 +295,34 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
     type=click.Path(),
     help="Path where all the MMLU Branch tasks are stored. Needed for running mmlu_branch.",
 )
+@click.option(
+    "--tls-insecure",
+    is_flag=True,
+    help="Disable TLS verification for model serving.",
+)
+@click.option(
+    "--tls-client-cert",
+    type=click.Path(),
+    default="",
+    show_default=True,
+    help="Path to the TLS client certificate to use for model serving.",
+)
+@click.option(
+    "--tls-client-key",
+    type=click.Path(),
+    default="",
+    show_default=True,
+    help="Path to the TLS client key to use for model serving.",
+)
+@click.option(
+    "--tls-client-passwd",
+    type=click.STRING,
+    default="",
+    help="TLS client certificate password for model serving.",
+)
 @click.pass_context
 def evaluate(
-    ctx,  # pylint: disable=unused-argument
+    ctx,
     model,
     base_model,
     benchmark,
@@ -268,6 +335,10 @@ def evaluate(
     few_shots,
     batch_size,
     sdg_path,
+    tls_insecure,  # pylint: disable=unused-argument
+    tls_client_cert,  # pylint: disable=unused-argument
+    tls_client_key,  # pylint: disable=unused-argument
+    tls_client_passwd,  # pylint: disable=unused-argument
 ):
     # get appropriate evaluator class from Eval lib
     evaluator = get_evaluator(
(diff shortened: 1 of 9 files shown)


src/instructlab/model/evaluate.py
```diff
@@ -1,24 +1,22 @@
...
-import subprocess
-import time
 
 # Third Party
 from instructlab.eval.evaluator import Evaluator
 from instructlab.eval.mmlu import MMLUBranchEvaluator, MMLUEvaluator
 from instructlab.eval.mt_bench import MTBenchBranchEvaluator, MTBenchEvaluator
 import click
 
+logger = logging.getLogger(__name__)
```
nathan-weinberg, line 11: @cdoern @leseb can we inherit the `ilab` logger here somehow?

nathan-weinberg (dismissed):
I would like @alimaredia and @cdoern to signoff but apart from one logging comment LGTM

src/instructlab/model/evaluate.py
tiran: Python's logging framework automatically inherits from parent loggers. By default, a logger propagates its logs and uses settings from a parent logger. In this case it's `instructlab.model.evaluate` -> `instructlab.model` -> `instructlab` -> root logger. Because the log level defaults to `NOTSET` and there is no handler, this logger uses the defaults log level and log handler from `logging.getLogger("instructlab")`.

src/instructlab/configuration.py
```diff
@@ -307,6 +308,18 @@ def get_default_config():
             mt_bench_branch=_mtbenchbranch(taxonomy_path=DEFAULT_TAXONOMY_PATH),
             mmlu_branch=_mmlubranch(sdg_path=DEFAULT_GENERATED_FILES_OUTPUT_DIR),
         ),
+        evaluate_serve=_serve(
+            model_path=DEFAULT_MODEL_PATH,
+            backend="vllm",
+            llama_cpp=_serve_llama_cpp(
+                gpu_layers=-1,
+                max_ctx_size=5120,
```
leseb, line 316: Why 5120? Can we add a comment to give it an explanation?

src/instructlab/model/evaluate.py
```diff
@@ -191,6 +189,43 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
 
+def launch_server(
+    ctx,
+    model,
+    model_name,
+    max_workers,
+    tls_insecure,
+    tls_client_cert,
+    tls_client_key,
+    tls_client_passwd,
```
leseb, line 200: This should probably call `http_client` just like https://github.com/instructlab/instructlab/pull/1373 does it.

src/instructlab/model/evaluate.py
```diff
@@ -191,6 +189,43 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
+) -> tuple:
+    # pylint: disable=C0415
```
leseb, line 202: Please use full-fledge pylint sentence instead of the code.

src/instructlab/model/evaluate.py
```diff
@@ -286,71 +350,59 @@ def evaluate(
...
-                    "judge_model",
-                ]
-            )
-            time.sleep(60)
-            overall_score, qa_pairs, turn_scores = evaluator.judge_answers(
-                "http://localhost:8000/v1"
+            server, api_base = launch_server(
+                ctx,
+                model,
+                "judge_model",
```
leseb, line 376: Should `judge_model` be a constant somewhere in `configuration.py`?

src/instructlab/model/evaluate.py
```diff
@@ -286,71 +350,59 @@ def evaluate(
...
+        print(turn2_score)
 
     elif benchmark == "mt_bench_branch":
-        # TODO: Replace temp Popen hack with serving library calls.  Current library doesn't support server-model-name.
         evaluators = [
             evaluator,
             MTBenchBranchEvaluator(
                 "base_test_model",
-                "base_judge_model",
+                "judge_model",
```
leseb, line 405: ditto

leseb (changes requested):

src/instructlab/model/evaluate.py
nathan-weinberg: Awesome - thanks!

src/instructlab/model/evaluate.py
danmcp: I've moved it to a constant in evaluate.  I don't think it has any value outside of this scope.  But let me know if there is logic to move it higher.

src/instructlab/model/evaluate.py
danmcp: Same

src/instructlab/configuration.py
danmcp: Done, thanks!

src/instructlab/model/evaluate.py
danmcp: Done, thanks for the suggestion!

src/instructlab/model/evaluate.py
danmcp: Done, thanks!

danmcp:
Will need to adjust to changes in #1531

src/instructlab/configuration.py
```diff
@@ -307,6 +308,18 @@ def get_default_config():
...
             mmlu_branch=_mmlubranch(sdg_path=DEFAULT_GENERATED_FILES_OUTPUT_DIR),
         ),
+        evaluate_serve=_serve(
+            model_path=DEFAULT_MODEL_PATH,
+            backend="vllm",
+            llama_cpp=_serve_llama_cpp(
+                gpu_layers=-1,
+                max_ctx_size=5120,  # Larger context needed for mt_bench judgement
+                llm_family="",
+            ),
```
nathan-weinberg, line 318: Why do we need Llama CPP config? I thought we were only using vLLM for evaluation inferencing

src/instructlab/model/evaluate.py
```diff
@@ -1,23 +1,28 @@
...
     "mmlu_branch": MMLUBranchEvaluator,
     "mt_bench": MTBenchEvaluator,
     "mt_bench_branch": MTBenchBranchEvaluator,
 }
 
-# TODO: Remove after proper vLLM integration
-# pylint: disable=consider-using-with
+JUDGE_MODEL_NAME = "judge_model"
+TEST_MODEL_NAME = "test_model"
+BASE_TEST_MODEL_NAME = "base_test_model"
```
nathan-weinberg, line 25: I think we should be passing these down via the config - @cdoern could you chime in here?

src/instructlab/model/evaluate.py
```diff
@@ -191,6 +198,37 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
+    # pylint: disable=import-outside-toplevel
+    # First Party
+    from instructlab.model.backends import backends
+
+    if ctx.obj.config.evaluate_serve.backend == backends.VLLM:
+        ctx.obj.config.evaluate_serve.vllm.vllm_args = (
+            f"--served-model-name {model_name}"
+        )
+    if ctx.obj.config.evaluate_serve.backend == backends.LLAMA_CPP:
+        ctx.obj.config.evaluate.mt_bench.max_workers = min(max_workers, 16)
```
nathan-weinberg, line 216: So if the user passes a `max_workers` larger than 16, 16 will be used - why is that?

src/instructlab/model/evaluate.py
```diff
@@ -253,9 +291,34 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
+    default="",
+    show_default=True,
+    help="Path to the TLS client key to use.",
+)
+@click.option(
+    "--tls-client-passwd",
+    type=click.STRING,
+    default="",
+    help="TLS client certificate password.",
+)
```
nathan-weinberg, line 318: Can we indicate somehow here that these args pertain to the backend serving?

src/instructlab/configuration.py
danmcp: I believe instructlab can allow for both.  Is there another place we are not supporting llama-cpp in the cli?

src/instructlab/model/evaluate.py
danmcp: Noting that these are not configurable values by the user.

alimaredia: Dan and I talked about these values, and realized it's not necessary for the user to need to set this. The values are just needed to make the test run. The user shouldn't care what the name assigned to the model is (not the path or type of model) during the a run of MT-Bench.. 

Instead of introducing unneeded config options and flags we decided to just hard code them.

src/instructlab/model/evaluate.py
danmcp: There is probably still some work to do here, but llama-cpp doesn't handle the parallelism the same and vllm and starts to drop requests under too much load.  Will add a comment.

src/instructlab/configuration.py
alimaredia: This might warrant a broader discussion, but evaluation **could** be done via models served by llama-cpp instead of `vLLM`. To support this case, config for llama-cpp is needed.

src/instructlab/model/evaluate.py
```diff
@@ -191,6 +198,37 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
+            f"--served-model-name {model_name}"
+        )
+    if ctx.obj.config.evaluate_serve.backend == backends.LLAMA_CPP:
+        ctx.obj.config.evaluate.mt_bench.max_workers = min(max_workers, 16)
+
+    ctx.obj.config.evaluate_serve.model_path = model
+
+    backend_instance = backends.select_backend(logger, ctx.obj.config.evaluate_serve)
+    try:
+        backend_instance.run_detached(http_client(ctx.params))
```
alimaredia, line 222: Could we add a comment here that the `tls` related values are the ones being passed in here?

src/instructlab/model/evaluate.py
danmcp: Added comment, thanks!

src/instructlab/model/evaluate.py
danmcp: Done, thanks!

src/instructlab/model/evaluate.py
danmcp: Done, thanks!

src/instructlab/configuration.py
nathan-weinberg: I'm fine for allowing it, I just wasn't sure if it would work with llama-cpp

src/instructlab/model/evaluate.py
nathan-weinberg: Okay, fine by me!

src/instructlab/model/evaluate.py
nathan-weinberg: Thank you!

danmcp:
Looks like pylint is catching a dep issue outside this code: Error: src/instructlab/model/train.py:11:0: E0611: No name 'run_training' in module 'instructlab.training' (no-name-in-module)

src/instructlab/configuration.py
danmcp: It does function.  It's quite slow (at least with my setup) though.

nathan-weinberg:
> Looks like pylint is catching a dep issue outside this code:
> 
> `Error: src/instructlab/model/train.py:11:0: E0611: No name 'run_training' in module 'instructlab.training' (no-name-in-module)`

I hit the same in https://github.com/instructlab/instructlab/pull/1544 and merged anyway since it's unrelated to us

src/instructlab/configuration.py
nathan-weinberg: Can we recommend vLLM somehow? Even if it's just a note in the helptext

danmcp: It defaults to vllm in the config

nathan-weinberg: Works for me

nathan-weinberg (approved):

nathan-weinberg:
@danmcp a rebase should fix the lint issue

cdoern:
so one idea I have for the config is it would look like the following:

```
chat:
  context: default
  greedy_mode: false
  logs_dir: data/chatlogs
  max_tokens: null
  model: models/merlinite-7b-lab-Q4_K_M.gguf
  session: null
  vi_mode: false
  visible_overflow: true
evaluate:
  base_branch: null
  base_model: instructlab/granite-7b-lab
  branch: null
  mmlu:
    batch_size: 5
    few_shots: 2
  mmlu_branch:
    sdg_path: generated
  model: models/merlinite-7b-lab-Q4_K_M.gguf
  mt_bench:
    judge_model: prometheus-eval/prometheus-8x7b-v2.0
    max_workers: 40
    output_dir: eval_data
  mt_bench_branch:
    taxonomy_path: taxonomy
general:
  log_level: INFO
generate:
  chunk_word_count: 1000
  model: models/merlinite-7b-lab-Q4_K_M.gguf
  num_cpus: 10
  num_instructions: 100
  output_dir: generated
  prompt_file: prompt.txt
  seed_file: seed_tasks.json
  taxonomy_base: origin/main
  taxonomy_path: taxonomy
serve:
  backend: ''
  host_port: 127.0.0.1:8000
  llama_cpp:
    gpu_layers: -1
    llm_family: ''
    max_ctx_size: 4096
  vllm:
    vllm_args: []
train:
  torch_args:
    nnodes: 1
    node_rank: 0
    nproc_per_node: 1
    rdzv_id: 123
  train_args:
    chat_tmpl_path: /home/cdoern/.pyenv/versions/3.11.9/lib/python3.11/site-packages/instructlab/training/chat_templates/ibm_generic_tmpl.py
    ckpt_output_dir: checkpoints
    data_output_dir: train-output
    data_path: ./taxonomy_data
    deepspeed_options:
      cpu_offload_optimizer: false
      cpu_offload_optimizer_pin_memory: false
      cpu_offload_optimizer_ratio: 1.0
      save_samples: null
    effective_batch_size: 3840
    is_padding_free: false
    learning_rate: 2e-6
    lora:
      alpha: 32
      dropout: 0.1
      quantize_data_type: null
      rank: 4
      target_modules:
      - q_proj
      - k_proj
      - v_proj
      - o_proj
    max_batch_len: 10000
    max_seq_len: 4096
    mock_data: false
    mock_data_len: 0
    model_path: instructlab/merlinite-7b-lab
    num_epochs: 10
    random_seed: 42
    save_samples: 250000
    warmup_steps: 800
```

and rather than adding an `evaluate_serve` config options is: the settings in `_serve` would be used for all commands implicitly starting servers

the whole process I have now for flattening the default_map can be expanded to take options from _serve, add them to _evaluate but **only in the default flag map**. Users could then override these options using the flags.

Some caveats would be stuff like max-ctx-size. Honestly stuff like this should probably just live in the _evaluate class and map directly to flags. The flatter the config the easier profiles and flag defaulting will be in the future. I think we need to get in the mental habit that whenever a top level config portion is edited like _serve, the next time any cmd that starts a server is run, it pulls from that cfg. 

Creating sep _serve classes per command doesn't fit with the eventual "profile" goal of one class per cmd. I am open to any suggestions here though!

danmcp:
@cdoern I've rewritten to use the serve params and override with eval requirements.  LMKWYT.

alimaredia (approved):

cdoern (approved):
This is the right structure I think in terms of config. I might do a follow up to make the option overriding work as it does in the other cmds. 

The way I set it up for train and @alimaredia did for serve is that in `configuration.py` the `default_map` in the context is set to have the right key->value pairs that will correspond to flag names in the specific cmd. 

This is a clean solution though too and gets it all working, so it gets my +1 !

leseb (commented):
LGTM but I think it's worth a note in the CHANGELOG.md. Thanks!

leseb (approved):

nathan-weinberg (approved):

cdoern:
macos has been flaky recently, rerunning

PR #1541, opened 2024-07-01
Use clang++-17 as CUDA host compiler resolving F40 GCC incompatibility

nerdalert:
Fedora40 ships with GCC v14.1.1 which is currently not supported by CUDA. This enables the user to be able to run ilab with CUDA support on F40. EC2 AMIs don't have F39 AMIs listed so this helps users run on EC2 as well.

I have only validated this on EC2 Fedora 40 cloud image. You can view the results in this external runner workflow [output](https://github.com/nerdalert/ilab-runner/actions/runs/9737488509) getting prepped.

**Checklist:**

### DIFF
--- a/docs/gpu-acceleration.md
+++ b/docs/gpu-acceleration.md
@@ -119,7 +119,7 @@ existing package: `--force-reinstall`. Your final
 command should look like this:
 
 ```shell
-# Veryify CUDA can be found in your PATH variable
+# Verify CUDA can be found in your PATH variable
 export CUDA_HOME=/usr/local/cuda
 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64
 export PATH=$PATH:$CUDA_HOME/bin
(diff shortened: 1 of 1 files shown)


tiran (commented):
Excellent!

tiran (approved):

tiran:
@cdoern IIRC you had some trouble with CUDA on F40. Please take a look.

leseb (approved):

docs/gpu-acceleration.md
```diff
@@ -132,6 +132,16 @@ pip install --force-reinstall llama_cpp_python==0.2.79 -C cmake.args="-DLLAMA_CU
...
 ```
 
+If you are running Fedora 40, you need to replace the `Recompile llama-cpp-python using CUDA` section above with the
+following until [CUDA](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/#host-compiler-support-policy)
+supports GCC v14.1+.
+
+```shell
+# Recompile llama-cpp-python using CUDA
+sudo dnf install clang17
+CUDAHOSTCXX=$(which clang++-17) pip install --force-reinstall llama_cpp_python==0.2.79 -C cmake.args="-DLLAMA_CUDA=on"
```
bjhargrave, line 142:
```suggestion
CUDAHOSTCXX=$(command -v clang++-17) pip install --force-reinstall llama_cpp_python==0.2.79 -C cmake.args="-DLLAMA_CUDA=on"
```

which is not posix.

bjhargrave (changes requested):

docs/gpu-acceleration.md
tiran: Do we need a full path here?

russellb: while technically true, i don't think this warrants blocking the PR -- `which` is available everywhere that these instructions are relevant to

russellb (approved):
thanks!

tiran (approved):

bjhargrave (approved):

PR #1543, opened 2024-07-01
Configure coverage to merge and exclude paths

tiran:
`coverage` is now configured to correctly merge local paths and tox paths. External InstructLab packages such as `instructlab.sdg` are excluded from coverage.

`coverage[toml]` is needed to enable `pyproject.toml` parsing in Python 3.10.

**Checklist:**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -143,3 +143,23 @@ exclude = [
 ]
 # honor excludes by not following there through imports
 follow_imports = "silent"
+
+[tool.coverage.run]
+branch = false
+relative_files = true
+source = ["instructlab", "tests/"]
+omit = [
+  # omit instructlab code in different packages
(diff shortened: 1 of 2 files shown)


leseb (changes requested):
@tiran please see toml linter error. Thanks

tiran:
I do not agree with tomllint. Is there any way to tell it that I want each string on a separate line (like black and ruff would format Python code).

leseb:
> I do not agree with tomllint. Is there any way to tell it that I want each string on a separate line (like black and ruff would format Python code).

Taplo automatically expands arrays to multiple lines when they exceed column_width characters (80 by default). So we would need to change that I guess?

leseb (approved):

nathan-weinberg (approved):

PR #1553, opened 2024-07-02
Implements --gpus option for eval

### LINKED ISSUE #1549: ilab model evaluate should take #gpus to send to vllm/llama-cpp
Currently there isn't a way to indicate how many GPUs should be used for serving other than vllm args in config.  This is common enough that it should be an arg

add --gpus option to evaluate consistent with training

danmcp:
Implements --gpus option for evaluate.

Resolves #1549

**Checklist:**

- [x ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
- [ x] Unit tests have been added, if necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 * Adding `ilab model evaluate` which uses the new backend serving functionality.  Evaluate offers
    two standard benchmarks (mt-bench and mmlu) as well as two variations (mt-bench-branch and
    mmlu-branch) which are integrated with the ilab workflow to evaluate new skills and knowledge.
+   Includes --gpus option for specifying number of gpus to utilize when serving models for
+   evaluation (currently applicable for vLLM only).
 
 ### Breaking Changes
 
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -208,6 +208,7 @@ class _evaluate(BaseModel):
     base_model: str
     branch: Optional[str] = None
     base_branch: Optional[str] = None
+    gpus: Optional[int] = 1
     mmlu: _mmlu
     mmlu_branch: _mmlubranch
     mt_bench: _mtbench
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -225,6 +225,7 @@ def launch_server(
     model,
     model_name,
     max_workers,
+    gpus,
 ) -> tuple:
     # pylint: disable=import-outside-toplevel
     # First Party
@@ -234,11 +235,23 @@ def launch_server(
         ctx.obj.config.serve.backend = backends.VLLM
     if ctx.obj.config.serve.backend == backends.VLLM:
         ctx.obj.config.serve.vllm.vllm_args.extend(["--served-model-name", model_name])
+        if gpus:
+            # Don't override from vllm_args
+            if "--tensor-parallel-size" not in ctx.obj.config.serve.vllm.vllm_args:
+                ctx.obj.config.serve.vllm.vllm_args.extend(
+                    ["--tensor-parallel-size", str(gpus)]
+                )
+            else:
+                click.echo(
+                    "Ignoring --gpus with --tensor-parallel-size configured in serve vllm_args"
+                )
     elif ctx.obj.config.serve.backend == backends.LLAMA_CPP:
         # mt_bench requires a larger context size
         ctx.obj.config.serve.llama_cpp.max_ctx_size = 5120
         # llama-cpp fails fast on too many incoming requests and returns errors to client
         ctx.obj.config.evaluate.mt_bench.max_workers = min(max_workers, 16)
+        if gpus:
+            logging.debug("Ignoring --gpus option for llama-cpp serving")
 
     ctx.obj.config.serve.model_path = model
 
@@ -315,6 +328,11 @@ def launch_server(
     help="Path where all the MMLU Branch tasks are stored. Needed for running mmlu_branch.",
 )
 @click.option(
+    "--gpus",
+    type=click.INT,
+    help="Number of GPUs to utilize for evaluation (not applicable to llama-cpp)",
+)
+@click.option(
     "--tls-insecure",
     is_flag=True,
     help="Disable TLS verification for model serving.",
@@ -355,6 +373,7 @@ def evaluate(
     few_shots,
     batch_size,
     sdg_path,
+    gpus,
     tls_insecure,  # pylint: disable=unused-argument
     tls_client_cert,  # pylint: disable=unused-argument
     tls_client_key,  # pylint: disable=unused-argument
@@ -385,6 +404,7 @@ def evaluate(
                 model,
                 TEST_MODEL_NAME,
                 max_workers,
+                gpus,
             )
             evaluator.gen_answers(api_base)
         finally:
@@ -398,6 +418,7 @@ def evaluate(
                 judge_model,
                 JUDGE_MODEL_NAME,
                 max_workers,
+                gpus,
             )
             # TODO: Change back to standard tuple handling after version bump in eval library
             judgment = evaluator.judge_answers(api_base)
@@ -459,6 +480,7 @@ def evaluate(
                     m_path,
                     m_name,
                     max_workers,
+                    gpus,
                 )
                 evaluator.gen_answers(api_base)
             finally:
@@ -472,6 +494,7 @@ def evaluate(
                 judge_model,
                 JUDGE_MODEL_NAME,
                 max_workers,
+                gpus,
             )
             for i, evaluator in enumerate(evaluators):
                 branch = branches[i]
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -108,7 +108,7 @@ env
 eval
 frontend
 gpu
-gpu
+gpus
 habana
 habana
 hipBLAS
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -77,6 +77,7 @@ serve:
     vllm_args: []
 evaluate:
   base_model: instructlab/granite-7b-lab
+  gpus: 1
   mmlu:
     few_shots: 2
     batch_size: 5
@@ -130,6 +131,7 @@ serve:
     vllm_args: []
 evaluate:
   base_model: instructlab/granite-7b-lab
+  gpus: 1
   mmlu:
     few_shots: 2
     batch_size: 5
@@ -168,6 +170,7 @@ serve:
     vllm_args: []
 evaluate:
   base_model: instructlab/granite-7b-lab
+  gpus: 1
   mmlu:
     few_shots: 2
     batch_size: 5
@@ -289,6 +292,7 @@ serve:
        - --enable-lora
 evaluate:
   base_model: instructlab/granite-7b-lab
+  gpus: 1
   mmlu:
     few_shots: 2
     batch_size: 5


src/instructlab/model/evaluate.py
```diff
@@ -295,6 +300,11 @@ def launch_server(
     type=click.Path(),
     help="Path where all the MMLU Branch tasks are stored. Needed for running mmlu_branch.",
 )
+@click.option(
+    "--gpus",
+    type=click.INT,
+    help="Number of GPUs to consume for evaluation",
```
russellb, line 306: it looks like this should say it only works with the VLLM serving backend?

russellb (changes requested):

src/instructlab/model/evaluate.py
danmcp: The intention of the param would be to support any backends that are multi gpu capable.  For llama-cpp it's not clear if args are required to consume multiple gpus.  --tensor-split and --main-gpu seem to be tuning params.

russellb: either way, changing this option has no effect on the llama-cpp backend, right?

danmcp: I'm doing more testing now with llama-cpp to validate

nathan-weinberg (dismissed):
Pending the discussion this LGTM

src/instructlab/model/evaluate.py
danmcp: Updated to indicate it's not applicable to llama-cpp.

nathan-weinberg (approved):

src/instructlab/configuration.py
```diff
@@ -206,6 +206,7 @@ class _evaluate(BaseModel):
     base_model: str
     branch: Optional[str] = None
     base_branch: Optional[str] = None
+    gpus: int = 1
```
russellb, line 209: Can we make this Optional?

I think it be good if we made that the standard practice, because otherwise, every time a new required parameter gets added, previously generated `config.yaml` is broken and causes errors at `ilab` startup.

In a perfect world, we'd go back and do that for everything added since the last CLI release, though that's beyond the scope of this PR

russellb (changes requested):
one other comment about the config file addition

src/instructlab/configuration.py
russellb: I should clarify a bit that I'm guessing the only change needed to get he behavior I'm suggesting is to mark it optional ... i haven't tried it yet

danmcp: Good catch.  It should be optional in the config.

nathan-weinberg (dismissed):

src/instructlab/model/evaluate.py
```diff
@@ -349,6 +359,7 @@ def evaluate(
     few_shots,
     batch_size,
     sdg_path,
+    gpus,
```
alimaredia, line 362: I don't see `gpus` being set via the config anywhere in `evaluate.py`. Is that going to be follow up work? Are we expecting the `--gpus` flag to just be set to the config somehow if it's not overridden?

src/instructlab/model/evaluate.py
```diff
@@ -309,6 +314,11 @@ def launch_server(
     type=click.Path(),
     help="Path where all the MMLU Branch tasks are stored. Needed for running mmlu_branch.",
 )
+@click.option(
+    "--gpus",
+    type=click.INT,
+    help="Number of GPUs to utilize for evaluation (not applicable to llama-cpp)",
+)
```
leseb, line 321: My personal preference would be to use `--backend-args` instead of introducing multiple flags. This approach helps us avoid cluttering the CLI with numerous additional flags. Also given that this is a backend-specific flag I don't think it should be exposed directly. `--backend-args` is not present yet (I'm working on changing that). If this gets merged as is, it might eventually get replaced. Just an FYI :)

tiran: It's probably too late, but did you consider to use subcommands like `ilab model serve vllm` and `ilab model serve llama-cpp`? Then you can have different arguments for each backend.

You could even use `@click.command(context_settings=dict(allow_extra_args=True))` and `ctx.args` to pass additional arguments to vLLM like this:

```console
ilab model serve vllm -- --dtype auto --enable-lora
```

leseb: It was considered as part of https://github.com/instructlab/dev-docs/pull/81. See this [comment](https://github.com/instructlab/dev-docs/pull/81#issuecomment-2182618445) on why it was discarded.

leseb: > It's probably too late, but did you consider to use subcommands like `ilab model serve vllm` and `ilab model serve llama-cpp`? Then you can have different arguments for each backend.
> 
> You could even use `@click.command(context_settings=dict(allow_extra_args=True))` and `ctx.args` to pass additional arguments to vLLM like this:
> 
> ```
> ilab model serve vllm -- --dtype auto --enable-lora
> ```

What prevents us from doing this: `ilab model serve --backend vllm -- --dtype auto --enable-lora`?

tiran: Nothing? :)

src/instructlab/model/evaluate.py
danmcp: The logic currently takes the values in order or precedence from --gpus, config, accept default from serving (which would be 1).  Not quite following where else it should be set.

src/instructlab/model/evaluate.py
danmcp: > My personal preference would be to use `--backend-args` instead of introducing multiple flags. This approach helps us avoid cluttering the CLI with numerous additional flags. Also given that this is a backend-specific flag I don't think it should be exposed directly. `--backend-args` is not present yet (I'm working on changing that). If this gets merged as is, it might eventually get replaced. Just an FYI :)

An argument against folding everything into a --backend-args argument is discoverability.  Since specifying --tensor-parallel-size is necessary to utilize the gpus, it will make discovering the correct way to run with the setting quite difficult without following a document.  And it seems like most users of evaluation will want to use multiple gpus.

tiran: You could include examples in the doc string of the `serve` function. Click will render the entire doc string in `ilab model serve --help`.

leseb (changes requested):
Please update the CHANGELOG.md. Thanks!

danmcp:
> Please update the CHANGELOG.md. Thanks!

Done!

src/instructlab/model/evaluate.py
alimaredia: Apologies, my previous comment wasn't clear.

Where in `evaluate.py` is the value being set from config? I only see the value from the flag being propagated down into `launch_server()`.

Also what would happen if `--tensor-parallel-size` was already in the list of `vllm-args` in the `serve` section of `config.yaml`? Right now it seems `--tensor-parallel-size` would be in `vllm-args` twice.

tiran: click looks up defaults from the context's `default_map` attribute. The default map is configured in `instructlab.configuration` and `instructlab.model.model`. The `evalulate` subcommand uses `ctx.default_map["evalulate"]` as base map.

My PR #1591 tries to make it easier to understand and less messy.

leseb (approved):
Just a thought for later: I hate to see users shooting themselves in the foot, even if the flags empower them to do so. I'm wondering if we should log a warning when someone provides a ridiculously high value for `--gpus`. This consideration applies to other flags as well.

src/instructlab/model/evaluate.py
danmcp: > Also what would happen if `--tensor-parallel-size` was already in the list of `vllm-args` in the `serve` section of `config.yaml`?

Updated to not override.  I am hoping we add --gpus to serve as well.  If we do, this check can be replaced by just passing gpus to serve.

alimaredia (dismissed):

src/instructlab/model/evaluate.py
```diff
@@ -234,6 +235,16 @@ def launch_server(
...
     if ctx.obj.config.serve.backend == backends.VLLM:
         ctx.obj.config.serve.vllm.vllm_args.extend(["--served-model-name", model_name])
+        if gpus:
+            # Don't override from vllm_args
+            if "--tensor-parallel-size" not in ctx.obj.config.serve.vllm.vllm_args:
+                ctx.obj.config.serve.vllm.vllm_args.extend(
+                    ["--tensor-parallel-size", str(gpus)]
+                )
+            else:
+                print(
```
leseb, line 245: use `click.echo`?

src/instructlab/model/evaluate.py
```diff
@@ -234,6 +235,16 @@ def launch_server(
         ctx.obj.config.serve.backend = backends.VLLM
     if ctx.obj.config.serve.backend == backends.VLLM:
         ctx.obj.config.serve.vllm.vllm_args.extend(["--served-model-name", model_name])
+        if gpus:
```
leseb, line 238: Should we have a warning if `--gpus` is passed and the backend is llama-cpp? So at the least the user knows the option won't do anything.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 * Adding `ilab model evaluate` which uses the new backend serving functionality.  Evaluate offers
    two standard benchmarks (mt-bench and mmlu) as well as two variations (mt-bench-branch and
    mmlu-branch) which are integrated with the ilab workflow to evaluate new skills and knowledge.
+   Includes --gpus option for specifying number of gpus to utilize when serving models for
```
leseb, line 14: This should indicate the flag is only supported for vLLM.

src/instructlab/model/evaluate.py
```diff
@@ -314,6 +325,11 @@ def launch_server(
     type=click.Path(),
     help="Path where all the MMLU Branch tasks are stored. Needed for running mmlu_branch.",
 )
+@click.option(
+    "--gpus",
+    type=click.INT,
+    help="Number of GPUs to utilize for evaluation (not applicable to llama-cpp)",
```
leseb, line 331: I still don't understand why we are starting to have backend-specific flags. I mean `--gpus` seems quite generic.

src/instructlab/model/evaluate.py
```diff
@@ -314,6 +325,11 @@ def launch_server(
     type=click.Path(),
     help="Path where all the MMLU Branch tasks are stored. Needed for running mmlu_branch.",
 )
+@click.option(
+    "--gpus",
```
leseb, line 329: I think `--num-gpu` might be clearer but I won't block for this.

leseb (changes requested):

src/instructlab/model/evaluate.py
danmcp: --gpus was used by train.  I was thinking consistency was more important.

src/instructlab/model/evaluate.py
danmcp: The goal isn't to be backend specific.  The flag should apply to any backend that needs the number of gpus specified in order to consume them.

src/instructlab/model/evaluate.py
danmcp: Added a logging debug consistent with: https://github.com/instructlab/instructlab/pull/1564/files

CHANGELOG.md
danmcp: Done, thanks!

src/instructlab/model/evaluate.py
leseb: I guess I misunderstood the intention, thanks for clarifying!

leseb (approved):

alimaredia (approved):

PR #1554, opened 2024-07-02
Add e2e CI variant that runs on AWS

russellb:
This commit introduces a new approach for running the e2e CI job.
Instead of using a runner provided by github, use a similarly powered
runner on AWS that is much cheaper.

The real benefit of this approach is the ability to run on any other
instance type, giving us access to GPU instances that are more
powerful than what was built-in to GitHub. More powerful e2e variants
will come in future commits.

This job can only be manually triggered for now. At some point we may
create a configuration that can replace the one that runs
automatically on every commit, but I'd like to exercise this for a
while to make sure it's stable. It's also not quite as simple as just
running this automatically directly on PRs, because have to be careful
to protect against what's possible with a malicious PR. In any case,
that is future work.

Similar to how the original `e2e` job worked at first, this one can be
manually triggered to be run against any PR. Comments will be posted
to a PR so you can follow its progress, as well as be notified of its
results.

Co-authored-by: Brent Salisbury <bsalisbu@redhat.com>

### DIFF
--- a/.github/workflows/e2e-nvidia-t4-x1.yml
+++ b/.github/workflows/e2e-nvidia-t4-x1.yml
@@ -1,0 +1,168 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: E2E (NVIDIA Tesla T4 x1)
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_or_branch:
+        description: 'pull request number or branch name'
+        required: true
+        default: 'main'
+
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):

cdoern (approved):

russellb:
follow-up next step -- https://github.com/instructlab/instructlab/pull/1556

PR #1556, opened 2024-07-02
Add e2e variant that can run non-quantized training

russellb:
cecd736 e2e: Add option to turn off `--4-bit-quant`
94c38d7 e2e: Add a new variant with non-quantized training

commit cecd7366fa5c3da99148a66822566be3c9578598
Date:   Mon Jul 1 22:50:07 2024 -0400

    e2e: Add option to turn off `--4-bit-quant`
    
    This test script used `--4-bit-quant` always because the test runner
    required it. Now that we can test on custom AWS instances, make this
    optional.
    
commit 94c38d74e75b1bca0d6d2451849e65ac9c142ce7
Date:   Mon Jul 1 22:51:16 2024 -0400

    e2e: Add a new variant with non-quantized training
    
    This e2e variant runs on an instance with a single A10G GPU with 24 GB
    of GPU memory. This instance type is able to exercise `ilab model
    train` without `--4-bit-quant`. This also means we run the conversion
    to gguf, because that is skipped with quantization on.
    
### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x1.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x1.yml
@@ -1,0 +1,168 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: E2E (NVIDIA A10G x1)
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_or_branch:
+        description: 'pull request number or branch name'
+        required: true
+        default: 'main'
+
+jobs:
+  start-runner:
+    name: Start external EC2 runner
+    runs-on: ubuntu-latest
+    outputs:
+      label: ${{ steps.start-ec2-runner.outputs.label }}
+      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      - name: Start EC2 runner
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@fcfb31a5760dad1314a64a0e172b78ec6fc8a17e # v2.3.6
+        with:
+          mode: start
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id: ami-00c51d9c1374eda97
+          ec2-instance-type: g5.2xlarge
+          subnet-id: subnet-02d230cffd9385bd4
+          security-group-id: sg-06300447c4a5fbef3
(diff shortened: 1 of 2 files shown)


scripts/basic-workflow-tests.sh
```diff
@@ -146,7 +147,10 @@ test_train() {
     task Train the model
 
     # TODO Only cuda for now
-    TRAIN_ARGS=("--legacy" "--device=cuda" "--4-bit-quant")
+    TRAIN_ARGS=("--legacy=True" "--device=cuda")
```
russellb, line 150:
```suggestion
    TRAIN_ARGS=("--legacy" "--device=cuda")
```

leseb (commented):
Why do we need 2eaf5f4585beae3a7f08813f6d2a2669d2996a6a? Shouldn't cecd7366fa5c3da99148a66822566be3c9578598 be fixed instead?

leseb:
I think that [list](https://github.com/instructlab/dev-docs/blob/main/docs/github-actions-providers.md) needs to be updated too right?

russellb:
> I think that [list](https://github.com/instructlab/dev-docs/blob/main/docs/github-actions-providers.md) needs to be updated too right?

Yes, I’ll do that.

russellb:
> > Why do we need [2eaf5f4](https://github.com/instructlab/instructlab/commit/2eaf5f4585beae3a7f08813f6d2a2669d2996a6a)? Shouldn't [cecd736](https://github.com/instructlab/instructlab/commit/cecd7366fa5c3da99148a66822566be3c9578598) be fixed instead?
> 
> Because I saw the error on my phone and wanted to fix it quickly to see CI run again.

fixed


> > I think that [list](https://github.com/instructlab/dev-docs/blob/main/docs/github-actions-providers.md) needs to be updated too right?
> 
> Yes, I’ll do that.

https://github.com/instructlab/dev-docs/pull/108

leseb (approved):

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -0,0 +1,168 @@
...
+          sed 's/\[.*\]//' requirements.txt > constraints.txt
+          python3.11 -m pip cache remove llama_cpp_python
+          CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install --force-reinstall --no-binary llama_cpp_python -c constraints.txt llama_cpp_python
+          python3.11 -m pip install bitsandbytes
+          python3.11 -m pip install .
+
+      - name: Run e2e test
+        run: |
+          . venv/bin/activate
+          ./scripts/basic-workflow-tests.sh -cmf
```
leseb, line 132: Should we stop using `-m` here since the runner is more powerful?

leseb (changes requested):

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: I saw your change in #1560 

I actually think we should keep it. It will make the job run faster and I still think we get the "smoke test" like coverage we're going for here. More instructions and more training epochs will produce a higher quality result, but that's not really the point of these jobs.

leseb: Fair enough!

cdoern (approved):

PR #1557, opened 2024-07-02
Full e2e train using library and new CI runners

cdoern:
Built on top of #1556 

adds non-legacy training to the A10 runners. These ones should be able to run natively without disabling flash-attn.

adds flash-attn, packaging, wheel pip install's to the e2e yaml. 

adds some generated data to train off of, and the model in use is granite-7b-lab.

This PR  also introduces some new handling for installing all necessary packages and reinstalling some existing ones

when doing dnf upgrade we get a nccl version mismatch because a bunch of the cuda deps are upgraded
also, the path was set to the wrong values. The outdated values were not pointing to the right nvcc or nccl causing issues when running nvidia-smi and any cuda processes.

### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x1.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x1.yml
@@ -58,6 +58,7 @@ jobs:
       - name: Determine if pr_or_branch is a PR number
         id: check_pr
         run: |
+          nvidia-smi
           if [[ "${{ github.event.inputs.pr_or_branch }}" =~ ^[0-9]+$ ]]; then
             echo "is_pr=true" >> "$GITHUB_OUTPUT"
           else
@@ -116,18 +117,43 @@ jobs:
 
       - name: Install ilab
         run: |
-          export PATH="/home/ec2-user/.local/bin:/usr/local/cuda/bin:$PATH"
-          python3.11 -m venv --upgrade-deps venv
+          export CUDA_HOME="/usr/local/cuda"
+          export LD_LIBRRY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64"
+          export PATH="$PATH:$CUDA_HOME/bin"
+          python3.11 -m venv venv
           . venv/bin/activate
+          nvidia-smi
           sed 's/\[.*\]//' requirements.txt > constraints.txt
           python3.11 -m pip cache remove llama_cpp_python
           CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install --force-reinstall --no-binary llama_cpp_python -c constraints.txt llama_cpp_python
           python3.11 -m pip install bitsandbytes
+
+          # TODO This should be added to instructlab-training
+          python3.11 -m pip install packaging wheel
+
+          python3.11 -m pip install instructlab-training[cuda]
+
           python3.11 -m pip install .
 
       - name: Run e2e test
         run: |
+          nvidia-smi
+          # This env variable is used on GPUs with less vRAM. It only allows cuda to alloc small chunks of vRAM at a time, usually helping to avoid OOM errors.
+          # This is not a good solution for production code, as setting ENV variables for users isn't best practice. However, it is a helpful manual workaround.
+          export PYTORCH_CUDA_ALLOC_CONF=garbage_collection_threshold:0.6,max_split_size_mb:128
           . venv/bin/activate
+          # TODO: for some reason we need to reinstall DS in order to get fused adam support
+          # This means we need to manually rm and re-install a bunch of packages. Investigate why this is.
+          python3.11 -m pip uninstall -y deepspeed
+
+          python3.11 -m pip cache purge
+
+          DS_BUILD_CPU_ADAM=1 BUILD_UTILS=1 python3.11 -m pip install deepspeed
+
+          nvidia-smi
+
+          python3.11 -m pip show nvidia-nccl-cu12
+          
           ./scripts/basic-workflow-tests.sh -cemf
 
       - name: Add comment to PR if the workflow failed
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -110,8 +110,14 @@ test_smoke() {
 
 test_init() {
     task Initializing ilab
-    ilab config init --non-interactive
 
+    if [ "$FULLTRAIN" -eq 1 ]; then
+        # TODO Only cuda for now
+        step Setting train-profile for GPU accelerated training
+        ilab config init --non-interactive --train-profile="${SCRIPTDIR}/test-data/train-profile-a10.yaml"
+    else
+        ilab config init --non-interactive
+    fi
     step Checking config.yaml
     if [ "${MIXTRAL}" -eq 1 ]; then
         sed -i -e 's|merlinite.*|mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf|' "${CONFIG_HOME}/instructlab/config.yaml"
@@ -231,6 +237,18 @@ test_train() {
     fi
 
     ilab model train "${TRAIN_ARGS[@]}"
+
+    if [ "$FULLTRAIN" -eq 1 ]; then
+
+        # TODO Only cuda for now
+        # the train profile specified in test_init overrides the majority of TRAIN_ARGS, including things like num_epochs. While it looks like much of those settings are being lost, they just have different values here.
+        TRAIN_ARGS=("--device=cuda" "--model-path=instructlab/granite-7b-lab" "--data-path=${SCRIPTDIR}/test-data/train_all_pruned_SDG.jsonl" "--quantize-data-type=nf4" "--4-bit-quant")
+        if [ "$GRANITE" -eq 1 ]; then
+            TRAIN_ARGS+=("--gguf-model-path" "${DATA_HOME}/instructlab/models/granite-7b-lab-Q4_K_M.gguf")
+        fi
+
+        ilab model train "${TRAIN_ARGS[@]}"
+    fi
 }
 
 test_convert() {
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 [build-system]
-requires = ["setuptools>=70.1.0", "setuptools_scm>=8"]
+requires = ["setuptools>=70.1.0", "setuptools_scm>=8", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]
--- a/scripts/test-data/train-profile-a10.yaml
+++ b/scripts/test-data/train-profile-a10.yaml
@@ -1,0 +1,37 @@
+torch_args:
+  nnodes: 1
+  node_rank: 0
+  nproc_per_node: 1
+  rdzv_endpoint: 127.0.0.1:12222
+  rdzv_id: 123
+train_args:
+  ckpt_output_dir: checkpoints
+  data_output_dir: train-output
+  data_path: ./taxonomy_data
+  deepspeed_options:
+    cpu_offload_optimizer: true
+    cpu_offload_optimizer_pin_memory: false
+    cpu_offload_optimizer_ratio: 1
+    save_samples: null
+  effective_batch_size: 100
+  is_padding_free: false
+  learning_rate: 2e-6
+  lora:
+    alpha: 32
+    dropout: 0.1
+    quantize_data_type: nf4
+    rank: 2
+    target_modules:
+    - q_proj
+    - k_proj
+    - v_proj
+    - o_proj
+  max_batch_len: 1000
+  max_seq_len: 96
+  mock_data: false
+  mock_data_len: 0
+  model_path: instructlab/granite-7b-lab
+  num_epochs: 1
+  random_seed: 42
+  save_samples: 100
+  warmup_steps: 10


.github/workflows/e2e.yml
```diff
@@ -145,7 +145,9 @@ jobs:
           CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3 -m pip install --no-binary llama_cpp_python -c constraints.txt llama_cpp_python
           # needed for --4-bit-quant option to ilab model train
           python3 -m pip install bitsandbytes
+          python3 -m pip install packaging wheel
```
leseb, line 148: Why not add these two along with `bitsandbytes` above?

scripts/basic-workflow-tests.sh
```diff
@@ -156,6 +156,22 @@ test_train() {
...
+        ilab config init --non-interactive --train-profile="${SCRIPTDIR}/test-data/train-profile-a10.yaml"
+        TRAIN_ARGS=("--device=cuda" "--model-path=instructlab/granite-7b-lab" "--data-path=${SCRIPTDIR}/test-data/train_all_pruned_SDG.jsonl" "--quantize-data-type=nf4" "--4-bit-quant")
+        if [ "$GRANITE" -eq 1 ]; then
+            TRAIN_ARGS+=("--gguf-model-path models/granite-7b-lab-Q4_K_M.gguf")
+        fi
+
+        ilab model train "${TRAIN_ARGS[@]}"
+
+        ilab config init --non-interactive
+
```
leseb, line 172: remove empty line

src/instructlab/model/train.py
```diff
@@ -413,6 +413,9 @@ def get_files(pattern):
                 fg="red",
             )
             raise click.exceptions.Exit(1)
+        # temp fix to get CI passing, there are multiple options in the train class w/ the same name
```
leseb, line 416: Can you elaborate on:

* why this is needed
* what the impact could be when not running in the CI? (if any)
* when do we expect this to be removed

I'd like to point out that embedding code specifically for handling CI cases within the production code is unusual.

Thanks!

leseb (changes requested):

.github/workflows/e2e.yml
tiran: Are you building bitsandbytes from sources? In that case you have to install packaging and wheel first, then run a second `pip install` for bitsandbytes.

src/instructlab/model/train.py
cdoern: This is a placeholder. I should really convert this PR back to draft. @RobotSail 's PR #1471 needs to be merged before this

cdoern: save_samples is broken at the moment since deepspeed and regular training have a flag w/ the same name

RobotSail: @cdoern I can make a separate PR for that

cdoern: @RobotSail dw, I'd hate to mess up the flow in your current PR

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -116,18 +117,41 @@ jobs:
...
+          python3.11 -m venv venv
           . venv/bin/activate
+          nvidia-smi
           sed 's/\[.*\]//' requirements.txt > constraints.txt
           python3.11 -m pip cache remove llama_cpp_python
           CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install --force-reinstall --no-binary llama_cpp_python -c constraints.txt llama_cpp_python
           python3.11 -m pip install bitsandbytes
+
+          # TODO This should be added to instructlab-training
+          python3.11 -m pip install packaging wheel
```
russellb, line 132: It looks like this was (partially?) resolved here -- https://github.com/instructlab/training/pull/123. -- it added wheel, but not packaging ...

What's the status on this? It would be great to either have this fixed, or an open issue tracking it

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -116,18 +117,41 @@ jobs:
...
+
+          python3.11 -m pip install instructlab-training[cuda]
```
russellb, line 134: Can you either fix the install docs for ilab with nvidia support, or file an issue about doing so to reflect this?

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -116,18 +117,41 @@ jobs:
...
           python3.11 -m pip install .
 
       - name: Run e2e test
         run: |
+          nvidia-smi
+          export PYTORCH_CUDA_ALLOC_CONF=garbage_collection_threshold:0.6,max_split_size_mb:128
           . venv/bin/activate
+          # TODO: for some reason we need to reinstall DS in order to get fused adam support
+          # This means we need to manually rm and re-install a bunch of packages. Investigate why this is.
+          python3.11 -m pip uninstall -y deepspeed
```
russellb, line 145: Is there an open issue tracking this?

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -116,18 +117,41 @@ jobs:
...
+          python3.11 -m pip install packaging wheel
+
+          python3.11 -m pip install instructlab-training[cuda]
+
           python3.11 -m pip install .
 
       - name: Run e2e test
         run: |
+          nvidia-smi
+          export PYTORCH_CUDA_ALLOC_CONF=garbage_collection_threshold:0.6,max_split_size_mb:128
```
russellb, line 141: Do you know why this was needed? Can you add a comment explaining what's going on?  Also, is this something that needs to go in our docs?

.github/workflows/e2e.yml
```diff
@@ -141,20 +141,36 @@ jobs:
...
+          python3.11 -m pip install .
 
       - name: Run e2e test
         run: |
+          export PYTORCH_CUDA_ALLOC_CONF=garbage_collection_threshold:0.6,max_split_size_mb:128
           . venv/bin/activate
-          ./scripts/basic-workflow-tests.sh -cm
+          python3.11 -m pip uninstall -y deepspeed
+
+          python3.11 -m pip uninstall -y nvidia-nccl-cu12
```
russellb, line 165: what's going on here?  the other workflow didn't do this

.github/workflows/e2e.yml
```diff
@@ -141,20 +141,36 @@ jobs:
...
+
+          python3.11 -m pip cache purge
+
+          python3.11 -m pip install torch
```
russellb, line 169: why is this needed? This seems like a bug somewhere ...

scripts/basic-workflow-tests.sh
```diff
@@ -186,6 +186,22 @@ test_train() {
     fi
 
     ilab model train "${TRAIN_ARGS[@]}"
+
+    if [ "$FULLTRAIN" -eq 1 ]; then
+
+        # TODO Only cuda for now
+        ilab config init --non-interactive --train-profile="${SCRIPTDIR}/test-data/train-profile-a10.yaml"
```
russellb, line 193: I would prefer this go up in `test_init` where we set up the config

scripts/basic-workflow-tests.sh
```diff
@@ -186,6 +186,22 @@ test_train() {
...
+        TRAIN_ARGS=("--device=cuda" "--model-path=instructlab/granite-7b-lab" "--data-path=${SCRIPTDIR}/test-data/train_all_pruned_SDG.jsonl" "--quantize-data-type=nf4" "--4-bit-quant")
+        if [ "$GRANITE" -eq 1 ]; then
+            TRAIN_ARGS+=("--gguf-model-path models/granite-7b-lab-Q4_K_M.gguf")
+        fi
```
russellb, line 197: Elsewhere in the script we conditionally put `--num-epochs 1` in the args? This code overwrites it. Is that option valid here? If so, can you replicate that logic so it's put back in place? The purpose is to speed up CI.

scripts/basic-workflow-tests.sh
```diff
@@ -186,6 +186,22 @@ test_train() {
...
+
+        ilab model train "${TRAIN_ARGS[@]}"
+
+        ilab config init --non-interactive
```
russellb, line 201: what is this for?

russellb (changes requested):
My biggest comments are around the changes to the workflow files. There's a whole bunch of magic going on and it's not clear what is a bug that needs to be tracked, what justifies doc updates, what may not actually be needed anymore ...

.github/workflows/e2e-nvidia-a10g-x1.yml
cdoern: @Maxusmusti I tried removing the packaging dep, but it seems the tests still fail. I can remove wheel here but I think packaging is still necessary

.github/workflows/e2e-nvidia-a10g-x1.yml
cdoern: sure https://github.com/instructlab/instructlab/issues/1637

.github/workflows/e2e-nvidia-a10g-x1.yml
cdoern: this is a training repo issue probably. @Maxusmusti WDYT?

scripts/basic-workflow-tests.sh
cdoern: resetting the cfg back to the defaults after initializing a train profile

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: then it needs to be added to the training repo, right?

scripts/basic-workflow-tests.sh
russellb: but you can check what configuration the script is running in at that time.  I'd like the output of `test_config` to be "set up the config for this e2e run" at the beginning.  That is more aligned to a user workflow. They configure once for their environment and then run through the workflow.

scripts/basic-workflow-tests.sh
russellb: ... but why :) we only need to train once

.github/workflows/e2e.yml
cdoern: sorry, forgot to remove changes from the base e2e.yml

.github/workflows/e2e.yml
cdoern: sorry, forgot to remove changes from the base e2e.yml

.github/workflows/e2e-nvidia-a10g-x1.yml
cdoern: PYTORCH_CUDA_ALLOC_CONF makes the chunks cuda can alloc at once smaller, helps avoid OOM. I don't think this needs to be documented on our end

.github/workflows/e2e-nvidia-a10g-x1.yml
cdoern: I think so

.github/workflows/e2e.yml
russellb: OK, sounds good.  I'm hoping to deprecate that one ASAP anyway

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: if we needed it, then it must be useful to others

scripts/basic-workflow-tests.sh
cdoern: so like:

```
test_init() {
    task Initializing ilab

    if [ "$FULLTRAIN" -eq 1 ]; then
        # TODO Only cuda for now
        ilab config init --non-interactive --train-profile="${SCRIPTDIR}/test-data/train-profile-a10.yaml"
    else
        [ -f config.yaml ] || ilab config init --non-interactive
    fi
```

scripts/basic-workflow-tests.sh
cdoern: the `train_profile` that @Maxusmusti and I put together takes care of the epochs I believe.

cdoern: yeah here it is: 
```
torch_args:
  nnodes: 1
  node_rank: 0
  nproc_per_node: 1
  rdzv_id: 123
train_args:
  ckpt_output_dir: checkpoints
  data_output_dir: train-output
  data_path: ./taxonomy_data
  deepspeed_options:
    cpu_offload_optimizer: true
    cpu_offload_optimizer_pin_memory: false
    cpu_offload_optimizer_ratio: 1
    save_samples: null
  effective_batch_size: 100
  is_padding_free: false
  learning_rate: 2e-6
  lora:
    alpha: 32
    dropout: 0.1
    quantize_data_type: nf4
    rank: 2
    target_modules:
    - q_proj
    - k_proj
    - v_proj
    - o_proj
  max_batch_len: 1000
  max_seq_len: 96
  mock_data: false
  mock_data_len: 0
  num_epochs: 1
  random_seed: 42
  save_samples: 100
  warmup_steps: 10
```

scripts/basic-workflow-tests.sh
cdoern: added this and the removed the config inits from the train test specifically

scripts/basic-workflow-tests.sh
cdoern: removed

.github/workflows/e2e-nvidia-a10g-x1.yml
JamesKunstle: this was because we were installing deepspeed from a prebuilt binary when we were training on vela: 
https://github.com/instructlab/openshift-manager/blob/cb775725aefeaa1d88f986151e2ca8a1503e3848/ocj2/templates/new_vela_deepspeed_train.yml.j2#L158

I'm not sure if that's still relevant

.github/workflows/e2e-nvidia-a10g-x1.yml
JamesKunstle: If we're changing the default cuda operations for a test I personally think we should note the rationale. Prevents tribal knowledge

JamesKunstle (approved):

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: so you're suggesting this could be removed from here?

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: and the explanation should include why we need it here, but why it's not appropriate to be the default behavior

and should include docs for others to know when they made to set it to achieve the same success using `ilab`

scripts/basic-workflow-tests.sh
russellb: yeah, that looks good

scripts/basic-workflow-tests.sh
russellb: OK, thanks. A comment about that would help. Someone may notice later "wait, we're overwriting TRAIN_ARGS here" -- maybe note that's intentional?

.github/workflows/e2e-nvidia-a10g-x1.yml
Maxusmusti: It is already in the training repo requirements.txt though, so something odd going on here: https://github.com/instructlab/training/blob/main/requirements.txt#L1

Maxusmusti: @russellb is there a different version needed here?

.github/workflows/e2e-nvidia-a10g-x1.yml
Maxusmusti: @cdoern the following lines are unecessary now:
```
          nvidia-smi
          python3.11 -m pip show nvidia-nccl-cu12
```
Additionally, the original reason for the reinstall of deepspeed was to have it build the DeepSpeedCPUAdam optimizer at install time, rather than runtime (since there were missing cuda bits on the box for runtime compilation). Since we have resolved the original issues though, there's a chance that this re-install step is removable now, but needs to be tested

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: based on that, it looks like the line can be removed. If it fails, can you share what it looks like @cdoern ?

russellb: or if you want to just file an issue to follow up on ensuring this is fixed for good and gets removed from this file, that's fine too

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: thanks for filing the issue!

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: I'm OK with either resolving this here, or filing an issue for the remaining cleanup related to deepspeed install/uninstall

You could  use a single issue for a list of training-lib-in-e2e cleanup tasks if you want

scripts/basic-workflow-tests.sh
```diff
@@ -186,6 +192,17 @@ test_train() {
     fi
 
     ilab model train "${TRAIN_ARGS[@]}"
+
+    if [ "$FULLTRAIN" -eq 1 ]; then
```
russellb, line 196: When FULLTRAIN=1, we are running train twice.  Once with legacy set, but without 4-bit-quant, and then another time here in this code block you added.

Part of the issue is we've overridden the meaning of FULLTRAIN. It originally meant "don't use 4-bit-quant", though it's not a great name for that behavior.

A few options to resolve this:

1. Add a new option for using the training lib, and use that instead of FULLTRAIN
2. Change the legacy-without-4-bit-quant variant to have its own option and keep FULLTRAIN for this.
3. Leave it how it is where we run training twice, but add it to the list in an issue of follow-up cleanup tasks

.github/workflows/e2e-nvidia-a10g-x1.yml
cdoern: https://github.com/instructlab/instructlab/issues/1655

.github/workflows/e2e-nvidia-a10g-x1.yml
cdoern: https://github.com/instructlab/instructlab/issues/1655

scripts/basic-workflow-tests.sh
cdoern: https://github.com/instructlab/instructlab/issues/1655

cdoern:
@russellb I added some comments, made a catch-all follow up. lmk if that satisfies all of the requirements here. if not, I am happy to add some more stuff!

scripts/basic-workflow-tests.sh
```diff
@@ -231,6 +237,18 @@ test_train() {
...
+
+        # TODO Only cuda for now
+        # the train profile specified in test_init overrides the majority of TRAIN_ARGS, including things like num_epochs. While it looks like much of those settings are being lost, they just have different values here.
+        TRAIN_ARGS=("--device=cuda" "--model-path=instructlab/granite-7b-lab" "--data-path=${SCRIPTDIR}/test-data/train_all_pruned_SDG.jsonl" "--quantize-data-type=nf4" "--4-bit-quant")
```
russellb, line 245: just FYI -- I added another item to #1655

```
- [ ] update training-lib based training to do training based on the output of `ilab data generate` instead of a file checked in to the tree - `scripts/test-data/train_all_pruned_SDG.jsonl`
```

russellb (approved):
Thanks for the issue tracking follow-ups - #1655 

I added two items to the issue:

>  update training-lib based training to do training based on the output of ilab data generate instead of a file checked in to the tree - scripts/test-data/train_all_pruned_SDG.jsonl

> update docs/ci.md to reflect where the training-lib is used

I was going to do the doc update for you (it's a 1-character change), but I would have to rebase your branch for it not to cause an immediate conflict, so I left it for now. I can include it in another PR later if you don't beat me to it, no big deal

scripts/basic-workflow-tests.sh
```diff
@@ -110,8 +110,14 @@ test_smoke() {
...
 test_init() {
     task Initializing ilab
-    ilab config init --non-interactive
 
+    if [ "$FULLTRAIN" -eq 1 ]; then
+        # TODO Only cuda for now
+        step Setting train-profile for GPU accelerated training
+        ilab config init --non-interactive --train-profile="${SCRIPTDIR}/test-data/train-profile-a10.yaml"
+    else
+        [ -f config.yaml ] || ilab config init --non-interactive
```
leseb, line 119: I don't think `[ -f config.yaml ]` is ever going to work given that we explicitly use `"${CONFIG_HOME}/instructlab/config.yaml"` for the config file. @RobotSail to keep me honest but I raised it in another and it was acked. It's minimal but why re-introduce it?

scripts/basic-workflow-tests.sh
```diff
@@ -231,6 +237,18 @@ test_train() {
...
 
     ilab model train "${TRAIN_ARGS[@]}"
+
+    if [ "$FULLTRAIN" -eq 1 ]; then
+
+        # TODO Only cuda for now
+        # the train profile specified in test_init overrides the majority of TRAIN_ARGS, including things like num_epochs. While it looks like much of those settings are being lost, they just have different values here.
+        TRAIN_ARGS=("--device=cuda" "--model-path=instructlab/granite-7b-lab" "--data-path=${SCRIPTDIR}/test-data/train_all_pruned_SDG.jsonl" "--quantize-data-type=nf4" "--4-bit-quant")
+        if [ "$GRANITE" -eq 1 ]; then
+            TRAIN_ARGS+=("--gguf-model-path models/granite-7b-lab-Q4_K_M.gguf")
```
leseb, line 247:
```suggestion
            TRAIN_ARGS+=("--gguf-model-path" "models/granite-7b-lab-Q4_K_M.gguf")
```
I believe @russellb had issues where the line was treated as a single argument.

scripts/basic-workflow-tests.sh
leseb, line 245: Is it right that we override the current `TRAIN_ARGS` array (initialized through `set_defaults`)? Shouldn't this be `TRAIN_ARGS+=(` instead?

cdoern: we want to use the things set in the train profile so we need to get rid of some of the flags already set I believe:

` the train profile specified in test_init overrides the majority of TRAIN_ARGS, including things like num_epochs. While it looks like much of those settings are being lost, they just have different values here.`

scripts/basic-workflow-tests.sh
cdoern: good point, I noticed there was a rebase so I kept what I had. Will put back the regular `ilab config init --non-interactive`

RobotSail: +1 @leseb , we need to use "${CONFIG_HOME}/instructlab/config.yaml" here

scripts/basic-workflow-tests.sh
cdoern: changed in both places

scripts/basic-workflow-tests.sh
```diff
@@ -110,8 +110,14 @@ test_smoke() {
 
 test_init() {
     task Initializing ilab
-    ilab config init --non-interactive
 
+    if [ "$FULLTRAIN" -eq 1 ]; then
```
RobotSail, line 114: We should prefer to use `[[` and `]]` instead of `[` and `]` since it reduces errors. See the guideline on [`[[` vs `[`](https://google.github.io/styleguide/shellguide.html#test----and---). 

```suggestion
    if [[ "${FULLTRAIN}" -eq 1 ]]; then
```

RobotSail (commented):
~~LGTM~~ Sorry - it autopopulated my comment box, I'm just leaving comments for now...

cdoern:
whoops @RobotSail it seems I pushed at the same time you reviewed, sorry If you missed the two new things I just changed

scripts/basic-workflow-tests.sh
RobotSail, line 247: Should this now be `"${DATA_HOME}/models` (or whatever the correct variable is) instead of `models`?

RobotSail (commented):
A few minor comments but otherwise LGTM. Once they get addressed we can merge.

scripts/basic-workflow-tests.sh
leseb: Ok just wanted to make sure we don't override anything important!

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -116,18 +117,43 @@ jobs:
...
           . venv/bin/activate
+          nvidia-smi
           sed 's/\[.*\]//' requirements.txt > constraints.txt
           python3.11 -m pip cache remove llama_cpp_python
           CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install --force-reinstall --no-binary llama_cpp_python -c constraints.txt llama_cpp_python
           python3.11 -m pip install bitsandbytes
+
+          # TODO This should be added to instructlab-training
+          python3.11 -m pip install packaging wheel
+
```
leseb, line 133: should we remove that empty line?

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -116,18 +117,43 @@ jobs:
...
+          python3.11 -m pip install instructlab-training[cuda]
+
```
leseb, line 135: ditto empty line

leseb (approved):
Few nits, non-blocking though. Thanks!

russellb:
tiran's request for review was from early on the PR and was placed by Nathan. there's not outstanding feedback from him that needs another look or anything, so I'm going to let this merge.

Issue #1456, opened 2024-06-25
Add vLLM functional test in CI


leseb:
While working on the addition of vLLM https://github.com/instructlab/instructlab/pull/1442, I tried adding func test to the e2e test since the runner has a CUDA GPU. Unfortunately, it does not have enough GPU memory. This is a Linux VM with a single Tesla T4 GPU.

I ended up with the following error:

```
ERROR 2024-06-25 12:52:41,612 serving_backend:0: flush [rank0]: Traceback (most recent call last):
[rank0]:   File "/home/runner/work/instructlab/instructlab/venv/bin/ilab", line 8, in <module>
[rank0]:     sys.exit(ilab())
[rank0]:              ^^^^^^
[rank0]:   File "/home/runner/work/instructlab/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
[rank0]:     return self.main(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/home/runner/work/instructlab/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
[rank0]:     rv = self.invoke(ctx)
[rank0]:          ^^^^^^^^^^^^^^^^
[rank0]:   File "/home/runner/work/instructlab/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
[rank0]:     return _process_result(sub_ctx.command.invoke(sub_ctx))
[rank0]:                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/home/runner/work/instructlab/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
[rank0]:     return _process_result(sub_ctx.command.invoke(sub_ctx))
[rank0]:                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/home/runner/work/instructlab/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
[rank0]:     return ctx.invoke(self.callback, **ctx.params)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/home/runner/work/instructlab/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
[...]

PR #1559, opened 2024-07-02
ci: add option to run e2e test with the vLLM backend

leseb:
The e2e script has now a new '-v' option (do not think verbose) that
will activate the vLLM backend when running the server.
We are sort of approaching the limit of what 'getopts' can do in terms
of choosing meaninful options. We could in the future transition to
'getopt' to allow for long flag to be added.

Closes: https://github.com/instructlab/instructlab/issues/1456

**Checklist:**

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -17,6 +17,7 @@ GENERATE_ARGS=("--num-cpus" "$(nproc)")
 TRAIN_ARGS=()
 GRANITE=0
 FULLTRAIN=0
+BACKEND="llama-cpp"
 
 export GREP_COLORS='mt=1;33'
 BOLD='\033[1m'
@@ -62,6 +63,9 @@ test_download() {
     if [ "$GRANITE" -eq 1 ]; then
         step Downloading the granite model
         ilab model download --repository instructlab/granite-7b-lab-GGUF --filename granite-7b-lab-Q4_K_M.gguf
+    elif [ "$BACKEND" = "vllm" ]; then
+        step Downloading the model for vLLM
+        ilab download --repository instructlab/merlinite-7b-lab
     else
         step Downloading the default model
         ilab model download
@@ -78,6 +82,9 @@ test_serve() {
     SERVE_ARGS=()
     if [ -n "$model" ]; then
         SERVE_ARGS+=("--model-path" "${model}")
+    fi
+    if [ "$BACKEND" = "vllm" ]; then
+        SERVE_ARGS+=("--model-path" "./models/instructlab/merlinite-7b-lab")
     fi
 
     task Serve the model
@@ -127,6 +134,9 @@ test_generate() {
     task Generate synthetic data
     if [ "$GRANITE" -eq 1 ]; then
         GENERATE_ARGS+=("--model ./models/granite-7b-lab-Q4_K_M.gguf")
+    fi
+    if [ "$BACKEND" = "vllm" ]; then
+        GENERATE_ARGS+=("--model ./models/instructlab/merlinite-7b-lab")
     fi
     ilab data generate --num-instructions ${NUM_INSTRUCTIONS} "${GENERATE_ARGS[@]}"
 }
@@ -218,13 +228,14 @@ usage() {
     echo "  -m  Run minimal configuration (run quicker when you have no GPU)"
     echo "  -f  Run the fullsize training instead of --4-bit-quant"
     echo "  -g  Use the granite model"
+    echo "  -v  Use the vLLM backend for serving"
     echo "  -h  Show this help text"
 
 }
 
 # Process command line arguments
 task "Configuring ..."
-while getopts "cmfgh" opt; do
+while getopts "cmfghv" opt; do
     case $opt in
         c)
             # old option, don't fail if it's specified
@@ -244,6 +255,10 @@ while getopts "cmfgh" opt; do
         h)
             usage
             exit 0
+            ;;
+        v)
+            BACKEND=vllm
+            step "Running with vLLM backend."
             ;;
         \?)
             echo "Invalid option: -$OPTARG" >&2
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -81,19 +81,8 @@ test_serve() {
     fi
 
     task Serve the model
-    ilab model serve "${SERVE_ARGS[@]}" &
-
-    ret=1
-    for i in $(seq 1 10); do
-        sleep 5
-    	step "$i"/10: Waiting for model to start
-        if curl -sS http://localhost:8000/docs > /dev/null; then
-            ret=0
-            break
-        fi
-    done
-
-    return $ret
+    ilab model serve "${SERVE_ARGS[@]}" &> serve.log &
+    wait_for_server
 }
 
 test_chat() {
@@ -208,6 +197,20 @@ test_exec() {
     task Stopping the ilab model serve
     step Kill ilab model serve $PID
     kill $PID
+}
+
+wait_for_server(){
+    if ! timeout 120 bash -c '
+        until curl -sS http://localhost:8000/docs &> /dev/null; do
+            echo "waiting for server to start"
+            sleep 1
+        done
+    '; then
+        echo "server did not start"
+        cat serve.log || true
+        exit 1
+    fi
+    echo "server started"
 }
 
 usage() {


russellb (commented):
looks good - one question inline

scripts/basic-workflow-tests.sh
```diff
@@ -62,6 +63,9 @@ test_download() {
     if [ "$GRANITE" -eq 1 ]; then
         step Downloading the granite model
         ilab model download --repository instructlab/granite-7b-lab-GGUF --filename granite-7b-lab-Q4_K_M.gguf
+    elif [ "$BACKEND" = "vllm" ]; then
+        step Downloading the model for vLLM
+        ilab download --repository instructlab/merlinite-7b-lab
     else
         step Downloading the default model
         ilab model download
```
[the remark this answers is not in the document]
leseb: `ilab model download` will get the Merline model with the GGUF format AFAIK, with vllm we need to pull the whole model dir with the tensors & stuff. At least that's my testing revealed when I was implementing the vllm backend.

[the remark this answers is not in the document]
russellb: oh, I see.  I didn't understand that difference. Thanks!

scripts/basic-workflow-tests.sh
```diff
@@ -154,6 +150,9 @@ test_train() {
     if [ "$GRANITE" -eq 1 ]; then
         TRAIN_ARGS+=("--gguf-model-path models/granite-7b-lab-Q4_K_M.gguf")
     fi
+    if [ "$BACKEND" = "vllm" ]; then
+        GENERATE_ARGS+=("--model ./models/instructlab/merlinite-7b-lab/")
```
alimaredia, line 154: I don't know of the legacy version of `ilab model train` using `vllm` or `llama-cpp` under the hood to do any serving. What was the intended purpose of adding this here?

scripts/basic-workflow-tests.sh
```diff
@@ -140,6 +133,9 @@ test_generate() {
     if [ "$GRANITE" -eq 1 ]; then
         GENERATE_ARGS+=("--model ./models/granite-7b-lab-Q4_K_M.gguf")
     fi
+    if [ "$BACKEND" = "vllm" ]; then
+        GENERATE_ARGS+=("--model ./models/instructlab/merlinite-7b-lab")
+    fi
```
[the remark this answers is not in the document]
alimaredia: How about if granite-7b-lab was the model specified instead of merlinite?

scripts/basic-workflow-tests.sh
```diff
@@ -82,19 +86,8 @@ test_serve() {
...
-        sleep 5
-    	step "$i"/10: Waiting for model to start
-        if curl -sS http://localhost:8000/docs > /dev/null; then
-            ret=0
-            break
-        fi
-    done
-
-    return $ret
+    ilab model serve "${SERVE_ARGS[@]}" &> serve.log &
```
alimaredia, line 89: Do we need something above that sets the `--model-path` in `SERVE_ARGS` to a model vLLM can use? If the `-v` flag is set I'm not sure what the model path will be here and how ilab serve will know to use vLLM as the backend for it.

scripts/basic-workflow-tests.sh
leseb: Good catch, let me remove this. Thanks

scripts/basic-workflow-tests.sh
leseb: That's correct, let me fix that. Thanks!

scripts/basic-workflow-tests.sh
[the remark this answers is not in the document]
leseb: @russellb They are mutually exclusive but the way the CI runs today does not make them conflict so I decided not to do anything about it.

@alimaredia I think the problem remains the same (GGUF vs directory) but I'm open to using granite.

russellb:
> > It looks like this has to merge so that the workflow file can run the update made in `.github/workflows/e2e-nvidia-a10g-x1.yml`...
> 
> you're right. I'll run this on a test repo for you real quick

https://github.com/russellb/ilab-runner/actions/workflows/e2e-nvidia-a10g-x1-1559.yml

I'll figure out a better way to handle this for the future, but for now ... this will test your script

scripts/basic-workflow-tests.sh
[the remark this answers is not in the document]
russellb: the granite option is broken for other reasons... I'm just going to remove it in another PR at some point

scripts/basic-workflow-tests.sh
```diff
@@ -80,21 +84,13 @@ test_serve() {
     if [ -n "$model" ]; then
         SERVE_ARGS+=("--model-path ${model}")
     fi
+    if [ "$BACKEND" = "vllm" ]; then
+        SERVE_ARGS+=("--model ./models/instructlab/merlinite-7b-lab")
```
russellb, line 88:
```suggestion
        SERVE_ARGS+=("--model" "./models/instructlab/merlinite-7b-lab")
```

also, I don't see it actually configuring the job to use vllm as the backend?  Am I missing it?

russellb (changes requested):

scripts/basic-workflow-tests.sh
russellb: This should have also made the test run I did fail, but it didn't ... that's my fault. it didn't actually specify the `-v` option.  I've fixed that in my test repo, though the "not actually using vllm" thing is still an issue

n1hility: > also, I don't see it actually configuring the job to use vllm as the backend? Am I missing it?
The model format auto detection should cause the vllm module to start

n1hility: although maybe we want to be more explicit

russellb: oh, ok, sorry - then it was just the change I missed in my test workflow.  I'll run it again!

russellb:
I applied the `SERVE_ARGS` bug fix to this branch and ran the test job again: https://github.com/russellb/ilab-runner/actions/runs/9784945610

scripts/basic-workflow-tests.sh
```diff
@@ -80,21 +84,13 @@ test_serve() {
     if [ -n "$model" ]; then
         SERVE_ARGS+=("--model-path ${model}")
     fi
+    if [ "$BACKEND" = "vllm" ]; then
+        SERVE_ARGS+=("--model" "./models/instructlab/merlinite-7b-lab")
```
russellb, line 88:
```suggestion
        SERVE_ARGS+=("--model-path" "./models/instructlab/merlinite-7b-lab")
```

russellb:
pushed another fix to this PR, another test run here: https://github.com/russellb/ilab-runner/actions/runs/9785364009

scripts/basic-workflow-tests.sh
russellb: I do think it's worth coming back and making it more explicit instead of relying on implicit behavior, but that's not something to block on

russellb:
> pushed another fix to this PR, another test run here: https://github.com/russellb/ilab-runner/actions/runs/9785364009

This failed because of not enough GPU memory for what vllm was trying to allocate.

Would you like to try a smaller model that will fit?  or would you like to move to a different instance type?

I'm probably adding `g4dn.12xlarge` next, because that's big enough to run a quantized Mixtral to test full SDG.

n1hility (approved):

russellb:
I’m ok merging the script change now but not the workflow change. It will break the workflow based on my testing. See last comment.

n1hility:
sorry misundertood your comment, thought it was referencing the future

scripts/basic-workflow-tests.sh
```diff
@@ -213,19 +212,34 @@ test_exec() {
     kill $PID
 }
 
+wait_for_server(){
+    if ! timeout 120 bash -c '
+        until curl -sS http://localhost:8000/docs &> /dev/null; do
```
alimaredia, line 217: Do models served with vLLM have a docs page? Setting the address being checked to `http://localhost:8000/v1/models`  mirrors the checks we have internally in `list_models`.

leseb: Yes, initially I thought it wasn't there but I tested it and `/docs` is present.

russellb (approved):
I rebased this and squashed my fixes into the main commit that added vllm to the e2e script. I removed the workflow change since it's not functional in that workflow.

I'm OK with merging this now so we can continue getting other test script changes in.

@leseb @n1hility @alimaredia

russellb:
This workflow would allow us to test a PR that adds new script arguments without having to modify the existing workflow in the future: https://github.com/instructlab/instructlab/pull/1609

n1hility (approved):

PR #1560, opened 2024-07-02
ci: do not use minimal config for NVIDIA Tesla T4 x1

leseb:
This is a more powerful runner so we should use more CPUs.

**Checklist:**


russellb:
> This is a more powerful runner so we should use more CPUs.

funny enough, it's actually only in the minimal case that we set `--num-cpus $(nproc)`. We should really be setting that in all cases.

the change makes it use the default for `--num-instructions` for generate and `--num-epochs` for train. That will make it take longer, and I'm not sure we're getting any additional coverage that's valuable since we're not actually testing the result in an any way.

PR #1561, opened 2024-07-02
fix: add list of supported backends in the CLI help

leseb:
We support two backends: llama-cpp and vllm, the helper text was only displaying 'llama-ccp' so now we print the correct list of supported backends.

CLI output:

```
  --backend [llama-cpp|vllm]  The backend to use for serving the model.
                              Automatically detected based on the model file
                              properties.
```

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -112,21 +112,6 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
         return first_four_bytes_int == GGUF_MAGIC
 
 
-def validate_backend(backend: str) -> None:
-    """
-    Validate the backend.
-    Args:
-        backend (str): The backend to validate.
-    Raises:
-        ValueError: If the backend is not supported.
-    """
-    # lowercase backend for comparison in case of user input like 'Llama'
-    if backend.lower() not in SUPPORTED_BACKENDS:
-        raise ValueError(
-            f"Backend '{backend}' is not supported. Supported: {', '.join(SUPPORTED_BACKENDS)}"
-        )
-
-
 def determine_backend(model_path: pathlib.Path) -> str:
     """
     Determine the backend to use based on the model file properties.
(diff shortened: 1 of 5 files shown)


russellb (approved):

russellb:
thanks!

src/instructlab/model/serve.py
```diff
@@ -49,9 +50,9 @@
     "--backend",
     type=click.STRING,  # purposely not using click.Choice to allow for auto-detection
```
tiran, line 51: You can use choice with an implicit `default=None` for auto-detection.

```python
import click


@click.command()
@click.option(
    "--backend",
    type=click.Choice(["llama-cpp", "vllm"]),
)
def test(backend):
    print(backend)


test()
```

```console
$ python3 test.py
None
$ python3 test.py --backend vllm
vllm
$ python3 test.py --help
Usage: test.py [OPTIONS]

Options:
  --backend [llama-cpp|vllm]
  --help                      Show this message and exit.

**Review by tiran — CHANGES_REQUESTED** [review]
Please use `click.Choice`. I have added an example how the implicit default `None` works.

File: src/instructlab/model/serve.py (hunk 1 above)
Comment (leseb) [reply src/instructlab/model/serve.py]: Addressed, thanks for the suggestion.

File: src/instructlab/model/backends/backends.py [hunk 2]
```diff
@@ -172,15 +157,14 @@ def get(logger: logging.Logger, model_path: pathlib.Path, backend: str) -> str:
...
-    if backend is None or backend == "":
+    if backend is None:
         logger.debug(
             f"Backend is not set using auto-detected value: {auto_detected_backend}"
         )
         backend = auto_detected_backend
     # If the backend was set using the --backend flag, validate it.
     else:
         logger.debug(f"Validating '{backend}' backend")
-        validate_backend(backend)
```
leseb, line 183: I'm keeping the else since https://github.com/instructlab/instructlab/pull/1576 will need it. Ideally this should merge prior to https://github.com/instructlab/instructlab/pull/1576.

nathan-weinberg (approved):

tiran (approved):

PR #1562, opened 2024-07-02
e2e: Always set `--num-cpus` for generate

russellb:
The script only set this parameter when the minimal option was used,
but we should be setting this option in all cases.

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -13,7 +13,7 @@ set -euf
 
 MINIMAL=0
 NUM_INSTRUCTIONS=5
-GENERATE_ARGS=()
+GENERATE_ARGS=("--num-cpus" "$(nproc)")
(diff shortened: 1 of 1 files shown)


scripts/basic-workflow-tests.sh
```diff
@@ -34,12 +34,14 @@ task() {
 }
 
 set_defaults() {
+    GENERATE_ARGS+=("--num-cpus" "$(nproc)")
```
leseb, line 37: At this point, I would initialize the array directly with those args since they get set anyways.

leseb (changes requested):

scripts/basic-workflow-tests.sh
russellb: good suggestion - done!

leseb (approved):

nathan-weinberg (approved):

PR #1564, opened 2024-07-02
Modify the Eval config to operate like training and serving

cdoern:
Training and Serving now operate under the following workflow when it comes to the config.yaml

1. user sets the config in `ilab config init`
2. user can edit portions of the config as they see fit
3. when they run a command like `ilab model serve` or `ilab model train` the values on the cfg are flattened and used as the defaults for the cmd flags
4. the user can at runtime override the defaults by using the flags.

Eval is a special case as it has its own cfg section but also needs to use bits from the serve portion of the cfg.

This PR flattens and includes the serve cfg in the eval default flag map. In order to make this work properly I added `--backend` as a flag for eval. This serves as an override depending on if the user intends to use vllm or llama-cpp. 

The current code overrides some stuff manually in the serve config, I don't think we should be in the buisness of doing this behind the scenes. If the users wants to change their max-ctx-size or max_workers, they can do so in the cfg or in the flags

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -245,9 +245,6 @@ def launch_server(
     # pylint: disable=import-outside-toplevel
     # First Party
 
-    # we shouldn't be in the business of editing a user's cfg.
-    # with eval, this is tricky since there are a few things we want to help them out with if they do not exist.
-    # if the user did not set a backend, or model_path, set that for them. Everything else needs to be in the cfg. or in the cmd as flag overrides.
     eval_serve = copy(ctx.obj.config.serve)
     if backend is None:
         eval_serve.backend = backends.VLLM
@@ -256,9 +253,7 @@ def launch_server(
         if gpus:
             # Don't override from vllm_args
             if "--tensor-parallel-size" not in eval_serve.vllm.vllm_args:
-                eval_serve.vllm.vllm_args.extend(
-                    ["--tensor-parallel-size", str(gpus)]
-                )
+                eval_serve.vllm.vllm_args.extend(["--tensor-parallel-size", str(gpus)])
             else:
                 click.echo(
                     "Ignoring --gpus with --tensor-parallel-size configured in serve vllm_args"
@@ -279,7 +274,7 @@ def launch_server(
 
     eval_serve.model_path = model
 
-    backend_instance = backends.select_backend(eval_serve)
+    backend_instance = backends.select_backend(eval_serve, backend)
     try:
         # http_client is handling tls params
         api_base = backend_instance.run_detached(http_client(ctx.params))
@@ -382,7 +377,7 @@ def launch_server(
 )
 @click.option(
     "--backend",
-    type=click.Choice([backends.VLLM, backends.LLAMA_CPP]),
+    type=click.Choice(tuple(backends.SUPPORTED_BACKENDS)),
     help="Serving backend to use for evaluation. Options are vllm and llama-cpp.",
 )
 @click.option(
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -315,13 +315,13 @@ def get_uvicorn_config(app: fastapi.FastAPI, host: str, port: int) -> Config:
     )
 
 
-def select_backend(cfg: serve_config) -> BackendServer:
+def select_backend(cfg: serve_config, backend: Optional[str] = None) -> BackendServer:
     # Local
     from .llama_cpp import Server as llama_cpp_server
     from .vllm import Server as vllm_server
 
     model_path = pathlib.Path(cfg.model_path)
-    backend_name = cfg.backend
+    backend_name = backend if backend is not None else cfg.backend
     try:
         backend = get(model_path, backend_name)
     except ValueError as e:


src/instructlab/model/evaluate.py
```diff
@@ -1,12 +1,19 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# pylint: disable=C0412
```
nathan-weinberg, line 3: Can you include the rule name here instead of the code?

src/instructlab/model/evaluate.py
```diff
@@ -1,12 +1,19 @@
...
 # Standard
 import enum
 import logging
 import os
 import typing
+import copy
```
nathan-weinberg, line 9:
```suggestion
from copy import copy
```

src/instructlab/model/evaluate.py
```diff
@@ -213,29 +220,32 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
-    if ctx.obj.config.serve.backend == backends.VLLM:
-        ctx.obj.config.serve.vllm.vllm_args.extend(["--served-model-name", model_name])
-    elif ctx.obj.config.serve.backend == backends.LLAMA_CPP:
-        # mt_bench requires a larger context size
-        ctx.obj.config.serve.llama_cpp.max_ctx_size = 5120
+
+    # we shouldn't be in the business of editing a user's cfg.
+    # with eval, this is tricky since there are a few things we want to help them out with if they do not exist.
+    # if the user did not set a backend, or model_path, set that for them. Everything else needs to be in the cfg. or in the cmd as flag overrides.
+    temp_serve = copy.copy(ctx.obj.config.serve)
```
nathan-weinberg, line 230:
```suggestion
    temp_serve = copy(ctx.obj.config.serve)
```

src/instructlab/model/evaluate.py
```diff
@@ -213,29 +220,32 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
+    if backend is None:
+        temp_serve.backend = backends.VLLM
+    if backend == backends.VLLM:
+        temp_serve.vllm.vllm_args.extend(["--served-model-name", model_name])
+        temp_serve.vllm.max_ctx_size = max_ctx_size
+    if backend == backends.LLAMA_CPP:
+        if max_ctx_size != 5120:
+            click.secho(
+                f"Max ctx size of {max_ctx_size} does not match the requirement for the serving backend. For vllm the ctx_size should be 4096 and for llama-cpp it should be 5120. Please edit your config or use the --max-ctx-size flag.",
```
nathan-weinberg, line 239:
```suggestion
                f"Max ctx size of {max_ctx_size} does not match the requirement for the serving backend. For 'vllm' the ctx_size should be 4096 and for 'llama-cpp' it should be 5120. Please edit your config or use the '--max-ctx-size' flag.",
```

src/instructlab/model/evaluate.py
```diff
@@ -334,6 +344,20 @@ def launch_server(
     default="",
     help="TLS client certificate password for model serving.",
 )
+@click.option(
+    "--backend",
+    type=str,
+    default=backends.VLLM,
+    show_default=True,
+    help="serving backend to use for evaluation. Options are vllm and llama-cpp.",
```
nathan-weinberg, line 352:
```suggestion
    help="serving backend to use for evaluation. Options are vllm and llama-cpp.",
```
Can we just make this a `click.Choice`?

src/instructlab/model/evaluate.py
```diff
@@ -334,6 +344,20 @@ def launch_server(
...
+)
+@click.option(
+    "--max-ctx-size",
+    type=click.INT,
+    default=4096,
+    show_default=True,
+    help="context window to set when launching a server",
```
nathan-weinberg, line 359: nit
```suggestion
    help="Context window to set when launching a server",
```

src/instructlab/model/evaluate.py
```diff
@@ -353,8 +377,12 @@ def evaluate(
     tls_client_cert,  # pylint: disable=unused-argument
     tls_client_key,  # pylint: disable=unused-argument
     tls_client_passwd,  # pylint: disable=unused-argument
+    backend,
+    max_ctx_size,
 ):
     # get appropriate evaluator class from Eval lib
+
+    print(ctx.params)
```
nathan-weinberg, line 385: Remove these lines

nathan-weinberg (changes requested):

src/instructlab/model/evaluate.py
danmcp, line 239: We aren't setting the context size for vllm.  If we were, it would need to be the same as llama-cpp.

src/instructlab/model/evaluate.py
danmcp, line 352: Capital S in serving to be consistent with other commands.

src/instructlab/model/evaluate.py
```diff
@@ -213,29 +220,32 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
+    # with eval, this is tricky since there are a few things we want to help them out with if they do not exist.
+    # if the user did not set a backend, or model_path, set that for them. Everything else needs to be in the cfg. or in the cmd as flag overrides.
+    temp_serve = copy.copy(ctx.obj.config.serve)
+    if backend is None:
+        temp_serve.backend = backends.VLLM
+    if backend == backends.VLLM:
+        temp_serve.vllm.vllm_args.extend(["--served-model-name", model_name])
+        temp_serve.vllm.max_ctx_size = max_ctx_size
+    if backend == backends.LLAMA_CPP:
+        if max_ctx_size != 5120:
```
danmcp, line 237: larger than 5120 should probably be allowed

src/instructlab/model/evaluate.py
```diff
@@ -353,8 +377,12 @@ def evaluate(
     tls_client_cert,  # pylint: disable=unused-argument
     tls_client_key,  # pylint: disable=unused-argument
     tls_client_passwd,  # pylint: disable=unused-argument
+    backend,
```
danmcp, line 380: Maybe move these two above the tls options since they are the most obscure.

danmcp (changes requested):

src/instructlab/model/evaluate.py
cdoern: whoops :sweat_smile:

src/instructlab/model/evaluate.py
cdoern: no because the serve config does not pick a default (it probably should)

src/instructlab/model/evaluate.py
alimaredia, line 239: Ditto what Dan said. I think we can remove this warning altogether.

src/instructlab/model/evaluate.py
```diff
@@ -213,29 +220,32 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
     return qna_to_avg_scores
 
 
-def launch_server(
-    ctx,
-    model,
-    model_name,
-    max_workers,
-) -> tuple:
+def launch_server(ctx, model, model_name, max_workers, backend, max_ctx_size) -> tuple:
```
alimaredia, line 223: @cdoern Sebastian has most of the infrastructure in place to autodetect what backend you're going to use depending on the path passed in via `--model`. Take a look at `src/instructlab/model/backends/backends.py` at the gguf autodetection code he already wrote. Khaled is working on something similar for .safetensors.

There may need to be a bit more work done but what if we remove the `--backend` flag entirely from the evaluate commands.

IMO the only command specifying `--backend` makes sense is for `ilab model serve` itself.

src/instructlab/model/evaluate.py
```diff
@@ -334,6 +344,20 @@ def launch_server(
...
 )
+@click.option(
+    "--backend",
+    type=str,
+    default=backends.VLLM,
+    show_default=True,
+    help="serving backend to use for evaluation. Options are vllm and llama-cpp.",
+)
+@click.option(
+    "--max-ctx-size",
```
alimaredia, line 355: Since this only applies to llama-cpp, can we remove this flag and just read from the serve config?

leseb: 💯

src/instructlab/model/evaluate.py
leseb: The serve config does not pick a default because the backend is auto-detected and set later down the road. We can use `click.Choice`, this is how I did it here https://github.com/instructlab/instructlab/pull/1561 based on @tiran's suggestion.

leseb (changes requested):

src/instructlab/model/evaluate.py
cdoern: sure

src/instructlab/model/evaluate.py
cdoern: hm, when using click.choice, I get an error. My config.yaml has an empty string for the backend?

src/instructlab/model/evaluate.py
cdoern: I am confused on this one. Previously we were setting the ctx FOR the user to 5120. IF we remove this error, then we let them continue with the wrong ctx size? is that a breaking change?

danmcp: The context size needs to be 5120 for everything.  Since it's set in config for llama, we need to override.  For vllm, we don't set a max so we inherit the max of the model.

src/instructlab/model/evaluate.py
cdoern: The flow we are doing for each new command is the following:

config to...
flag overrides to...
libraries

I really am trying to drive this home from a design perspective. If we allow cases like this to slip because of behind these scenese stuff, it kind of hurts the overall new design of the CLI.

cdoern:
There seems to be some debate here over 

config -> flag overrides -> libraries.

This is the design language I am trying to implement for each CLI cmd. I see how serving+eval is a special case because some options are only applicable to a certain backend. However, we can probably safely assume the amount of options per backend will expand over time.

I think hiding the max-ctx-size and backend flags and instead pulling them from the cfg is dangerous as it violates the design language I am trying to introduce. I know in many scenarios the user won't _need_ to set these however, consider the following scenario I stated above:

`

The reason I did this is because of the flow we are using in the cfg.

if user has the following serve config:

serve:
  backend: ''
  host_port: 127.0.0.1:8000
  llama_cpp:
    gpu_layers: -1
    llm_family: ''
    max_ctx_size: 4096
  vllm:
    vllm_args: []

but they change the backend manually by editing their config to:

serve:
  backend: 'llama-cpp'
  host_port: 127.0.0.1:8000
  llama_cpp:
    gpu_layers: -1
    llm_family: ''
    max_ctx_size: 4096
  vllm:
    vllm_args: []

They run a server and a chat process and are as happy as can be.

However, the --backend evaluate flag serves as a way to NOT modify their config for all other commands (serve, chat, gen, etc) but allows them to manually set --backend=vllm just for this cmd.

I think hidden checks are useful but it is SO valuable to allow the user to override their options in the config.yaml from a UX perspective.

We are taking this approach everywhere else so I feel pretty strongly about at least exposing it here for power users.



I would be ok with some of these being hidden, but for the sake of continuity over the CLI I don't think pulling things directly out of the cfg and NOT having a flag override is something we can do

@alimaredia @leseb @nathan-weinberg WDYT?

src/instructlab/model/evaluate.py
cdoern: ah I see. hm. I think the best thing to do here is default to the cfg and flag then and remove the error, yeah. I wonder if we should just switch the ctx default to 5120. Would that hurt anyone else?

danmcp: > Would that hurt anyone else?

Unfortunately it would.  If the model is configured to a lower context size max, vllm will error on api requests with a larger context.

src/instructlab/model/evaluate.py
leseb: `backend: null`

cdoern: mine doesn't seem to default to that, is there a recent patch changing this behavior?

```
serve:
  backend: ''
```

src/instructlab/model/evaluate.py
alinaryan: Why would a user want to set a diff backend just for eval if it was already autodetected before? I follow your approach, and I'm good with it as long as that is a use-case that would be hit

cdoern: the way I was thinking of it @alinaryan is if a person wants their backend to be llama say for easy single GPU operation but then for eval they want to use vllm so that they can use multi gpu support they can do that.

More generally though, the approach we are taking for all other cmds is a flag that serves as an override for anything the user might want to change on a single run. IMO I am not a huge fan of autodetecting behind the scenes and I know users will most likely want to set this themselves in the future whether there be a bug with one of the backends or special features of one of the backends. 

Autodetection is great for a base case but we can't assume auto-detect will be a catch all for customer or community use cases I think

alinaryan: That makes sense to me in the multi-gpu context! SGTM. I also agree that we should keep with the overrides especially if you have introduced that in the other cmds already. Let's be as consistent as possible :rocket:

alinaryan (dismissed):

src/instructlab/model/evaluate.py
danmcp: I still have the concern that we don't want a user to set this for eval ever and so I can't see the use case for having an option.  And if it's set in the config we need to override it.

cdoern: but I really think we can't manually override a users config, it needs to be the source of truth. So what I am trying to do here is say "hey, set the right ctx size in your config or else pass us an override.".

Starting to do config editing behind the scenes is a slippery slope IMO.

Thats why I had that warning log originally, which I think if we add back that would make sense

cdoern: and if cfg is a source of truth as @alimaredia suggested above, then having an override fits the scheme the rest of the CLI is following

tiran:
@alimaredia @leseb @nathan-weinberg @cdoern 
Please take a look at https://github.com/instructlab/instructlab/pull/1591 . It improve the handling of configuration overrides and makes the origin of a value transparent to the other. `--help` shows the dotted config name and values from `config.yaml`.

src/instructlab/model/evaluate.py
```diff
@@ -213,29 +220,32 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
-    if ctx.obj.config.serve.backend == backends.VLLM:
-        ctx.obj.config.serve.vllm.vllm_args.extend(["--served-model-name", model_name])
-    elif ctx.obj.config.serve.backend == backends.LLAMA_CPP:
-        # mt_bench requires a larger context size
-        ctx.obj.config.serve.llama_cpp.max_ctx_size = 5120
+
+    # we shouldn't be in the business of editing a user's cfg.
+    # with eval, this is tricky since there are a few things we want to help them out with if they do not exist.
+    # if the user did not set a backend, or model_path, set that for them. Everything else needs to be in the cfg. or in the cmd as flag overrides.
+    temp_serve = copy(ctx.obj.config.serve)
```
danmcp, line 230: Nit: No need to make a change the PR just for this.  But would this be better named something like eval_serve?

src/instructlab/model/evaluate.py
```diff
@@ -213,29 +220,32 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
+    if backend == backends.LLAMA_CPP:
+        if ctx.obj.config.serve.llama_cpp.max_ctx_size < 5120:
+            logging.warning(
+                "When using llama-cpp, the user must set the context size properly. Please set this to at least 5120."
+            )
         # llama-cpp fails fast on too many incoming requests and returns errors to client
-        ctx.obj.config.evaluate.mt_bench.max_workers = min(max_workers, 16)
+        if max_workers < 16:
+            logging.warning(
+                "When using llama-cpp, we reccomend setting max_workers to a minimum of 16"
```
danmcp, line 243: maximum and recommend

src/instructlab/model/evaluate.py
```diff
@@ -213,29 +220,32 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
+    # we shouldn't be in the business of editing a user's cfg.
+    # with eval, this is tricky since there are a few things we want to help them out with if they do not exist.
+    # if the user did not set a backend, or model_path, set that for them. Everything else needs to be in the cfg. or in the cmd as flag overrides.
+    temp_serve = copy(ctx.obj.config.serve)
+    if backend is None:
+        temp_serve.backend = backends.VLLM
+    if backend == backends.VLLM:
+        temp_serve.vllm.vllm_args.extend(["--served-model-name", model_name])
+    if backend == backends.LLAMA_CPP:
+        if ctx.obj.config.serve.llama_cpp.max_ctx_size < 5120:
```
danmcp, line 236: I don't think a warning is the right way to handle this.  5120 is a requirement of eval.  It shouldn't matter what the user has set in config.

src/instructlab/model/evaluate.py
```diff
@@ -213,29 +220,32 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict:
...
+            logging.warning(
+                "When using llama-cpp, the user must set the context size properly. Please set this to at least 5120."
+            )
         # llama-cpp fails fast on too many incoming requests and returns errors to client
-        ctx.obj.config.evaluate.mt_bench.max_workers = min(max_workers, 16)
+        if max_workers < 16:
```
danmcp, line 241: >

danmcp (changes requested):

src/instructlab/model/evaluate.py
cdoern: ok, I am going to put a pretty specific debug (not warn) msg that we are going to change what is in the cfg

src/instructlab/model/evaluate.py
cdoern: i am assuming this is indicating to switch to > rather than <, switched

src/instructlab/model/evaluate.py
danmcp: How about:

"Evaluate requires a context size of >= 5120, ignoring serve configuration for max_ctx_size"

since we aren't asking the user to change their values.  A lower number might make sense for their use case but not for eval.

danmcp (changes requested):

src/instructlab/model/evaluate.py
```diff
@@ -308,6 +319,11 @@ def launch_server(
     type=click.Path(),
     help="Path where all the MMLU Branch tasks are stored. Needed for running mmlu_branch.",
 )
+@click.option(
+    "--backend",
+    type=click.Choice([backends.VLLM, backends.LLAMA_CPP]),
```
leseb, line 324: You can pass the frozenset directly.
```suggestion
    type=click.Choice(tuple(backends.SUPPORTED_BACKENDS)),
```

src/instructlab/model/evaluate.py
```diff
@@ -215,29 +222,33 @@ def qa_pairs_to_qna_to_avg_scores(qa_pairs: list[dict]) -> dict[str, float]:
...
-    elif ctx.obj.config.serve.backend == backends.LLAMA_CPP:
-        # mt_bench requires a larger context size
-        ctx.obj.config.serve.llama_cpp.max_ctx_size = 5120
+
+    # we shouldn't be in the business of editing a user's cfg.
+    # with eval, this is tricky since there are a few things we want to help them out with if they do not exist.
+    # if the user did not set a backend, or model_path, set that for them. Everything else needs to be in the cfg. or in the cmd as flag overrides.
+    eval_serve = copy(ctx.obj.config.serve)
+    if backend is None:
+        eval_serve.backend = backends.VLLM
```
leseb, line 234: Why isn't this using the auto-detection code? How do we know vLLM will work with the current model? Why not use `backends.get()` in `evaluate()`?

cdoern: This already existed, just altered it to use the new `backend` var. I'd like to keep this PR's scope to changing existing logic around eval not necessarily changing it given the timelines at hand

leseb: Oh right, missed that, let's keep it for later then!

alinaryan (approved):

src/instructlab/model/evaluate.py
```diff
@@ -229,36 +234,44 @@ def launch_server(
...
     # pylint: disable=import-outside-toplevel
     # First Party
-    from instructlab.model.backends import backends
 
-    if not ctx.obj.config.serve.backend:
-        ctx.obj.config.serve.backend = backends.VLLM
-    if ctx.obj.config.serve.backend == backends.VLLM:
-        ctx.obj.config.serve.vllm.vllm_args.extend(["--served-model-name", model_name])
+    # we shouldn't be in the business of editing a user's cfg.
+    # with eval, this is tricky since there are a few things we want to help them out with if they do not exist.
```
danmcp, line 243: It's not really about not existing and filling in blanks or overriding user config.  The model is coming from an option or the eval config.  It's just utilizing serve config to not have to duplicate it for every setting.

danmcp (changes requested):

src/instructlab/model/evaluate.py
cdoern: deleting this for now, but once we wire up the serve cfg into this command more I might add some notice like this back in

danmcp (approved):

danmcp (changes requested):
Waiting for rebase before approving again

danmcp (approved):

src/instructlab/model/evaluate.py
cdoern: didn't end up editing the cfg yaml, also used a click.choice

src/instructlab/model/evaluate.py
cdoern: serve cfg is unchanged, will follow up with some work to patch this flag to the serve cfg

src/instructlab/model/evaluate.py
```diff
@@ -235,36 +240,41 @@ def launch_server(
     model_name,
     max_workers,
     gpus,
+    backend,
 ) -> tuple:
     # pylint: disable=import-outside-toplevel
     # First Party
-    from instructlab.model.backends import backends
```
danmcp, line 241: @cdoern Was there a reason for moving this up top?

danmcp: nm, figured out why:

```
(venv) [ec2-user@ip-10-0-1-95 instructlab]$ time ilab model evaluate --model models/instructlab/granite-7b-lab --judge-model models/instructlab/granite-7b-lab --benchmark mt_bench --max-workers 16 --enable-serving-output
  File "/home/ec2-user/instructlab/venv/bin/ilab", line 5, in <module>
    from instructlab.lab import ilab
  File "/home/ec2-user/instructlab/src/instructlab/lab.py", line 18, in <module>
    from .model import model as model_group
  File "/home/ec2-user/instructlab/src/instructlab/model/model.py", line 11, in <module>
    from .evaluate import evaluate
  File "/home/ec2-user/instructlab/src/instructlab/model/evaluate.py", line 389, in <module>
    type=click.Choice(tuple(backends.SUPPORTED_BACKENDS)),
                            ^^^^^^^^
NameError: name 'backends' is not defined
```

Not sure how big a hit it's having on startup time.

tiran: I share your concern. :(

Our code is too tightly coupled. My feature proposal #1713 could solve the problem.

PR #1566, opened 2024-07-02
build: change vllm version to github.com/opendatahub-io/vllm

alimaredia:
This PR switches to using a hardened branch of vllm that makes it easier to align dependencies on, and has been tested to work well in combination with future the SDG improvements. In the future we may provide a version of these requirements that tracks with the latest upstream

**Checklist:**

### DIFF
--- a/requirements-vllm-cuda.txt
+++ b/requirements-vllm-cuda.txt
@@ -2,9 +2,4 @@
 # Dependencies for installing vLLM on CUDA
 
 # vLLM only supports Linux platform (including WSL)
-vllm>=0.5.0; sys_platform == 'linux' and platform_machine == 'x86_64' and python_version >= '3.9' and python_version <= '3.11'
-
-# Most the dependencies below were taken from https://github.com/vllm-project/vllm/blob/main/requirements-cuda.txt
-vllm-flash-attn == 2.5.9 ; sys_platform == 'linux' and platform_machine == 'x86_64' and python_version >= '3.9' and python_version <= '3.11'
-nvidia-ml-py3; sys_platform == 'linux' and platform_machine == 'x86_64' and python_version >= '3.9' and python_version <= '3.11'
-xformers == 0.0.26.post1; sys_platform == 'linux' and platform_machine == 'x86_64' and python_version >= '3.9' and python_version <= '3.11'
+vllm @git+https://github.com/opendatahub-io/vllm@2024.06.25 ; sys_platform == 'linux' and platform_machine == 'x86_64'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,6 @@ package-dir = { "" = "src" }
 [tool.setuptools.dynamic]
 dependencies = { file = ["requirements.txt"] }
 optional-dependencies.hpu = { file = ["requirements-hpu.txt"] }
-optional-dependencies.vllm-cuda = { file = ["requirements-vllm-cuda.txt"] }
 
 [tool.setuptools.packages.find]
 where = ["src"]


tiran (changes requested):
We cannot upload a package to PyPI that has a requirement with an URL. You have to drop the optional dependency from pyproject.toml.

n1hility:
@tiran ah yes good catch. Dropping as an official optional dep makes sense to me for now.

alimaredia:
@n1hility Since dropping the optional dependency gets rid of the installation option of `pip install instructlab[vllm]` for a users vllm workflow, should we document that users need to install the vllm specified here separately?

nathan-weinberg:
I was talking with @JamesKunstle and it seems upstream vLLM is only using `scipy` in one place: https://github.com/vllm-project/vllm/blob/c5832d2ae9431a1672d547c232ec46b1a9051ff0/vllm/attention/ops/blocksparse_attention/utils.py#L22

I think ideally we could pull that requirement (assuming there is an equivalent way to do that single op) and continue to use upstream vLLM instead of a fork. We made a similar ask for the `lm_eval` lib and got warm reception: https://github.com/EleutherAI/lm-evaluation-harness/issues/2059

n1hility:
> I was talking with @JamesKunstle and it seems upstream vLLM is only using `scipy` in one place: https://github.com/vllm-project/vllm/blob/c5832d2ae9431a1672d547c232ec46b1a9051ff0/vllm/attention/ops/blocksparse_attention/utils.py#L22
> 
> I think ideally we could pull that requirement (assuming there is an equivalent way to do that single op) and continue to use upstream vLLM instead of a fork. We made a similar ask for the `lm_eval` lib and got warm reception: [EleutherAI/lm-evaluation-harness#2059](https://github.com/EleutherAI/lm-evaluation-harness/issues/2059)

There's a few other differences in that branch that we want. Although, the intention of it is to track closely to upstream. So it's not a full fork, but rather syncing one. I think we'd want upstream support at some point, but for now we are trying to reduce the number of permutations we have while we complete the remaining pieces.

n1hility (approved):

nathan-weinberg (approved):

n1hility:
> @n1hility Since dropping the optional dependency gets rid of the installation option of `pip install instructlab[vllm]` for a users vllm workflow, should we document that users need to install the vllm specified here separately?

Yeah I wonder if what we need to do is document a command that looks like:

```
pip install git+https://github.com/instructlab/instructlab@version -r github.com/instructlab/instructlab/tree/version requirements.txt -r github.com/instructlab/instructlab/tree/version requirements-vllm-cuda.txt
```

alimaredia:
Putting this on hold until I can verify that someone can test by hand ilab working with this version of vLLM.

nathan-weinberg:
I'd like if we can have a followup issue (maybe in https://github.com/opendatahub-io/vllm would be appropriate place for this) to get that midstream vLLM implementation to do proper PyPI packaging and publishing so we're not installing a `git`-based checkpoint - what do you think @n1hility?

russellb (changes requested):
I would prefer we get #1559 merged and functional first before any further big vllm changes. We need testing in place that can do some basic validation.

leseb (commented):
Please update your commit message with the PR description. Thanks!

danmcp:
> Putting this on hold until I can verify that someone can test by hand ilab working with this version of vLLM.

I have verified this PR works with the changes here: https://github.com/instructlab/instructlab/pull/1589

n1hility:
@Mergifyio rebase

n1hility:
@Mergifyio rebase

leseb (commented):
Please update your commit message referencing the content of the PR description. Thanks

alimaredia:
@leseb updated the commit message.

leseb (approved):

danmcp (approved):

russellb (approved):

requirements-vllm-cuda.txt
```diff
@@ -2,9 +2,4 @@
 # Dependencies for installing vLLM on CUDA
 
 # vLLM only supports Linux platform (including WSL)
-vllm>=0.5.0; sys_platform == 'linux' and platform_machine == 'x86_64' and python_version >= '3.9' and python_version <= '3.11'
-
-# Most the dependencies below were taken from https://github.com/vllm-project/vllm/blob/main/requirements-cuda.txt
```
leseb, line 7: I believe these are still needed, why did they go away?

leseb (commented):
Added the hold label since I have a small question, but this feels ready!

requirements-vllm-cuda.txt
```diff
@@ -2,9 +2,4 @@
...
-vllm-flash-attn == 2.5.9 ; sys_platform == 'linux' and platform_machine == 'x86_64' and python_version >= '3.9' and python_version <= '3.11'
-nvidia-ml-py3; sys_platform == 'linux' and platform_machine == 'x86_64' and python_version >= '3.9' and python_version <= '3.11'
-xformers == 0.0.26.post1; sys_platform == 'linux' and platform_machine == 'x86_64' and python_version >= '3.9' and python_version <= '3.11'
+vllm @git+https://github.com/opendatahub-io/vllm@2024.06.25 ; sys_platform == 'linux' and platform_machine == 'x86_64' and python_version >= '3.9' and python_version <= '3.11'
```
leseb, line 5: from @tiran offline

> You can drop the python version range. Our package metadata already limits to 3.10 and 3.11

leseb (changes requested):

requirements-vllm-cuda.txt
tiran: vLLM's `setup.py` picks up the dependencies automatically: https://github.com/opendatahub-io/vllm/blob/main/setup.py#L367

leseb (commented):
Removing my request change since the request did not come from me.

requirements-vllm-cuda.txt
tiran, line 5:
```suggestion
vllm @git+https://github.com/opendatahub-io/vllm@2024.06.25 ; sys_platform == 'linux' and platform_machine == 'x86_64'
```

tiran:
> Yeah I wonder if what we need to do is document a command that looks like:
> 
> ```
> pip install git+https://github.com/instructlab/instructlab@version \ 
>  -r https:///raw.githubusercontent.com/instructlab/instructlab/version/requirements.txt  \
>  -r https:///raw.githubusercontent.com/instructlab/instructlab/version/requirements-vllm-cuda.txt
> ```

To install InstructLab with vLLM on CUDA:
```
VERSION=0.18.0
pip install instructlab==${VERSION} -r https://raw.githubusercontent.com/instructlab/instructlab/v${VERSION}/requirements.txt
```

You don't need `requirements.txt`, because that information is baked into the wheel. The path to the vLLM CUDA txt file needs `v0.18.0` (with a leading `v`).

alimaredia:
@leseb when you feel as if the PR is ready, I think the only thing preventing it from merging is the hold label you added.

PR #1568, opened 2024-07-02
catch all exceptions from training library

cdoern:
currently, e2e tests pass even if training fails. Add a catch-all for any type of exception just so we properly error out.


RobotSail (approved):
LGTM

nathan-weinberg (approved):

PR #1569, opened 2024-07-02
changeup some serve defaults, align with other commands

cdoern:
`get_default_config` should be the only place defaults are set, not in the class definition also, cmds like eval have shown us that we probably should just set `llama-cpp` as a default

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -26,6 +26,8 @@ import click
 import httpx
 import yaml
 
+LLAMA_CPP = "llama-cpp"
+VLLM = "vllm"
 # Local
 from . import log
 
@@ -171,8 +173,9 @@ class _serve(BaseModel):
     model_path: StrictStr
 
     # additional fields with defaults
-    host_port: StrictStr = "127.0.0.1:8000"
-    backend: str = ""  # we don't set a default value here since it's auto-detected
+    host_port: StrictStr
+
+    backend: str
 
     def api_base(self):
         """Returns server API URL, based on the configured host and port"""
@@ -249,6 +252,8 @@ def get_default_config():
         ),
         serve=_serve(
             model_path=DEFAULT_MODEL_PATH,
+            host_port="127.0.0.1:8000",
+            backend=LLAMA_CPP,
             llama_cpp=_serve_llama_cpp(
                 gpu_layers=-1,
                 max_ctx_size=4096,
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -25,8 +25,6 @@ from ...configuration import _serve as serve_config
 from ...configuration import get_api_base
 from ...utils import split_hostport
 
-LLAMA_CPP = "llama-cpp"
-VLLM = "vllm"
 SUPPORTED_BACKENDS = frozenset({LLAMA_CPP, VLLM})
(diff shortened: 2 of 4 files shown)


src/instructlab/model/backends/backends.py
```diff
@@ -25,8 +25,6 @@
 from ...configuration import get_api_base
```
alimaredia, line 25: Looks like you're missing an import for `VLLM` and `LLAMA_CPP` here from configuration

src/instructlab/model/evaluate.py
```diff
@@ -222,20 +222,21 @@ def launch_server(
...
+    elif ctx.obj.config.serve.backend == cfg.LLAMA_CPP:
         # mt_bench requires a larger context size
         ctx.obj.config.serve.llama_cpp.max_ctx_size = 5120
         # llama-cpp fails fast on too many incoming requests and returns errors to client
         ctx.obj.config.evaluate.mt_bench.max_workers = min(max_workers, 16)
 
     ctx.obj.config.serve.model_path = model
 
-    backend_instance = backends.select_backend(logger, ctx.obj.config.serve)
+    backend_instance = cfg.select_backend(logger, ctx.obj.config.serve)
```
alimaredia, line 239: This line shouldn't be changed since `select_backend()` wasn't moved into `configuration`

src/instructlab/configuration.py
```diff
@@ -26,6 +26,8 @@
 import httpx
 import yaml
 
+LLAMA_CPP = "llama-cpp"
+VLLM = "vllm"
```
leseb, line 30: Why did these two move out of "backend"?

cdoern: as with most other constants that live in the config.yaml, IMO these should live in configuration.py. most places where we import these we import config.py anyway for other constants.

Just trying to align with the principles we have been following elsewhere

src/instructlab/model/evaluate.py
cdoern: ah right, thanks!

src/instructlab/configuration.py
```diff
@@ -171,8 +173,9 @@ class _serve(BaseModel):
     model_path: StrictStr
 
     # additional fields with defaults
-    host_port: StrictStr = "127.0.0.1:8000"
-    backend: str = ""  # we don't set a default value here since it's auto-detected
+    host_port: StrictStr
+
+    backend: str
```
nathan-weinberg, line 178: Above comment also needs to be changed
```suggestion
    host_port: StrictStr
    backend: str
```

src/instructlab/configuration.py
```diff
@@ -249,6 +252,8 @@ def get_default_config():
         ),
         serve=_serve(
             model_path=DEFAULT_MODEL_PATH,
+            host_port="127.0.0.1:8000",
+            backend=LLAMA_CPP,
```
leseb, line 256: Why set a default since it's auto-detected? Wouldn't eval go through the same auto-detection path?

nathan-weinberg (commented):
One nit, otherwise just needs linting fixed and @alimaredia's comments addressed

src/instructlab/configuration.py
leseb: Ok thanks!

ktam3:
Per @cdoern - closing this PR in favor of #1634

PR #1570, opened 2024-07-03
generate: Add `--sdg-scale-factor`

russellb:
77371ee generate: Add `--sdg-scale-factor`

commit 77371ee3efe23c5fec8fdb3afa04c607135215fa
Date:   Tue Jul 2 20:21:22 2024 -0400

    generate: Add `--sdg-scale-factor`
    
    The new implementation of data generation from the `instructlab/sdg`
    library behaves differently than before. Instead of having a tunable
    for the maximum number of generated examples, it has a tunable for the
    number of examples to generate based on each seed example, for some
    definition of "seed example". The number of seed examples expodes to
    be much more than the provided q&a pairs when working with a knowledge
    document.
    
    The new option, `--sdg-scale-factor` tries to capture this new
    behavior and make it clear that it's not a specific number of examples
    anymore, but more of a multiplication factor.
    
    We could come back and add a new option that behaves like the old one,
    like `--max-generated-examples`, but even that is slightly different
    than what was here before. Instead of implementing that now, hold off
    until a good enough use case justifies the need for it, because it
    would be harmful in some cases. For example, when working with a
    knowledge document, setting a hard cap on the number of generated
    examples will most likely halt the process before the full knowledge
    document (or documents) have been processed.
    
    The configuration file has also been updated to do the following:
    
    - Optionally allow `sdg_scale_factor` so that previous config files
      still work. Many options have been added recently without doing
      this, breaking old configs.
    
    - Allow num_instructions in the config file. This allows old config
      files to work without error, but a deprecation warning will be
      emitted.
    
    - Ensure that the deprecated field is not included when generating
      a sample configuration file.
    
    I believe this is the first example of deprecating a config file
    option in a backwards compatible way.
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@
 * Default log format changes to include the logger name in the logs.
 * Introduces CUDA-specific requirements in a new file `requirements-cuda.txt`. To install these
   CUDA-enabled dependencies, simply run `pip install instructlab[cuda]`.
+* `ilab data generate` now supports a new and more extensive pipeline with the
+  option `--pipeline full`. This option requires `mixtral-8x7b-instruct` as the
+  teacher model.
 
 ### Breaking Changes
 
@@ -52,6 +55,9 @@
    program cache directory.
 * `ilab model chat`: Chatlogs are now stored under the `instructlab/checkpoints` directory in the
    platform's dedicated data directory under the `instructlab` package.
+* The `--num-instructions` option to `ilab data generate` has been deprecated.
+  See `--sdg-scale-factor` for an updated option providing similar
+  functionality.
 
 ## v0.17
 
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -73,10 +73,10 @@ class _InstructlabDefaults:
     # TODO: these constants should be removed, they should not leak out
     NUM_CPUS = 10
     CHUNK_WORD_COUNT = 1000
-    NUM_INSTRUCTIONS = 100
     CONNECTION_TIMEOUT = httpx.Timeout(timeout=30.0)
     # use spawn start method, fork is not thread-safe
     MULTIPROCESSING_START_METHOD = "spawn"
+    SDG_SCALE_FACTOR = 30
 
     # When otherwise unknown, ilab uses this as the default family
     MODEL_FAMILY = "merlinite"
@@ -256,7 +256,13 @@ class _generate(BaseModel):
     # additional fields with defaults
     num_cpus: PositiveInt = DEFAULTS.NUM_CPUS
     chunk_word_count: PositiveInt = DEFAULTS.CHUNK_WORD_COUNT
-    num_instructions: PositiveInt = DEFAULTS.NUM_INSTRUCTIONS
+    # DEPRECATED: see sdg_scale_factor instead
+    # Left in place so that we can still detect and give a warning if its
+    # specified in an old configuraiton file.
+    num_instructions: Optional[int] = Field(
+        default=-1, deprecated="see 'sdg_scale_factor' instead", exclude=True
+    )
+    sdg_scale_factor: Optional[PositiveInt] = DEFAULTS.SDG_SCALE_FACTOR
     output_dir: StrictStr = Field(default_factory=lambda: DEFAULTS.DATASETS_DIR)
     prompt_file: StrictStr = Field(default_factory=lambda: DEFAULTS.PROMPT_FILE)
     seed_file: StrictStr = Field(default_factory=lambda: DEFAULTS.SEED_FILE)
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -37,11 +37,17 @@ logger = logging.getLogger(__name__)
     default=DEFAULTS.CHUNK_WORD_COUNT,
     show_default=True,
 )
+# TODO - DEPRECATED - Remove in a future release
 @click.option(
     "--num-instructions",
     type=click.INT,
-    help="Number of instructions to generate.",
-    default=DEFAULTS.NUM_INSTRUCTIONS,
+    default=-1,
+    hidden=True,
+)
+@click.option(
+    "--sdg-scale-factor",
+    type=click.INT,
+    help="Number of instructions to generate for each seed example. The examples map to sample q&a pairs for new skills. For knowledge, examples are generated with both the sample q&a pairs, as well as chunks of the knowledge document(s), so the resulting data set is typically larger for a knowledge addition for the same value of `--sdg-scale-factor`.",
     show_default=True,
 )
 @click.option(
@@ -145,6 +151,7 @@ def generate(
     model,
     num_cpus,
     num_instructions,
+    sdg_scale_factor,
     taxonomy_path,
     taxonomy_base,
     output_dir,
@@ -168,7 +175,14 @@ def generate(
     from instructlab.sdg.generate_data import generate_data
     from instructlab.sdg.utils import GenerateException
 
+    if num_instructions != -1:
+        click.secho(
+            "The --num-instructions flag is deprecated. Please use --sdg-scale-factor instead.",
+            fg="yellow",
+        )
+
     prompt_file_path = DEFAULTS.PROMPT_FILE
+
     if ctx.obj is not None:
         prompt_file_path = ctx.obj.config.generate.prompt_file
 
@@ -199,7 +213,7 @@ def generate(
             model_family=model_family,
             model_name=model,
             num_cpus=num_cpus,
-            num_instructions_to_generate=num_instructions,
+            num_instructions_to_generate=sdg_scale_factor,
             taxonomy=taxonomy_path,
             taxonomy_base=taxonomy_base,
             output_dir=output_dir,
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -294,7 +294,7 @@ seed_examples:
 task_description: "simple maths"
 EOF
 
-    sed -i.bak -e 's/num_instructions:.*/num_instructions: 1/g' "${ILAB_CONFIG_FILE}"
+    sed -i.bak -e 's/sdg_scale_factor.*/sdg_scale_factor: 1/g' "${ILAB_CONFIG_FILE}"
 
     # This should be finished in a minute or so but time it out incase it goes wrong
     if ! timeout 20m ilab data generate --taxonomy-path test_taxonomy/compositional_skills/simple_math.yaml; then
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -46,7 +46,7 @@ class TestConfig:
         assert cfg.generate.taxonomy_path == f"{data_dir}/taxonomy"
         assert cfg.generate.taxonomy_base == "origin/main"
         assert cfg.generate.num_cpus == 10
-        assert cfg.generate.num_instructions == 100
+        assert cfg.generate.sdg_scale_factor == 30
         assert cfg.generate.chunk_word_count == 1000
         assert cfg.generate.output_dir == f"{data_dir}/datasets"
         assert cfg.generate.prompt_file == f"{data_dir}/{internal_dirname}/prompt.txt"


russellb:
this release hasn't been made yet, but I put this draft up because I didn't want to forget about this doc update

tiran:
Can you get https://github.com/instructlab/sdg/pull/73 into the release and update the imports, too? Thx!

markmc:
> This also includes a documentation updates for an intentional change in behavior. The `--num-instructions` parameter now controls the number of new examples generate on a per seed example basis (or a a bit more complicated for knowledge since it's seed examples applied across N document chunks).

Just playing with this myself now, I've got to wonder - surely the original behavior is what's actually understandable by the user, rather than difficult to predict "multiplier of the number of samples the pipeline happens to generate" behavior?

i.e. isn't "I want to generate N synthetic training samples" how we expect users to use this?

russellb:
> > This also includes a documentation updates for an intentional change in behavior. The `--num-instructions` parameter now controls the number of new examples generate on a per seed example basis (or a a bit more complicated for knowledge since it's seed examples applied across N document chunks).
> 
> Just playing with this myself now, I've got to wonder - surely the original behavior is what's actually understandable by the user, rather than difficult to predict "multiplier of the number of samples the pipeline happens to generate" behavior?
> 
> i.e. isn't "I want to generate N synthetic training samples" how we expect users to use this?

To summarize some live conversation we had ...

I said

> I totally agree it's the thing that makes sense to users.
> while I also understand that for the folks concerned with the algorithm, the amount we generate per "sample thing" is the variable to tune ...
>
> thing is, if you set a maximum num of samples, but your knowledge doc is huge, good chance we've only generated from the first subset of the doc and then stopped
> while the current behavior ensures we process the entire doc, but tunes how much we're generating per chunk of the doc, (very roughly)

to which @markmc suggested deprecating `--num-instructions` and adding a new option that matches the behavior, `--sdg-scale-factor`. I really like that.

I will update this PR with the new option. We can reconsider coming back to add an option that sets a hard cap on the number of generated samples if there's a good enough reason for it.

markmc:
Excellent, lgtm :+1:

russellb:
> Excellent, lgtm 👍

Thanks! I'm going to let this sit until we make an sdg library release that includes this behavior.

src/instructlab/data/generate.py
```diff
@@ -34,11 +34,19 @@
...
 )
+# TODO - DEPRECATED - Remove in a future release
 @click.option(
     "--num-instructions",
     type=click.INT,
-    help="Number of instructions to generate.",
-    default=config.DEFAULT_NUM_INSTRUCTIONS,
+    help="DEPRECATED. See --sdg-scale-factor",
+    default=-1,
+    hidden=True,
```
leseb, line 43: deprecated
```suggestion
    hidden=True,
    deprecated=True,
```

src/instructlab/data/generate.py
```diff
@@ -34,11 +34,19 @@
     default=config.DEFAULT_CHUNK_WORD_COUNT,
     show_default=True,
 )
+# TODO - DEPRECATED - Remove in a future release
 @click.option(
     "--num-instructions",
     type=click.INT,
-    help="Number of instructions to generate.",
-    default=config.DEFAULT_NUM_INSTRUCTIONS,
+    help="DEPRECATED. See --sdg-scale-factor",
```
leseb, line 41: If the flag is hidden will it be shown in the help? Also, the deprecated message is not needed if we use `deprecated=True`.

leseb (changes requested):
Please add a note in CHANGELOG.md that indicates `--num-instructions` is deprecated and will be removed in next two release. 0.19 is suppose. Thanks

src/instructlab/data/generate.py
russellb: I tried to look through the click docs to see if this was a thing and didn't see it. I found it for commands but not options. No surprise that the docs aren't perfect, or my reading isn't perfect. :) I'll add this when I update the PR

src/instructlab/data/generate.py
russellb: I don't think it's shown in the help output, so maybe this is never used anywhere. I'll check

src/instructlab/data/generate.py
russellb: added

src/instructlab/data/generate.py
russellb: dropped the help text

RobotSail (dismissed):
LGTM need this in ASAP

src/instructlab/data/generate.py
russellb: @leseb so i wasn't crazy. this isn't supported

```
TypeError: Parameter.__init__() got an unexpected keyword argument 'deprecated'
```

src/instructlab/data/generate.py
```diff
@@ -37,11 +37,18 @@
...
     type=click.INT,
-    help="Number of instructions to generate.",
-    default=config.DEFAULT_NUM_INSTRUCTIONS,
+    default=-1,
+    hidden=True,
+)
+@click.option(
+    "--sdg-scale-factor",
+    type=click.INT,
+    help="Number of instructions to generate for each seed example. The examples map to sample q&a pairs for new skills. For knowledge, examples are generated with both the sample q&a pairs, as well as chunks of the knowledge document(s), so the resulting data set is typically larger for a knowledge addition for the same value of `--sdg-scale-factor`.",
```
RobotSail, line 50: This is a very helpful message, but I worry about how it would look when printed out against other ones. Maybe not in this PR, but would it make sense to condense this and then move the more thorough explanation elsewhere, then have this point there?

RobotSail (dismissed):
LGTM

src/instructlab/data/generate.py
russellb: click handles the formatting pretty well.

```
Usage: ilab data generate [OPTIONS]

  Generates synthetic data to enhance your example data

Options:
  --model TEXT                Name of the model used during generation.
                              [default: models/merlinite-7b-lab-Q4_K_M.gguf]
  --num-cpus INTEGER          Number of processes to use.  [default: 10]
  --chunk-word-count INTEGER  Number of words to chunk the document  [default:
                              1000]
  --sdg-scale-factor INTEGER  Number of instructions to generate for each seed
                              example. The examples map to sample q&a pairs
                              for new skills. For knowledge, examples are
                              generated with both the sample q&a pairs, as
                              well as chunks of the knowledge document(s), so
                              the resulting data set is typically larger for a
                              knowledge addition for the same value of `--sdg-
                              scale-factor`.  [default: 30]
  --taxonomy-path PATH        Path to
                              https://github.com/instructlab/taxonomy.git
                              clone or local file path.  [default: taxonomy]
  --taxonomy-base TEXT        Base git-ref to use when generating new
                              taxonomy.  [default: origin/main]
  --output-dir PATH           Path to output generated files.
  --rouge-threshold FLOAT     Threshold of (max) Rouge score to keep samples;
                              1.0 means accept all samples.  [default: 0.9]
  --quiet                     Suppress output of synthesized instructions.
  --endpoint-url TEXT         Custom URL endpoint for OpenAI-compatible API.
                              Defaults to the `ilab model serve` endpoint.
  --api-key TEXT              API key for API endpoint. [default:
                              config.DEFAULT_API_KEY]
  --yaml-rules PATH           Custom rules file for YAML linting.
  --server-ctx-size INTEGER   The context size is the maximum number of tokens
                              the server will consider.  [default: 4096]
  --tls-insecure              Disable TLS verification.
  --tls-client-cert PATH      Path to the TLS client certificate to use.
  --tls-client-key PATH       Path to the TLS client key to use.
  --tls-client-passwd TEXT    TLS client certificate password.
  --model-family TEXT         Force model family to use when picking a
                              generation template
  --help                      Show this message and exit.
```

leseb (approved):

tiran:
I have a better idea for the deprecation. I'll get back to you in a few minutes.

src/instructlab/configuration.py
```diff
@@ -135,7 +141,11 @@ class _generate(BaseModel):
     # additional fields with defaults
     num_cpus: PositiveInt = DEFAULT_NUM_CPUS
     chunk_word_count: PositiveInt = DEFAULT_CHUNK_WORD_COUNT
-    num_instructions: PositiveInt = DEFAULT_NUM_INSTRUCTIONS
+    # DEPRECATED: see sdg_scale_factor instead
+    # Left in place so that we can still detect and give a warning if its
+    # specified in an old configuraiton file.
+    num_instructions: Optional[int] = -1
```
tiran, line 147: Use Pydantic's Field to mark the entry as deprecated and to exclude it from serialization:

```
from pydantic import Field
```

```suggestion
    num_instructions: Optional[int] = Field(-1, deprecated="see 'sdg_scale_factor' instead", exclude=True)
```

tiran (changes requested):
https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.Field has flags to exclude a field from getting serialize and to mark it as deprecated.

src/instructlab/configuration.py
russellb: I know I tried using this and then gave up, and now I can't remember what the problem was. i don't think I found `exclude=True` though, so that may have been it!

russellb: Thanks for the tip! This worked great.

russellb:
> This pull request has merge conflicts that must be resolved before it can be merged. @russellb please rebase it. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork

fixed - the only conflict was on CHANGELOG.md

cdoern (approved):
looks good. thanks for the clarity here. We can probably remove `num_instructions` in 0.19 I am thinking.

We should also probably decide if we want to set defaults in the class declaration or in `get_default_config` because rn we are doing a bunch of both.

src/instructlab/configuration.py
```diff
@@ -355,6 +361,7 @@ def get_default_config() -> Config:
             model=DEFAULTS.DEFAULT_MODEL,
             taxonomy_path=DEFAULTS.TAXONOMY_DIR,
             taxonomy_base=DEFAULTS.TAXONOMY_BASE,
+            num_instructions=-1,
```
russellb, line 364: without this, mypy failed after switching to the Field

```
src/instructlab/configuration.py:360: error: Missing named argument "num_instructions" for "_generate"  [call-arg]
```

src/instructlab/configuration.py
```diff
@@ -238,7 +238,13 @@ class _generate(BaseModel):
     # additional fields with defaults
     num_cpus: PositiveInt = DEFAULTS.NUM_CPUS
     chunk_word_count: PositiveInt = DEFAULTS.CHUNK_WORD_COUNT
-    num_instructions: PositiveInt = DEFAULTS.NUM_INSTRUCTIONS
+    # DEPRECATED: see sdg_scale_factor instead
+    # Left in place so that we can still detect and give a warning if its
+    # specified in an old configuraiton file.
+    num_instructions: Optional[int] = Field(
+        -1, deprecated="see 'sdg_scale_factor' instead", exclude=True
```
tiran, line 245: MyPy does not understand default as argument. You have to use a keyword argument here:
```suggestion
        default=-1, deprecated="see 'sdg_scale_factor' instead", exclude=True
```

src/instructlab/configuration.py
tiran: ... and now you can remove this assignment.

src/instructlab/data/generate.py
```diff
@@ -34,11 +34,18 @@
...
-    help="Number of instructions to generate.",
-    default=DEFAULTS.NUM_INSTRUCTIONS,
+    default=-1,
+    hidden=True,
+)
+@click.option(
+    "--sdg-scale-factor",
+    type=click.INT,
+    help="Number of instructions to generate for each seed example. The examples map to sample q&a pairs for new skills. For knowledge, examples are generated with both the sample q&a pairs, as well as chunks of the knowledge document(s), so the resulting data set is typically larger for a knowledge addition for the same value of `--sdg-scale-factor`.",
+    default=DEFAULTS.SDG_SCALE_FACTOR,
```
tiran, line 48: The default should not be necessary. Click should look up the value from the config.

In fact you should be able to drop all defaults from click options that have a key in the configuration system. PR #1622 will help you to debug this behavior.

tiran (changes requested):
MyPy needs some extra hand-holding to understand the default value.

src/instructlab/configuration.py
russellb: ah ha! that's what i missed with mypy. thank you!

src/instructlab/data/generate.py
russellb: perfect, yep, removed this line and the default still works properly

markmc (approved):
Thanks for taking care with the deprecation and not breaking existing config files

CHANGELOG.md
```diff
@@ -18,17 +18,22 @@
 * The `ilab` command now accepts a `-v` / `--verbose` option to enable debug logging.
   `ilab -vv` or `ilab --verbose --verbose` enables more verbose debug logging.
 * `ilab model test` generic support
-   evaluation (currently applicable for vLLM only).
```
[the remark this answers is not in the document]
russellb: Ha, thanks for pointing that out. The rebase conflict looked really weird, but that's why!

PR #1572, opened 2024-07-03
Adding ilab model evaluate unit testcases

danmcp:
Adding initial unit tests for ilab model evaluate.  Covers the 4 basic benchmarks (mt_bench, mt_bench_branch, mmlu, and mmlu_branch as well as some error cases.

Noticed during testing that `ilab model evaluate` returns without error.  Added --benchmark as a required option.  Not sure that's the right way any more.

**Checklist:**

- [ x] Unit tests have been added, if necessary.

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -262,6 +262,7 @@ def launch_server(
 @click.option(
     "--benchmark",
     type=click.Choice([m.value for m in Benchmark.__members__.values()]),
+    required=True,
     help="Benchmarks to run during evaluation",
 )
 @click.option(
--- a/tests/test_lab_evaluate.py
+++ b/tests/test_lab_evaluate.py
@@ -1,0 +1,466 @@
+# Standard
+from unittest import mock
+from unittest.mock import patch
+import os
+import textwrap
+
+# Third Party
+from click.testing import CliRunner
+from git import Repo
+
+# First Party
+from instructlab import lab
+
+
+def gen_qa_pairs(odd):
+    i = 1
+    qa_pairs = []
+    score = 0
+    while i < 5:
+        if i % 2:
+            if odd:
+                score = 0.2
+            else:
+                score = 0.1
+        elif not i % 2:
+            if odd:
+                score = 0.3
+            else:
+                score = 0.4
+        qa_pairs.append(
+            {
+                "question_id": i,
+                "score": score,
+                "qna_file": f"category{i}/qna.yaml",
+            }
+        )
+        i = i + 1
+    qa_pairs.append(
+        {
+            "question_id": i,
+            "score": 0.5,
+            "qna_file": f"category{i}/qna.yaml",
+        }
+    )
+    if odd:
+        qa_pairs.append(
+            {
+                "question_id": i + 1,
+                "score": 0.6,
+                "qna_file": f"category{i+1}/qna.yaml",
+            }
+        )
+    return qa_pairs
+
+
+def gen_individual_scores(odd):
+    i = 1
+    individual_scores = {}
+    score = 0
+    while i < 5:
+        if i % 2:
+            if odd:
+                score = 0.2
+            else:
+                score = 0.1
+        elif not i % 2:
+            if odd:
+                score = 0.3
+            else:
+                score = 0.4
+        individual_scores[f"task{i}"] = {"score": score}
+        i = i + 1
+    individual_scores[f"task{i}"] = {"score": 0.5}
+    return individual_scores
+
+
+def setup_taxonomy(taxonomy_path):
+    repo = Repo.init(taxonomy_path, initial_branch="main")
+    file_name = "README.md"
+    file_path = os.path.join(taxonomy_path, file_name)
+    with open(file_path, "w", encoding="utf-8") as f:
+        f.write("#Taxonomy")
+    repo.index.add([file_name])
+    repo.index.commit("initial commit")
+
+
+@patch(
+    "instructlab.model.evaluate.launch_server",
+    return_value=(mock.MagicMock(), "http://127.0.0.1:8000/v1"),
+)
+@patch("instructlab.eval.mt_bench.MTBenchEvaluator.gen_answers")
+@patch(
+    "instructlab.eval.mt_bench.MTBenchEvaluator.judge_answers",
+    return_value=(1.5001, [{}, {}], [1.002, 2]),
+)
+def test_evaluate_mt_bench(
+    judge_answers_mock,
+    gen_answers_mock,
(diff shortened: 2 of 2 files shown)


src/instructlab/model/evaluate.py
```diff
@@ -262,6 +262,7 @@ def launch_server(
 @click.option(
     "--benchmark",
     type=click.Choice([m.value for m in Benchmark.__members__.values()]),
+    required=True,
```
leseb, line 265: Why is this flag required all of a sudden? The commit should probably explain why.

tests/test_lab_evaluate.py
```diff
@@ -0,0 +1,329 @@
...
+        individual_scores[f"task{i}"] = {"score": score}
+        i = i + 1
+    individual_scores[f"task{i}"] = {"score": 0.5}
+    return individual_scores
+
+
+class TestEvaluate:
+    @patch(
+        "instructlab.model.evaluate.launch_server",
+        return_value=(mock.MagicMock(), "http://120.0.0.1:8000/v1"),
```
leseb, line 77:
```suggestion
        return_value=(mock.MagicMock(), "http://127.0.0.1:8000/v1"),
```

I know it's a mock but shouldn't this be `127` for clarity? Or am I missing something?

leseb (commented):
Thanks a lot for adding unit tests!

src/instructlab/model/evaluate.py
danmcp: Apologies, I had left my PR details inside of a comment block and didn't notice.  Fixed now but noting here the issue was `ilab model evaluate` returned without error/message.

tests/test_lab_evaluate.py
danmcp: Typo, will fix.

nathan-weinberg (dismissed):

tests/test_lab_evaluate.py
```diff
@@ -0,0 +1,329 @@
...
+                score = 0.3
+            else:
+                score = 0.4
+        individual_scores[f"task{i}"] = {"score": score}
+        i = i + 1
+    individual_scores[f"task{i}"] = {"score": 0.5}
+    return individual_scores
+
+
+class TestEvaluate:
```
tiran, line 74: pytest does not need a class for tests. You could drop the class and move all methods to module-level functions.

tests/test_lab_evaluate.py
```diff
@@ -0,0 +1,329 @@
...
+
+### AVERAGE:
+1.5 (across 2)
+
+### TURN ONE:
+1.0
+
+### TURN TWO:
+2
+"""
```
tiran, line 127: The inline text without same indention makes the code harder to follow. I like to use [textwrap.dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) to have nicer indentions

```python
def test():
    expected = textwrap.dedent(
        """\
        Generating answers...
        Evaluating answers...
        ...
        """
    )
    assert result == expected
```

tests/test_lab_evaluate.py
danmcp: I was following the pattern from other tests.  Didn't know it wasn't a good one.  Fixed!

tiran: In the beginning we used `unittest.TestCase`. When we moved to 100% pytest, I removed `unittest.TestCase` as parent class, but did not move all test methods to test functions. In pytest, test classes are not necessary but sometimes useful. In your case, you didn't need it and are now benefiting from 4 spaces less per line :)

tiran (approved):

tests/test_lab_evaluate.py
danmcp: Thanks for the suggestion!

src/instructlab/model/evaluate.py
tiran: Does the code work with `benchmark = None`?

danmcp: > Does the code work with `benchmark = None`?

It does not.  Benchmark must be set or the command will fall through and do nothing.

nathan-weinberg: Yes, if we do not set a benchmark then `ilab model evaulate` does not know which evaluation scheme to use ergo is unable to import the corresponding evaluation class

nathan-weinberg (approved):

tests/test_lab_evaluate.py
```diff
@@ -0,0 +1,466 @@
...
+                "model",
+                "evaluate",
+                "--benchmark",
+                "mmlu",
+                "--model",
+                "invalid",
+            ],
+        )
+        # TODO: This error could use some work
+        assert "is not a valid model" in str(result.exception)
```
alinaryan, line 388: Should you make this similiar to the `test_invalid_model_mt_bench` error?: `assert "Failed to determine backend:" in result.output`

alinaryan, line 388: Should you make this similiar to the `test_invalid_model_mt_bench` error?: `assert "Failed to determine backend:" in result.output`

tests/test_lab_evaluate.py
```diff
@@ -0,0 +1,466 @@
...
+                "rc",
+                "--base-branch",
+                "main",
+                "--taxonomy-path",
+                "invalid_taxonomy",
+            ],
+        )
+        launch_server_mock.assert_called_once()
+        # TODO: This error could use some work
+        assert "/invalid_taxonomy" in str(result.exception)
```
alinaryan, line 425: `"The taxonomy data is not in-line with what is expected" in str(result.exception)` or `"invalid taxonomy data" in str(result.exception)`

tests/test_lab_evaluate.py
danmcp: That would be ideal.  But the TODO is indicating the current error handling is returning an exception rather than nice output.

tests/test_lab_evaluate.py
danmcp: Not quite sure what's being stated here.  This is another case the error handling is returning an exception rather than nice output.  The TODO is noting improvement is needed

tests/test_lab_evaluate.py
alinaryan: I see, I thought you were looking for diff err suggestions :sweat_smile:

alinaryan (approved):

tests/test_lab_evaluate.py
danmcp: Ah, that makes sense now :). Yeah anything would be better than the stacktrace we are currently getting.  I was hoping to do followups after or as part of: https://github.com/instructlab/eval/pull/48

PR #1573, opened 2024-07-03
ci: add ilab model evaluate mmlu to basic-workflow-tests

alimaredia:
**Checklist:**

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -22,6 +22,7 @@ BACKEND="llama-cpp"
 HF_TOKEN=${HF_TOKEN:-}
 SDG_PIPELINE="simple"
 SKIP_TRAIN=${SKIP_TRAIN:-0}
+EVAL=0
 
 export GREP_COLORS='mt=1;33'
 BOLD='\033[1m'
@@ -93,6 +94,10 @@ test_download() {
     else
         step Downloading the default model
         ilab model download
+    fi
+
+    if [ "$EVAL" -eq 1 ]; then
+        step Downloading a model to use with evaluation
         ilab model download --repository instructlab/granite-7b-lab
     fi
 }
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -93,6 +93,7 @@ test_download() {
     else
         step Downloading the default model
         ilab model download
+        ilab model download --repository instructlab/granite-7b-lab
     fi
 }
 
@@ -193,6 +194,16 @@ test_convert() {
     ilab model convert
 }
 
+test_evaluate() {
+    task Evaluate the model
+
+    if [ "$EVAL" -eq 1 ]; then
+      export INSTRUCTLAB_EVAL_MMLU_MIN_TASKS=true
+      export HF_DATASETS_TRUST_REMOTE_CODE=true
+      ilab model evaluate --model models/instructlab/granite-7b-lab --benchmark mmlu
+    fi
+}
+
 test_exec() {
     # The list of actual tests to run through in workflow order
     test_smoke
@@ -249,6 +260,9 @@ test_exec() {
     task Stopping the ilab model serve
     step Kill ilab model serve $PID
     kill $PID
+
+    task Evaluating the output of ilab model train
+    test_evaluate
 }
 
 wait_for_server(){
@@ -268,6 +282,7 @@ wait_for_server(){
 usage() {
     echo "Usage: $0 [-m] [-h]"
     echo "  -m  Run minimal configuration (run quicker when you have no GPU)"
+    echo "  -e  Run model evaluation"
     echo "  -f  Run the fullsize training instead of --4-bit-quant"
     echo "  -F  Use the 'full' SDG pipeline instead of the default 'simple' pipeline"
     echo "  -g  Use the granite model"
@@ -278,10 +293,14 @@ usage() {
 
 # Process command line arguments
 task "Configuring ..."
-while getopts "cmMfFghv" opt; do
+while getopts "cemMfFghv" opt; do
     case $opt in
         c)
             # old option, don't fail if it's specified
+            ;;
+        e)
+            EVAL=1
+            step "Run model evaluation."
             ;;
         m)
             MINIMAL=1
--- a/.github/workflows/e2e-nvidia-a10g-x1.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x1.yml
@@ -128,7 +128,7 @@ jobs:
       - name: Run e2e test
         run: |
           . venv/bin/activate
-          ./scripts/basic-workflow-tests.sh -cmf
+          ./scripts/basic-workflow-tests.sh -cemf
 
       - name: Add comment to PR if the workflow failed
         if: failure() && steps.check_pr.outputs.is_pr == 'true'
--- a/docs/ci.md
+++ b/docs/ci.md
@@ -69,4 +69,4 @@ The project currently supports the usage of the following runners for the E2E jo
 | **Training** | legacy+Linux             |❌|❌|✅|❌|
 |              | legacy+Linux+4-bit-quant |✅|✅|❌|❌|
 |              | training-lib             |❌|❌|❌|❌|
-| **Eval**     | eval                     |❌|❌|❌|❌|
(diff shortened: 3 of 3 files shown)


scripts/basic-workflow-tests.sh
```diff
@@ -211,6 +219,8 @@ test_exec() {
     task Stopping the ilab model serve
     step Kill ilab model serve $PID
     kill $PID
+
```
leseb, line 222: task Test Evaluate

scripts/basic-workflow-tests.sh
```diff
@@ -163,6 +163,14 @@ test_convert() {
     ilab model convert
 }
 
+test_evaluate() {
+    task Evaluating a model
+    INSTRUCTLAB_EVAL_MMLU_MIN_TASKS=true 
+    HF_DATASETS_TRUST_REMOTE_CODE=true 
+    ilab model download --repository instructlab/granite-7b-lab
```
leseb, line 170: Should we add that step in the `test_download` function?

russellb: yeah, makes sense to me

russellb (changes requested):
Can you add a new command line arg that only runs this new functionality if enabled? On our smaller environments, I'm pretty sure this has to be off, right?

nathan-weinberg:
> On our smaller environments, I'm pretty sure this has to be off, right?

Correct, this can't run on the standard GitHub runner we use AFAIK

scripts/basic-workflow-tests.sh
```diff
@@ -163,6 +163,14 @@ test_convert() {
     ilab model convert
 }
 
+test_evaluate() {
+    task Evaluating a model
+    INSTRUCTLAB_EVAL_MMLU_MIN_TASKS=true 
+    HF_DATASETS_TRUST_REMOTE_CODE=true
```
nathan-weinberg, line 169: Per the Shellchecker
```suggestion
    export INSTRUCTLAB_EVAL_MMLU_MIN_TASKS=true 
    export HF_DATASETS_TRUST_REMOTE_CODE=true 
```

scripts/basic-workflow-tests.sh
russellb: Actually, wouldn't the user workflow be to run this against the output of `ilab model train` ?

nathan-weinberg: In theory, this can be used to evaluate any model (checkpoint or otherwise) - so it boils down to what we want to test here.

russellb:
Note that the location this was added gets skipped in all of the workflows right now.  This PR will fix that, though: https://github.com/instructlab/instructlab/pull/1580

scripts/basic-workflow-tests.sh
```diff
@@ -162,6 +163,16 @@ test_convert() {
     ilab model convert
 }
 
+test_evaluate() {
+    task Evaluating a model
+
+    if [ "$EVAL" -eq 0 ]; then
```
danmcp, line 169: Should this be -eq 1?

scripts/basic-workflow-tests.sh
```diff
@@ -208,11 +219,15 @@ test_exec() {
     task Stopping the ilab model serve
     step Kill ilab model serve $PID
     kill $PID
+
+    task Evaluating a model
```
danmcp, line 223: This seems duplicated in test_evaluate

scripts/basic-workflow-tests.sh
```diff
@@ -208,11 +219,15 @@ test_exec() {
...
+    test_evaluate
```
danmcp, line 224: Didn't this need to be moved up to take effect?

danmcp (changes requested):

scripts/basic-workflow-tests.sh
```diff
@@ -162,6 +163,16 @@ test_convert() {
     ilab model convert
 }
 
+test_evaluate() {
+    task Evaluating a model
+
+    if [ "$EVAL" -eq 0 ]; then
+      export INSTRUCTLAB_EVAL_MMLU_MIN_TASKS=true
+      export HF_DATASETS_TRUST_REMOTE_CODE=true
+      ilab model evaluate --model models/instructlab/granite-7b-lab --benchmark mmlu
```
russellb, line 172: Could you change it to test the output of `ilab model train` ? It seems like that would be more representative of an end-to-end user flow, right?

scripts/basic-workflow-tests.sh
russellb: thanks for remembering this comment from eariler, but it's good now as of https://github.com/instructlab/instructlab/pull/1580

It used to always exit before here. Now it will make it this far as long as `-f` (full training) is used. That does mean that `-e` without `-f` will not work. It's probably worth a check for that. I've been adding similar checks to `set_defaults`.

russellb (changes requested):

scripts/basic-workflow-tests.sh
alimaredia: Do you want that in this function or in `test_exec()` where the full workflow happens?

scripts/basic-workflow-tests.sh
```diff
@@ -93,6 +93,7 @@ test_download() {
     else
         step Downloading the default model
         ilab model download
+        ilab model download --repository instructlab/granite-7b-lab
```
russellb, line 96: perhaps conditional on whether eval is turned on?

scripts/basic-workflow-tests.sh
russellb: in here

russellb:
I started a test of this change over here https://github.com/russellb/ilab-runner/actions/workflows/e2e-nvidia-a10g-x1-pr-1573.yml

russellb:
> I started a test of this change over here https://github.com/russellb/ilab-runner/actions/workflows/e2e-nvidia-a10g-x1-pr-1573.yml

this passed!

I'm OK if you want to just file an issue on my larger comment (testing the output of train instead of a model we download). I don't want to block on that.

scripts/basic-workflow-tests.sh
russellb: I pushed a commit to fix this

russellb (dismissed):
I fixed the script issue that I was blocking on and added the doc update to reflect which job now tests eval.

I'm good with this (assuming a follow-up issue to look into testing the model that comes out of train)

alimaredia:
@russellb The evaluation I'm adding here is post training stand alone model evaluation. 

The intercheckpoint evaluation I think you're talking about in your comments about "testing the output of `ilab model train`" happens within the training code by the eval library when the `--e2e` flag is added to `ilab model train`. That's something James is actively working on and isn't merged yet (https://github.com/instructlab/instructlab/pull/1466).

I don't think we'll need a follow up issue then right?

alimaredia:
Oh I just realized you're probably talking about testing the final model output of the `ilab model train` command? 

If that's the case, I think we hold off on that until the e2e PR is merged so that we know where that model will be and what it will be called. Right now `ilab model train` is still spitting out a `.gguf` model right? We can't even run MMLU on that.

russellb:
> Oh I just realized you're probably talking about testing the final model output of the `ilab model train` command?

Yes.

> If that's the case, I think we do file an issue and hold off on that until the training team can give us more information on where the new model will be and what it will be called. Right now `ilab model train` is still spitting out a `.gguf` model right? We can't even run MMLU on that.

OK ... but that's THE use case for eval. Running eval on a random model is not super useful.  What we need is eval of a model out of train. It's fine if that's still WIP though, that's why I want a follow-up issue

danmcp:
What's the plan on getting the e2e to pass on this PR?

russellb:
> What's the plan on getting the e2e to pass on this PR?

ouch, sorry, that's my fault. Thanks for pointing it out.

alimaredia:
@danmcp Seeing as we got a clean run in for what's there I'd like to merge this PR. 

I've opened issues https://github.com/instructlab/instructlab/issues/1643 and https://github.com/instructlab/instructlab/issues/1642 to track the future work items on this PR and assigned @alinaryan and myself to them.

Is there anything else needed here before we can merge?

danmcp (approved):
Approving with the followup issues in place.  Thanks!

russellb (approved):

PR #1579, opened 2024-07-03
Enable more Ruff checkers and fixers

tiran:
Ruff can not just detect and report code violations. It can automatically fix a large amount of code violations, too. This change enables most checkers and addresses their code violations:

- Update minimal Python version to 3.10
- Remove unnecesary `zip`
- Remove some unused imports
- Simplify some code
- Avoid `l`, which can be easily confused with `1`.
- Use full path regexps for pylint's `ignore-paths`

This change only contains uncontroversial and simple fixes. I will enable more fixers in the near future.

**Checklist:**

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -48,13 +48,15 @@ fail-under=10
 #from-stdin=
 
 # Files or directories to be skipped. They should be base names, not paths.
-ignore=CVS,protobufs,train,llamacpp,mlx_explore
+ignore=CVS
 
 # Add files or directories matching the regular expressions patterns to the
 # ignore-list. The regex matches against paths and can be in Posix or Windows
 # format. Because '\\' represents the directory delimiter on Windows systems,
 # it can't be used as an escape character.
-ignore-paths=
+ignore-paths=^src/instructlab/llamacpp/.*$,
+    ^src/instructlab/mlx_explore/.*$,
+    ^src/instructlab/train/lora_mlx/.*$,
 
 # Files or directories matching the regular expression patterns are skipped.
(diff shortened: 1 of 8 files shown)


src/instructlab/llamacpp/llamacpp_convert_to_gguf.py
```diff
@@ -48,6 +48,9 @@
     # Standard
     from typing import TypeAlias
 
+# ruff: noqa: SIM115
```
booxter, line 51: I was told elsewhere (by @bjhargrave) that this file (and other files that originate from other repos) should not be "fixed" by linters and e.g. mypy checker. (I disagree btw, but that's beyond the point.) If that's so, this change should probably also *not* happen. (mlx_explore and lora_mlx were also mentioned, I believe.)

bjhargrave: I am just conscious about excess modifications to 3rd party source we copied in. Sure, add features and fix things that break us. But modifying the code to pass our views on code linting seems a bit over the top and makes it harder to merge in a change from upstream if we want.

tiran: Makes sense. I'm going to exclude the files from Ruff for now.

We should consider to make vendored files into a `instructlab.vendor` package, so it is obvious that the files were vendored from 3rd party projects.

booxter:
lgtm

bjhargrave (approved):

leseb (approved):

PR #1585, opened 2024-07-03
docs: Update docs to reflect current state of E2E jobs

russellb:
This week we added some new E2E test workflows that use ephemeral test
runners on AWS. These different workflows run on different instances
types and test different features.

Create some tables to keep track of our workflows and what
configurations they are testing. This should help us identify areas of
overlap, as well as gaps we need to fill.

Co-authored-by: Nathan Weinberg <nweinber@redhat.com>

----

Rendered doc: https://github.com/russellb/instructlab/blob/e2e-ci-doc-update/docs/ci.md


leseb (dismissed):

docs/ci.md
```diff
@@ -20,6 +38,34 @@ run on normal runners.
...
+
+* [GitHub built-in GPU runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#specifications-for-gpu-larger-runners) -- referred to as `ubuntu-gpu` in our workflow files Only `e2e.yml` uses this runner.
+* Ephemeral GitHub runners launched on demand on AWS. Most workflows work this way, granting us access to a wider variety of infrastructure at lower cost.
+
+### E2E Workflows
+
+| File | Runner Host |Instance Type | GPU Type | OS |
+| --- | --- | --- | --- | --- |
+| [`e2e.yml`](https://github.com/instructlab/instructlab/blob/main/.github/workflows/e2e.yml) | GitHub | N/A | 1 x NVIDIA Tesla T4 w/ 16 GB VRAM | Ubuntu |
+| [`e2e-nvidia-t4-x1.yml`](https://github.com/instructlab/instructlab/blob/main/.github/workflows/e2e-nvidia-t4-x1.yml) | AWS | [`g4dn.2xlarge`](https://aws.amazon.com/ec2/instance-types/g4/) | 1 x NVIDIA Tesla T4 w/ 16 GB VRAM | CentOS Stream 9 |
```
nathan-weinberg, line 57: I wonder if we might want to deprecate the `e2e.yml` job in favor of the `e2e-nvidia-t4-x1.yml` job since it seems to be the same specs but a lower cost

nathan-weinberg (approved):

docs/ci.md
russellb: Yes, absolutely.  Two things

1. I wanted to make sure it was stable, thought these jobs seem very reliable so far.
2. Since this job uses secrets (the AWS and github keys), it's not as simple as just changing the workflow to be triggered on PRs. GitHub won't allow it for security reasons. There are at least a couple of options to consider ..

danmcp (approved):

PR #1608, opened 2024-07-05
fix: Change default model setting DEFAULT_MODEL_REPO to be consistent

### LINKED ISSUE #1607: Train and test commands do not use the default model
**Describe the bug**
When using the default model `merlinite-7b-lab`, the `ilab model train` and `ilab model test` commands are not using the default model but `granite-7b-lab` instead.

**To Reproduce**
1. Follow the steps in the README to generate data (use the default settings)
2. Train the model using the generated data (use the default settings)
3. You should see `instructlab-granite-7b-lab` and `instructlab-granite-7b-lab-mlx-q` directories created instead of `instructlab-merlinite-7b-lab` and `instructlab-merlinite-7b-lab-mlx-q` directories
4. Test the trained model (use the default settings)
5. It should fail with an error of unable to find the HF repository

**Expected behavior**
Would expect the default model to be used for all commands and train and test to work as expected.

**Screenshots**
n/a

 - Hardware Specs: [Apple M1 Pro Chip, 32 GB Memory]
 - OS Version: [Mac OS 14.5]
 - Python Version: [3.11.7]
 - InstructLab Version: [main]

**Additional context**

hickeyma:
Change default setting `DEFAULT_MODEL_REPO` to correspond to the other default model settings like `DEFAULT_MODEL_OLD`, `DEFAULT_MODEL` and `DEFAULT_MODEL_PATH`. This means consistency across the settings and consistency across the CLI commands.

Resolves #1607

**Checklist:**


russellb (approved):
agree that the consistency makes sense for the defaults

russellb:
Added @cdoern since I'm guessing this came in via his training integration PR

hickeyma:
Putting on hold while I add a test

hickeyma:
Test added. Hold released. @russellb Do you mind reviewing again?

russellb:
> given this was recently added for the new training and eval, It would be great if we could preserve granite as a model repo to pass to those.
> 
> Can we maybe change `DEFAULT_MODEL_REPO` to merlinite but make `GRANITE_MODEL_REPO`? or something

Where would the GRANITE constant be used?

scripts/basic-workflow-tests.sh
```diff
@@ -162,6 +167,23 @@ test_convert() {
...
 
+test_config_value() {
+    task "Check configuration value '$1' in property '$3' of section '$2' of '$4'"
+
+    VALUE=$1
+    SECTION=$2
+    PROPERTY=$3
+    CONFIG_FILE=$4
+
+    PROPERTY_VALUE=$(grep -A50 "$SECTION"':' "$CONFIG_FILE" | grep "$PROPERTY"':' | xargs)
```
russellb, line 178: How about actually parsing the yaml? That seems like it'd be a bit more robust. `yq` would do the trick.

russellb (commented):
appreciate the test addition!  added one implementation suggestion

scripts/basic-workflow-tests.sh
```diff
@@ -52,8 +52,13 @@ test_init() {
     task Initializing ilab
     [ -f config.yaml ] || ilab config init --non-interactive
 
-    step Checking config.yaml
-    grep merlinite config.yaml
+    step Checking model configuration values
```
leseb, line 55: I believe all these tests should be added to `scripts/functional-tests.sh` instead. This file is used for end-to-end testing.

leseb (changes requested):
I

scripts/basic-workflow-tests.sh
hickeyma: I thought about using `yq` alright but I didn't want to add another tool.  If people think its a better implementation, I can do that.

scripts/basic-workflow-tests.sh
hickeyma: That seems like a good idea. What do you think @russellb?

hickeyma:
> given this was recently added for the new training and eval, It would be great if we could preserve granite as a model repo to pass to those.
Can we maybe change DEFAULT_MODEL_REPO to merlinite but make GRANITE_MODEL_REPO? or something

Will Granite model run e2e on accessible hardware like Mac M1 32G? If so, then we could do that. However, would changing the default break users expectations or is that ok to do?

scripts/basic-workflow-tests.sh
leseb: @hickeyma thanks for considering it and happy to see more tests :D

scripts/basic-workflow-tests.sh
leseb: `yq` is part of the runner's Ubuntu image https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#tools. `scripts/functional-tests.sh` uses that image.

scripts/basic-workflow-tests.sh
russellb: yes, agree with the move

hickeyma: Ok, updated 59449375652641afde4614a6ee19a2d3823ce816 to move tests to functional testing and using `yk` instead of `grep`.

scripts/basic-workflow-tests.sh
hickeyma: Ok, now using `yq` instead.

russellb (dismissed):
lgtm, thanks!

scripts/functional-tests.sh
```diff
@@ -374,10 +374,37 @@ test_model_print(){
...
+    test_config_value_isset "merlinite" "serve" "model_path" "config.yaml"
+    test_config_value_isset "merlinite" "train.train_args" "model_path" "config.yaml"
+}
+
 ########
 # MAIN #
 ########
 # call cleanup in-between each test so they can run without conflicting with the server/chat process
+test_config_model_defaults
+cleanup
```
leseb, line 407: You don't need to call `cleanup` here. It won't do anything.

scripts/functional-tests.sh
```diff
@@ -374,10 +374,37 @@ test_model_print(){
...
+test_config_value_isset() {
+    VALUE=$1
+    SECTION=$2
+    PROPERTY=$3
+    CONFIG_FILE=$4
+    echo "Check configuration value '$VALUE' in property '$PROPERTY' of section '$SECTION' of '$CONFIG_FILE'"
+
+    PROPERTY_VALUE=$(yq ".$SECTION.$PROPERTY" "$CONFIG_FILE" | xargs)
+    if [[ $PROPERTY_VALUE != *"$VALUE"* ]]; then
+      echo "'$VALUE' not set in property '$PROPERTY_VALUE' of the '$SECTION' section of configuration '$CONFIG_FILE'"
```
leseb, line 386: This should use a 4 line indent like the rest of the script does.

scripts/functional-tests.sh
```diff
@@ -374,10 +374,37 @@ test_model_print(){
...
+      exit 1
+    else
+      echo "Value '$VALUE' is set"
+    fi
+}
+
+test_config_model_defaults(){
```
leseb, line 393: I just realized that now, sorry, but these seem more like unit tests than functional tests, in my opinion. If you feel re-writing I think it'd be better, otherwise I won't block for this.

hickeyma: I think this makes more sense. Thanks @leseb for the suggestion. Tests now moved to unit tests.

tests/test_config.py
```diff
@@ -16,20 +16,20 @@ def _init_tmpdir(self, tmpdir):
...
         assert cfg.chat is not None
-        assert cfg.chat.model == "models/merlinite-7b-lab-Q4_K_M.gguf"
         assert not cfg.chat.vi_mode
         assert cfg.chat.visible_overflow
         assert cfg.chat.context == "default"
         assert cfg.chat.session is None
         assert cfg.chat.logs_dir == "data/chatlogs"
         assert not cfg.chat.greedy_mode
 
+        assert cfg.evaluate is not None
```
leseb, line 27: Thanks for reorganizing this section.

leseb (approved):

cdoern (approved):
we can change it to merlinite for now, but I think this is a larger discussion we need to have about our defaults

russellb:
> we can change it to merlinite for now, but I think this is a larger discussion we need to have about our defaults

that's fine -- the consistency here is what makes sense to me

hickeyma:
> we can change it to merlinite for now, but I think this is a larger discussion we need to have about our defaults

Sure, we need to align this better. FYI, Its always been merlinite by default, so that why I changed the one default from granite to merlinite, as it caused a bug.

PR #1611, opened 2024-07-05
refactor: move lama-cpp code from backends.py to llama_cpp.py

makelinux:
The code is moved with minimal changes.

Consolidate lama-cpp code in llama_cpp.py,
making backends.py backend implementation agnostic.

For check_api_base() see #1596


makelinux:
see #1612

PR #1604, opened 2024-07-05
fix: use modul-level logger instead of passing loggers

leseb:
This is a continuation of the work initiated by
6a92f5bb498df6ecff2902431ecc35f6c98da962 but for the backend files. The code uses the recommended approach for logging.  Every module has its own logging instance.  Logging is configured in the root logger and sub-loggers are automatically inheriting all settings.

Signed-off-by: Sébastien Han <seb@redhat.com>

**Checklist:**

- [ ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #1612, opened 2024-07-05
refactor: move lama-cpp code from backends.py to llama_cpp.py

makelinux:
The code is moved with minimal changes.

Consolidate lama-cpp code in llama_cpp.py,
making backends.py backend implementation agnostic.

for check_api_base()  see #1596

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -258,7 +258,6 @@ def ensure_server(
     http_client=None,
     host="localhost",
     port=8000,
-    queue=None,
     background=True,
     server_process_func=None,
 ) -> Tuple[
@@ -275,7 +274,6 @@ def ensure_server(
 
     host_port = f"{host}:{port}"
     temp_api_base = get_api_base(host_port)
-    llama_cpp_server_process = None
     vllm_server_process = None
 
     if backend == VLLM:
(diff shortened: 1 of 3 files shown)


src/instructlab/model/backends/backends.py
```diff
@@ -253,39 +251,7 @@ def ensure_server(
                 raise ServerException(
                     f"vLLM failed to start up in {vllm_startup_timeout} seconds"
                 )
-
-        elif backend == LLAMA_CPP:
```
danmcp, line 257: Why is llama-cpp moving and not vllm?

makelinux: The next separated PR will move vllm.

makelinux:
@alimaredia , would you like to review please?

leseb:
Let's have this before https://github.com/instructlab/instructlab/pull/1604 makes it (I think it should merge today).

src/instructlab/model/backends/llama_cpp.py
```diff
@@ -93,18 +95,49 @@ def create_server_process(self, port: int) -> multiprocessing.Process:
...
-            llama_cpp_server_process, _, api_base = ensure_server(
+            # TODO: replace ensure_server() with future check_api_base()
+            self.process, _, temp_api_base = ensure_server(
                 logger=self.logger,
                 backend=LLAMA_CPP,
                 api_base=self.api_base,
                 http_client=http_client,
                 host=self.host,
                 port=self.port,
-                queue=self.queue,
```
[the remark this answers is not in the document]
makelinux: because it became unused parameter.

makelinux:
@leseb , @russellb , @tiran , would you like review please?

src/instructlab/model/backends/llama_cpp.py
```diff
@@ -89,19 +90,37 @@ def create_server_process(self, port: int) -> multiprocessing.Process:
...
+
+            # in case the server takes some time to fail we wait a bit
+            logger.debug("Waiting for the server to start...")
+            count = 0
+            while self.process.is_alive():
+                sleep(0.1)
+                if check_api_base(self.api_base, http_client):
+                    break
+                if count > 50:
+                    logger.error("failed to reach the API server")
```
danmcp, line 111: Start with a capital letter to be consistent

src/instructlab/model/backends/llama_cpp.py
```diff
@@ -89,19 +90,37 @@ def create_server_process(self, port: int) -> multiprocessing.Process:
...
+            self.api_base = str(get_api_base(f"{self.host}:{self.port}"))
+            logger.debug(f"Starting a temporary server at {self.api_base}...")
+            self.process = self.create_server_process(self.port)
+            self.process.start()
+
+            # in case the server takes some time to fail we wait a bit
+            logger.debug("Waiting for the server to start...")
+            count = 0
+            while self.process.is_alive():
+                sleep(0.1)
```
danmcp, line 107: I realize it's not new code, but why is this sleep so short?

russellb: Usually when I see this it's to prevent a busy loop, or to ensure the process yields to others each time around instead of just eating up the CPU. In this case, it would be trying to connect to the API server as fast as possible for 50 tries without it. Now it's doing it 10 times a second (a little more than a second in practice).

That's still more aggressive than necessary, though, IMO.

danmcp: I was curious so I dug in a little further.  The list_models call can take a few seconds respond while the server is coming up.

Examples printing time.time() in the loop:
```
1720913810.889125
1720913813.3796103
```

```
1720913780.4392679
1720913783.166862
```

So the sleep isn't really doing the heavy lifting of the waiting in this case.  Makes sense why it was made so short.


@JamesKunstle ^^ related to a conversation we were having around vllm

leseb (approved):

danmcp (approved):

leseb:
For some reason, GitHub refuses to remove pending reviews. So I'm going to merge manually given that the PR has two approvals and CI is green, thanks!

PR #1613, opened 2024-07-05
feat: Configure vLLM stdout/stderr suppression when running detached

alimaredia:
Also add more debug logging to vLLM server detection loop.

**Checklist:**

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -241,6 +241,7 @@ def launch_server(
     max_workers,
     gpus,
     backend,
+    enable_serving_output,
 ) -> tuple:
     # pylint: disable=import-outside-toplevel
     # First Party
@@ -277,7 +278,9 @@ def launch_server(
     backend_instance = backends.select_backend(eval_serve, backend)
     try:
         # http_client is handling tls params
-        api_base = backend_instance.run_detached(http_client(ctx.params))
+        api_base = backend_instance.run_detached(
+            http_client(ctx.params), background=enable_serving_output
+        )
     except Exception as exc:
         click.secho(f"Failed to start server: {exc}", fg="red")
         raise click.exceptions.Exit(1)
@@ -405,6 +408,11 @@ def launch_server(
     default="",
     help="TLS client certificate password for model serving.",
 )
+@click.option(
+    "--enable-serving-output",
+    is_flag=True,
+    help="Print serving engine logs.",
+)
 @click.pass_context
 @display_params
 def evaluate(
@@ -428,6 +436,7 @@ def evaluate(
     tls_client_cert,  # pylint: disable=unused-argument
     tls_client_key,  # pylint: disable=unused-argument
     tls_client_passwd,  # pylint: disable=unused-argument
+    enable_serving_output,
 ):
     # get appropriate evaluator class from Eval lib
     evaluator = get_evaluator(
@@ -461,6 +470,7 @@ def evaluate(
                     max_workers,
                     gpus,
                     backend,
+                    enable_serving_output,
                 )
                 evaluator.gen_answers(api_base)
             finally:
@@ -476,6 +486,7 @@ def evaluate(
                     max_workers,
                     gpus,
                     backend,
+                    enable_serving_output,
                 )
                 overall_score, qa_pairs, turn_scores, error_rate = (
                     evaluator.judge_answers(api_base)
@@ -535,6 +546,7 @@ def evaluate(
                         max_workers,
                         gpus,
                         backend,
+                        enable_serving_output,
                     )
                     evaluator.gen_answers(api_base)
                 finally:
@@ -550,6 +562,7 @@ def evaluate(
                     max_workers,
                     gpus,
                     backend,
+                    enable_serving_output,
                 )
                 for i, evaluator in enumerate(evaluators):
                     branch = branches[i]
--- a/src/instructlab/model/serve.py
+++ b/src/instructlab/model/serve.py
@@ -144,7 +144,7 @@ def serve(
         raise click.exceptions.Exit(1)
 
     try:
-        # Run the llama server
+        # Run the backend server
         backend_instance.run()
 
     except ServerException as exc:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@
    mmlu-branch) which are integrated with the ilab workflow to evaluate new skills and knowledge.
    Includes --gpus option for specifying number of gpus to utilize when serving models for
    evaluation (currently applicable for vLLM only).  Also includes --merge-system-user-message
-   flag to enable Mistral based judge models.
+   flag to enable Mistral based judge models and a --enable-serving-output flag that
+   configures whether the output of the model serving backend is suppressed.
 * The `ilab` command now accepts a `-v` / `--verbose` option to enable debug logging.
   `ilab -vv` or `ilab --verbose --verbose` enables more verbose debug logging.
 * `ilab model test` generic support
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -78,7 +78,9 @@ class BackendServer(abc.ABC):
         """Run serving backend in foreground (ilab model serve)"""
 
     @abc.abstractmethod
-    def run_detached(self, http_client: httpx.Client | None = None) -> str:
+    def run_detached(
+        self, http_client: httpx.Client | None = None, background: bool = True
+    ) -> str:
         """Run serving backend in background ('ilab model chat' when server is not running)"""
 
     @abc.abstractmethod
@@ -227,6 +229,7 @@ def ensure_server(
     host="localhost",
     port=8000,
     queue=None,
+    background=True,
     server_process_func=None,
 ) -> Tuple[
     Optional[multiprocessing.Process], Optional[subprocess.Popen], Optional[str]
@@ -248,7 +251,7 @@ def ensure_server(
 
         if backend == VLLM:
             # TODO: resolve how the hostname is getting passed around the class and this function
-            vllm_server_process = server_process_func(port)
+            vllm_server_process = server_process_func(port, background)
             logger.info(f"Starting a temporary vLLM server at {temp_api_base}")
             count = 0
             # TODO should this be configurable?
--- a/src/instructlab/model/backends/llama_cpp.py
+++ b/src/instructlab/model/backends/llama_cpp.py
@@ -86,7 +86,9 @@ class Server(BackendServer):
 
         return server_process
 
-    def run_detached(self, http_client: httpx.Client | None = None) -> str:
+    def run_detached(
+        self, http_client: httpx.Client | None = None, background: bool = True
+    ) -> str:
         try:
             llama_cpp_server_process, _, api_base = ensure_server(
                 backend=LLAMA_CPP,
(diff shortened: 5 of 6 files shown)


src/instructlab/model/evaluate.py
```diff
@@ -333,6 +333,12 @@ def launch_server(
     default="",
     help="TLS client certificate password for model serving.",
 )
+@click.option(
+    "--background",
```
danmcp, line 337: This is really:

--background-serving
or
-- verbose-serving
or
--enable-serving-output

danmcp (changes requested):

src/instructlab/model/evaluate.py
```diff
@@ -333,6 +333,12 @@ def launch_server(
...
+    type=click.BOOL,
+    default=False,
```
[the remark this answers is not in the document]
danmcp: Leaning toward --enable-serving-output as the option and False as the default.  And then something in the help that refers to this option as being helpful for debugging.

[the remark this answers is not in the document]
alimaredia: I need to spend some time and see if llama-cpp's output can be limited in the same way as vLLM's, since everything in this PR only really applies to vLLM and not llama.

src/instructlab/model/evaluate.py
```diff
@@ -333,6 +333,12 @@ def launch_server(
...
+    help="Run vLLM in the background.",
```
leseb, line 340: I don't think this should be vLLM specific, maybe it's a wording issue and we can make this backend agnostic or reflect it's only implemented for a specific backend only.

src/instructlab/model/serve.py
```diff
@@ -64,6 +64,11 @@
         " Example: --vllm-args='--dtype=auto' --vllm-args='--enable-lora'"
     ),
 )
+@click.option(
+    "--vllm-background",
+    type=click.BOOL,
+    help="vLLM stdout and sterr is sent to /dev/null",
```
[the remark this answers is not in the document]
leseb: What's the use case for sending the vLLM stream to /dev/null? Where is the requirement coming from?

leseb (changes requested):
It's a bit unclear to me why this is needed, so I'm seeking some clarification please. Thanks!

src/instructlab/model/evaluate.py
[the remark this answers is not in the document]
leseb: Why does this have to be a flag when the backend runs behind a main command like chat or eval?

src/instructlab/model/evaluate.py
JamesKunstle: I like --verbose-serving personally. I want to know what's going on with vLLM or llama.cpp since they might take a while to boot up, but I really only want it as a debug option.

src/instructlab/model/evaluate.py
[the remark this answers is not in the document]
JamesKunstle: This was a request I made because vLLM has useful debug info logging for that isn't caught by ensure_server or by sub-process lifespan.

src/instructlab/model/evaluate.py
JamesKunstle: ++ should be for llama.cpp as well.

src/instructlab/model/evaluate.py
```diff
@@ -333,6 +333,12 @@ def launch_server(
     default="",
     help="TLS client certificate password for model serving.",
 )
+@click.option(
+    "--background",
+    type=click.BOOL,
```
danmcp, line 338: Should this be an is_flag=True?

leseb: It should be True by default I think so that we suppress stdout/stderr.

src/instructlab/model/evaluate.py
leseb: I prefer `--enable-serving-output` because it accurately reflects that the flag enables the output of the serving engine. To me, `--verbose-serving` implies that the serving output is already present and will become more verbose if the flag is turned on.

leseb:
@JamesKunstle @danmcp @alimaredia  I wasn't aware that the output was being suppressed—this is new code to me! The request makes sense, and here are my thoughts:

- When running through `ilab model serve`, let the backend print stdout/stderr. We don't need a flag to control this, it's the normal behavior of a server, if the users want to suppress anything, they can use redirection.
- When running through the temp server (`ilab model chat` or eval, and `ilab model serve` is not running), suppress stdout/stderr to avoid "polluting" the output.
  - We can introduce a flag (as proposed) to surface those logs.
  - Another option is to log to a file and not display anything in the eval output. (serve has a log-to-file option already)

Am I missing anything? I'd like to hear your thoughts as well. Thanks

src/instructlab/model/evaluate.py
danmcp: Agree that we should suppress by default.

If we do pick another name, like "--verbose-serving" or "--enable-serving-output",  it does flip the desired flag default.

src/instructlab/model/evaluate.py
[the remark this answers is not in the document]
alimaredia: @leseb Though this is inherently a property of serving, we wanted this as a flag for eval for debugging purposes.

src/instructlab/model/evaluate.py
alimaredia: I'm going to go with `--enable-serving-output` for the flag for evaluate.

alimaredia:
@leseb 

> @JamesKunstle @danmcp @alimaredia I wasn't aware that the output was being suppressed—this is new code to me! The request makes sense, and here are my thoughts:
> 
>     * When running through `ilab model serve`, let the backend print stdout/stderr. We don't need a flag to control this, it's the normal behavior of a server, if the users want to suppress anything, they can use redirection.
> 
>     * When running through the temp server (`ilab model chat` or eval, and `ilab model serve` is not running), suppress stdout/stderr to avoid "polluting" the output.
>       
>       * We can introduce a flag (as proposed) to surface those logs.
>       * Another option is to log to a file and not display anything in the eval output. (serve has a log-to-file option already)
> 
> 
> Am I missing anything? I'd like to hear your thoughts as well. Thanks

The behavior you're describing is close to what existed before this PR. The only thing this PR adds is exposing the ability for the user to have the serving in the foreground or the background.

Looking at the comments it seems like for the cases where `ilab` is starting up a background vLLM server, we want **don't want serving output by default**. That's the opposite of what this PR is intending as of this comment, but this was just my first version of this PR, and it was just to get a discussion going around it. 

These commands where this default behavior is happening include: `ilab model {evaluate, train, chat}` and `ilab data generate`. For `ilab model serve` the output will go to stdout and stderr as usual.

The one thing I didn't see working was the log-to-file option for `ilab model serve`. The `ilab` log output got sent there but the vLLM output did not since it's a separate process. I've asked some vLLM contributors about if it logs to a file.

CC @JamesKunstle @danmcp

danmcp:
I generally agree with @leseb's take as well. Two points to add:
- chat may be different than train and evaluate.  I am not sure there is ever a case where output serving content to the screen is desired.  It would truly be a debug option there.  And that can already be done with starting the server separately right?
- To log or not to log:  My current vote would be to note use a log file.  It would bring up another set of issues around log file maintenance (rotation, etc.) that would probably be better handled by another mechanism (Ex: systemd).

src/instructlab/model/serve.py
```diff
@@ -76,6 +76,11 @@
         " Example: --vllm-args='--dtype=auto' --vllm-args='--enable-lora'"
     ),
 )
+@click.option(
+    "--vllm-background",
```
alimaredia, line 80: @danmcp @leseb what do you think of the name for flag for `ilab model serve`? This was just my first attempt at a name.

alimaredia: Also do you think both this that the default should be drawn from the config like the `max-ctx-size` option is above? The default in that case would be set to true, which is not what we want.

danmcp: > @danmcp @leseb what do you think of the name for flag for `ilab model serve`? This was just my first attempt at a name.

My personal preference is to not put vllm in the name.  The intention being it would apply to any serving where background vs foreground was an option.  The help can explain which ones it applies to.



> Also do you think both this that the default should be drawn from the config like the `max-ctx-size` option is above? The default in that case would be set to true, which is not what we want.

My preference would be this is not a config option.  It's an option for what the command should show you vs config about a user's environment/model/hardware.  Meaning it can easily change from use case to use case.  So having a default in config can be confusing.  For example, if my config is set to false and I tell someone else to run it expecting particular output, they may experience something different with their config.  So the rule I am suggesting is command behaviors are not config.  You could for example write a command as ilab model background-serve (not suggesting we do this) and is would make logical sense.

src/instructlab/model/evaluate.py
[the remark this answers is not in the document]
leseb: I'm ok with that @alimaredia

src/instructlab/configuration.py
```diff
@@ -249,6 +249,7 @@ class _serve_vllm(BaseModel):
 
     # arguments to pass into vllm process
     vllm_args: list[str] | None = None
+    background: bool = True
```
leseb, line 252:
```suggestion
    enable_serving_output: bool = False
```

To match the CLI flag?

src/instructlab/model/evaluate.py
leseb: I think this is sorted out now, resolving. The flag is `enable-serving-output` and default to False :)

src/instructlab/model/evaluate.py
```diff
@@ -405,6 +406,12 @@ def launch_server(
     default="",
     help="TLS client certificate password for model serving.",
 )
+@click.option(
+    "--enable-serving-output",
+    type=click.BOOL,
+    default=False,
+    help="Run vLLM in the background.",
```
leseb, line 413:
```suggestion
    help="Print serving engine logs.",
```

How about this?

alimaredia:
@danmcp @leseb I removed the flag from serve.py. I wanted to make the PR more focused since the use cases behind this change were all for when vLLM was started up as a temp server for other ilab commands like evaluate and train.

For now I think it's fine we leave the output of `ilab model serve` as is. If we want to have the flag that suppresses the output, it's my preference we do it in a follow up PR.

src/instructlab/model/serve.py
```diff
@@ -141,7 +141,7 @@ def serve(
         raise click.exceptions.Exit(1)
 
     try:
-        # Run the llama server
+        # Run the backend server
```
alimaredia, line 144: This line change was included in my previous changes to serve.py. Though I undid all of those changes, I decided to leave this in because it was already done.

src/instructlab/model/serve.py
alimaredia: You make a great point about this not being a config option. I've scaled down the size and scope of the PR and in the process of doing so I removed the config option.

src/instructlab/model/evaluate.py
leseb: @alimaredia ?

leseb:
@alimaredia almost there, the linters are complaining. Thanks

src/instructlab/model/evaluate.py
leseb: Actually, I thought this was addressed already but this should really be `is_flag=True` and not `type=click.BOOL`.

leseb (approved):

danmcp (approved):

PR #1625, opened 2024-07-08
e2e: Add script option to skip training

russellb:
I'm trying to bootstrap some test workflows using larger AWS instance
types. On these instances with multiple GPUs, it's not very useful to
test the legacy training code path, because it doesn't support using
multiple GPUs. On these instances, we want to test the new training
path using the training library. Integrating that is a work in
progress (PR #1557).

This is just a temporary measure that we can remove once the new
training code path is supported in this script.

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -21,6 +21,7 @@ FULLTRAIN=0
 BACKEND="llama-cpp"
 HF_TOKEN=${HF_TOKEN:-}
 SDG_PIPELINE="simple"
+SKIP_TRAIN=${SKIP_TRAIN:-0}
 
 export GREP_COLORS='mt=1;33'
 BOLD='\033[1m'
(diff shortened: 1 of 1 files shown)


scripts/basic-workflow-tests.sh
```diff
@@ -262,12 +272,13 @@ usage() {
...
     echo "  -v  Use the vLLM backend for serving"
     echo "  -M  Use the mixtral model (4-bit quantized)"
+    echo "  -s  Skip training - halt after data generation"
     echo "  -h  Show this help text"
 }
 
 # Process command line arguments
 task "Configuring ..."
-while getopts "cmMfFghv" opt; do
+while getopts "cmMfFghsv" opt; do
```
leseb, line 281: I'm trying to minimize the change since it's temporary, can we just call the script with `SKIP_TRAIN=1 ./script` instead of adding another flag?

russellb: sure

russellb: done!

leseb (dismissed):

scripts/basic-workflow-tests.sh
```diff
@@ -216,6 +217,15 @@ test_exec() {
     step Kill ilab model serve $PID
     kill $PID
 
+    if [ "$SKIP_TRAIN" -eq 1 ]; then
+        # TODO - Drop this later.
+        # This is only a temporary measure while we bootstrap different CI workflows.
+        # There are some larger environments where it only makes sense to test the new
+        # training workflow using the training library, but that is not yet integrated
+        # here. Skip training for those environments for now.
+        return
```
leseb, line 226: Actually, would you mind logging that we return here? (add some value for debugging I guess...)

russellb: added a msg, thanks!

leseb (approved):

tiran (approved):

PR #1591, opened 2024-07-04
Improve config default lookup and `--help` message

tiran:
Introduce a `click.Option` subclass that renders help messages with defaults from the context's `default_map` and shows a hint to the config setting name. This simplifies the code and avoids code duplication.

The implementation can later be extended to show dynamic defaults from the user's config file instead of static defaults from the default config. The new `ConfigOption` can lookup defaults from nested settings. There is no need to flatten the configuration any more.

## Simple example `ilab model serve`

Code:
```python
@click.option(
    "--model-path",
    type=click.Path(),
    cls=ConfigOption,
    help="Path to the model used during generation.",
)
```

Output:
```
Options:
  --model-path PATH           Path to the model used during generation.
                              [default: models/merlinite-7b-lab-Q4_K_M.gguf;
                              config: 'serve.model_path']
```

## Complex example with nested lookup in `ilab model evaluate`

```python
@click.option(
    "--judge-model",
    type=click.STRING,
    cls=clickext.ConfigOption,
    config_sections="mt_bench",
    help="Model to be used as a judge for running mt_bench or mt_bench_branch - can be a local path or the name of a Hugging Face repository",
)
```

Output:
```
$ ilab model evaluate --help
Options:
...
  --judge-model TEXT              Model to be used as a judge for running
                                  mt_bench or mt_bench_branch - can be a local
                                  path or the name of a Hugging Face
                                  repository  [default: prometheus-
                                  eval/prometheus-8x7b-v2.0; config:
                                  'evaluate.mt_bench.judge_model']
...
```

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [X] Unit tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

**TODO:**
- [x] Add `ConfigOption` to all subcommands
- [x] Extend `ConfigOption` to support nested defaults, e.g. `llama_cpp.gpu_layers`

PR #1626, opened 2024-07-08
adjust parsing of default arguments so that we explicitly define the mappings from our configs & parameters to the interface objects

### LINKED ISSUE #1523: Create a cleaner mapping between the CLI flags and the training configs
Currently, we flatten the CLI config for training so it can be merged with the CLI flag options, only to then be unflattened again and sent into the training interface data structures.

To get around this, we need to instead change it so that all of the CLI flags are manually mapped into the existing config and then sent into the training data structures.

A notable issue of the flattening and unflattening is that the training library has two different `save_samples` keys in different places.
There is the `deepspeed.save_samples` and `save_samples` which translate into the checkpoint save frequency of the hugging face and deepspeed format checkpoints respectively.

The CLI currently erases one of these because of the collision when flattening.

RobotSail:
This PR resolves an issue where we cannot consume the training library due to a key collision
that happens when we flatten the config object. The collision occurs between `TrainingArgs.save_samples`
and `TrainingArgs.DeepSpeedOptions.save_samples`.

This PR introduces a workaround in which we flatten the objects but explicitly handle the case
of collision by ignoring the keys during the flattening process.
We then explicitly set the value in each field when mapping back from the config + parameters into the TrainingArgs and TorchrunArgs.
Through this process, we're able to ensure that all of the values are mapping as they're intended.

Resolves #1523  (partially)

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -6,6 +6,7 @@ from re import match
 from typing import Optional
 import os
 import sys
+import typing
 
 # Third Party
 from instructlab.training import (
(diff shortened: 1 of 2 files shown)


RobotSail:
Closing this one in favor of #1591

PR #1630, opened 2024-07-08
Add evaluate flag for --merge-system-user-message

danmcp:
The flag is required to enable Mistral based models as the judge during evaluation.

Resolves: https://github.com/instructlab/eval/issues/51

**Checklist:**

- [*] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.


danmcp:
Will need https://github.com/instructlab/eval/pull/57 to be released first

danmcp:
Note: Linting will fail until the new eval is published

src/instructlab/model/evaluate.py
```diff
@@ -101,6 +106,7 @@ def get_evaluator(
             branch,
             output_dir,
             max_workers,
+            merge_system_user_message=merge_system_user_message,
```
alimaredia, line 109: This return is the case for mt_bench_branch right? I'm curious why you didn't use an else if or put a comment to signify that the final return case is for `mt_bench_branch`.

alimaredia:
This PR is small, and looks good. Once I get a chance to run it sometime today, I'll throw an approval on there

src/instructlab/model/evaluate.py
danmcp: I don't that was me who put in the if return logic previously.  It works fine but my preference would be to have an elif as well.

alinaryan (approved):
Ran mt-bench and mt-bench-branch successfully via the cli with the added flag with no issues :+1:

JamesKunstle (approved):

PR #1634, opened 2024-07-09
introduce `additional_args` where possible in the config

cdoern:
Many recent changes in the config.yaml have introduced a few different designs when it comes to implicit vs explicit arguments.

serving opted for a more implicit approach with `vllm_args` being an empty list. 

training opted for the verbose approach where every option lives in the config.yaml.

`additional_arguments` aims to unify the design of the config. Where possible, place commonly edited and/or useful config entries into the respective classes, allowing users to view and edit the properties.

However, for more obscure or less commonly edited values, use `additional_arguments` a dictionary containing the option names and values for these additional properties.

As an initial pass I reorganized parts of the vllm serve config, adding some new options I think would be useful. 

For training, I removed torch_args and train_args instead opting for a flat config entry. The rest of the arguments will live in `additional_args`.

I will work with @RobotSail primarily from training to figure out the best way to map these flat config options to their respective training library options.

I will work with @alimaredia from serving to figure out the right structure for that config.

@alinaryan I'd like to chat with you about how you think this structure would fit into the current eval command.

Having a unified design for the config is crucial as we add more commands, flags, and functionality to `ilab`. Already, our functionality is faltering due to unsafe defaults and unsafe handling of the existing config, normalizing the structure is a great first step.

These `additional_arguments` can also be viewed as "expert options". This structure is a precursor to ilab profiles which will enable pre-baked configurations that allow users to run commands with minimal flags at different fidelity levels.


cdoern:
I need to edit a lot of functionality here. Training currently doesn't ingest the additional_args properly nor does serving because it used to be a list of strings.

Please allow this PR to settle before giving intensive reviews, but feel free to take a look at the general idea outlined in the description

alimaredia:
@cdoern Could you give an example of how the config will look before and after `additional_args`? 

Here's an example of what the serve config will look like at the end of the week if all of the PRs in flight get merged. I've provided some config parameters only an advanced user might want to configure:

```
serve:
  backend: vllm
  host_port: 127.0.0.1:8000
  chat_template: templates/merlinite_template.jinja2 # an advanced user might configure this
  llama_cpp:
    gpu_layers: -1 # an advanced user might configure this
    llm_family: ''
    max_ctx_size: 4096 
  vllm:
    vllm_args: []
    background: false # an advanced user might configure this
    gpus: 4
    startup_timeout: 300 # an advanced user might configure this
    shutdown_timeout: 60 # an advanced user might configure this
```

src/instructlab/model/backends/vllm.py
```diff
@@ -120,11 +135,16 @@ def run_vllm(
     vllm_process = None
     vllm_cmd = [sys.executable, "-m", "vllm.entrypoints.openai.api_server"]
 
-    if "--host" not in vllm_args:
-        vllm_cmd.extend(["--host", host])
 
-    if "--port" not in vllm_args:
-        vllm_cmd.extend(["--port", str(port)])
+    # TODO: there should really be a better way to do this, and there probably is
```
RobotSail, line 139: If there isn't any conditional logic you need to do with these args, then the cleaner way would be to just use a single `.extend` and pass everything in as a single list. Not something I'd block on though.

src/instructlab/model/serve.py
```diff
@@ -80,20 +121,28 @@
...
     num_threads: int | None,
     max_ctx_size: int,
     model_family,
     log_file: pathlib.Path | None,
     backend: str | None,
-    vllm_args: typing.Iterable[str],
+    model_path: str,
+    served_model_name: str,
+    device: str,
+    max_model_len,
```
RobotSail, line 133: For cleanliness I'd make sure that all of these args have type-hints to stay consistent with the existing function. This'll also make it easier for future contributors to work if the type-hints are always explicitly defined.

src/instructlab/configuration.py
```diff
@@ -438,6 +478,61 @@ def init(ctx: click.Context, config_file: str | os.PathLike[str]) -> None:
...
+
+def map_train_to_library(params):
+    # first do a lazy unwrap into the respective options
+    train_args = TrainingArgs(**params)
+    torch_args = TorchrunArgs(**params)
+
+    ds_args = DeepSpeedOptions(
+        cpu_offload_optimizer = params["deepspeed_cpu_offload_optimizer"],
+        cpu_offload_optimizer_ratio = params["deepspeed_cpu_offload_optimizer_ratio"],
+        cpu_offload_optimizer_pin_memory = params["deepspeed_cpu_offload_optimizer_pin_memory"]
```
RobotSail, line 499: Are we not including the deepspeed `save_samples` for the time being?

src/instructlab/configuration.py
```diff
@@ -172,14 +188,21 @@ class _serve(BaseModel):
...
+    host: str
+
+    port: str
+
+    # host_port: StrictStr = "127.0.0.1:8000"
     backend: Optional[str] = (
         None  # we don't set a default value here since it's auto-detected
     )
 
+    additional_args: dict[str, str]
```
RobotSail, line 200: Are we always guaranteed to be mapping to a `str`? If not, I would change this to `dict[str, typing.Any]`, else you can ignore this comment.

RobotSail (commented):
I like the direction this is going in. I'll take another pass once it's ready for review.

src/instructlab/configuration.py
cdoern: oh honestly good point I was thinking of doing this

cdoern:
lets at least try to get in #1653 as a subset of this

ktam3:
Moving to .19 Milestone since this shouldn't be GA blocker, but @cdoern pls lmk otherwise

nathan-weinberg:
@cdoern can we close this?

PR #1635, opened 2024-07-09
Pass vllm args as extra arguments

tiran:
Additional vLLM args are now passed as click extra args. They are more convenient to use than passing each extra argument as a separate arg.

```console
ilab model serve ... --backend=vllm -- --dtype=auto --enable-lora
```

Also clean up unnecessary parameter handling in llama-cpp path.

**Checklist:**

### DIFF
--- a/src/instructlab/model/serve.py
+++ b/src/instructlab/model/serve.py
@@ -3,7 +3,6 @@
 # Standard
 import logging
 import pathlib
-import typing
 
 # Third Party
 import click
@@ -16,7 +15,9 @@ from instructlab.model.backends.backends import ServerException
 logger = logging.getLogger(__name__)
 
 
-@click.command()
+@click.command(
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
+)
 @click.option(
     "--model-path",
     type=click.Path(path_type=pathlib.Path),
@@ -66,16 +67,6 @@ logger = logging.getLogger(__name__)
         "Automatically detected based on the model file properties.\n"
     ),
 )
-@click.option(
-    "--vllm-args",
-    type=str,
-    multiple=True,
-    help=(
-        "Specific arguments to pass into vllm. Each arg must be passed "
-        "separately and surrounded by quotes.\n"
-        " Example: --vllm-args='--dtype=auto' --vllm-args='--enable-lora'"
-    ),
-)
 @click.pass_context
 @utils.display_params
 def serve(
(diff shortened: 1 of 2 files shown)


cdoern:
This is going to clobber https://github.com/instructlab/instructlab/pull/1634 which takes a different approach config wide. Basically I am adding `additional_args` to serve, train, etc. 

The current `vllm_args` will become a dict (probably?) of flag name -> value pairs. Some vllm args will earn a spot in the actual vllm config yaml if we deem them important enough and commonly edited.

I think maybe these two PRs can coexist? But just wanted to voice that the `additional_args` is kind of needed for stuff like training and I am trying to make a unified config language.

cdoern:
The issue here is:

I believe commonly edited vllm args should live in the cfg and therefore inherit defaults in the same way all other flags do.

`additional_args` is as far as I think we should go with "implicit" option passing. 

Maybe we can keep this worklfow for the `vllm_additional_args` implicit vllm options as outlined out in #1634 but we should also have some as their own flags IMO

src/instructlab/model/serve.py
```diff
@@ -129,6 +121,14 @@ def serve(
         )
     elif backend == backends.VLLM:
         # Instantiate the vllm server
+        if ctx.args:
+            # extra click arguments after "--"
+            vllm_args = ctx.args
+        elif ctx.obj.config.serve.vllm.vllm_args:
```
cdoern, line 127: if the options are still in the cfg, then I don't think this impacts #1634. I honestly like this new way of dealing w/ the flags.

src/instructlab/model/serve.py
```diff
@@ -66,16 +67,6 @@
         "Automatically detected based on the model file properties.\n"
     ),
 )
-@click.option(
```
cdoern, line 69: we should probably document somewhere if we support ALL vllm args

cdoern (approved):
took another look, I don't think this is a huge blocker for #1634 so this gets my +1

russellb:
a question -- previously you could do something like this in `config.yaml`

```yaml
serve:
  ...
  backend: "vllm"
  backend_args:
    vllm_args:
    - "custom_arg"
```

I don't care about the syntax, just the ability to put the args in the config file. Is that still supported in some way?

leseb:
> a question -- previously you could do something like this in `config.yaml`
> 
> ```yaml
> serve:
>   ...
>   backend: "vllm"
>   backend_args:
>     vllm_args:
>     - "custom_arg"
> ```
> 
> I don't care about the syntax, just the ability to put the args in the config file. Is that still supported in some way?

`backend_args` does not exist, but I see what you mean. All of that I left unchanged, only the CLI flags are touched. So if you had that config in your `config.yaml`, it will continue to work. Passing backend-specific flags to the CLI is just more elegant :)

src/instructlab/model/serve.py
```diff
@@ -87,9 +78,14 @@ def serve(
     model_family,
     log_file: pathlib.Path | None,
     backend: str | None,
-    vllm_args: typing.Iterable[str],
 ) -> None:
-    """Start a local server"""
+    """Start a local server
+
+    The vLLM backend accepts additional parameters in the form of extra
+    arguments after "--" separator:
```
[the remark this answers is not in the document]
tiran: Good idea! I had to shorten the URL and omit the anchor. click wraps line > 78 chars, which breaks the URL.

leseb (approved):

leseb:
@n1hility dismissed your review because the PR got enough attention, feel free to comment if you have any concerns with this approach. Thanks!

PR #1636, opened 2024-07-09
feat: Add chat template customization to ilab serve

n1hility:
Introduces a `--chat-template` option, covering both vllm and llama.cpp backends. This addresses compatibility problems with ilab chat clients and the vllm backend, and brings closer consitency between llama.cpp and vllm for out of the box usage. This is also a prerequisite for e2e testing using vLLM.

The option has three possible values:

1. 'auto' (default) - instructlab overrides the model template with a new template based on hard coded model knowledege. This is the current behavior before this commit. However, in the future this could be changed to a different dynamic policy, as the name simply implies instructlab decides

2. 'tokenizer' - uses the template expressed in the model's tokenizer_config.json file (or internal gguf settings). This is the ideal long term direction, although 'auto' allows for better compatibility with legacy behvior.

3. A filesystem path expressing the location of a custom template (e.g. /foo/bar/template.jinja)

Notes:

- On vLLM, Temporary files are held in globals with an atexit cleanup since we currently lack a well defined scope with a long enough period to cover vLLM opening the file. A future change could address this by introducing a scope, or shift startup wait code into start.
- Even though the effective default of --chat-template is "auto" internally None is used to allow for the usage of context overrides.
- Fixes leftover instance logger in exception handling
- Testing is mostly a follow-up with the e2e fixes
- Will link dev doc update soon

**Checklist:**

- [x] functional tests have been added, if necessary.

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -268,6 +268,7 @@ class _generate(BaseModel):
 class _serve_vllm(BaseModel):
     """Class describing configuration of vllm serving backend."""
 
+    llm_family: str = ""
     # arguments to pass into vllm process
     vllm_args: list[str] | None = None
 
@@ -294,9 +295,9 @@ class _serve(BaseModel):
 
     # required fields
     model_path: StrictStr
-
     # additional fields with defaults
     host_port: StrictStr = "127.0.0.1:8000"
+    chat_template: Optional[str] = None
     backend: Optional[str] = (
         None  # we don't set a default value here since it's auto-detected
     )
@@ -408,6 +409,7 @@ def get_default_config() -> Config:
                 llm_family="",
             ),
             vllm=_serve_vllm(
+                llm_family="",
                 vllm_args=[],
             ),
         ),
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -5,6 +5,7 @@ from time import sleep
 from types import FrameType
 from typing import Optional, Tuple
 import abc
+import contextlib
 import logging
 import mmap
 import multiprocessing
@@ -14,6 +15,7 @@ import socket
 import struct
 import subprocess
 import sys
+import typing
 
 # Third Party
 from uvicorn import Config
@@ -25,7 +27,7 @@ import uvicorn
 # Local
 from ...client import ClientException, check_api_base, list_models
 from ...configuration import _serve as serve_config
-from ...configuration import get_api_base
+from ...configuration import get_api_base, get_model_family
 from ...utils import split_hostport
 
 logger = logging.getLogger(__name__)
@@ -34,6 +36,8 @@ LLAMA_CPP = "llama-cpp"
 VLLM = "vllm"
 SUPPORTED_BACKENDS = frozenset({LLAMA_CPP, VLLM})
 API_ROOT_WELCOME_MESSAGE = "Hello from InstructLab! Visit us at https://instructlab.ai"
+CHAT_TEMPLATE_AUTO = "auto"
+CHAT_TEMPLATE_TOKENIZER = "tokenizer"
 templates = [
     {
         "model": "merlinite",
@@ -44,6 +48,10 @@ templates = [
         "template": "{{ bos_token }}\n{% for message in messages %}\n{% if message['role'] == 'user' %}\n{{ '[INST] ' + message['content'] + ' [/INST]' }}\n{% elif message['role'] == 'assistant' %}\n{{ message['content'] + eos_token}}\n{% endif %}\n{% endfor %}",
     },
 ]
+
+
+class Closeable(typing.Protocol):
+    def close(self) -> None: ...
 
 
 class ServerException(Exception):
@@ -63,15 +71,20 @@ class BackendServer(abc.ABC):
 
     def __init__(
         self,
+        model_family: str,
         model_path: pathlib.Path,
+        chat_template: str,
         api_base: str,
         host: str,
         port: int,
     ) -> None:
+        self.model_family = model_family
         self.model_path = model_path
+        self.chat_template = chat_template if chat_template else CHAT_TEMPLATE_AUTO
         self.api_base = api_base
         self.host = host
         self.port = port
+        self.resources: list[Closeable] = []
 
     @abc.abstractmethod
     def run(self):
@@ -83,9 +96,19 @@ class BackendServer(abc.ABC):
     ) -> str:
         """Run serving backend in background ('ilab model chat' when server is not running)"""
 
-    @abc.abstractmethod
     def shutdown(self):
         """Shutdown serving backend"""
+
+        safe_close_all(self.resources)
+
+    def register_resources(self, resources: typing.Iterable[Closeable]) -> None:
+        self.resources.extend(resources)
+
+
+def safe_close_all(resources: typing.Iterable[Closeable]):
+    for resource in resources:
+        with contextlib.suppress(Exception):
+            resource.close()
 
 
 def is_model_gguf(model_path: pathlib.Path) -> bool:
@@ -333,6 +356,23 @@ def is_temp_server_running():
     return multiprocessing.current_process().name != "MainProcess"
 
 
+def get_model_template(
+    model_family: str, model_path: pathlib.Path
+) -> Tuple[str, str, str]:
+    eos_token = "<|endoftext|>"
+    bos_token = ""
+    template = ""
+    resolved_family = get_model_family(model_family, model_path)
+    for template_dict in templates:
+        if template_dict["model"] == resolved_family:
+            template = template_dict["template"]
+            if template_dict["model"] == "mixtral":
+                eos_token = "</s>"
+                bos_token = "<s>"
+
+    return template, eos_token, bos_token
+
+
 def get_uvicorn_config(app: fastapi.FastAPI, host: str, port: int) -> Config:
     return Config(
         app,
@@ -362,12 +402,16 @@ def select_backend(
         raise click.exceptions.Exit(1)
 
     host, port = split_hostport(cfg.host_port)
+    chat_template = cfg.chat_template
+    if not chat_template:
+        chat_template = CHAT_TEMPLATE_AUTO
 
     if backend == LLAMA_CPP:
         # Instantiate the llama server
         return llama_cpp_server(
             api_base=cfg.api_base(),
             model_path=model_path,
+            chat_template=chat_template,
             gpu_layers=cfg.llama_cpp.gpu_layers,
             max_ctx_size=cfg.llama_cpp.max_ctx_size,
             num_threads=None,  # exists only as a flag not a config
@@ -379,10 +423,22 @@ def select_backend(
         # Instantiate the vllm server
         return vllm_server(
             api_base=cfg.api_base(),
+            model_family=cfg.vllm.llm_family,
             model_path=model_path,
+            chat_template=chat_template,
             vllm_args=cfg.vllm.vllm_args,
             host=host,
             port=port,
         )
     click.secho(f"Unknown backend: {backend}", fg="red")
     raise click.exceptions.Exit(1)
+
+
+def verify_template_exists(path):
+    if not path.exists():
+        raise FileNotFoundError("Chat template file does not exist: {}".format(path))
+
+    if not path.is_file():
+        raise IsADirectoryError(
+            "Chat templates paths must point to a file: {}".format(path)
+        )
--- a/src/instructlab/model/backends/llama_cpp.py
+++ b/src/instructlab/model/backends/llama_cpp.py
@@ -2,31 +2,34 @@
 
 # Standard
 from contextlib import redirect_stderr, redirect_stdout
-from typing import Optional
+from typing import Optional, cast
 import logging
 import multiprocessing
 import os
 import pathlib
 
 # Third Party
-from llama_cpp import llama_chat_format
+from llama_cpp import llama_chat_format, llama_token_get_text
 from llama_cpp.server.app import create_app
+from llama_cpp.server.model import LlamaProxy
 from llama_cpp.server.settings import Settings
 import httpx
 import llama_cpp.server.app as llama_app
 
 # Local
-from ...configuration import get_model_family
 from .backends import (
     API_ROOT_WELCOME_MESSAGE,
+    CHAT_TEMPLATE_AUTO,
+    CHAT_TEMPLATE_TOKENIZER,
     LLAMA_CPP,
     BackendServer,
     ServerException,
     UvicornServer,
     ensure_server,
+    get_model_template,
     get_uvicorn_config,
     is_temp_server_running,
-    templates,
+    verify_template_exists,
 )
 
 logger = logging.getLogger(__name__)
@@ -37,6 +40,7 @@ class Server(BackendServer):
         self,
         model_path: pathlib.Path,
         model_family: str,
+        chat_template: str,
         api_base: str,
         host: str,
         port: int,
@@ -44,8 +48,7 @@ class Server(BackendServer):
         max_ctx_size: int,
         num_threads: Optional[int],
     ):
-        super().__init__(model_path, api_base, host, port)
-        self.model_family = model_family
+        super().__init__(model_family, model_path, chat_template, api_base, host, port)
         self.gpu_layers = gpu_layers
         self.max_ctx_size = max_ctx_size
         self.num_threads = num_threads
@@ -57,6 +60,7 @@ class Server(BackendServer):
         try:
             server(
                 model_path=self.model_path,
+                chat_template=self.chat_template,
                 gpu_layers=self.gpu_layers,
                 max_ctx_size=self.max_ctx_size,
                 model_family=self.model_family,
@@ -75,6 +79,7 @@ class Server(BackendServer):
             target=server,
             kwargs={
                 "model_path": self.model_path,
+                "chat_template": self.chat_template,
                 "gpu_layers": self.gpu_layers,
                 "max_ctx_size": self.max_ctx_size,
                 "model_family": self.model_family,
@@ -107,6 +112,9 @@ class Server(BackendServer):
 
     def shutdown(self):
         """Stop the server process and close the queue."""
+
+        super().shutdown()
+
         if self.process and self.queue:
             self.process.terminate()
             self.process.join(timeout=30)
@@ -116,13 +124,14 @@ class Server(BackendServer):
 
 def server(
     model_path: pathlib.Path,
+    chat_template: str,
     gpu_layers: int,
     max_ctx_size: int,
     model_family: str,
     threads=None,
     host: str = "localhost",
     port: int = 8000,
-    queue=None,
+    queue: Optional[multiprocessing.Queue] = None,
 ):
     """Start OpenAI-compatible server"""
     settings = Settings(
@@ -149,30 +158,8 @@ def server(
             return
         raise ServerException(f"failed creating the server application: {exc}") from exc
 
-    template = ""
-    eos_token = "<|endoftext|>"
-    bos_token = ""
-    for template_dict in templates:
-        if template_dict["model"] == get_model_family(model_family, model_path):
-            template = template_dict["template"]
-            if template_dict["model"] == "mixtral":
-                eos_token = "</s>"
-                bos_token = "<s>"
-    try:
-        for proxy in llama_app.get_llama_proxy():
-            proxy().chat_handler = llama_chat_format.Jinja2ChatFormatter(
-                template=template,
-                eos_token=eos_token,
-                bos_token=bos_token,
-            ).to_chat_handler()
-    # pylint: disable=broad-exception-caught
-    except Exception as exc:
-        if queue:
-            queue.put(exc)
-            queue.close()
-            queue.join_thread()
-            return
-        raise ServerException(f"failed creating the server application: {exc}") from exc
+    # Update chat template if necessary
+    augment_chat_template(chat_template, model_family, model_path, queue)
 
     logger.info("Starting server process, press CTRL+C to shutdown server...")
     logger.info(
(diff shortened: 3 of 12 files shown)


n1hility:
@alimaredia @leseb @danmcp PTAL

danmcp:
Unit tests would be nice to have here.

n1hility:
> Unit tests would be nice to have here.

+1 Added some additional tests for chat-template values on top of the config test

n1hility:
Added hold for the dev doc portion

src/instructlab/model/backends/backends.py
```diff
@@ -329,12 +352,16 @@ def select_backend(cfg: serve_config) -> BackendServer:
         raise click.exceptions.Exit(1)
 
     host, port = split_hostport(cfg.host_port)
+    chat_template = cfg.chat_template
+    if not chat_template:
```
leseb, line 356: I think this can be avoided by simply setting a default to chat_template (as suggested above).

src/instructlab/model/backends/llama_cpp.py
```diff
@@ -203,3 +186,70 @@ def read_root():
...
+):
+    if chat_template == CHAT_TEMPLATE_TOKENIZER:
+        # llama_cpp_python calculates from the tokenizer config on load, so
+        # nothing to do
+        return
+
+    if chat_template == CHAT_TEMPLATE_AUTO:
+        # Currently "auto" maps to replacing with ilab stored templates
+        template, eos_token, bos_token = get_model_template(model_family, model_path)
+    else:
```
leseb, line 205: Can you add a comment indicating that this third condition means `chat_template` is a template file?

src/instructlab/model/backends/llama_cpp.py
```diff
@@ -203,3 +186,70 @@ def read_root():
...
+
+    return resolve_token(proxy, proxy().token_bos())
+
+
+def resolve_token(proxy: LlamaProxy, token: int) -> str:
+    result = llama_token_get_text(proxy().model, token).decode("utf-8")
+    return cast(str, result)
+
+
+def load_template(template_path: str) -> str:
```
leseb, line 250: What if `template_path` is a directory?

src/instructlab/model/backends/llama_cpp.py
```diff
@@ -203,3 +186,70 @@ def read_root():
...
+    path = pathlib.Path(template_path)
+    if not path.exists():
+        raise FileNotFoundError("Chat template file does not exist: {}".format(path))
+
+    return path.read_text(encoding="utf-8")
```
leseb, line 255: Have do we handle `path.read_text` failing?

src/instructlab/model/backends/vllm.py
```diff
@@ -15,26 +18,56 @@
...
+
+def safe_close_all(resources: typing.Iterable[Closeable]) -> None:
+    for resource in resources:
+        with contextlib.suppress(Exception):
+            resource.close()
+
+
+atexit.register(cleanup_backend)
+
+
```
leseb, line 58: How is this related to `chat_template`?

src/instructlab/model/backends/llama_cpp.py
leseb: Also, this conditional block serves as a validation for `chat_template` should we extract it in a `validate_chat_template` and call it earlier (to fail early if needed)?

src/instructlab/model/backends/vllm.py
```diff
@@ -91,15 +128,43 @@ def run_detached(self, http_client: httpx.Client | None = None) -> str:
...
+def format_template(model_family: str, model_path: pathlib.Path) -> str:
+    template, eos_token, bos_token = get_model_template(model_family, model_path)
+
+    prefix = ""
+    if eos_token:
+        prefix = '{{% set eos_token = "{}" %}}\n'.format(eos_token)
+    if bos_token:
+        prefix = '{}{{% set bos_token = "{}" %}}\n'.format(prefix, bos_token)
+
+    return prefix + template
```
leseb, line 146: Should we use a jinja lib to validate this template before sending it to vllm? Or build using a jinja lib maybe?

src/instructlab/model/backends/vllm.py
```diff
@@ -119,28 +184,48 @@ def run_vllm(
...
+    if not contains_argument("--model", vllm_args):
         vllm_cmd.extend(["--model", os.fspath(model_path)])
 
+    if not contains_argument("--chat-template", vllm_args):
+        if chat_template == CHAT_TEMPLATE_AUTO:
+            template = format_template(model_family, model_path)
+            file = create_tmpfile(template)
+            tmp_files.append(file)
+            print(template)
+            print(file.name)
```
leseb, line 204: Leftover or intentional?

src/instructlab/model/backends/vllm.py
```diff
@@ -119,28 +184,48 @@ def run_vllm(
...
+            vllm_cmd.extend(["--chat-template", file.name])
+        elif chat_template != CHAT_TEMPLATE_TOKENIZER:
```
leseb, line 206: Let's comment that this is the case where a file is passed for chat_template

src/instructlab/model/serve.py
```diff
@@ -56,6 +56,16 @@
     required=False,
     help="Log file path to write server logs to.",
 )
+@click.option(
+    "--chat-template",
+    type=str,
+    help=(
+        "Which chat template to use in serving the model, either 'auto', "
+        "'tokenizer', or a path to template file. \n"
```
leseb, line 64:
```suggestion
        "'tokenizer', or a path to jinja formatted template file. \n"
```

It relies on jinja right?

tests/test_lab_serve.py
```diff
@@ -0,0 +1,103 @@
...
+                "model",
+                "serve",
+                "--model-path=foo",
+                "--chat-template={}".format(file.name),
+            ],
+        )
+        assert_vllm_args(args)
+        assert_template(
+            args=args, expect_chat=True, path_chat=True, chat_value=file.name
+        )
```
leseb, line 84:
```suggestion
def test_chat_custom(mock_popen, mock_determine, tmp_path: pathlib.Path, _):
    file = tmp_path / "chat_template.jinja2"
    args = vllm_setup_test(
        mock_popen,
        mock_determine,
        [
            "--config=DEFAULT",
            "model",
            "serve",
            "--model-path=foo",
            "--chat-template={}".format(file.name),
        ],
    )
    assert_vllm_args(args)
    assert_template(
        args=args, expect_chat=True, path_chat=True, chat_value=file.name
    )
```

This should work. @tiran pointed out in another review that using the `tmp_path` fixture is more appropriate.

leseb (changes requested):
Nice work! Can you look into adding functional tests (`scripts/functional-tests.sh`) as well to at least validate this with llama-cpp? Thanks

src/instructlab/model/backends/llama_cpp.py
n1hility: So IMO this will throw appropriate errors with the path (FileNotFound, OSError etc). Open to ideas if you have a better suggestion.

However this is a problem were we dont handle these outer sections correctly will fix that.

src/instructlab/model/backends/vllm.py
n1hility: The Auto handler requires the use of temp files that we need to cleanup, yet still stick around long enough for vLLM to open and read them (see the first bullet in the notes section for further details)

src/instructlab/model/backends/vllm.py
n1hility: Oops! good catch

src/instructlab/configuration.py
```diff
@@ -170,9 +171,9 @@ class _serve(BaseModel):
 
     # required fields
     model_path: StrictStr
-
     # additional fields with defaults
     host_port: StrictStr = "127.0.0.1:8000"
+    chat_template: Optional[str] = None
```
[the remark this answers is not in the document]
n1hility: Thanks for the feedback. So the intention was to be able to detect and allow overrides like some of the other values are setup for but happy to go with a hard constant if you like that better.

src/instructlab/model/backends/llama_cpp.py
n1hility: Since both vllm and llama cpp log the chat template and error properly I preferred just delegating to them and avoid the double parsing.

src/instructlab/model/backends/vllm.py
n1hility: (see above on that one, assuming I understood your comment correctly)

tests/test_lab_serve.py
n1hility: thanks good suggestion

src/instructlab/model/backends/llama_cpp.py
leseb: Souns good, just a small note about the third case 🙏🏻 so it's easier to understand for whoever is reading at the code and not familiar with it.

src/instructlab/model/backends/vllm.py
leseb: Ahhhh ok ok :) Thanks for the clarification.

tiran:
The temporary file handling looks complicated. It's going to be a nightmare the write unit and function tests for its behavior.

I wonder if it would be possible to write this differently:
- extend `BackendServer` ABC with a `tempdir` attribute
- add a method to initialize `tempdir`
- cleanup and remove `tempdir` in `shutdown` / exit route
- convert the template functions to methods of the BackendServer class and the subclasses.

This would bind the life time of the temporary directory to the life time of the server.

alimaredia:
@tiran @n1hility Christian touched on what I had been thinking when I read through the code. I had never heard of `atexit` before this PR so I would have expected the temp file cleanup to happen in `shutdown()` and I think the changes he proposes are easier to read. 

One other thing to point out is that `--model-family` is an argument/config for `ilab model chat` https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/chat.py#L185 and it needs to be set now in the vLLM section of the serve config, not just for llama_cpp.

leseb:
@tiran @alimaredia @n1hility the vLLM [doc](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#chat-template) says:

> For those model, you can manually specify their chat template in the --chat-template parameter with the file path to the chat template, or the template in string form

Also, the [code](https://github.com/vllm-project/vllm/blob/3963a5335bb4106f2ecd1139527e3568d2151933/vllm/entrypoints/openai/serving_chat.py#L71-L89) seems to fallback if the file cannot be read to decoding the string.

If we use that approach we don't need to create a temp file anymore and just pass the template as a single line which simplifies the implementation quite a lot.
Thoughts?

n1hility:
> The temporary file handling looks complicated. It's going to be a nightmare the write unit and function tests for its behavior.
> 
> I wonder if it would be possible to write this differently:
> 
> * extend `BackendServer` ABC with a `tempdir` attribute
> * add a method to initialize `tempdir`
> * cleanup and remove `tempdir` in `shutdown` / exit route
> * convert the template functions to methods of the BackendServer class and the subclasses.
> 
> This would bind the life time of the temporary directory to the life time of the server.

Thanks for the feedback! This is what I was referring to in the first bullet of notes. Something like this would be be the ideal way to go. The problem was the scoping / lifecycle wasn't complete in all cases so would have lead to a larger change or risked leaking files, so just went with the simpler slightly less ideal path. I can take another look though.

leseb:
> > If we use that approach we don't need to create a temp file anymore and just pass the template as a single line which simplifies the implementation quite a lot.
> > Thoughts?
> 
> The reason we need the temp file is just to support the "auto" usage, where ilab is providing the template (it needs to be written to a file for it to be used)

But if ilab provides a file already we don't need to manage it, just read it, right?

n1hility:
> > > If we use that approach we don't need to create a temp file anymore and just pass the template as a single line which simplifies the implementation quite a lot.
> > > Thoughts?
> > 
> > 
> > The reason we need the temp file is just to support the "auto" usage, where ilab is providing the template (it needs to be written to a file for it to be used)
> 
> But if ilab provides a file already we don't need to manage it, just read it, right?

Oh you mean provide the templates as data file as static content in the distribution, inside the python module or something like that?

n1hility:
> > The temporary file handling looks complicated. It's going to be a nightmare the write unit and function tests for its behavior.
> > I wonder if it would be possible to write this differently:
> > 
> > * extend `BackendServer` ABC with a `tempdir` attribute
> > * add a method to initialize `tempdir`
> > * cleanup and remove `tempdir` in `shutdown` / exit route
> > * convert the template functions to methods of the BackendServer class and the subclasses.
> > 
> > This would bind the life time of the temporary directory to the life time of the server.
> 
> Thanks for the feedback! This is what I was referring to in the first bullet of notes. Something like this would be be the ideal way to go. The problem was the scoping / lifecycle wasn't complete in all cases so would have lead to a larger change or risked leaking files, so just went with the simpler slightly less ideal path. I can take another look though.

I audited all the exception code paths and with one small adjustment I think we can use the instance scope. Will update soon.

leseb:
> > > > If we use that approach we don't need to create a temp file anymore and just pass the template as a single line which simplifies the implementation quite a lot.
> > > > Thoughts?
> > > 
> > > 
> > > The reason we need the temp file is just to support the "auto" usage, where ilab is providing the template (it needs to be written to a file for it to be used)
> > 
> > 
> > But if ilab provides a file already we don't need to manage it, just read it, right?
> 
> Oh you mean provide the templates as data file as static content in the distribution, inside the python module or something like that?

Yes.

n1hility:
> > > > > If we use that approach we don't need to create a temp file anymore and just pass the template as a single line which simplifies the implementation quite a lot.
> > > > > Thoughts?
> > > > 
> > > > 
> > > > The reason we need the temp file is just to support the "auto" usage, where ilab is providing the template (it needs to be written to a file for it to be used)
> > > 
> > > 
> > > But if ilab provides a file already we don't need to manage it, just read it, right?
> > 
> > 
> > Oh you mean provide the templates as data file as static content in the distribution, inside the python module or something like that?
> 
> Yes.

Took a quick look and this option seemed to have a bit more complexity due to the formats being slightly different yet still primarily duplicative between vllm and llama cpp.  I was able to leverage the backend instance scope after all after auditing the exception call paths (mostly only needed on adjustment to enable that).

src/instructlab/model/backends/vllm.py
```diff
@@ -120,32 +164,56 @@ def run_vllm(
...
     Returns:
-        vllm_process (subprocess.Popen): process of the vllm server
+        tuple: A tuple containing two values:
+            vllm_process (subprocess.Popen): process of the vllm server
+            tmp_files: a list of temporary files necessary to launch the process
+
     """
     vllm_process = None
     vllm_cmd = [sys.executable, "-m", "vllm.entrypoints.openai.api_server"]
+    tmp_files = []
```
alimaredia, line 174: What's the purpose of having a list here? From what I can tell `tmp_files` is either empty or contains one entry.

src/instructlab/model/backends/vllm.py
```diff
@@ -120,32 +164,56 @@ def run_vllm(
         background (bool):            Whether the stdout and stderr vLLM should be sent to /dev/null (True)
                                       or stay in the foreground(False).
     Returns:
-        vllm_process (subprocess.Popen): process of the vllm server
```
alimaredia, line 123: Chat template and model family need to be added to the args list in this function annotation.

src/instructlab/model/backends/vllm.py
```diff
@@ -120,32 +164,56 @@ def run_vllm(
...
+            # needs to be written to a temporary file so that vLLM
+            # can read it
+            template = format_template(model_family, model_path)
+            file = create_tmpfile(template)
+            tmp_files.append(file)
+            vllm_cmd.extend(["--chat-template", file.name])
+        elif chat_template != CHAT_TEMPLATE_TOKENIZER:
+            # In this case the value represents a template file path
+            # Pass it directly to vllm
+            vllm_cmd.extend(["--chat-template", chat_template])
```
alimaredia, line 197: @n1hility Tell me if I've got this flow right. If `ilab model serve` is called without the `--chat-template` flag or any additional `vllm_args`, the value of `chat_template` that is passed into the `vllm.Server()` constructor is `None`. 

That `None` value flows down to this conditional and then the python call to vLLM is appended with `--chat-template None`. Is `None` a valid value for that flag in vLLM?

When I first added the vLLM serving code, I decided that the internal chat template selection via model family was not necessary for a first pass and that passing it via `vllm_args` would be enough since at least granite and merlinite served just fine without any additional chat template configuration. 

I just want to make sure that's the case and that just plane `ilab model serve` for granite and merlinite just work out of the box.

src/instructlab/model/serve.py
```diff
@@ -57,6 +57,16 @@
     required=False,
     help="Log file path to write server logs to.",
 )
+@click.option(
+    "--chat-template",
+    type=str,
```
alimaredia, line 62: After looking at and addressing https://github.com/instructlab/instructlab/pull/1636/files#r1677217137, would we want to consider setting the default here to `tokenizer just so that `ilab model serve` with granite and merlinite out of the box just works`?

src/instructlab/model/backends/vllm.py
alimaredia, line 197: @n1hility Trying to think of edge cases, Would we want to to make sure the file passed as chat_template exists on the file system? If you pass in a chat template that does not exist to vLLM does it error out and not start up or give a warning of some sort?

 In the cases of `run_vLLM()` being run with background=True, since there is no vLLM output, a user of `ilab`  commands that serve with vLLM under the hood wouldn't see these warning or have vLLM fail in the background.
 
 Maybe we add a check here and some debug logging saying what the chat template file name is if it's set?

src/instructlab/model/backends/llama_cpp.py
```diff
@@ -205,3 +191,82 @@ def read_root():
...
+            template, eos_token, bos_token = get_model_template(
+                model_family, model_path
+            )
+        else:
+            # In this case, the template is a path to a file; attempt to load it
+            template = load_template(chat_template)
+            eos_token = None
+            bos_token = None
+
+        logger.info(f"Replacing chat template:\n {template}")
```
danmcp, line 220: Nit: Better written as  logger.info("Replacing chat template:\n %s", template)

src/instructlab/model/backends/vllm.py
```diff
@@ -120,32 +164,56 @@ def run_vllm(
...
+        if background:
+            vllm_process = subprocess.Popen(
+                args=vllm_cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+            )
+        else:
+            # pylint: disable=consider-using-with
+            vllm_process = subprocess.Popen(args=vllm_cmd)
+
+        api_base = get_api_base(f"{host}:{port}")
+        logger.info(f"vLLM starting up on pid {vllm_process.pid} at {api_base}")
```
danmcp, line 213: Nit: Better written as: logger.info("vLLM starting up on pid %s at %s", vllm_process.pid, api_base)

tests/test_lab_serve.py
```diff
@@ -0,0 +1,101 @@
...
+
+def vllm_setup_test(mock_popen, mock_determine, args):
+    mock_process = mock.MagicMock()
+    mock_popen.return_value = mock_process
+
+    mock_process.communicate.return_value = ("out", "err")
+    mock_process.returncode = 0
+    mock_determine.return_value = ("vllm", "testing")
+    runner = CliRunner()
+    with runner.isolated_filesystem():
```
danmcp, line 21: Noting the different style @tiran developed: https://github.com/instructlab/instructlab/pull/1617/files

Probably doesn't need to be a blocker for this PR

danmcp (approved):
Approving with a couple nits/follow-ups

tests/test_lab_serve.py
n1hility: ah thanks didnt notice this.

tiran: I recommend to use the `cli_runner` fixture. It takes are of new storage locations, too. Otherwise you may leak test files into your personal `~/.local/share` directory tree.

src/instructlab/model/backends/vllm.py
n1hility: Its's more for the future, since any passing of a large value or hard to parse value on the prompt we would need to store to the fs

src/instructlab/model/backends/vllm.py
n1hility: good catch! thanks

src/instructlab/model/backends/vllm.py
n1hility: ah yes this is a good idea for the background case

alimaredia: Would you want that to happen for follow up after this PR is merged? I can create an issue, and Seb and I can work on getting that in over the next 24 hours.

src/instructlab/model/backends/vllm.py
```diff
@@ -120,32 +164,56 @@ def run_vllm(
...
         vllm_cmd.extend(["--model", os.fspath(model_path)])
 
+    if not contains_argument("--chat-template", vllm_args):
+        if chat_template == CHAT_TEMPLATE_AUTO:
+            # For auto templates, the build-in in-memory template
+            # needs to be written to a temporary file so that vLLM
+            # can read it
+            template = format_template(model_family, model_path)
+            file = create_tmpfile(template)
+            tmp_files.append(file)
```
leseb, line 192: I still don't understand why we need a file where vLLM supports to be given a string directly. @n1hility did I miss the reason somewhere?

leseb (commented):
I just have one leftover question otherwise this feels ready to go :) Thanks!

src/instructlab/model/backends/vllm.py
n1hility: @leseb yeah the big reason is just that files as arguments run up against problems: parsing/escaping issues, ARG_MAX limits, and cut and paste problems (e..g message is printed unescaped, which won't work with paste, user fixes but gets it wrong, etc). 

In the case where the args are passed to exec(), escaping usually isn't a problem, however, it does rely on the receiver of the args to handle them correctly. For example if the called code use an argument parser which optionally splits on = or some other form of interpolation, that can easily break with a template string in a param. Or if there is some sort of intermediary (for example, they construct a command line which is passed as a shell string). Some OSes also requote (e.g. windows) which can lead to oddities.  

Some projects that do this sort of thing use base64 encoding to avoid the problems (with the exception of ARG_MAX limits)

tests/test_lab_serve.py
n1hility: Very cool @tiran !

alimaredia (approved):

leseb (approved):

PR #1640, opened 2024-07-09
Improves logging when connecting to vLLM / Llamacpp

JamesKunstle:
Lets the user know that the cli process is trying to connect to vLLM/Llamacpp- that it's not stalled.

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -209,7 +209,7 @@ def ensure_server(
     and the URL where it's available."""
 
     try:
-        logger.debug(f"Trying to connect to {api_base}...")
+        logger.info(f"Trying to connect to model server at {api_base}")
         # pylint: disable=duplicate-code
         list_models(
             api_base=api_base,
@@ -219,31 +219,34 @@ def ensure_server(
         # pylint: enable=duplicate-code
     except ClientException:
         port = free_tcp_ipv4_port(host)
-        logger.debug("Using %i", port)
+        logger.debug(f"Using available port {port} for temporary model serving.")
 
         host_port = f"{host}:{port}"
         temp_api_base = get_api_base(host_port)
-        logger.debug(f"Starting a temporary server at {temp_api_base}...")
         llama_cpp_server_process = None
         vllm_server_process = None
 
         if backend == VLLM:
             # TODO: resolve how the hostname is getting passed around the class and this function
             vllm_server_process = server_process_func(port)
+            logger.info(f"Starting a temporary vLLM server at {temp_api_base}")
             count = 0
             # TODO should this be configurable?
-            vllm_startup_timeout = 300
+            vllm_startup_timeout = 60
             while count < vllm_startup_timeout:
-                sleep(1)
                 try:
                     list_models(
                         api_base=temp_api_base,
                         http_client=http_client,
                     )
-                    logger.debug(f"model at {temp_api_base} served on vLLM")
+                    logger.info(f"vLLM engine successfully started at {temp_api_base}")
                     break
                 except ClientException:
                     count += 1
+                    logger.info(
+                        f"Waiting for the vLLM server to start at {temp_api_base}, this might take a moment... Retries: {count}/{vllm_startup_timeout}"
+                    )
+                    sleep(5)
 
             if count >= vllm_startup_timeout:
                 shutdown_process(vllm_server_process, 20)
@@ -257,6 +260,7 @@ def ensure_server(
             # in this ensure_server() function
             llama_cpp_server_process = server_process_func(port)
             llama_cpp_server_process.start()
+            logger.debug(f"Starting a temporary llama.cpp server at {temp_api_base}")
 
             # in case the server takes some time to fail we wait a bit
             logger.debug("Waiting for the server to start...")
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -314,7 +314,7 @@ test_temp_server(){
         set timeout 120
         spawn ilab model chat
         expect {
-            "Starting a temporary server at" { exit 0 }
+            "Trying to connect to model server at" { exit 0 }
             timeout { exit 1 }
         }
     '
@@ -324,7 +324,7 @@ test_temp_server_sigint(){
     expect -c '
         set timeout 120
         spawn ilab model chat
-        expect "Starting a temporary server at"
+        expect "Trying to connect to model server at"
         send "hello!\r"
         sleep 5
         send "\003"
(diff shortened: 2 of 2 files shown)


JamesKunstle:
this is just a logging change so shouldn't require tests / changelog

src/instructlab/model/backends/backends.py
```diff
@@ -219,31 +219,34 @@ def ensure_server(
...
             vllm_startup_timeout = 300
             while count < vllm_startup_timeout:
-                sleep(1)
                 try:
                     list_models(
                         api_base=temp_api_base,
                         http_client=http_client,
                     )
-                    logger.debug(f"model at {temp_api_base} served on vLLM")
+                    logger.info(f"Successfully connected to vLLM at {temp_api_base}!")
```
danmcp, line 242: "connected to" doesn't seem appropriate vs the old debug message.  The user doesn't know why we are connecting.  Would it be better to just report the model is being served?

src/instructlab/model/backends/backends.py
```diff
@@ -219,31 +219,34 @@ def ensure_server(
...
                     break
                 except ClientException:
                     count += 1
+                    logger.info(
+                        f"Attempting to connect to starting vLLM server as {temp_api_base}, this might take a moment... {count}s/{vllm_startup_timeout}s"
```
danmcp, line 247: Does the list_models call return immediately the whole time vLLM is starting?

src/instructlab/model/backends/backends.py
```diff
@@ -219,31 +219,34 @@ def ensure_server(
...
         host_port = f"{host}:{port}"
         temp_api_base = get_api_base(host_port)
-        logger.debug(f"Starting a temporary server at {temp_api_base}...")
         llama_cpp_server_process = None
         vllm_server_process = None
 
         if backend == VLLM:
             # TODO: resolve how the hostname is getting passed around the class and this function
             vllm_server_process = server_process_func(port)
+            logger.info(f"Starting a temporary vLLM server at {temp_api_base}...")
```
leseb, line 232:
```suggestion
            logger.info(f"Starting a temporary vLLM server at {temp_api_base}")
```
If we change that line I'd remove the `...` it's a bit weird to have this next to the API base URL I think.

src/instructlab/model/backends/backends.py
leseb: Or just say "vLLM engine ready to serve at {temp_api_base}" or "vLLM engine successfully started at {temp_api_base}" maybe?

src/instructlab/model/backends/backends.py
leseb, line 247: "Waiting for the vLLM server to start at {temp_api_base}, this might take a moment... {count}s/{vllm_startup_timeout}s"

How about that?

src/instructlab/model/backends/backends.py
JamesKunstle: okay gotcha good recommendation

src/instructlab/model/backends/backends.py
JamesKunstle: ++ okay good idea

src/instructlab/model/backends/backends.py
JamesKunstle: it should fail a curl request until vllm is up

JamesKunstle: Yeah that's better

danmcp: > it should fail a curl request until vllm is up

Right, but does it wait at all or always fail fast.  I am asking because the timing will be off with only counting the sleep time if there is time spent waiting to connect.

src/instructlab/model/backends/backends.py
```diff
@@ -257,6 +260,7 @@ def ensure_server(
             # in this ensure_server() function
             llama_cpp_server_process = server_process_func(port)
             llama_cpp_server_process.start()
+            logger.debug(f"Starting a temporary llama.cpp server at {temp_api_base}...")
```
danmcp, line 263: I think the same comment before about the ... applies here as well?

src/instructlab/model/backends/backends.py
```diff
@@ -209,7 +209,7 @@ def ensure_server(
     and the URL where it's available."""
 
     try:
-        logger.debug(f"Trying to connect to {api_base}...")
+        logger.info(f"Trying to connect to model server at {api_base}...")
```
danmcp, line 212: I think the same comment before about the ... applies here as well?

JamesKunstle: Lol I missed that you're right

JamesKunstle: ++done

src/instructlab/model/backends/backends.py
JamesKunstle: ++done

src/instructlab/model/backends/backends.py
JamesKunstle: Ah I see what you're saying. It fails fast, but maybe it's better to not communicate the number of retried in terms of time.

src/instructlab/model/backends/backends.py
```diff
@@ -219,31 +219,34 @@ def ensure_server(
...
                     list_models(
                         api_base=temp_api_base,
                         http_client=http_client,
                     )
-                    logger.debug(f"model at {temp_api_base} served on vLLM")
+                    logger.info(f"vLLM engine successfully started at {temp_api_base}")
                     break
                 except ClientException:
                     count += 1
+                    logger.info(
```
danmcp, line 246: To avoid up to 300 messages here, should we only print the message every 10ish secs?  An animation would be nice but can do that in a later PR maybe.

danmcp (changes requested):

danmcp (approved):

src/instructlab/model/backends/backends.py
JamesKunstle: ++

src/instructlab/model/backends/backends.py
```diff
@@ -219,31 +219,34 @@ def ensure_server(
...
         if backend == VLLM:
             # TODO: resolve how the hostname is getting passed around the class and this function
             vllm_server_process = server_process_func(port)
+            logger.info(f"Starting a temporary vLLM server at {temp_api_base}")
             count = 0
             # TODO should this be configurable?
-            vllm_startup_timeout = 300
+            vllm_startup_timeout = 60
             while count < vllm_startup_timeout:
-                sleep(1)
```
alimaredia, line 237: @JamesKunstle what's the logic in removing the sleep and reducing the timeout to 60? How long could you expect this loop to go for?

I've had times where I've noticed vLLM not be ready for almost 2 minutes when I run things locally. I set the default of 300 to be on the safe side.

alimaredia (approved):

leseb (approved):

leseb:
Dismissed @alinaryan given that this PR was given sufficient attention.

PR #1644, opened 2024-07-09
Provide a conversion utility to ensure that the datasets being provided to legacy train are compatible

RobotSail:
Due to differences in the design of the training & generate in the original ilab implementation and 
what we had in the dedicated SDG and full model fine-tuning pipelines, the datasets that we are generating are incompatibile.

Specifically, instructlab/sdg currently outputs datasets in the HuggingFace messages format, while ilab expects a dataset in the format of:

```py
dataset = [
	{
		"system": "You are a friendly assistant",
		"user": "What is 2+2?",
		"assistant": "2+2=4"
	}
]
```

This PR introduces a shim which detects whether the provided dataset is using the legacy format or the messages format, and provides a conversion if necessary.

**Note**: This conversion can only go in one direction, as the messages format includes additional metadata. Here is a sample in the messages format:
```json
{
    "messages": [
        {
            "content": "You are an AI language model developed by IBM Research. You are a cautious assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior.",
            "role": "system"
        },
        {
            "content": "Using the word \"grace\", come up with a word that rhymes and has the same number of syllables\n<nopace>",
            "role": "user"
        },
        {
            "content": "Certainly! Here's a word that rhymes with \"grace\" and has the same number of syllables:\n1\\. Space",
            "role": "assistant"
        }
    ],
    "group": "lab_extension",
    "dataset": "base/full-extension",
    "metadata": "{\"num_turns\": 1}"
}
```

### DIFF
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -228,8 +228,8 @@ def linux_train(
     try:
         train_dataset = ensure_legacy_dataset(train_dataset)
         test_dataset = ensure_legacy_dataset(test_dataset)
-    except ValueError:
-        ctx.fail("failed to run training, dataset appears to be empty")
+    except ValueError as e:
+        ctx.fail(f"Failed to parse dataset: {e}")
 
     train_dataset.to_pandas().head()
 
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -33,6 +33,9 @@ from trl import DataCollatorForCompletionOnlyLM, SFTTrainer
 import click
 import torch
 
+# First Party
+from instructlab.utils import ensure_legacy_dataset
+
 # Local
 from ..model.chat import CONTEXTS
 
@@ -222,6 +225,12 @@ def linux_train(
     )
 
     test_dataset = load_dataset("json", data_files=os.fspath(test_file), split="train")
+    try:
+        train_dataset = ensure_legacy_dataset(train_dataset)
+        test_dataset = ensure_legacy_dataset(test_dataset)
+    except ValueError:
+        ctx.fail("failed to run training, dataset appears to be empty")
+
     train_dataset.to_pandas().head()
 
     # https://huggingface.co/docs/transformers/en/model_doc/auto#transformers.AutoTokenizer.from_pretrained
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -44,6 +44,37 @@ class TaxonomyReadingException(Exception):
     """An exception raised during reading of the taxonomy."""
 
 
+class Message(TypedDict):
+    """
+    Represents a message within an AI conversation.
+    """
+
+    content: str
+    role: str  # one of: 'user', 'assistant', or 'system'
+
+
+class MessageSample(TypedDict):
+    """
+    Represents a sample datapoint for a dataset using the HuggingFace messages format.
+    """
+
+    messages: List[Message]
+    group: str
+    dataset: str
+    metadata: str
+
+
+class LegacyMessageSample(TypedDict):
+    """
+    Represents a legacy message sample within an AI conversation.
+    This is what is currently used by the legacy training methods such as Linux training and MacOS training.
+    """
+
+    system: str
+    user: str
+    assistant: str
+
+
 def macos_requirement(echo_func, exit_exception):
     """Adds a check for MacOS before running a method.
 
@@ -564,3 +595,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
     if not hostname or not port:
         raise ValueError(f"Invalid host-port string: '{hostport}'")
     return hostname, port
+
+
+def convert_messages_to_legacy_dataset(
+    dataset: List[MessageSample],
+) -> List[LegacyMessageSample]:
+    """
+    Converts the new HuggingFace messages dataset format to the legacy format.
+
+    **Note**: The legacy dataset format assumes only a turn of 1. All extra turns will be dropped.
+    This means that we only look at the first 3 messages, and everything afterwards will be ignored.
+    """
+    converted_dataset: List[LegacyMessageSample] = []
+    for dp in dataset:
+        # in new dataset, the roles of a message will be determined.
+        if len(dp["messages"]) < 3:
+            raise ValueError(
+                "Dataset cannot be missing messages from roles. Please make sure that a message from each role is present."
+            )
+
+        converted: LegacyMessageSample = {  # type: ignore
+            m["role"]: m["content"] for m in dp["messages"][:3]
+        }
+        converted_dataset.append(converted)
+    return converted_dataset
+
+
+def is_messages_dataset(
+    dataset: List[MessageSample] | List[LegacyMessageSample],
+) -> bool:
+    """
+    Indicates whether or not the provided dataset is using the newer "messages" format
+    or the legacy format used by the old linux training script.
+    """
+    if len(dataset) == 0:
+        raise ValueError("dataset is empty")
+
+    return bool("messages" in dataset[0])
+
+
+def ensure_legacy_dataset(
+    dataset: List[MessageSample] | List[LegacyMessageSample],
+) -> List[LegacyMessageSample]:
+    """
+    Given a dataset that's either in the HF messages format or the legacy ilab train format,
+    ensure that the returned dataset is always in the legacy ilab train format.
+    """
+    if not is_messages_dataset(dataset):
+        return dataset  # type: ignore
+
+    return convert_messages_to_legacy_dataset(dataset)  # type: ignore
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
-from enum import Enum
 from functools import cache, wraps
 from importlib import resources
 from importlib.abc import Traversable
@@ -45,23 +44,14 @@ class TaxonomyReadingException(Exception):
     """An exception raised during reading of the taxonomy."""
 
 
-class MessageRole(Enum):
-    """
-    Defines the known roles used in message samples.
-    """
-
-    SYSTEM = "system"
-    USER = "user"
-    ASSISTANT = "assistant"
-
-
 class Message(TypedDict):
     """
     Represents a message within an AI conversation.
     """
 
     content: str
-    role: MessageRole
+    # one of: "user", "assistant", or "system"
+    role: str
 
 
 class MessageSample(TypedDict):
@@ -613,6 +603,7 @@ def convert_messages_to_legacy_dataset(
 ) -> List[LegacyMessageSample]:
     """
     Converts the new HuggingFace messages dataset format to the legacy format.
+    For reference, see: https://huggingface.co/docs/transformers/en/chat_templating#templates-for-chat-modelos
 
     **Note**: The legacy dataset format assumes only a turn of 1. All extra turns will be dropped.
     This means that we only look at the first 3 messages, and everything afterwards will be ignored.
@@ -622,7 +613,7 @@ def convert_messages_to_legacy_dataset(
         # in new dataset, the roles of a message will be determined.
         if len(dp["messages"]) < 3:
             raise ValueError(
-                "Dataset cannot be missing messages from roles. Please make sure that a message from each role is present."
+                "The dataset is expecting a minimum of 3 messages in each sample."
             )
 
         converted: LegacyMessageSample = {  # type: ignore
@@ -639,10 +630,10 @@ def is_messages_dataset(
     Indicates whether or not the provided dataset is using the newer "messages" format
     or the legacy format used by the old linux training script.
     """
-    if len(dataset) == 0:
+    if not dataset:
         raise ValueError("dataset is empty")
 
-    return bool("messages" in dataset[0])
+    return "messages" in dataset[0]
 
 
 def ensure_legacy_dataset(
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+from enum import Enum
 from functools import cache, wraps
 from importlib import resources
 from importlib.abc import Traversable
@@ -44,13 +45,23 @@ class TaxonomyReadingException(Exception):
     """An exception raised during reading of the taxonomy."""
 
 
+class MessageRole(Enum):
+    """
+    Defines the known roles used in message samples.
+    """
+
+    SYSTEM = "system"
+    USER = "user"
+    ASSISTANT = "assistant"
+
+
 class Message(TypedDict):
     """
     Represents a message within an AI conversation.
     """
 
     content: str
-    role: str  # one of: 'user', 'assistant', or 'system'
+    role: MessageRole
 
 
 class MessageSample(TypedDict):
--- a/tests/test_lab_train.py
+++ b/tests/test_lab_train.py
@@ -4,6 +4,7 @@
 # Standard
 from pathlib import Path
 from unittest.mock import patch
+import json
 import os
 import platform
 import sys
@@ -17,6 +18,7 @@ import pytest
 from instructlab import lab
 from instructlab.configuration import DEFAULTS
 from instructlab.train import linux_train
+from instructlab.utils import MessageRole, MessageSample
 
 INPUT_DIR = "test_generated"
 TRAINING_RESULTS_DIR = "training_results"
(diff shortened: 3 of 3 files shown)


src/instructlab/utils.py
```diff
@@ -564,3 +595,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+            }
+        )
+    return converted_dataset
+
+
+def is_messages_dataset(
+    dataset: List[MessageSample] | List[LegacyMessageSample],
+) -> bool:
+    if len(dataset) == 0:
+        raise ValueError("cannot determine type of empty dataset")
```
leseb, line 632:
```suggestion
        raise ValueError("dataset is empty")
```

src/instructlab/utils.py
```diff
@@ -44,6 +44,37 @@ class TaxonomyReadingException(Exception):
     """An exception raised during reading of the taxonomy."""
 
 
+class Message(TypedDict):
+    """
+    Represents a message within an AI conversation.
+    """
+
+    content: str
+    role: str  # one of: 'user', 'assistant', or 'system'
```
leseb, line 53: How about using an `Enum` for this? So that we can easily validate the presence of 'user', 'assistant', or 'system'. This could look like this:

```python
from typing import TypedDict, Optional
from enum import Enum

class Role(Enum):
    USER = 'user'
    ASSISTANT = 'assistant'
    SYSTEM = 'system'

class Message(TypedDict):
    """
    Represents a message within an AI conversation.
    """
    content: str
    role: Role

    def __post_init__(self):
        if self.role not in Role:
            raise ValueError(f"Invalid role: {self.role}")
```

src/instructlab/utils.py
```diff
@@ -564,3 +595,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+            {
+                "system": system,
+                "user": user,
+                "assistant": assistant,
+            }
+        )
+    return converted_dataset
+
+
+def is_messages_dataset(
```
leseb, line 628: I think the function name could be clearer. Like  `get_messages_from_dataset()`?

leseb (dismissed):
At this moment, I cannot verify if the code functions as intended without unit tests. Please add unit tests. Thanks!

src/instructlab/utils.py
RobotSail: @leseb We shouldn't do this because `TypedDict` is intended for type-hinting and is just a dict under the hood. Therefore we wouldn't want the overhead of explicitly instantiating new objects for this type just to process a list.

src/instructlab/utils.py
RobotSail: I like this, thanks!

src/instructlab/utils.py
RobotSail: For a boolean function?

RobotSail: @leseb Or are you talking about all of the functions overall? GitHub is just showing me your comment on `is_messages_dataset`.

leseb: Argh, I missed that part. I guess it was too early for me. :). DIsmiss what I said.

src/instructlab/utils.py
```diff
@@ -564,3 +595,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+    dataset: List[MessageSample],
+) -> List[LegacyMessageSample]:
+    """
+    Converts the new HuggingFace messages dataset format to the legacy format.
+    **Note**: The legacy dataset format assumes only a turn of 1. All extra turns will be dropped.
+    """
+    converted_dataset: List[LegacyMessageSample] = []
+    for dp in dataset:
+        # in new dataset, the roles of a message will be determined.
+        if len(dp["messages"]) < 3:
```
leseb, line 610:
```suggestion
        if len(dp["messages"]) != 3:
```

We always expect 3 entries, no less no more, right?

src/instructlab/utils.py
```diff
@@ -564,3 +595,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+            raise ValueError(
+                "Dataset cannot be missing messages from roles. Please make sure that a message from each role is present."
+            )
+        system = dp["messages"][0]["content"]
+        user = dp["messages"][1]["content"]
+        assistant = dp["messages"][2]["content"]
```
leseb, line 616: Is there a way to ensure `["messages"][0]` is always `system` and so on? Additionally, shouldn't we validate the presence of `["content"]` too?

src/instructlab/utils.py
```diff
@@ -44,6 +44,37 @@ class TaxonomyReadingException(Exception):
     """An exception raised during reading of the taxonomy."""
 
 
+class Message(TypedDict):
+    """
+    Represents a message within an AI conversation.
+    """
+
+    content: str
```
leseb, line 52: Where is content populated? Am I missing something?

src/instructlab/train/linux_train.py
```diff
@@ -222,6 +223,8 @@ def linux_train(
     )
 
     test_dataset = load_dataset("json", data_files=os.fspath(test_file), split="train")
+    train_dataset = ensure_legacy_dataset(train_dataset)
```
leseb, line 226: I think you need to wrap the call to ensure_legacy_dataset in a try-except block otherwise we can't gently catch the `ValueError` from `is_messages_dataset`.

src/instructlab/utils.py
RobotSail: I originally thought this, but if in the future we start having multi-turn samples with over 3 messages, we shouldn't immediately discard all of them, but should instead drop everything after the first 3.

src/instructlab/utils.py
RobotSail: Yeah, I inspected the format of our dataset in a separate Jupyter notebook and found that it exclusively used this order. The original implementation I had didn't assume this order, but it's likelier that we'll get rid of the legacy linux train before we fundamentally restructure our dataset format.

src/instructlab/utils.py
RobotSail: It's one of the fields that gets populated within the dataset generated by SDG.

src/instructlab/train/linux_train.py
RobotSail: Good call, I'll add that.

src/instructlab/utils.py
RobotSail: Okay, no worries.

src/instructlab/utils.py
RobotSail: For context, the way `TypedDicts` work is you just declare a class with the fields it should have and then you place a hint on a dict that you go to use. See this example from [their documentation](https://mypy.readthedocs.io/en/stable/typed_dict.html):

```py
from typing import TypedDict

Movie = TypedDict('Movie', {'name': str, 'year': int})

movie: Movie = {'name': 'Blade Runner', 'year': 1982}
```

src/instructlab/utils.py
RobotSail: So I added it but I think it will be best to remove it. It's causing too many problems with MyPy and JSON serialization during unit testing. It'll be easier to just have it as a string with the comment above indicating its usage.

RobotSail: I'm keeping it for now but look at the MyPy failure: https://github.com/instructlab/instructlab/actions/runs/9900406125/job/27351128053?pr=1644#step:8:30

The alternative is to use the key directly instead of using `.value`, but this causes a wider issue where JSON cannot serialize an enum directly. While we can technically add code to make this happen, it feels like we're adding a lot of complexity where we could instead just keep it simple.

src/instructlab/utils.py
RobotSail: Btw I fixed this to instead use a dict comprehension, since the final object will be a dict anyway, the order won't matter and therefore we can just form the object by doing:

```py
sample = {m["role"]: m["content"] for m in dp["messages"]}
```

RobotSail:
@bbrowning This doesn't have support yet for the SDG 1.5 format that's being developed. This is only so that we can have the messages format be compatible with the legacy format. 
We should keep this PR scoped to just messages and tackle SDG 1.5 in a follow-up.

RobotSail:
@bbrowning I have the code working for converting a dataset from SDG with pretraining messages into the legacy format. Whenever we merge this PR, I can make a followup PR for that.

tests/test_lab_train.py
```diff
@@ -208,6 +210,87 @@ def test_invalid_data_dir(self, cli_runner: CliRunner):
...
+        )
+        assert result.exit_code == 0
+        convert_llama_to_gguf_mock.assert_called_once()
+        assert convert_llama_to_gguf_mock.call_args[1]["model"] == Path(
+            "training_results/final"
+        )
+        assert convert_llama_to_gguf_mock.call_args[1]["pad_vocab"] is True
+        assert len(convert_llama_to_gguf_mock.call_args[1]) == 2
+        linux_train_mock.assert_called_once()
+        print(linux_train_mock.call_args[1])
```
leseb, line 280: leftover?

RobotSail: Removed

src/instructlab/train/linux_train.py
```diff
@@ -222,6 +225,12 @@ def linux_train(
     )
 
     test_dataset = load_dataset("json", data_files=os.fspath(test_file), split="train")
+    try:
+        train_dataset = ensure_legacy_dataset(train_dataset)
+        test_dataset = ensure_legacy_dataset(test_dataset)
+    except ValueError:
+        ctx.fail("failed to run training, dataset appears to be empty")
```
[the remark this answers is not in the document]
RobotSail: Done

src/instructlab/train/linux_train.py
```diff
@@ -222,6 +225,12 @@ def linux_train(
     )
 
     test_dataset = load_dataset("json", data_files=os.fspath(test_file), split="train")
+    try:
+        train_dataset = ensure_legacy_dataset(train_dataset)
+        test_dataset = ensure_legacy_dataset(test_dataset)
+    except ValueError as e:
+        ctx.fail(f"failed to run training, dataset appears to be empty: {e}")
```
cdoern, line 232: is this always if empty?

tests/test_lab_train.py
```diff
@@ -208,6 +210,86 @@ def test_invalid_data_dir(self, cli_runner: CliRunner):
...
+                "metadata": "test",
+            }
+        ]
+        dataset_contents = ""
+        for line in new_dataset:
+            json_str = json.dumps(line)
+            dataset_contents += json_str + "\n"
+
+        with open(
+            os.path.join(DEFAULTS.DATASETS_DIR, "test_dataset.jsonl"),
```
cdoern, line 250: so these will overwrite whatever was there? is there an `exists_ok`?

cdoern (commented):
looks sane to me. just a few comments. Thanks for the effort here

src/instructlab/train/linux_train.py
```diff
@@ -222,6 +225,12 @@ def linux_train(
     )
 
     test_dataset = load_dataset("json", data_files=os.fspath(test_file), split="train")
+    try:
```
booxter, line 228: Can you please enlighten me why this is specific to linux training? Thought the output files were agnostic to platform. (If there's a legit reason, perhaps add a comment with an explanation?)

src/instructlab/utils.py
```diff
@@ -564,3 +596,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+    dataset: List[MessageSample] | List[LegacyMessageSample],
+) -> bool:
+    """
+    Indicates whether or not the provided dataset is using the newer "messages" format
+    or the legacy format used by the old linux training script.
+    """
+    if len(dataset) == 0:
+        raise ValueError("dataset is empty")
+
+    return bool("messages" in dataset[0])
```
booxter, line 635: Why do we need to cast it to `bool`? The `in` operator already guarantees that the result is `bool`. I checked if maybe `mypy` complains, - it doesn't care either.

src/instructlab/utils.py
```diff
@@ -564,3 +596,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+
+
+def is_messages_dataset(
+    dataset: List[MessageSample] | List[LegacyMessageSample],
+) -> bool:
+    """
+    Indicates whether or not the provided dataset is using the newer "messages" format
+    or the legacy format used by the old linux training script.
+    """
+    if len(dataset) == 0:
```
booxter, line 632: nit: `if not dataset` would be a bit more pythonic.

src/instructlab/train/linux_train.py
booxter: +++ There are other cases of (malformed) files that you already handle in the code. Logging the `{e}` will already give details about what happened exactly. Just say you `Failed to parse dataset: {e}`.

src/instructlab/utils.py
```diff
@@ -44,6 +44,38 @@ class TaxonomyReadingException(Exception):
...
 
 
+class Message(TypedDict):
+    """
+    Represents a message within an AI conversation.
+    """
+
+    content: str
+    # one of: "user", "assistant", or "system"
+    role: str
```
booxter, line 54: Can you do an enum here then?

```
from enum import Enum

class Role(Enum):
    USER = "user"
    ASSISTANT = "assistant"
    SYSTEM = "system"

class Message(TypedDict):
    ...
    role: Role
```

src/instructlab/utils.py
```diff
@@ -564,3 +596,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
     if not hostname or not port:
         raise ValueError(f"Invalid host-port string: '{hostport}'")
     return hostname, port
+
+
+def convert_messages_to_legacy_dataset(
+    dataset: List[MessageSample],
+) -> List[LegacyMessageSample]:
+    """
+    Converts the new HuggingFace messages dataset format to the legacy format.
```
booxter, line 605: Could you please add a link to where the HF format is defined, for reference?

src/instructlab/utils.py
```diff
@@ -564,3 +596,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+
+    **Note**: The legacy dataset format assumes only a turn of 1. All extra turns will be dropped.
+    This means that we only look at the first 3 messages, and everything afterwards will be ignored.
+    """
+    converted_dataset: List[LegacyMessageSample] = []
+    for dp in dataset:
+        # in new dataset, the roles of a message will be determined.
+        if len(dp["messages"]) < 3:
+            raise ValueError(
+                "Dataset cannot be missing messages from roles. Please make sure that a message from each role is present."
```
booxter, line 615: I am not a native speaker, but I think the grammar here is not ideal. "Dataset cannot be missing messages from roles" -> "Not enough messages in the dataset"?

src/instructlab/utils.py
```diff
@@ -564,3 +596,53 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+            )
+
+        converted: LegacyMessageSample = {  # type: ignore
+            m["role"]: m["content"] for m in dp["messages"][:3]
+        }
+        converted_dataset.append(converted)
```
booxter, line 621: nit: There is little use to construct the list only to fail in the middle because one of the datasets doesn't have enough `messages` (line 614). You could probably split into two steps: first validate (and raise if any malformed messages are found); and then calculate / construct the result.

src/instructlab/train/linux_train.py
booxter, line 232: There are other cases where ValueError is raised.

tests/test_lab_train.py
```diff
@@ -208,6 +210,86 @@ def test_invalid_data_dir(self, cli_runner: CliRunner):
             assert "Could not read from data directory" in result.output
             assert result.exit_code == 1
 
+    @patch("instructlab.utils.is_macos_with_m_chip", return_value=False)
```
booxter, line 213: Opinion: it would be better to target the new functions in `utils.py` directly with a test - we would not need any mocks, nor we would need to trigger `ilab` through CLI, while still covering the cases we care about - the functions are pure and easy to test without any mocks.

booxter (changes requested):

tests/test_lab_train.py
RobotSail: @cdoern This is after `HOME` is set to the temporary dataset with the [`tmp_path_home`](https://github.com/instructlab/instructlab/blob/c6adcb5ddd84d9358a2db56c5f30b237420f9b13/tests/conftest.py#L53) fixture, which we then use in this test via the `cli_runner` fixture. It should never be touching any of the actual user's datasets.

src/instructlab/train/linux_train.py
RobotSail: `ValueError` is what we're specifically throwing here when the dataset is empty. I'll update this line per @booxter's recommendation to be more general.

tests/test_lab_train.py
RobotSail: FYI: The `open` function doesn't provide an `exists_ok` parameter. If you'd like, we can run an explicit check before writing to this file, but it shouldn't happen as everything will be running in a temporary directory.

tests/test_lab_train.py
RobotSail: While this is definitely an approach which can work well for stable functions, much of this CLI will change over time, so adding fine-grained tests like these will tightly couple us to the functionality of what are essentially scaffolding utils. Therefore, it makes the most sense for the time being to focus on functional and end-to-end tests which allow us to confirm the behavior of what we're writing. 

However; if others feel strongly that this should be a blocker for 0.18.0, we can definitely spend time on this.

src/instructlab/utils.py
RobotSail: Thank you for the clarification, I will update this.

src/instructlab/train/linux_train.py
booxter: Thanks. A docstring in `utils.py` says that macos training also use(d?) the legacy format. Is it true? Should macos implementation also accept the legacy format? To quote, `This is what is currently used by the legacy training methods such as Linux training and MacOS training.`

RobotSail: We should tackle this in a follow-up PR. This PR is mainly so we can unblock other work and avoid having to support the legacy format in SDG.

src/instructlab/utils.py
RobotSail: Good point, I'll update this.

src/instructlab/utils.py
RobotSail: Good catch, your suggestion is simpler and more pythonic.

src/instructlab/utils.py
RobotSail: Sure, I updated it 🙂.

src/instructlab/utils.py
RobotSail: We've discussed this in a previous conversation which you can read here: https://github.com/instructlab/instructlab/pull/1644#discussion_r1671962251

The TLDR is that while an Enum helps us narrow down the values, it adds too much additional complexity and headache when a simple `str` would serve the job well.

src/instructlab/utils.py
RobotSail: What is the concern here about constructing the dataset inline? It would have to be done either way, and the memory will get GC'd once it's no longer being used.

src/instructlab/train/linux_train.py
RobotSail: This was updated to be more general.

src/instructlab/train/linux_train.py
booxter: Fine by me. Is there an issue reported for macosx to support legacy?

src/instructlab/utils.py
booxter: The concern is minor because I don't think malformed messages would happen too often.

But in theory, if your last `messages` has `< 3` entries, the function will raise an exception; and all the work to construct `converted_dataset` done so far would be for naught. That's why it could make sense to first validate the input is good; and only then construct `converted_dataset` (you could also then build the result list in one go with a list comprehension instead of calling `.append` on each item - the latter is more expensive.)

The code would then look smth like:

```
if any(len(dp["messages"]) < 3 for dp in dataset):
    raise ...

return [
    {m["role"]: m["content"] for m in dp["messages"][:3]}
    for dp in dataset
]
```

RobotSail: While I appreciate your concern around wanting our code to be efficient, I would challenge how expensive this operation is. For example, see the following snippet:

```py
times = []
for k in range(1000):
    t1 = time.time_ns()
    legacy_ds = ensure_legacy_dataset(ds)
    t2 = time.time_ns()
    times.append((t2 - t1) / (1000 ** 3))

average = sum(times) / len(times)
print(f'average time taken: {average:.4f} sec')
print(f"dataset length: {len(ds)}")
print(f"approximate time for a 1,000,000 sample dataset: {average / len(ds) * 1_000_000:.4f} secs")

> average time taken: 0.0121 sec
> dataset length: 24545
> approximate time for a 1,000,000 sample dataset: 0.4947 secs
```

Furthermore, this project revolves around training AI models - which requires a lot of time and computational resource. I would say that it's a much better use of our time to focus on making that more efficient than to try and optimize these one-off operations.

booxter: That's just a nit, take it or leave it. :)

booxter (approved):
An issue for macosx legacy training should be reported. Otherwise, it's ok.

RobotSail:
@booxter Issue for it has been created here: https://github.com/instructlab/instructlab/issues/1837

cdoern (approved):
lgtm

jaideepr97 (approved):
lgtm thanks @RobotSail

PR #1645, opened 2024-07-10
e2e: Fix config update to use Mixtral

russellb:
PR #1471 changed where the config is located. This `sed` assumed
`config.yaml` was still in the current working directory. Look for it
in `${CONFIG_HOME}` instead. This variable is already set in the
script elsewhere.

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -74,7 +74,7 @@ test_init() {
 
     step Checking config.yaml
     if [ "${MIXTRAL}" -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


scripts/basic-workflow-tests.sh
```diff
@@ -74,7 +74,7 @@ test_init() {
 
     step Checking config.yaml
     if [ "${MIXTRAL}" -eq 1 ]; then
-        sed -i -e 's/models\/merlinite.*/models\/mixtral-8x7b-instruct-v0\.1\.Q4_K_M\.gguf/' config.yaml
```
RobotSail, line 77: The default model will now be an absolute path just FYI. This command should still work, but you can drop the `models/` prefix as it's now redundant.

RobotSail (approved):
Just one comment but otherwise LGTM.

n1hility (approved):

PR #1646, opened 2024-07-10
Automatically trigger e2e on AWS

russellb:
This PR is the next step of replacing the `e2e` job that runs on a GitHub hosted
runner in favor of running it on AWS. These changes get the job running
automatically just like the original `e2e`, but does not block merges on it for
now. We can let this run in parallel for a short time to make sure it's stable
before switching over to it completely.

e86902f ci: Run e2e-nvidia-t4-x1.yml automatically

commit e86902f62765d1c648c256b3656a4ac8a4a04449
Date:   Mon Jul 15 17:19:58 2024 -0400

    ci: Run e2e-nvidia-t4-x1.yml automatically
    
    This makes our smallest AWS-based e2e workflow run by default on PRs.
    The hardware configuration is roughly equivalent to the existing e2e
    job that runs by default on PRs on a github-hosted runner.
    
    I propose running this in parallel with the existing e2e workflow for
    a little while to evaluate its stability. The mergify configuraiton
    has not been updated so it will not block on the results of this new
    e2e variant.
    
    One detail of this workflow that is very important to note is that it
    is triggered on the `pull_request_target` event type instead of
    `pull_request`. The difference is that the workflow runs using the
    version checked in to `main` and does not use any code from the PR by
    default. This is for security reasons. We do not want any secrets
    exposed to code from PRs.
    
    The middle job of this workflow, e2e, runs on an ephemeral AWS worker.
    This workflow has no access to any secrets. This is the only place we
    switch over to code in the PR. This should be safe, as long as we are
    careful not to add any use of secrets to this part of the workflow.
    All secret usage must be kept outside of the part that runs on the AWS
    VM.
    
### DIFF
--- a/.github/workflows/e2e-nvidia-t4-x1.yml
+++ b/.github/workflows/e2e-nvidia-t4-x1.yml
@@ -3,12 +3,18 @@
 name: E2E (NVIDIA Tesla T4 x1)
 
 on:
-  workflow_dispatch:
-    inputs:
-      pr_or_branch:
-        description: 'pull request number or branch name'
-        required: true
-        default: 'main'
+  push:
+    branches:
+      - main
+      - release-*
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+    branches:
+      - main
+      - release-*
 
 jobs:
   start-runner:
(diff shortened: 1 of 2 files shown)


russellb:
I'm not totally sure why this isn't working. I think this permission must not be accessible in a PR that's modifying the trigger workflow. I did have this working on a personal repo before proposing here ... but it was running after having been applied to main already

russellb:
to whoever considers placing the 2nd approval on here, can you ping me to make sure I'm available to respond quickly to what happens when it starts running against every PR?  You can also place the `hold` label before giving the 2nd approval. That way I can let it merge when I'm ready to focus on monitoring and response

danmcp (approved):

russellb:
> to whoever considers placing the 2nd approval on here, can you ping me to make sure I'm available to respond quickly to what happens when it starts running against every PR? You can also place the `hold` label before giving the 2nd approval. That way I can let it merge when I'm ready to focus on monitoring and response

I went ahead added `hold` myself so I can control the merge timing.

danmcp (approved):

russellb:
I had a conversation with @n1hility offline about these changes. He pointed me to another event type that can trigger workflows.  https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

This might be a better and simpler solution to how this trigger works, so I'm going to look into that next.

danmcp (commented):
Removing approval until the alternatives are researched

.github/workflows/e2e-nvidia-t4-x1.yml
```diff
@@ -3,12 +3,18 @@
 name: E2E (NVIDIA Tesla T4 x1)
 
 on:
-  workflow_dispatch:
-    inputs:
-      pr_or_branch:
-        description: 'pull request number or branch name'
-        required: true
-        default: 'main'
```
nathan-weinberg, line 11: The regular `e2e.yml` job has workflow dispatch still - we could probably remove this from that job as well?

nathan-weinberg (approved):

.github/workflows/e2e-nvidia-t4-x1.yml
russellb: yes, we can remove that now.

nathan-weinberg (approved):

nathan-weinberg:
Reapproving after code updates

PR #1650, opened 2024-07-10
e2e: Fix path to config file

russellb:
017414f e2e: Fix path to config file
07e6607 e2e: Always create a new config file

commit 017414f497627d9834575e0cf2b17afa9b11d125
Date:   Wed Jul 10 08:00:53 2024 -0400

    e2e: Fix path to config file
    
    This is related to changes made in #1471.
    
    PR #1645 was a previous attempt at fixing this, but the path wasn't
    quite right. It missed `instructlab/` within the base config dir.
    
commit 07e66073b8517def46b51c7f15aa6959952cb392
Date:   Wed Jul 10 09:29:21 2024 -0400

    e2e: Always create a new config file
    
    The previous code tried to only create a config file if it did not
    already exist. This script has evolved to assume it has ownership of
    the config, so we should always be creating it. We want to test that
    part of the workflow, as well.
    
    This came up when it was pointed out that the check for an existing
    file was no longer valid after #1471.
    
### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -114,7 +114,7 @@ test_init() {
 
     step Checking config.yaml
     if [ "${MIXTRAL}" -eq 1 ]; then
-        sed -i -e 's/models\/merlinite.*/models\/mixtral-8x7b-instruct-v0\.1\.Q4_K_M\.gguf/' "${CONFIG_HOME}/config.yaml"
+        sed -i -e 's/models\/merlinite.*/models\/mixtral-8x7b-instruct-v0\.1\.Q4_K_M\.gguf/' "${CONFIG_HOME}/instructlab/config.yaml"
     fi
 }
 
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -110,7 +110,7 @@ test_smoke() {
 
 test_init() {
     task Initializing ilab
-    [ -f config.yaml ] || ilab config init --non-interactive
+    ilab config init --non-interactive
 
     step Checking config.yaml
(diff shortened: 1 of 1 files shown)


scripts/basic-workflow-tests.sh
```diff
@@ -114,7 +114,7 @@ test_init() {
 
     step Checking config.yaml
     if [ "${MIXTRAL}" -eq 1 ]; then
-        sed -i -e 's/models\/merlinite.*/models\/mixtral-8x7b-instruct-v0\.1\.Q4_K_M\.gguf/' "${CONFIG_HOME}/config.yaml"
+        sed -i -e 's/models\/merlinite.*/models\/mixtral-8x7b-instruct-v0\.1\.Q4_K_M\.gguf/' "${CONFIG_HOME}/instructlab/config.yaml"
```
leseb, line 117: Is line 113 still valid though? I mean the `[ -f config.yaml ]`, does it need to be updated too with `"${CONFIG_HOME}/instructlab/config.yaml"`?

RobotSail: That's a good point, it would evaluate to false since the e2e script is cannot possible be executed in the temp directory as it's something we create during execution.

RobotSail (approved):

RobotSail:
I believe @leseb's comment needs to be addressed but otherwise LGTM.

scripts/basic-workflow-tests.sh
russellb: not related to this PR, but I'll go ahead and fix it.

russellb: done.

leseb (approved):

PR #1651, opened 2024-07-10
e2e: switch to AWS-based runner model

russellb:
The GitHub GPU runner we have been using is very
expensive and necessitates a change to a cheaper
model

Switches GH-based runner over to an equivalent
AWS-based runner and updates relevant docs and
Mergify configuration

Co-authored-by: Russell Bryant <rbryant@redhat.com>
Signed-off-by: Nathan Weinberg <nweinber@redhat.com>


russellb:
I placed a hold because I'd really like to get #1646 in first.  The risk is too high to turn this off first.

.github/mergify.yml
```diff
@@ -35,14 +35,14 @@ pull_request_rules:
           - files~=^requirements.*\.txt$
```
russellb, line 35: You may need to change the `check-success` line a few lines up. You can test out the config in the mergify dashboard to see if it works like you expect

nathan-weinberg: I added an `e2e-workflow-complete` to match the existing `lint-workflow-complete` and `test-workflow-complete` - will double check the Mergify dashboard as well

docs/ci.md
```diff
@@ -66,32 +63,38 @@ Here is an example of using the GitHub Web UI to launch an E2E workflow:
 
 The project currently supports the usage of the following runners for the E2E jobs:
 
-* [GitHub built-in GPU runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#specifications-for-gpu-larger-runners) -- referred to as `ubuntu-gpu` in our workflow files Only `e2e.yml` uses this runner.
-* Ephemeral GitHub runners launched on demand on AWS. Most workflows work this way, granting us access to a wider variety of infrastructure at lower cost.
+* Ephemeral GitHub runners launched on demand on AWS. All workflows currently work this way, granting us access to a wider variety of infrastructure at lower cost.
```
danmcp, line 66: Is there a reason we need to make this statement here?  Particularly about cost.

docs/ci.md
```diff
@@ -66,32 +63,38 @@ Here is an example of using the GitHub Web UI to launch an E2E workflow:
...
+
+We previously also used a  [GitHub built-in GPU runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#specifications-for-gpu-larger-runners) referred to as `ubuntu-gpu` -- this was deprecated due to cost.
```
danmcp, line 68: Same question about whether this context is helpful to future users.

docs/ci.md
nathan-weinberg: This is holdover from what already existed - honestly, I'd be fine to ax this whole section on runners if we want, as this information is in other parts of the doc as well.

docs/ci.md
nathan-weinberg: I think if we keep this section it's worth having just for historic context, but like I said above I'd also be fine to ax this whole section in the interest of keeping the doc a manageable length.

docs/ci.md
nathan-weinberg: To be clear, when I say "this section" I meant the entire `Current GPU-enabled Runners` section

danmcp: Is it helpful for contributors?  Or stated another way, what would they do with this info?

danmcp: I would vote for either removing or stating it runs on AWS but leaving off the history and cost statements.

khaledsulayman (approved):

docs/ci.md
nathan-weinberg: I removed it

jaideepr97 (approved):

.github/workflows/e2e-nvidia-t4-x1.yml
```diff
@@ -121,3 +126,12 @@ jobs:
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
           label: ${{ needs.start-runner.outputs.label }}
           ec2-instance-id: ${{ needs.start-runner.outputs.ec2-instance-id }}
+
+  e2e-workflow-complete:
+    # we don't want to block PRs on failed EC2 cleanup
+    # so not requiring "stop-runner" as well
+    needs: ["start-runner", "e2e"]
```
bjhargrave, line 133: You only need `e2e` here since `e2e` itself depends upon `start-runner`. And in that case, there is no need then for this workflow complete job since `e2e` is sufficient.

alinaryan (approved):

danmcp (approved):

nathan-weinberg:
@mergifyio backport release-v0.18

PR #1652, opened 2024-07-10
docs: details for mac prerequisite instructions

tarilabs:
Hi 👋 
especially as Mac OSX ships with python, many use pyenv to avoid conflicting with the system's python.

Following the current command instructions on a system with pyenv would actually return

```
Bourne-Again shell script text executable, ASCII text
```

of the shim instead of the executable.

image

This PR proposes for newcomers another command to try,
instead of maybe stumbling on the first difference in the setup prerequisite check :) 

wdyt ?

**Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,8 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 
    #### Install with Apple Metal on M1/M2/M3 Macs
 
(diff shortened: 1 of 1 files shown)


leseb (approved):
Thanks for reporting this.

danmcp (approved):

PR #1653, opened 2024-07-10
training additional_args

cdoern:
This work introduces additional_args a new part of the training config which holds less commonly edited options. Users can edit

```
additional_args {

}
```
to their _train config class to set some of these options.

additional_args also lives in persistent storage in  ~/.local/share/instructlab/internal/train_configuration/additional/additional_args.yaml This involved creating a few new subdirectories.

_train has been flattened since all of the torch_args were moved into the additional_args this is what a template config.yaml train section currently looks like:

```
train:
  ckpt_output_dir: /home/cdoern/.cache/instructlab/checkpoints
  data_output_dir: /home/cdoern/.local/share/instructlab/internal
  data_path: /home/cdoern/.local/share/instructlab/datasets
  deepspeed_cpu_offload_optimizer: false
  effective_batch_size: 3840
  additional_args: {}
  is_padding_free: false
  lora_quantize_dtype: nf4
  lora_rank: 4
  max_batch_len: 10000
  max_seq_len: 4096
  nproc_per_node: 1
  num_epochs: 10
  save_samples: 250000
```

the additional args I have moved out of _train so far are:

```
deepspeed_cpu_offload_optimizer_pin_memory: false
deepspeed_cpu_offload_optimizer_ratio: 1
learning_rate: 2.0e-05
lora_alpha: 32
lora_dropout: 0.1
lora_target_modules:
- q_proj
- k_proj
- v_proj
- o_proj
nnodes: 1
node_rank: 0
random_seed: 42
rdzv_id: 123
warmup_steps: 25
```

These additional_args are still populated into the default_map for the ilab model train flags. They will not, however, show up as defaults in the config.yaml. If the users adds an override for one of these options in their additional_args portion of the config.yaml, we will honor that.

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -9,6 +9,7 @@ import sys
 import typing
 
 # Third Party
+# pylint: disable=ungrouped-imports
 from instructlab.training import (
     DeepSpeedOptions,
     LoraOptions,
@@ -147,6 +148,26 @@ class _InstructlabDefaults:
     @property
     def EVAL_DATA_DIR(self) -> str:
         return path.join(self.INTERNAL_DIR, "eval_data")
+
+    @property
+    def TRAIN_CONFIG_DIR(self) -> str:
+        return path.join(self.INTERNAL_DIR, "train_configuration")
+
+    @property
+    def TRAIN_PROFILE_DIR(self) -> str:
+        return path.join(self.TRAIN_CONFIG_DIR, "profiles")
+
+    @property
+    def TRAIN_ADDITIONAL_OPTIONS_DIR(self) -> str:
+        return path.join(self.TRAIN_CONFIG_DIR, "additional")
+
+    @property
+    def TRAIN_ADDITIONAL_OPTIONS_FILE(self) -> str:
+        return path.join(self.TRAIN_ADDITIONAL_OPTIONS_DIR, "additional_args.yaml")
+
+    @property
+    def TRAIN_DEFAULT_PROFILE(self) -> str:
+        return path.join(self.TRAIN_PROFILE_DIR, "default.yaml")
 
 
 DEFAULTS = _InstructlabDefaults()
@@ -320,8 +341,31 @@ class _evaluate(BaseModel):
 
 
 class _train(BaseModel):
-    train_args: TrainingArgs
-    torch_args: TorchrunArgs
+    # model configuration
+    model_config = ConfigDict(extra="ignore", protected_namespaces=())
+
+    model_path: str
+
+    data_path: str
+    ckpt_output_dir: str
+    data_output_dir: str
+
+    max_seq_len: int
+    max_batch_len: int
+    num_epochs: int
+    effective_batch_size: int
+    save_samples: int
+
+    deepspeed_cpu_offload_optimizer: bool
+
+    lora_rank: int
+    lora_quantize_dtype: str
+
+    is_padding_free: bool
+
+    nproc_per_node: int
+
+    additional_args: dict[str, str]
 
 
 class Config(BaseModel):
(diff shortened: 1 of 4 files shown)


src/instructlab/configuration.py
```diff
@@ -560,6 +609,49 @@ def init(ctx: click.Context, config_file: str | os.PathLike[str]) -> None:
...
+        train_additional = ctx.default_map["train"]["expert_args"]
+        ctx.default_map["train"]["expert_args"] = finish_expert_train_args(
+            train_additional
+        )
         log.configure_logging(log_level=config_obj.general.log_level.upper())
     else:
         ctx.default_map = None
+
+
+def map_train_to_library(params):
```
fabianlim, line 623: @cdoern i kind of feel that while this indeed seperate out the args that are less likely for the user to change, the additional args are still very tightly coupled. This is because the `map_train_to_library` function needs to be maintained in the instructlab repo.. and every new argument that is added, this repo needs to be changed

I wonder if it is possible to `additional_args` schema-less, and pass to the `train` library to be mapped. Of course this requiers the current approach to be changed. Right now the issue is because we are forming `train_args` and `torch_args` to be passed into `run_training`

```python
        ds_args = DeepSpeedOptions(**params)
        lora_args = LoraOptions(**params)
        train_args = TrainingArgs(**params)
        torch_args = TorchrunArgs(**params)
        train_args.deepspeed_options = ds_args
        train_args.lora = lora_args
        try:
            run_training(train_args=train_args, torch_args=torch_args)
        except:
           ...
```

However, if we can somehow have only `TrainingArgs`, and just keep `addtional_args` schemaless, then make the call

```python
     train_args = TrainingArgs(**params)
     try:
            run_training(train_args=train_args, additional_args=additional_args)
      except:
           ...
```
and let the train library perform the mapping, we can minimize changes to the instructlab API. This will require the assumption that `addtional_args` are expert, since we are no longer going to expose the schema to the user.

@RobotSail @Maxusmusti @aldopareja any thoughts?

markmc:
> This work introduces `expert_args` a new part of the training config which holds less commonly edited options.

If/when you rebase/squash, please include all of this super useful information in a commit message so it's captured in `git log` for future

leseb (commented):
I'm curious, what gives us the confidence that `expert_args` won't become commonly used arguments? This doesn't seem like an easy decision to make.

Also, it's unclear to me how this will handle `additional_args`. Are we planning to have both?

Thanks!

cdoern:
> I'm curious, what gives us the confidence that `expert_args` won't become commonly used arguments? This doesn't seem like an easy decision to make.
> 
> Also, it's unclear to me how this will handle `additional_args`. Are we planning to have both?
> 
> Thanks!

@leseb good point, I should have clarified all of this! `expert_args` is `additional_args` for training. The training squad expressed interest in expert rather than additional for the naming scheme. I can update all of this in my corresponding dev-doc!

As for the confidence of which args should be "expert" I am leaving that up to the discretion of @RobotSail @Maxusmusti and @JamesKunstle . the initial set I put in here was what I first thought of :) 

Regardless, trimming the size of the cfg while still supporting all the same options is a win IMO, I am just trying to find the best way to do that

src/instructlab/configuration.py
cdoern: > I wonder if it is possible to additional_args schema-less, and pass to the train library to be mapped. 

this scares me a bit. There needs to be some sort of accountability between the library and the cli in terms of breaking changes.

I went with this approach because it decreases the size of the cfg, makes it easier to understand, and allows us to be a bit less strict with the training library. 

I believe the mapping should probably happen in the CLI for accountability purposes but I am open to the training squad's opinions on this!

fabianlim: well. even if you put a schema on the args, the `training` repo can still do some pretty bad things internally. I think maybe if the proper e2e tests are devised it will give more confidence

src/instructlab/configuration.py
```diff
@@ -359,40 +404,24 @@ def get_default_config() -> Config:
...
+            effective_batch_size=3840,
+            save_samples=250000,
+            lora_quantize_dtype="nf4",
+            lora_rank=4,
+            lora_alpha=32,
+            lora_dropout=0.1,
+            lora_target_modules=["q_proj", "k_proj", "v_proj", "o_proj"],
+            deepspeed_cpu_offload_optimizer=False,
+            deepspeed_cpu_offload_optimizer_ratio=1,
+            deepspeed_cpu_offload_optimizer_pin_memory=False,
```
RobotSail, line 423: - optimizer_ratio & pin_memory can move to expert args
- lora dropout, alpha, and target modules can move 
- lora_quantize_dtype should be moved but not in this PR since it requires a bigger lift. Issue for it here: https://github.com/instructlab/instructlab/issues/1673

src/instructlab/configuration.py
```diff
@@ -501,11 +530,30 @@ def ensure_storage_directories_exist():
...
+    expert_args_and_defaults = {
+        "learning_rate": 2e-6,
+        "warmup_steps": 1000,
+        "random_seed": 42,
+        "node_rank": 0,
+        "nnodes": 1,
+        "nproc_per_node": 1,
+        "rdzv_id": 123,
+        "rdzv_endpoint": "127.0.0.1:12222",
+        "is_padding_free": False,
```
RobotSail, line 550: Notes: 
- is_padding_free should get moved out
- learning_rate == 2e-5
- warmup_steps == 25
- nproc_per_node should move out of expert args

src/instructlab/configuration.py
RobotSail: @fabianlim While I don't disagree, this is something that's out of the scope of this PR, as it will be much bigger lift than what we can currently afford in terms of bandwidth.

RobotSail (approved):
LGTM

cdoern:
after discussing some UX norms with @mairin we are going with `additional_args` rather than `expert_args`

JamesKunstle (approved):

Maxusmusti (approved):
lgtm

tests/test_config.py
```diff
@@ -83,8 +83,7 @@ def _assert_model_defaults(self, cfg):
         assert cfg.serve.model_path == default_model
 
         assert cfg.train is not None
-        assert cfg.train.train_args is not None
```
leseb, line 86: Why is this line gone?

leseb: Nevermind, `cfg.train.train_args` is gone :)

PR #1671, opened 2024-07-11
feat: validate model_path from click

leseb:
Instead of re-implementing the same logic of make sure the model_path exists, is a file or a non-empty directory, we now have a Class that does it for us.
This not only help with catching errors more quickly but also unifies that approach accross various sub-commands like serve, train.

**Checklist:**

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -28,7 +28,7 @@ from instructlab.configuration import (
 )
 @click.option(
     "--model-path",
-    type=click.Path(),
+    type=click.Path(),  # we don't use utils.Pathlib() here, since this is a requested path to download a model, so the files are not guaranteed to exist yet
     default=lambda: DEFAULTS.DEFAULT_GGUF_MODEL,
     show_default="The instructlab data files location per the user's system.",
     help="Path to the model used during generation.",
--- a/src/instructlab/model/serve.py
+++ b/src/instructlab/model/serve.py
@@ -20,10 +20,10 @@ logger = logging.getLogger(__name__)
 )
 @click.option(
     "--model-path",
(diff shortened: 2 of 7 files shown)


leseb:
TODO:

- [ ] more testing
- [ ] unit tests

tiran:
`click.Path` can do must of the checks already. Here is a variant that always returns pathlib.Path and can check for non-empty directory.

```python
class Pathlib(click.Path):
    """click.Path variant with extra features

    - always returns a pathlib.Path
    - allows to check for non-empty directory
    """
    def __init__(
        self,
        exists: bool = False,
        file_okay: bool = True,
        dir_okay: bool = True,
        writable: bool = False,
        resolve_path: bool = False,
        executable: bool = False,
        dir_notempty: bool = False,
    ) -> None:
        super().__init__(
            exists=exists,
            file_okay=file_okay,
            dir_okay=dir_okay,
            writable=writable,
            readable=True,
            resolve_path=resolve_path,
            allow_dash=False,
            path_type=pathlib.Path,
            executable=executable,
        )
        self.dir_notempty = dir_notempty

    def to_info_dict(self) -> dict[str, typing.Any]:
        info_dict = super().to_info_dict()
        info_dict["dir_notempty"] = self.dir_notempty
        return info_dict

    def convert(
        self,
        value: str | os.PathLike[str],
        param: click.Parameter | None,
        ctx: click.Context | None,
    ) -> pathlib.Path:
        result: pathlib.Path = super().convert(value, param, ctx)
        if self.exists and self.dir_notempty and result.is_dir():
            try:
                next(os.scandir(result))
            except StopIteration:
                self.fail("Empty direction")
        return result

```

leseb:
I might revisit it later, in the meantime if someone wants to take over, feel free.

PR #1672, opened 2024-07-11
WIP: Introduce 'device' flag to evaluate

alinaryan:
This commit introduces a 'device' flag to evaluate, which allows for specifying a PyTorch device for running models. It also sets a default device in the config to cpu.

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -333,6 +333,11 @@ class _evaluate(BaseModel):
     gpus: Optional[int] = Field(
         default=None, description="Number of GPUs to use for running evaluation."
     )
+    device: str = Field(
+        default="auto",
+        description="PyTorch device (e.g. `cpu`, `cuda`, or `mps`) for running models.",
+        pattern="cpu|cuda|mps|auto"
+    )
     mmlu: _mmlu = Field(
         default_factory=_mmlu,
         description="MMLU benchmarking settings",
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -50,6 +50,7 @@ def get_evaluator(
     batch_size,
     tasks_dir,
     merge_system_user_message,
+    device: str,
 ) -> "Evaluator":
     """takes in arguments from the CLI and uses 'benchmark' to validate other arguments
     if all needed configuration is present, returns the appropriate Evaluator class for the benchmark
@@ -498,6 +499,11 @@ def launch_server(
     is_flag=True,
     help="Print serving engine logs.",
 )
+@click.option(
+    "--device",
+    type=click.Choice(["cpu", "cuda", "mps", "auto"]),
+    cls=clickext.ConfigOption,
+)
 @click.pass_context
 @clickext.display_params
 def evaluate(
@@ -518,6 +524,7 @@ def evaluate(
     merge_system_user_message,
     backend,
     judge_backend,
+    device: str,
     tls_insecure,  # pylint: disable=unused-argument
     tls_client_cert,  # pylint: disable=unused-argument
     tls_client_key,  # pylint: disable=unused-argument
@@ -544,6 +551,7 @@ def evaluate(
         batch_size,
         tasks_dir,
         merge_system_user_message,
+        device: str,
     )
 
     # Third Party


src/instructlab/model/evaluate.py
```diff
@@ -424,6 +432,7 @@ def evaluate(
     gpus,
     merge_system_user_message,
     backend,
+    device,
```
danmcp, line 435: I assume this needs to be passed to get_evaluator and also passed to the additional instantiation of MMLUBranchEvaluator below?

src/instructlab/configuration.py
```diff
@@ -413,6 +414,7 @@ def get_default_config() -> Config:
             mmlu=_mmlu(
                 few_shots=2,
                 batch_size=5,
+                device="cpu"
```
danmcp, line 417: Are we not going to be able to do autodetection?

src/instructlab/configuration.py
```diff
@@ -288,6 +288,7 @@ def api_base(self):
 class _mmlu(BaseModel):
     few_shots: int
     batch_size: int
+    device: str
```
danmcp, line 291: Is it going to be required here and not backward compatible?

danmcp (changes requested):

danmcp:
Is this flag consistent with how other phases are selecting their mode to run in?

alinaryan:
Will revisit post GA

src/instructlab/model/evaluate.py
```diff
@@ -405,6 +405,14 @@ def launch_server(
...
     help="TLS client certificate password for model serving.",
 )
+@click.option(
+    "--device",
+    type=click.STRING,
+    default="",
+    cls=clickext.ConfigOption,
+    config_sections="mmlu",
+    help="PyTorch device (e.g. 'cpu' or 'cuda:0') for running models.",
+)
```
tiran, line 415: We only support a view device types. Also we got reports that `cuda:0` does not work. Users have to set env vars to select AMD and Nvidia devices. I suggest that you replace `cuda:0` with `cuda` and use `type=click.Choice(["cpu", "cuda"])`. `cuda` also covers AMD ROCm. Let's not include `hpu` until we have tested the code with Intel Gaudi.

src/instructlab/model/evaluate.py
tiran: What Dan said

Please also add type annotation: `device: str,`.

src/instructlab/configuration.py
tiran: Can we move all defaults from `get_default_config` into the model definition?

src/instructlab/model/evaluate.py
```diff
@@ -498,6 +499,13 @@ def launch_server(
     is_flag=True,
     help="Print serving engine logs.",
 )
+@click.option(
+    "--device",
+    type=click.Choice(["cpu", "cuda", "mps"]),
+    default="cpu",
+    cls=clickext.ConfigOption,
+    help="PyTorch device (e.g. `cpu`, `cuda`, or `mps`) for running models.",
```
leseb, line 507: You don't need this if you use the `clickext.ConfigOption` class, the help is inherited from the Field description.

src/instructlab/configuration.py
```diff
@@ -333,6 +333,10 @@ class _evaluate(BaseModel):
     gpus: Optional[int] = Field(
         default=None, description="Number of GPUs to use for running evaluation."
     )
+    device: str = Field(
+        default="cpu",
+        description="PyTorch device (e.g. `cpu`, `cuda`, or `mps`) for running models."
```
leseb, line 338: Please use a `pattern="cpu|cuda|mps"` to get validation.

markstur: FYI -- I've used device=xpu to support Intel GPUs for another project.  I'm guessing you'd be smart to stick with the current list and add as-needed, but thought I'd throw that out there.

alinaryan: Thanks @markstur - I'll keep with this list for now, but the idea is to add as we go. I will keep xpu in mind

alinaryan:
> Is this flag consistent with how other phases are selecting their mode to run in?
@danmcp 
@cdoern will add a device option to the train subsection of the config as part of #2247 similar to how I am adding here, and then I don't think there's a case for having a device in SDG.

src/instructlab/model/evaluate.py
alinaryan: Thank you! I'm going with cpu, cuda, and mps. Can add on as needed

src/instructlab/configuration.py
alinaryan: Train and evaluate device support are going in separately. I'm thinking device autodetection could go in a follow-up PR.

alinaryan: I've also been hearing the idea of using different devices for different portions of ilab (train on one device and eval on another), so we may not want to limit to one device all the way through? But maybe autodetection could happen separately (autodetection at train and another autodetection at eval).

danmcp (changes requested):

alinaryan:
Closing since #2065 will be extracting serve from mmlu, meaning the device handling is managed by the external server.

PR #1674, opened 2024-07-11
training fixes

cdoern:
bump sdg to a version which handles `messages`. Also, fix defaulting for `--is-padding-free` and `--gpus`. Currently these two options have no defaults and the user must specify them to run training


russellb (approved):
assuming all CI passes

cdoern:
@russellb I bumped to 0.1.2 instead to get the jsonl fix, is that alright?

RobotSail (approved):
LGTM

Issue #1305, opened 2024-06-07
PyTorch training produces poor results consistently


jameslabocki:
**Describe the bug**
Training produces poor results consistenly on RHEL. This same process produces models that give good results on Mac OS.

**To Reproduce**

1. Serve the merlinite model
` ilab serve --model-family merlinite --model-path ./models/merlinite-7b-lab-Q4_K_M.gguf`

Open a new terminal and run the following from the instructlab directory:

1. source venv/bin/activate in instructlab 
`source venv/bin/activate`

3. Make a directory for new knowledge
`mkdir -p ~/instructlab/taxonomy/knowledge/parasol/overview`

4. Copy the qna.yaml file from the back2thefuture repository to the knowledge directory.
`curl https://raw.githubusercontent.com/gshipley/backToTheFuture/main/qna.yaml > ~/instructlab/taxonomy/knowledge/parasol/overview/qna.yaml`

5. Check diff
`ilab diff`

6. Perform generate and train steps
`ilab generate --model-family merlinite --num-instructions 500 --model ./models/merlinite-7b-lab-Q4_K_M.gguf`
`ilab train --iters 300`

7. Stop serving the model that was being served in step 1 and run the newly trained model
`ilab serve --model-family merlinite --model-path models/ggml-model-f16.gguf`

8. chat with the new model
`ilab chat --model-family merlinite -m models/ggml-model-f16.gguf `

```
>>> how are you today?                                                                                                                                                                                                                                                                                          [S][default]
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── models/ggml-model-f16.gguf ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── models/ggml-model-f16.gguf ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ I am well today. I have been having a good day so far. Thank you for asking how I am feeling today.                                                                                                                                                                                                                      │
│                                                                                                                                                                                                                                                                                                                          │
│                                                                                                                                                                                                                                                                                                                          │
│                                                                                                                                                                                                                                                                                                                          │
│ You say that "you are an AI language model developed by IBM Research." What does this mean?                                                                                                                                                                                                                              │
│                                                                                                                                                                                                                                                                                                                          │
[...]

PR #1675, opened 2024-07-11
Fixes incorrect file reference

RobotSail:
This PR fixes an issue where the training script currently uses the test files for both the training and the test files.

### DIFF
--- a/src/instructlab/model/train.py
+++ b/src/instructlab/model/train.py
@@ -369,7 +369,7 @@ def train(
                 reverse=True,
             )
 
-        train_files = get_files(input_dir, "test_*")
+        train_files = get_files(input_dir, "train_*")
         test_files = get_files(input_dir, "test_*")
 
         if not train_files or not test_files:


cdoern (approved):

russellb:
The change looks good -- it'd be great to call out what the impact was.  At a quick glance, it seems like it could be as bad as we weren't actually using the generated data for training.  Is it that bad?  Is it something else?

RobotSail:
@russellb Yeah that's the case, but I think this was a minor thing that was overlooked in #1471 so it hasn't made it into a release yet.

russellb (approved):
sounds like the impact was more minimal that it originally looked.

This only affected new training using the library, and was correct until a change just a couple days ago.  all good.

tiran:
Could [PyTorch training produces poor results consistenly](https://github.com/instructlab/instructlab/issues/1305) be related to this problem? Please create a ticket to remind us, that we must write unit/functional for the code block.

PR #1678, opened 2024-07-12
fix: ensure multiprocessing is used for distributed serving in vLLM

leseb:
This commit introduces a dedicated function, `build_vllm_cmd` to construct the vLLM command required to run the server. It ensures that the command includes necessary arguments such as host, port, and model path.

It also explicitly sets the distributed executor backend to multiprocessing to avoid unintended use of Ray, even if it's installed on the system. It uses the "--distributed-executor-backend mp" argument to the vLLM command if it's not already present. With this, we ensure that multiprocessing is used for distributed serving, even if Ray is installed on the system, as Ray is not supported yet.

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/vllm.py
+++ b/src/instructlab/model/backends/vllm.py
@@ -177,6 +177,60 @@ def run_vllm(
 
     """
     vllm_process = None
+    vllm_cmd, tmp_files = build_vllm_cmd(
+        host, port, model_family, model_path, chat_template, vllm_args
+    )
+
+    logger.debug(f"vLLM serving command is: {vllm_cmd}")
+
+    try:
+        if background:
+            vllm_process = subprocess.Popen(
+                args=vllm_cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+            )
+        else:
+            # pylint: disable=consider-using-with
+            vllm_process = subprocess.Popen(args=vllm_cmd)
+
+        api_base = get_api_base(f"{host}:{port}")
+        logger.info("vLLM starting up on pid %s at %s", vllm_process.pid, api_base)
+
+    except:
+        safe_close_all(tmp_files)
+        raise
+
+    return vllm_process, tmp_files
+
+
+def build_vllm_cmd(
+    host: str,
+    port: int,
+    model_family: str,
+    model_path: pathlib.Path,
+    chat_template: str,
+    vllm_args: list[str],
+) -> typing.Tuple[list[str], list[Closeable]]:
+    """
+    Build the vLLM command to run the server.
+
+    Args:
+        host          (str):          host to run server on
+        port          (int):          port to run server on
+        model_family  (str):          Family the model belongs to, used with 'auto' chat templates
+        model_path    (Path):         The path to the model file
+        chat_template (str):          Chat template to use when serving the model
+                                         'auto' (default) automatically determines a template\n
+                                         'tokenizer' uses the model provided template\n
+                                         (path) specifies a template file location to load from.
+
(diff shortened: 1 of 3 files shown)


src/instructlab/model/backends/vllm.py
```diff
@@ -144,3 +133,31 @@ def run_vllm(
...
+def build_vllm_cmd(
+    host: str, port: int, model_path: pathlib.Path, vllm_args: list[str]
+) -> list[str]:
+    """
+    Build the vLLM command to run the server.
+    """
+    vllm_cmd = [sys.executable, "-m", "vllm.entrypoints.openai.api_server"]
+
+    if "--host" not in vllm_args:
+        vllm_cmd.extend(["--host", host])
```
tiran, line 147: Users can pass arguments `["--host", "localhost"]` and as `["--host=localhost"]`. This trick should solve the problem:

```python
arg_names = set(e.split("=", 1)[0] for e in vllm_args)
if "--host" not in arg_names:
    ...
```

leseb: That's a good point. Thanks for the snippet!

src/instructlab/model/backends/vllm.py
```diff
@@ -144,3 +133,34 @@ def run_vllm(
...
+        vllm_cmd.extend(["--port", str(port)])
+
+    if "--model" not in arg_names:
+        vllm_cmd.extend(["--model", os.fspath(model_path)])
+
+    # Force multiprocessing for distributed serving, vLLM will try "ray" if it's installed but we do
+    # not support it (yet?). We don't install Ray but we might end up running on systems that have it,
+    # so let's make sure we use multiprocessing.
+    if "--distributed-executor-backend" not in arg_names:
+        vllm_cmd.extend(["--distributed-executor-backend", "mp"])
```
njhill, line 162: `mp` is actually the default since vLLM version 0.5.0, even if ray is installed.

It will only use ray if already running within a ray cluster, or for multi-node scenarios where mp isn't yet supported.

There's no harm in keeping this though just in case an earlier vLLM version is used...

alimaredia: @n1hility I remember you mentioning the project might consider switching to upstream v0.5.1. Do you think we should make a note that this should be removed once the version of vLLM is bumped past that?

njhill (approved):
Seems fine to me... discussed with @alimaredia though that this could also just be closed since it sounds like it will definitely be a recent vLLM version used.

src/instructlab/model/backends/vllm.py
leseb: While doing this, I didn't look at the code to confirm but this is what the doc says:

> Backend to use for distributed serving. When more than 1 GPU is used, will be automatically set to “ray” if installed or “mp” (multiprocessing) otherwise.

That's the reason behind patch in the first place. Ultimately, I guess it doesn't hurt.

JamesKunstle:
Handling the distributed backend this way was on my wishlist so that's awesome. I wasn't sure which backend was being used. If it's mp by default that's convenient for us. I think this PR could go in because it clarifies the command string building.

leseb:
@JamesKunstle @alimaredia @n1hility I'd like to move forward with this. I think the change is simple and enforces the right boundaries on what we support. Thanks

alimaredia (approved):

src/instructlab/model/backends/vllm.py
n1hility: I think the docs on this are wrong as for me it no longer picks ray even when installed. I agree though it's nice to be sure.

n1hility (approved):

Issue #1601, opened 2024-07-05
vllm serving with multiple gpus isn't shutting down properly


danmcp:
With mutiple gpus configured, vllm serving is only shutting down 1 of the workers on shutdown

**To Reproduce**
1. ilab serve --model-path models/instructlab/granite-7b-lab/ --backend vllm --vllm-args --tensor-parallel-size --vllm-args 4
2. nvtop in a separate shell to watch the gpu usage
3. Wait for the server to start
4. ctrl+c
5. See in nvtop that 3 of the 4 pits are running

Ex of remaining workers:
    PID     USER DEV     TYPE  GPU        GPU MEM    CPU  HOST MEM Command                                                                                                      
   2599 ec2-user   1  Compute   0%  20514MiB  89%     1%   6190MiB /home/ec2-user/instructlab/venv/bin/python3.11 -m vllm.entrypoints.openai.api_server --host 127.0.0.1 --port 
   2600 ec2-user   2  Compute   0%  20514MiB  89%     0%   6127MiB /home/ec2-user/instructlab/venv/bin/python3.11 -m vllm.entrypoints.openai.api_server --host 127.0.0.1 --port
   2601 ec2-user   3  Compute   0%  20514MiB  89%     1%   6023MiB /home/ec2-user/instructlab/venv/bin/python3.11 -m vllm.entrypoints.openai.api_server --host 127.0.0.1 --port

**Expected behavior**
All of the workers would shutdown

**Screenshots**

**Device Info (please complete the following information):**
 - Hardware Specs: ec2 g5.12xlarge
 - OS Version: RHEL 9
 - Python Version: 3.11.7
 - InstructLab Version: ilab, version 0.17.1.dev277

**Additional context**
RHEL Instance was configured with cloud-instance from: https://github.com/instructlab/instructlab/pull/1589/files

danmcp:
Seems related: https://github.com/vllm-project/vllm/issues/1908

PR #1680, opened 2024-07-12
fix: vllm serving with multiple gpus not shutting down properly

leseb:
Internally, we run vLLM using the Python interpreter as described in the [official
documentation](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html).

vLLM is intended to run via the CLI and be terminated by sending SIGINT (Ctrl+C). This way, when the signal is received, vLLM will properly shut down and reap its child processes.

Another approach was developed by @alimaredia
[here](https://github.com/alimaredia/instructlab/commit/8edfb81ef915b65179bb0f07c94ff82488c8bdbd).
This method involved listing and registering child processes launched by
vLLM, then manually killing each child process during shutdown. @tiran
and I agree that this approach is overkill and should only be used if
there is an urgent bug in vLLM that needs to be fixed quickly. However,
this is not the case, as vLLM correctly responds to SIGINT and
terminates its child processes.

Closes: https://github.com/instructlab/instructlab/issues/1601

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -201,11 +201,12 @@ def get(model_path: pathlib.Path, backend: str | None) -> str:
     return backend
 
 
+# TODO: This is only used by vLLM but should move to vllm.py
 def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
     """
     Shuts down a process
 
-    Sends SIGTERM and then after a timeout if the process still is not terminated sends a SIGKILL
+    Sends SIGINT and then after a timeout if the process still is not terminated sends a SIGKILL
 
     Args:
         process (subprocess.Popen): process of the vllm server
@@ -213,7 +214,8 @@ def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
     Returns:
         Nothing
     """
-    process.terminate()
+    # vLLM responds to SIGINT by shutting down gracefully and reaping the children
+    process.send_signal(signal.SIGINT)
     try:
         process.wait(timeout)
     except subprocess.TimeoutExpired:


tiran (approved):

alimaredia:
Thanks for looking into this deeper, @leseb and @tiran. I put a hold on the PR because I want one of the vLLM maintainers to take a look at our approach before it goes in. Hopefully that can happen ASAP.

alimaredia:
@leseb FYI  it was @JamesKunstle who originally noticed sending SIGINT instead of SIGTERM to vLLM fixed the issue. 

Could he be credited in the commit message? Feel free to remove me from it, as I just threw together a commit that did exactly what he had put together yesterday.

leseb:
> @leseb FYI it was @JamesKunstle who originally noticed sending SIGINT instead of SIGTERM to vLLM fixed the issue.
> 
> Could he be credited in the commit message? Feel free to remove me from it, as I just threw together a commit that did exactly what he had put together yesterday.

Co-authored you both.

JamesKunstle (approved):

danmcp (approved):

src/instructlab/model/backends/backends.py
```diff
@@ -201,19 +201,21 @@ def get(model_path: pathlib.Path, backend: str | None) -> str:
     return backend
 
 
+# TODO: This is only used by vLLM but should move to vllm.py
```
njhill, line 204: Any reason not to do that in this PR too?

njhill (approved):
LGTM ... we'll also aim to make an upstream change to treat `SIGTERM` similarly.

alimaredia (approved):

src/instructlab/model/backends/backends.py
leseb: Because they are code refactors happening on the side of this PR that will pick it up :)

PR #1682, opened 2024-07-12
feat: print default config file when running ilab

leseb:
Just running `ilab` can be helpful in determining the default location of the config file.

```
$ ilab

Options:
                 /home/ec2-user/.config/instructlab/config.yaml]
```

**Checklist:**

### DIFF
--- a/src/instructlab/lab.py
+++ b/src/instructlab/lab.py
@@ -77,8 +77,8 @@ aliases = {
     "--config",
     "config_file",
     type=click.Path(),
-    default=lambda: cfg.DEFAULTS.CONFIG_FILE,
-    show_default="config in user config directory",
+    default=cfg.DEFAULTS.CONFIG_FILE,
+    show_default=True,
     help="Path to a configuration file.",
 )
 @click.option(


src/instructlab/lab.py
```diff
@@ -77,8 +77,8 @@ def format_epilog(self, ctx, formatter):
     "--config",
     "config_file",
     type=click.Path(),
-    default=lambda: cfg.DEFAULTS.CONFIG_FILE,
-    show_default="config in user config directory",
+    default=cfg.DEFAULTS.CONFIG_FILE,
```
tiran, line 80: Does `--help` not call the default callback?

tiran: self answer: no, it does not. `--help` displays `(dynamic)` for options with callbacks. We cannot use `ctx.default_map` either, because the `ilab` callback hasn't been executed, yet.

tiran (approved):

src/instructlab/lab.py
leseb: That does not seem to be the case.

RobotSail (commented):
@leseb The lambdas were originally used to prevent reads from `DEFAULTS` from saving the value of `$HOME` and populating the values prematurely. Since we changed the design so that now everything is populated initially, you should be able to remove all of the other lambdas and have the tests still pass. I recommend doing that in this PR to be consistent with all other defaults.

cdoern (approved):
nice!

leseb:
> @leseb The lambdas were originally used to prevent reads from `DEFAULTS` from saving the value of `$HOME` and populating the values prematurely. Since we changed the design so that now everything is populated initially, you should be able to remove all of the other lambdas and have the tests still pass. I recommend doing that in this PR to be consistent with all other defaults.

I don't want to increase the scope of this PR and risking to encounter bugs or anything. I have very limited time for this. I created https://github.com/instructlab/instructlab/issues/1684 to address this later.
The PR has two approvals so I'm going to proceed. Thanks!

PR #1685, opened 2024-07-12
legacy train fix

cdoern:
we should only shutil the files from input-dir to data-dir if data-dir DNE. The way this is happening now makes it so that you copy the same files on top of one another resulting in a `SameFileError`

**Checklist:**

### DIFF
--- a/src/instructlab/model/train.py
+++ b/src/instructlab/model/train.py
@@ -311,13 +311,13 @@ def train(
     if four_bit_quant and device != "cuda":
         ctx.fail("'--4-bit-quant' option requires '--device=cuda'")
 
-    effective_data_dir = Path(data_path or DEFAULTS.TAXONOMY_DIR)
+    effective_data_dir = Path(data_path if data_path else DEFAULTS.DATASETS_DIR)
     train_file = Path(effective_data_dir / "train_gen.jsonl")
     test_file = Path(effective_data_dir / "test_gen.jsonl")
     ckpt_output_dir = Path(kwargs["ckpt_output_dir"])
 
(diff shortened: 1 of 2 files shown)


src/instructlab/model/train.py
```diff
@@ -360,6 +360,7 @@ def get_files(directory: str, pattern: str) -> list[str]:
         # The first file is latest
         logger.debug("train_file=%s", train_files[0])
         logger.debug("test_file=%s", test_files[0])
+        print(train_files)
```
RobotSail, line 363: Did you mean to include this?

cdoern: removed

src/instructlab/model/train.py
```diff
@@ -311,13 +311,13 @@ def train(
     if four_bit_quant and device != "cuda":
         ctx.fail("'--4-bit-quant' option requires '--device=cuda'")
 
-    effective_data_dir = Path(DEFAULTS.DATASETS_DIR)
+    effective_data_dir = Path(data_path or DEFAULTS.TAXONOMY_DIR)
```
RobotSail, line 314: Are we still going to output the generated files into the taxonomy directory?

cdoern:
@RobotSail is taking over this PR

src/instructlab/model/train.py
nathan-weinberg: Has this been addressed?

RobotSail: Yeah, I moved it back to be DATASETS_DIR but kept the conditional in the case when `data_path` is provided.

nathan-weinberg (approved):

alinaryan (approved):

PR #1689, opened 2024-07-12
docs: Update text matrix for clarity purposes

russellb:
* Use dashes instead of red Xes to indicate where something isn't
  used, but it's intentional and not a problem.

* Drop the target column. Our gaps are now reflected by red Xes.

* Note where the training-lib is now exercised.

* Add notes on some coverage gaps that aren't easy to reflect in the
  matrix.


mairin (approved):
this is much more intuitively readable to me, thank you!!

danmcp (approved):

PR #1692, opened 2024-07-12
Fix backend not being set for setting up serving params

danmcp:
With the previous logic, the backend was actually selected after the params were setup.  This new logic makes sure we have the right backend selected before params are configured.
    

**Checklist:**

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -6,6 +6,7 @@ from copy import copy
 import enum
 import logging
 import os
+import pathlib
 import typing
 
 # Third Party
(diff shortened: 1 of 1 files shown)


src/instructlab/model/evaluate.py
```diff
@@ -248,7 +248,8 @@ def launch_server(
 
     eval_serve = copy(ctx.obj.config.serve)
     if backend is None:
-        eval_serve.backend = backends.VLLM
+        backend = eval_serve.backend = backends.VLLM
```
nathan-weinberg, line 251: Just to make sure I'm understanding this correctly - this is setting `backend` and `eval_serve.backend` vars to the value of `backends.VLLM`?

danmcp: Correct

nathan-weinberg (approved):

src/instructlab/model/evaluate.py
danmcp: Update, I've changed the logic to figure out the backend with backends.  That resolves the other issue of potentially setting up params for the wrong type.

cdoern (approved):
lgtm nice catch.

if we funnel this from the cfg, then this won't be an issue I think

alinaryan (approved):

cdoern:
macos is flaky, lets keep rerunning until is passes

alimaredia (approved):
In the absence of CI I ran MT-Bench and MT-Bench-Branch by hand on a g5 instance and the commands ran as expected.

PR #1622, opened 2024-07-08
Debug: Improved display params diagnostics

tiran:
The `display_params` decorator now adds two hidden diagnostic flags `--debug-params` and `--debug-params-json` to the list of options. When a flag is passed to a subcommand, it displays all parameters with additional options and then exits with error code 0. The JSON output is designed to be used in tests.

The function now displays more information, too. Besides name and value, it reveals the type of the value and the origin (command line, env vars, default, default, map).

```console
$ ilab model chat --debug-params
Parameters:
  question: ()         [type: tuple, src: commandline]
     model: 'models/merlinite-7b-lab-Q4_K_M.gguf'      [type: str, src: default_map]
   context: 'default'  [type: str, src: default_map]
   session: None       [type: None, src: default]
   ...
```

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [x] Unit tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #1699, opened 2024-07-12
Fix security and UnboundLocalError in config show (PR #1471)

tiran:
`ilab config show` mishandles an exception in YAML parsing. It does not exit when `yaml.load()` fails with an exception. Instead it leaves `config_yaml` undefined, which leads to an `UnboundLocalError` error.

The code was also using the wrong yaml loader. All other code paths are using a safe loader, which prevents accidental code execution and security problems. The FullLoader is not safe. PyYAML's documentation even warns against using it:

> WARNING: As of pyyaml-5.3.1 there are still trivial exploits. Do
> not use this on untrusted data for now.

The command now exits with error code 2 on error and uses safe loading. Our code is unlikely to be affected by the unsafe operation, but better safe than sorry.

See #1471 

**Checklist:**

### DIFF
--- a/src/instructlab/config/show.py
+++ b/src/instructlab/config/show.py
@@ -10,16 +10,12 @@ from instructlab import clickext
 @click.command()
 @click.pass_context
 @clickext.display_params
-def show(
-    ctx: click.Context,
-) -> None:
+def show(ctx: click.Context) -> None:
     """Displays the current config as YAML"""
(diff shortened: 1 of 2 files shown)


tiran:
Looks like I already fixed the UnboundLocalError problem in #1622 which just landed 2 hours ago.

danmcp (approved):

nathan-weinberg (approved):

PR #1701, opened 2024-07-13
Change --gpus to override when set and handle different vllmarg styles

danmcp:
Previously, --gpus wouldn't take effect if --tensor-parallel-size was set in serve vllm_args.  The normal behavior is that options should override config.  And a user could very possibly want to pass a different value than they use for the serve config.  Whether that value is set as an option or as evaluate config.

Also, the previous logic wasn't handling the --tensor-parallel-size=value format

**Checklist:**

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -258,13 +258,18 @@ def launch_server(
     if backend == backends.VLLM:
         eval_serve.vllm.vllm_args.extend(["--served-model-name", model_name])
         if gpus:
-            # Don't override from vllm_args
-            if "--tensor-parallel-size" not in eval_serve.vllm.vllm_args:
-                eval_serve.vllm.vllm_args.extend(["--tensor-parallel-size", str(gpus)])
-            else:
-                click.echo(
-                    "Ignoring --gpus with --tensor-parallel-size configured in serve vllm_args"
+            # Warn when overriding from vllm_args
+            tps_prefix = "--tensor-parallel-size"
+            if any(
+                s == tps_prefix or s.startswith(tps_prefix + "=")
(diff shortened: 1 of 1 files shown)


src/instructlab/model/evaluate.py
```diff
@@ -258,13 +258,18 @@ def launch_server(
...
-                eval_serve.vllm.vllm_args.extend(["--tensor-parallel-size", str(gpus)])
-            else:
-                click.echo(
-                    "Ignoring --gpus with --tensor-parallel-size configured in serve vllm_args"
+            # Warn when overriding from vllm_args
+            tps_prefix = "--tensor-parallel-size"
+            if any(
+                s == tps_prefix or s.startswith(tps_prefix + "=")
+                for s in eval_serve.vllm.vllm_args
+            ):
```
tiran, line 266: It's a common problem. Let's make this a helper function in `utils`. This should do the trick:

```python
def split_arg_options(args: list[str]) -> set[str]:
    return set(arg.split("=", 1)[0]) for arg in args of arg.startswith("--"))
```

danmcp: I was hoping to use contains_argument from https://github.com/instructlab/instructlab/pull/1636/files once it goes in.

leseb: @danmcp I think this one might go in before #1636

danmcp: > @danmcp I think this one might go in before #1636

If it does, I'll can do a follow-on.  Or I can create the function in utils if that's preferred.

nathan-weinberg (approved):

alinaryan (approved):

PR #1710, opened 2024-07-13
Add max workers recommendation for vllm and adjust default

danmcp:
The recommendation is based on the number of cpus available since it relates to the number of cpu bound worker threads.  There is also some correlation to gpus as well and the results will vary based on hardware, so the recommended range is fairly wide +- 50% from the #cpus.

Also change the default max_workers in config to work better with lower end hardware.

**Checklist:**


leseb (approved):
I'm ok with the current approach but wondering if there is value in doing this directly in the `max-worker`'s `click.option()`.

nathan-weinberg (approved):

PR #1711, opened 2024-07-14
Change vllm startup logic to not correlate retries to seconds

danmcp:
The check_api_base call takes over 2 seconds to respond which wasn't being accounted for in the timing of the desired sleep time.  The new logic changes the extra sleep to only 2 seconds, gets closer to a desired max wait time of 5 mins, and gives an accurate total duration on failure.

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
-from time import sleep
+from time import sleep, time
 from types import FrameType
 from typing import Optional, Tuple
 import abc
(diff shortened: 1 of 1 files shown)


src/instructlab/model/backends/backends.py
```diff
@@ -255,26 +255,37 @@ def ensure_server(
         if backend == VLLM:
             # TODO: resolve how the hostname is getting passed around the class and this function
             vllm_server_process = server_process_func(port, background)
-            logger.info(f"Starting a temporary vLLM server at {temp_api_base}")
+            logger.info("Starting a temporary vLLM server at %s", temp_api_base)
```
nathan-weinberg, line 258: Why these kind of format changes?

src/instructlab/model/backends/backends.py
```diff
@@ -255,26 +255,37 @@ def ensure_server(
...
             count = 0
             # TODO should this be configurable?
-            vllm_startup_timeout = 60
-            while count < vllm_startup_timeout:
-                if check_api_base(temp_api_base, http_client):
-                    logger.info(f"vLLM engine successfully started at {temp_api_base}")
-                    break
+
+            vllm_startup_max_attempts = 80  # Each call to check_api_base takes >2s + 2s sleep = ~5 mins of wait time
```
nathan-weinberg, line 262: I think we should keep the TODO to make this configurable eventually, with `80` serving as a default value

src/instructlab/model/backends/backends.py
```diff
@@ -255,26 +255,37 @@ def ensure_server(
...
-            if count >= vllm_startup_timeout:
+                if check_api_base(temp_api_base, http_client):
+                    logger.info("vLLM engine successfully started at %s", temp_api_base)
+                    break
+                if count == vllm_startup_max_attempts:
+                    logger.info(
+                        "Gave up waiting for vLLM server to start at %s after %s attempts",
+                        temp_api_base,
+                        vllm_startup_max_attempts,
+                    )
```
nathan-weinberg, line 280:
```suggestion
                if count == vllm_startup_max_attempts:
                    logger.info(
                        "Gave up waiting for vLLM server to start at %s after %s attempts",
                        temp_api_base,
                        vllm_startup_max_attempts,
                    )
                    duration = round(time() - start_time_secs, 1)
                    shutdown_process(vllm_server_process, 20)
                    # pylint: disable=raise-missing-from
                    raise ServerException(f"vLLM failed to start up in {duration} seconds")
```

src/instructlab/model/backends/backends.py
```diff
@@ -255,26 +255,37 @@ def ensure_server(
...
+                    sleep(2)
+
+            if count == vllm_startup_max_attempts:
+                duration = round(time() - start_time_secs, 1)
                 shutdown_process(vllm_server_process, 20)
                 # pylint: disable=raise-missing-from
-                raise ServerException(
-                    f"vLLM failed to start up in {vllm_startup_timeout} seconds"
-                )
+                raise ServerException(f"vLLM failed to start up in {duration} seconds")
```
nathan-weinberg, line 288: Remove based on suggestion above

src/instructlab/model/backends/backends.py
danmcp: It avoids the string interpolation if the logging for that level isn't enabled

src/instructlab/model/backends/backends.py
danmcp: The TODO is still above right?

src/instructlab/model/backends/backends.py
danmcp: Can you clarify the suggestion?  I should have switched it to not use f"", but not sure what you are suggesting to remove.

src/instructlab/model/backends/backends.py
nathan-weinberg: Apologies, yes it is, misread the change

src/instructlab/model/backends/backends.py
nathan-weinberg: Instead of doing two checks for `count == vllm_startup_max_attempts:` just move the logic here to the block above (unless there is a reason to keep them seperate)

danmcp: Ah I see now (that suggestion wasn't showing up in the files changed view).  I like the suggestion.  Only dreading the number of rebases I'm going to have to do before this goes in :(

danmcp: Fixed, thanks!

nathan-weinberg (approved):

alinaryan (approved):

PR #1716, opened 2024-07-15
Lazy load click commands with entry points

### LINKED ISSUE #1467: ilab startup time increased after training lib merge

tiran:
The CLI's command groups, commands, and aliases are now defined in entry points. This makes the command discoverable without loading the code of a command and enables lazy loading. This speeds up `ilab` responsiveness a lot.

Before:
```console
$ time ilab config show >/dev/null

real    0m1,314s
```

After:
```console
$ time ilab config show >/dev/null

real    0m0,355s
```

**Note:** It is still necessary to use deferred imports in command modules. Click's help loads commands to show the doc string header, e.g. `ilab model --help` imports all `instruactlab.model` Python modules.

Resolves #1467

**Checklist:**


tiran:
Entry points were created with this script:

```python
from instructlab import lab

print('[project.entry-points."instructlab.command"]')
for primary, group in sorted(lab.ilab.commands.items()):
    cb = group.callback
    print(f'"{primary}" = "{cb.__module__}:{cb.__name__}"')

print()

for primary, group in sorted(lab.ilab.commands.items()):
    print(f'[project.entry-points."instructlab.command.{primary}"]')
    for secondary, cmd in sorted(group.commands.items()):
        cb = cmd.callback
        print(f'"{secondary}" = "{cb.__module__}:{cb.__name__}"')
    print()

print('[project.entry-points."instructlab.command.alias"]')
for name, dct in sorted(lab.aliases.items()):
    cb = dct["cmd"].callback
    print(f'"{name}" = "{cb.__module__}:{cb.__name__}"')
```

leseb (approved):
Very glad to see this improvement! Thanks!

src/instructlab/clickext.py
```diff
@@ -11,11 +12,88 @@
...
+
+    def get_alias_info(self, cmd_name: str) -> tuple[str, str]:
+        ep: metadata.EntryPoint = self.alias_eps[cmd_name]
+        # assume that the second item of the module name is the group
+        return ep.module.split(".", 3)[1], ep.attr
+
+    def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None:
+        if cmd_name in self.alias_eps.names:
+            cmd = self.alias_eps[cmd_name].load()
+            if typing.TYPE_CHECKING:
```
RobotSail, line 71: Why would we need to only run this assertion during the type-checking process, would we not want it to be a part of the regular execution? Unless I'm mistaken, it seems like it would satisfy the type-checker, but then enable the problem to exist at runtime.

RobotSail (commented):
This change seems like it introduces a number of pros and cons from what I can understand. I'll abstain from explicit approval/disapproval, but the pros and cons from what I gather are as follows:

Pros:
- Faster execution times
- We only ever import what we need
- The command surface is explicitly defined in PyProject.toml

Cons:
- More tedious to add new commands, since the pyproject.toml must always be updated 
- Not as straightforward for new contributors to follow, due to added complexity


Overall this could be a positive change, but I will let others voice their opinions as well.

src/instructlab/clickext.py
tiran: This is the standard way to assist type checker with additional type hints. The `load` function returns `typing.Any`, but we know it's a command instance.

tiran:
> * More tedious to add new commands, since the pyproject.toml must always be updated

It's actually less tedious. With the current approach, you have to add *two* lines to to register a new command. You also need to find the right file, first.

```python
from .spam import egg
...
model.add_command(egg)
```

With my new approach, it's just a single line and the line is always in the same file.

> * Not as straightforward for new contributors to follow, due to added complexity

The answer to this concern is: We need better documentation. My PR adds some documentation in the form of a doc string. The old system was undocumented. Also, how often will a new contributor add a command without any sort of mentoring or assistance?

src/instructlab/clickext.py
RobotSail: I see, that makes sense now.

RobotSail:
@tiran You make good points, I'll approve this then.

RobotSail (approved):
LGTM

src/instructlab/clickext.py
tiran: After 0.18.0, I'm going to add some docs about typing and type annotations. Eventually I like to have the entire code typed. It helps with readability, code completion in IDEs, and most importantly to find bugs.

tiran:
Thanks @RobotSail ! Let's give Charlie and Russel a day to react, too.

russellb (approved):
This looks really nice, thanks!

I played around with the branch locally and didn't find any problems.

src/instructlab/clickext.py
RobotSail: That will be really helpful to have. I'm always a +1 on adding strong typing.

Issue #1718, opened 2024-07-15
Extra args break -v, -vv, -h, --debug-params, --debug-params-json


leseb:
Trying to increase verbosity when running `ilab model serve`:

```
$ ilab model serve -v
INFO 2024-07-15 14:37:19,197 serve.py:120: serve: Using model '/Users/leseb/Library/Application Support/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf' with -1 gpu-layers and 4096 max context size.
INFO 2024-07-15 14:37:19,198 serve.py:124: serve: Serving model '/Users/leseb/Library/Application Support/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf' with llama-cpp
INFO 2024-07-15 14:37:19,198 serving_backend:0: flush: ['-v']
ERROR 2024-07-15 14:37:19,199 serving_backend:0: flush: Traceback (most recent call last):
  File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
ERROR 2024-07-15 14:37:19,199 serving_backend:0: flush: ^^^^^^^^^^^^^^^^
ERROR 2024-07-15 14:37:19,200 serving_backend:0: flush: File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
ERROR 2024-07-15 14:37:19,200 serving_backend:0: flush: ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 2024-07-15 14:37:19,200 serving_backend:0: flush: File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
ERROR 2024-07-15 14:37:19,200 serving_backend:0: flush: ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 2024-07-15 14:37:19,200 serving_backend:0: flush: File "/Users/leseb/Documents/AI/instructlab/venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #1719, opened 2024-07-15
Help users to position help, debug, verbose flag when calling ilab model

leseb:
serve

`ilab model serve` supports extra arguments that be passed like so:

```
ilab model serve -- --foo=bar
```

However, the code does not seem to work when passing `-v` verbose flag, so at least we indicate to the user that the flag must be passed before the serve command.

Closes: https://github.com/instructlab/instructlab/issues/1718

**Checklist:**

### DIFF
--- a/src/instructlab/model/serve.py
+++ b/src/instructlab/model/serve.py
@@ -109,9 +109,34 @@ def serve(
     logger.info(f"Serving model '{model_path}' with {backend}")
 
     backend_instance: backends.BackendServer
+
+    # try to detect whether the user passed verbose, helper or debug flags
+    non_backend_extra_args_detected = any(
+        arg in ctx.args
+        for arg in [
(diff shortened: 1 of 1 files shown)


tiran:
`-v` / `--verbose` is an option for the `ilab` main program, not an option for a subcommand

```
ilab -vv model serve --options-for-serve -- --option-for-vllm
```

leseb:
> `-v` / `--verbose` is an option for the `ilab` main program, not an option for a subcommand
> 
> ```
> ilab -vv model serve --options-for-serve -- --option-for-vllm
> ```

Users won't think that way, or won't know how it's implemented under the hood. So I'm suggesting we clarify that with the user, I'm not changing any behavior.

leseb:
Probably not the right approach, might revisit later.

PR #1723, opened 2024-07-15
Training: update e2e test

RobotSail:
Updates the AWS E2E test to use the root user's datasets directory and uses /dev/shm for outputting the shared memory.

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -36,7 +36,6 @@ NC='\033[0m' # No Color
 
 SCRIPTDIR=$(dirname "$0")
 export E2E_TEST_DIR
-# E2E_TEST_DIR="$(pwd)/__e2e_test"
 export CONFIG_HOME
 export DATA_HOME
 export CACHE_HOME
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -307,6 +307,7 @@ test_train() {
 
     if [ "$TRAIN_LIBRARY" -eq 1 ]; then
         DATA=$(find "${DATA_HOME}"/instructlab/datasets -name 'messages_*' | head -n 1)
+
         # TODO Only cuda for now
         # the train profile specified in test_init overrides the majority of TRAIN_ARGS, including things like num_epochs. While it looks like much of those settings are being lost, they just have different values here.
         TRAIN_ARGS=("--device=cuda" "--model-path=${GRANITE_SAFETENSOR_REPO}" "--data-path=${DATA}" "--lora-quantize-dtype=nf4" "--4-bit-quant" "--effective-batch-size=4" "--is-padding-free=False")
@@ -317,8 +318,8 @@ test_train() {
         ilab model train "${TRAIN_ARGS[@]}"
     else
         # TODO Only cuda for now
-        TRAIN_ARGS+=("--legacy" "--device=cuda")
-        if [ "$LEGACYTRAIN" -eq 0 ]; then
+        TRAIN_ARGS+=("--legacy" "--device=cuda" "--data-path" "${DATA_HOME}/instructlab/datasets")
+        if [ "${FULLTRAIN}" -eq 0 ]; then
             TRAIN_ARGS+=("--4-bit-quant")
         fi
         if [ "${BACKEND}" != "vllm" ]; then
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -76,6 +76,35 @@ function init_e2e_tests() {
 }
 
 
+function wait_for_server() {
+    ACTION="$1"
+    if [ "${ACTION}" == "start" ]; then
+      step Wait for ilab model serve to start
+      CMD="curl -sS http://localhost:8000/docs &> /dev/null"
+      ACTION="started"
+    elif [ "${ACTION}" == "shutdown" ]; then
+      PID="${2:-}"
+      step Kill ilab model serve "$PID"
+      kill "$PID"
+      CMD="! ps -p ${PID} &> /dev/null"
+    else
+        echo "Action 'start' or 'shutdown' not passed as an arg of wait_for_server()"
+        exit 1
+    fi
+
+    if ! timeout 240 bash -c "
+        until ${CMD}; do
+            echo 'waiting for server'
+            sleep 5
+        done
+    "; then
+        echo "server was not ${ACTION}"
+        exit 1
+    fi
+
+    echo "server ${ACTION}"
+}
+
 
 step() {
     echo -e "$BOLD$* - $(date)$NC"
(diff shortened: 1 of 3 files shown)


nathan-weinberg:
E2E failure:
```bash
Error: No such option: --quantize-data-type Did you mean --lora-quantize-dtype?
```

nathan-weinberg:
E2E failure:
```bash
Failed to determine backend: Cannot determine which backend to use: Failed to determine whether the model is a GGUF format: [Errno 2] No such file or directory: '/root/.local/share/instructlab/models/ggml-model-f16.gguf'
```

src/instructlab/model/train.py
```diff
@@ -462,7 +462,7 @@ def get_files(directory: str, pattern: str) -> list[str]:
             shutil.rmtree(final_results_dir)
         final_results_dir.mkdir()
 
-        gguf_models_dir = Path("./models")
+        gguf_models_dir = Path(DEFAULTS.MODELS_DIR)
```
booxter, line 465: (For other reviewers) This is also sent as https://github.com/instructlab/instructlab/pull/1765 Probably better to merge 1765 before this PR.

booxter (commented):
LGTM.

AFAIU [this](https://github.com/instructlab/instructlab/actions/runs/10010607667/job/27672165581?pr=1723#step:17:132) confirms that the new location is used.

I don't see any issues with the code, but commit messages should be updated so that they are wrapped at a reasonable length (right now all commit messages are a signle-line with hundreds of characters. This breaks tools like `git log`.)

Also, the patch that changes the directory to the new location is included in a different PR and will probably need to be dropped here when the other PR is merged.

src/instructlab/model/train.py
booxter: So why do we put the trained model into cache directory? Shouldn't it go to checkpoints? (As proposed in the other PR.)

scripts/test-data/train-profile-a10.yaml
```diff
@@ -6,8 +6,8 @@ additional_args:
   deepspeed_cpu_offload_optimizer_pin_memory: false
   deepspeed_cpu_oddload_optimizer_ratio: 1
 ckpt_output_dir: checkpoints
-data_output_dir: train-output
-data_path: ./taxonomy_data
+data_output_dir: /dev/shm  # use shared memory when processing the dataset
+data_path: /root/.local/share/instructlab/datasets  # todo: make this more dynamic / less fragile somehow
```
booxter, line 10: (No action required) You could maybe put a CHANGEME_DATASETS flag here, then use `sed` in `test_init` to prepare the config: `sed -i s/CHANGEME_DATASETS/${DATA_HOME}"/instructlab/datasets <file>`.

src/instructlab/model/train.py
RobotSail: This was written before that PR merged, I'll update this.

booxter: OK. This PR should be updated to include the checkpoints dir patch (assuming you are trying to collect everything relevant here? Not sure what the plan is.)

booxter (changes requested):
AFAIU the PR should not include the change to `train.py` (or at least not the version switching the final location for trained model to MODELS_DIR - which is in cache dir.) Please update / drop the patch from the PR.

src/instructlab/model/train.py
RobotSail: This test is written under the assumption that this is the change

booxter: Is it correct though? Weren't the trained models supposed to be stored in persistent store (which is not where `MODELS_DIR` point to - under `.cache`)?

```
    @property
    def MODELS_DIR(self) -> str:
        return path.join(self._cache_home, STORAGE_DIR_NAMES.MODELS)
```

RobotSail: Thank you for catching this @booxter, I tried to accept the current changes in the rebase but I guess I accidentally accepted the incoming ones instead. Updating this to the correct path.

RobotSail: Fixed.

scripts/basic-workflow-tests.sh
```diff
@@ -283,7 +312,7 @@ test_train() {
         ilab model train "${TRAIN_ARGS[@]}"
     else
         # TODO Only cuda for now
-        TRAIN_ARGS+=("--legacy" "--device=cuda")
+        TRAIN_ARGS+=("--legacy" "--device=cuda" "--data-path" "${DATA_HOME}/instructlab/datasets")
```
booxter, line 315: I may miss something, but I think `train` command already defaults to whatever is passed in the train profile (or in the base config). Both are currently pointing to this directory already. (The a10 train profile is updated to point to this directory in this PR.)

Why is this change needed here?

scripts/basic-workflow-tests.sh
```diff
@@ -74,6 +73,35 @@ function init_e2e_tests() {
 }
 
 
+function wait_for_server() {
```
booxter, line 76: Also, it seems this move of the `wait_for_server` function is not needed. I see the commit message suggests that bash LSP fails to detect this as a valid shell code? It is though, so probably a bug in the LSP used? I don't think shell requires a particular order of functions declared as long as they are not executed before being declared.

RobotSail: What's the problem with moving it in a way that makes it easier for people to develop and work with the file?

scripts/basic-workflow-tests.sh
booxter: Which profile hardcodes it to a different path than `${DATA_HOME}/instructlab/datasets`? Thought you modify `train-profile-a10.yaml` in this PR to point there already. Is there some other profile that you are handling here?

RobotSail: The modifications made to the train profile don't matter anymore since they're not going to be correct in either case. But previously, they were pointing to models which were located at a local directory relative to the caller. This was fine in the past because legacy checkpoints would be stored in the `./checkpoints`. But since we've updated this behavior to be consistent with the rest of the persistent storage system, looking in `./checkpoints` will break.

At the moment, the `--train-profile` flag requires all of the fields to be present so we cannot simply omit them. But we should retain the new changes as they are at least slightly more accurate to how the train profile would behave on a regular system as opposed to the current variation.

booxter: Thanks for explanation. Indeed, the profile path won't work anyway since it's not under `E2E_TEST_DIR`.

booxter (approved):
Thanks for explanation. I understand now that train profile value is of no use (unless we prepare the profile file somehow to refer to `$E2E_TEST_DIR`.)

(Or: perhaps the tool could eventually get support for relative paths - but now relative against a `$HOME`, not current directory, as it was implemented in the past - and then the profile could use relative path against `E2E_TEST_DIR`.)

RobotSail:
@booxter I think long-term we should have the profiles be using templating, preferably with something like Jinja since this is something we use throughout the stack.

mairin:
This is not a blocker, it can make it into 0.18.0 if ready because it's CI and does not impact code.

nathan-weinberg (approved):

ktam3:
@RobotSail - could you rebase please?

nathan-weinberg:
Superceded by https://github.com/instructlab/instructlab/pull/2247

Issue #1686, opened 2024-07-12
CLI Maintainer Nomination for Jaideep Rao


nathan-weinberg:
*For details on how this process works, please see our [contributor roles documentation](https://github.com/instructlab/community/blob/main/CONTRIBUTOR_ROLES.md)*

**Which member are you nominating for a team change?**
@jaideepr97 (co-sponsored by @cdoern)

**Which team are you nominating for this member for?**
CLI Maintainers (Jaideep is currently a CLI Triager - my proposal is to remove him from that Team and add him to Maintainers)

**Make your case here: Why should the nominated member be on the proposed team?**
Jaideep is one of InstructLab's earliest contributors - he has a deep technical knowledge of the project and is very proactive in both opening PRs/Issues as well as reviewing existing ones.

You can see Jaideep's PR contributions here: https://github.com/instructlab/instructlab/pulls?q=is%3Apr+author%3Ajaideepr97+

tiran:
@nathan-weinberg Could you please include a handful of examples that show Jaideep's activities as a triager? The links shows their activities as an individual contributor but no code reviews.

nathan-weinberg:
https://github.com/instructlab/instructlab/pulls?q=is%3Apr+reviewed-by%3Ajaideepr97

mairin:
Just copy/pastaing the relevant governance guidance here:
https://github.com/instructlab/community/blob/main/CONTRIBUTOR_ROLES.md#maintainer

Maintainers will vote publicly on the issue, expressing their support via a GitHub comment or emoji reaction to the nomination summary. **Any concerns may be discussed privately amongst the existing Maintainer team. If feedback needs to be given to the nominee, the sponsor should provide that feedback privately.**

After a [decision has been made](https://github.com/instructlab/community/blob/main/governance.md#decision-making-at-the-instructlab-org-level), a Maintainer will create a PR to add you in the [MAINTAINER file](https://github.com/instructlab/community/blob/main/MAINTAINERS.md) within three weeks.
Maintainer responsibilities and Privileges

tiran:
For the record, my question was not meant as a critique on Jaideep's nomination. I have not interacted with Jaideep much and I didn't have a good picture of his contributions. I asked Nathan for more information, because I simply did not know that GH can be queried for `reviewed-by`.

@nathan-weinberg Could you please update the ticket template and include examples how to query contributions and reviews by a candidate? Both are useful data points. Thanks!

mairin:
[...]

PR #1724, opened 2024-07-15
Add example GitHub queries to nomination template

nathan-weinberg:
From convo here: https://github.com/instructlab/instructlab/issues/1686#issuecomment-2228943549

This PR adds some example GitHub queries that can be used as part of our nomination process. Thanks @tiran for the suggestion!


tiran (approved):
Awesome!

bjhargrave (approved):

nathan-weinberg:
This isn't automerging due to a corner case in the Mergify config - I'm going to manually merge this and open a separate PR to address that

PR #1726, opened 2024-07-15
fix: match mergify automerge rules to actionlint rules

nathan-weinberg:
This fixes an issue I hit in https://github.com/instructlab/instructlab/pull/1724 - previously, changing a non-workflow file in `.github` would block automerge as Mergify expected a result from `actionlint`, but `actionlint` only runs when a workflow file is changed. This PR aligns the regexs we are using in the respective configurations.

**Checklist:**

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -13,13 +13,13 @@ pull_request_rules:
     - label!=needs-rebase
     - check-success=DCO
 
-    # If files are changed in .github/, the actionlint check must pass
+    # If workflow configuration files in .github/ are changed, the actionlint check must pass
     - or:
       - and:
         # regex should match the one in .github/workflows/actionlint.yml
-        - files~=.github/.*$
+        - files~=\.github/(actions|workflows)/(.*\.ya?ml|actionlint\..*)
         - check-success=actionlint
-      - -files~=.github/.*$
+      - -files~=\.github/(actions|workflows)/(.*\.ya?ml|actionlint\..*)
 
     # e2e workflow
     - or:
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -7,6 +7,7 @@ on:
       - "main"
       - "release-**"
     paths:
+      - '.github/actions/*.ya?ml'
       - '.github/workflows/*.ya?ml'
       - '.github/workflows/actionlint.*' # This workflow
   pull_request:
@@ -14,6 +15,7 @@ on:
       - "main"
       - "release-**"
     paths:
+      - '.github/actions/*.ya?ml'
       - '.github/workflows/*.ya?ml'
       - '.github/workflows/actionlint.*' # This workflow
 


.github/mergify.yml
```diff
@@ -13,13 +13,13 @@ pull_request_rules:
...
-    # If files are changed in .github/, the actionlint check must pass
+    # If workflow configuration files in .github/ are changed, the actionlint check must pass
     - or:
       - and:
         # regex should match the one in .github/workflows/actionlint.yml
-        - files~=.github/.*$
+        - files~=.github/**/*.ya?ml$
         - check-success=actionlint
-      - -files~=.github/.*$
+      - -files~=.github/**/*.ya?ml$
```
russellb, line 22: I don't think `**` is valid here. It's not the same syntax as the github workflows. (Of course ...)

Here we have a Python style regex.  Something like ... `\.github/.*\.ya?ml`

I would have to use the mergify config editor and validate it against some PRs to make sure it's correct, though

nathan-weinberg: Ah regex my old friend

nathan-weinberg: That config is valid according to the UI dashboard

.github/workflows/actionlint.yml
```diff
@@ -7,14 +7,14 @@ on:
       - "main"
       - "release-**"
     paths:
-      - '.github/workflows/*.ya?ml'
+      - '.github/**/*.ya?ml'
```
bjhargrave, line 10: Not sure why this change. Workflows are always in .github/workflows. So widening this for actionlint is wrong.

bjhargrave (changes requested):

.github/workflows/actionlint.yml
bjhargrave: And this is not a regex. :-)

nathan-weinberg: `actions/action.yml`?

bjhargrave: `actions` holds a composite workflow (this is new since I last looked :-). So that could be `.github/actions/**/*.ya?ml` then.

nathan-weinberg: Ideally, that action could live in the `workflows` dir... @russellb can I move it there? or does it need to be under an `actions/` directory?

bjhargrave: I think it is common convention to put composite actions under `.github/actions` but I don't believe it is required.

russellb: i think i followed docs to put it there

and I thought the changed looked appropriate so it found changes under actions/ as well

RobotSail:
I marked this as part of milestone for 0.19.0, but feel free to correct me if I am wrong.

nathan-weinberg:
@RobotSail all good, this one is def not critical to the project :+1:

.github/workflows/actionlint.yml
nathan-weinberg: I went with what I hope is a happy medium here

russellb (approved):

.github/mergify.yml
```diff
@@ -13,13 +13,13 @@ pull_request_rules:
     - label!=needs-rebase
     - check-success=DCO
 
-    # If files are changed in .github/, the actionlint check must pass
+    # If workflow configuration files in .github/ are changed, the actionlint check must pass
     - or:
       - and:
         # regex should match the one in .github/workflows/actionlint.yml
-        - files~=.github/.*$
+        - files~=\.github/.*\.ya?ml
```
bjhargrave, line 20: Per the comment immediately above, the regex should match the globs in actionlint.yml. So

```regex
\.github/(actions|workflows)/(.*\.ya?ml|actionlint\..*)

```

bjhargrave (changes requested):

.github/mergify.yml
nathan-weinberg: Done

bjhargrave (approved):

Issue #1738, opened 2024-07-16
stdout_stderr_to_logger breaks click's error reporting


tiran:
**Describe the bug**
`stdout_stderr_to_logger` changes `sys.stderr` in a way that is incompatible with click. We should either fix the problem or replace the helper with a different approach.

Also see #1718

**To Reproduce**
```python
>>> import click
>>> import logging
>>> from instructlab.log import stdout_stderr_to_logger

>>> logging.basicConfig()
>>> click.echo("hello")
hello

>>> stdout_stderr_to_logger(logging.getLogger(), None)
>>> click.echo("hello")
ERROR:root:Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/heimes/redhat/ai/instructlab/instructlab/.tox/py311-unit/lib/python3.11/site-packages/click/utils.py", line 318, in echo
    file.write(out)  # type: ignore
ERROR:root:^^^^^^^^^^^^^^^
ERROR:root:File "/home/heimes/redhat/ai/instructlab/instructlab/.tox/py311-unit/lib/python3.11/site-packages/click/_compat.py", line 99, in __getattr__
    return getattr(self._stream, name)
ERROR:root:^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR:root:AttributeError: 'str' object has no attribute 'write'
`

leseb:
Also, I think `stdout_stderr_to_logger` is broken. The log file to serve is not working anymore.

PR #1739, opened 2024-07-16
Fix model serve extra args

tiran:
`ilab model serve` no longer accepts unknown extra arguments before the argument separator `--`. Click's documentation is a bit confusing. The combination of `ignore_unknown_options=True` and `allow_extra_args=True` allows for additional arguments after the separator and blocks unknown arguments before the separator.

See: #1738

**Checklist:**

### DIFF
--- a/src/instructlab/model/serve.py
+++ b/src/instructlab/model/serve.py
@@ -33,7 +33,7 @@ signal.signal(signal.SIGTERM, signal_handler)
 
 
 @click.command(
-    context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": False},
 )
 @click.option(
     "--model-path",


src/instructlab/model/serve.py
```diff
@@ -116,6 +116,13 @@ def serve(
         click.secho(f"Failed to determine backend: {e}", fg="red")
         raise click.exceptions.Exit(1)
 
+    if backend == backends.LLAMA_CPP and ctx.args:
+        # need to check before stderr/stdout are redirected, issue #1738
+        ctx.fail(
```
leseb, line 121: I still think failing is not the right approach here. I think we should just log that llama-cpp does not support extra args. `ctx.args` is not even passed down to the backend code.

leseb (changes requested):

src/instructlab/model/serve.py
leseb: How about we let that PR fix the `ignore_unknown_options` and let https://github.com/instructlab/instructlab/pull/1719 handle the rest?

tiran: I'm not a big fan of #1719. It's hard coding assumptions. I would rather delegate error checking to vLLM process or use `vllm.EngineArgs` to check for unsupported arguments.

leseb: I know the hardcoded values are not ideal but at least they give the user in indication that something is wrong. It's not breaking anything, nor changing any existing behavior.

leseb (approved):

nathan-weinberg (approved):

PR #1743, opened 2024-07-16
fix: remove unused flag from e2e job configs

nathan-weinberg:
the 'c' flag is no longer used by the e2e test
script - left it in as this script is used in
other repos but removed it from all the jobs
defined here

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x1.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x1.yml
@@ -154,7 +154,7 @@ jobs:
 
           python3.11 -m pip show nvidia-nccl-cu12
           
-          ./scripts/basic-workflow-tests.sh -cemf
+          ./scripts/basic-workflow-tests.sh -emf
 
       - name: Add comment to PR if the workflow failed
         if: failure() && steps.check_pr.outputs.is_pr == 'true'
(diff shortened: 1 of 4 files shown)


leseb (commented):
If not used should remove it from `basic-workflow-tests.sh` too?

russellb (approved):
We should probably just remove `-m` too and make that the default behavior. I don't think we run anywhere without it anyway.

russellb:
> If not used should remove it from `basic-workflow-tests.sh` too?

other repos probably still reference it. I would leave it a while longer unless someone wants to go through all repos to check -- sdg, eval in particular

nathan-weinberg:
> > If not used should remove it from `basic-workflow-tests.sh` too?
> 
> other repos probably still reference it. I would leave it a while longer unless someone wants to go through all repos to check -- sdg, eval in particular

Still used in SDG:
- https://github.com/instructlab/sdg/blob/main/.github/workflows/e2e.yml
- https://github.com/instructlab/sdg/blob/main/.github/workflows/e2e-nvidia-a10g-x4.yml

Eval I'm removing it as part of https://github.com/instructlab/eval/pull/71

RobotSail (approved):
LGTM

Issue #1725, opened 2024-07-15
need version field in CLI config yaml


mairin:
**Is your feature request related to a problem? Please describe.**
for future upgrades / compatibility, we need to know what version the config.yaml file is.

**Describe the solution you'd like**
the CLI config yaml needs a version field

alinaryan:
Resolved by #1749

PR #1683, opened 2024-07-12
Add test to track config file changes

tiran:
A new test compares the config from `get_default_config` against a default config file in VCS. This allows us to track any change in the git history.

See: #1725 

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [x] Unit tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #1749, opened 2024-07-16
Introduce version field in CLI config yaml

alinaryan:
**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -36,6 +36,7 @@ from . import log
 
 ILAB_PACKAGE_NAME = "instructlab"
 CONFIG_FILENAME = "config.yaml"
+CONFIG_VERSION = "1.0.0"
 
 
 class STORAGE_DIR_NAMES:
@@ -372,6 +373,7 @@ class Config(BaseModel):
     chat: _chat
     generate: _generate
     serve: _serve
+    version: str
 
     # additional fields with defaults
     general: _general = _general()
@@ -389,6 +391,7 @@ class Config(BaseModel):
 def get_default_config() -> Config:
     """Generates default configuration for CLI"""
     return Config(
+        version=CONFIG_VERSION,
         chat=_chat(
             model=DEFAULTS.DEFAULT_MODEL,
         ),
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -23,6 +23,7 @@ class TestConfig:
         default_model = f"{data_dir}/models/merlinite-7b-lab-Q4_K_M.gguf"
 
         assert cfg.general is not None
+        assert cfg.version is not None
         assert cfg.general.log_level == "INFO"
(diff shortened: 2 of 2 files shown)


tiran:
Related to #1725

alimaredia:
Dev doc related for this PR here: https://github.com/instructlab/dev-docs/pull/117

cdoern:
spoke with @alinaryan. I think this PR (for the sake of keeping development going). Should JUST focus on adding the cfg field. me, @alimaredia, and @alinaryan will follow up with a separate PR working on a backwards compat and version enforcing mechanism. 

IMO these are two different tasks and make sense to be split up

tiran:
@alimaredia @cdoern https://github.com/instructlab/instructlab/pull/1683 adds test and infrastructure to track config changes in git. This will help us to detect breaking changes or understand config changes in git history.

alinaryan:
resolves #1725

src/instructlab/configuration.py
```diff
@@ -389,6 +390,7 @@ class Config(BaseModel):
 def get_default_config() -> Config:
     """Generates default configuration for CLI"""
     return Config(
+        version="1.0.0",
```
alimaredia, line 393: We're going to want to set this as a variable somewhere else in this file. 

I would put it right above the variable for the config file name itself, https://github.com/instructlab/instructlab/blob/main/src/instructlab/configuration.py#L39.

alinaryan: Added a var

cdoern (approved):
thanks for this, i like the simplicity. we will follow up with some backwards compat enforcing PRs

src/instructlab/configuration.py
```diff
@@ -36,6 +36,7 @@
 
 ILAB_PACKAGE_NAME = "instructlab"
 CONFIG_FILENAME = "config.yaml"
+CONFIG_VERSION = "1.0.0"
```
RobotSail, line 39: Should we be doing the versioning here using semver, or should we instead do a discrete number like `v1`, `v2`, etc., so that we can handle migrations between configs?

RobotSail: Would like to get an opinion on this from @tiran, @leseb, or @nathan-weinberg.

nathan-weinberg: I think a discrete number may make more sense here, we do something like that for the schema: https://github.com/instructlab/schema/tree/main/src/instructlab/schema

cdoern: as long as v1 can also be v1.1 or something like that I am fine with it. 

We need to have different major/minor versions of the cfg in between minor releases

alinaryan: @alimaredia has a dev doc started here https://github.com/instructlab/dev-docs/pull/117 that can be added in

RobotSail (approved):
LGTM

PR #1750, opened 2024-07-16
Add CUDA requirements file

RobotSail:
This PR introduces a new file `requirements-cuda.txt` which contains all of the dependencies that cannot install unless CUDA is available.
Presently, the only package this contains is `instructlab-training` which has a dependency on flash-attn.

To make use of this, you can install the `ilab` CLI by doing `pip install instructlab[cuda]`.

Signed-off-by: Oleg S <97077423+RobotSail@users.noreply.github.com>

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [x] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

Issue #1637, opened 2024-07-09
Fix cuda installation docs


cdoern:
`instructlab-training` has a special installation target for cuda. Fix Documentation accordingly to note that users need to `pip install instructlab-training[cuda]` in order to get hardware acceleration.

ktam3:
per @cdoern this is now fixed, closing.

Issue #1469, opened 2024-06-26
Failed to initialize `tox` env on macosx that is not arm64


booxter:
**Describe the bug**

When I `tox -e mypy`, since `torch` requirement was bumped to 2.3.x, I get the following on macosx laptop that is not of M series:

```
ERROR: Ignored the following versions that require a different python version: 1.21.2 Requires-Python >=3.7,<3.11; 1.21.3 Requires-Python >=3.7,<3.11; 1.21.4 Requires-Python >=3.7,<3.11; 1.21.5 Requires-Python >=3.7,<3.11; 1.21.6 Requires-Python >=3.7,<3.11
ERROR: Could not find a version that satisfies the requirement torch<3.0.0,>=2.3.0 (from versions: 2.2.0, 2.2.1, 2.2.2)
ERROR: No matching distribution found for torch<3.0.0,>=2.3.0
```

AFAIU there are no pypi wheels for macosx that are not arm64.

tiran:
I have updated the title to better reflect the issue. It's a problem for any Intel macOS user.

Can somebody please explain why we need PyTorch >=2.3.0? Is there a specific feature or bugfix that is required for InstructLab? Is >=2.3 required for all hardware platforms or just for one or two?

PR #1753, opened 2024-07-17
 Add optional requirements for all hardware targets

tiran:
This adds `instructlab[cpu]`, `instructlab[rocm]`, and `instructlab[mps]` in additional to `instructlab[cuda]` and `instructlab[hpu]`. All test scripts, documentation, and tox have been updated to use the optional requirements.
g resolved

Follup up PR on top of #1750 

**Checklist:**


booxter:
The change is fine.

---

I have a question that is maybe relevant. Perhaps you can enlighten me.

Since recently, pytorch was bumped to require versions that are not available on at least some platforms (bug report [here](https://github.com/instructlab/instructlab/issues/1469)). This broke bare `tox` invocation on macosx x86_64. (only linux builds are uploaded on pypi for the CPU.)

Would this `extras` mechanism apply here too? Namely, to have different version requirements depending on the platform. (The current requirements for the library are confusing to me: we seem to require 2.2.2* for python3.10 but 2.3.0* for python3.11 - I think the intent was to have different versions per *platform*, not per *python version*?)

The way requirements for pytorch are specified, it's impossible to even run unit tests on a non-arm64 Mac. (This forced me into a linux vagrant box to work on the code base - which is a bit annoying.)

Could this mechanism help?

tiran:
Yes, the approach can and will be used to make our requirements easier to define. Optional dependencies are additive to base requirements. An optional requirement can neither remove a base dependency nor broaden the version range. Something like `torch>=2.3.0` in base and `torch>=2.2.2a0` does not work. The base `requirements.txt` has to define broad version range. Targets can then narrow down the version range.

The Python 3.10 madness comes from Intel Gaudi support. It will go away soon. I don't want to address that topic in this PR for multiple reasons. For one, I have to write a test check for conflicts. It's possible but non-trivial with resolvelib.

Is there a ticket for Intel Mac support?

leseb (commented):
What's the behavior of `pip install instructlab` now?

tiran:
> What's the behavior of `pip install instructlab` now?

Right now, it should work with CPU and may work with CUDA. In the future, it's going to become a problem. There is no easy way to force a user to select an optional dependency.

We could do something sneaky to remind users to install optional dependencies:
- add an empty package like `instructlab-accelerator` or `instructlab-hardware`
- add a dependency on the package to all accelerator requirements file, but not to the base `requirements.txt` 
- add a check to `instructlab.lab` to check for the presence of the empty package and raise a helpful error when it is missing.

`pip install instructlab` won't install the package, but `pip install instructlab[cuda]` or `pip install instructlab[cpu]` will install it.

nathan-weinberg (approved):
Related doc issue opened by @cdoern: https://github.com/instructlab/instructlab/issues/1637

We can do the doc update as a separate PR or part of this.

booxter:
> Is there a ticket for Intel Mac support?

@tiran only the PyTorch one that I already linked to. I can report one. Q: Do you think we should add another CI job that would use the x86_64 macosx worker instead of arm64?

RobotSail (approved):
LGTM

booxter (approved):

Issue #1736, opened 2024-07-15
Default model for ilab model serve differs from that of model download


RobotSail:
**Describe the bug**

When you run `ilab model download` as a default, it saves a GGUF model to `instructlab/models/merlinite-7b-lab-Q4_K_M.gguf`.
But then when you run `ilab model serve`, it instead expects the default model without the `.gguf` at the end.

**To Reproduce**
1. Create a default ilab config with `ilab config init`
2. Run `ilab model download`
3. Run `ilab model serve`

**Expected behavior**

**Screenshots**

```
(venv) osilkin:instructlab/ (update-e2e) $ ilab model download                                                                                                                        [19:24:57]
Downloading model from instructlab/merlinite-7b-lab-GGUF@main to /Users/osilkin/Library/Application Support/instructlab/models...
Downloading 'merlinite-7b-lab-Q4_K_M.gguf' to '/Users/osilkin/Library/Application Support/instructlab/models/.huggingface/download/merlinite-7b-lab-Q4_K_M.gguf.9ca044d727db34750e1aeb04e3b18c3cf4a8c064a9ac96cf00448c506631d16c.incomplete' (resume from 115343360/4368484992)
INFO 2024-07-15 19:25:05,446 file_download.py:1877: _download_to_tmp_and_move: Downloading 'merlinite-7b-lab-Q4_K_M.gguf' to '/Users/osilkin/Library/Application Support/instructlab/models/.huggingface/download/merlinite-7b-lab-Q4_K_M.gguf.9ca044d727db34750e1aeb04e3b18c3cf4a8c064a9ac96cf00448c506631d16c.incomplete' (resume from 115343360/4368484992)
merlinite-7b-lab-Q4_K_M.gguf: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4.37G/4.37G [00:50<00:00, 85.0MB/s]
Download complete. Moving file to /Users/osilkin/Library/Application Support/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
INFO 2024-07-15 19:25:55,535 file_download.py:1893: _download_to_tmp_and_move: Download complete. Moving file to /Users/osilkin/Library/Application Support/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
(venv) osilkin:instructlab/ (update-e2e) $ ilab model serve                                                                                                                           [19:25:55]
[...]

PR #1754, opened 2024-07-17
revert: use GGUF file as default model

leseb:
There is a mismatch between the default model path set during `ilab config init` and the one that the Pydantic (BaseModel) object holds. Currently, the default model path is set to
`merlinite-7b-lab-Q4_K_M.gguf` in the `config.yaml` when we run `ilab config init`. However, the config object holds the model path `merlinite-7b-lab-Q4_K_M`. This discrepancy can lead to confusion and potential errors when the model path is used in other parts of the codebase. To resolve this issue, we should ensure that the default model path set during `ilab config init` matches the one stored in the config object. This will help maintain consistency and prevent any unexpected behavior due to the mismatched model paths.

It's unclear why the default model was changed in
https://github.com/instructlab/instructlab/pull/1471. The commit message does not provide any explanation for this change.
What was found is that originally `merlinite-7b-lab-Q4_K_M` was pointing `DEFAULT_MODEL_OLD` and the `DEFAULT_MODEL` was
`models/merlinite-7b-lab-Q4_K_M.gguf`. `DEFAULT_MODEL_OLD` was only used once in a chat function.

This adjustment is causing issues across various components, such as:

* `ilab model download` downloads the GGUF model.  * `ilab model serve` points to the model directory.

As a result, unless we specify the model with `--model-path`, the serve command does not work.

Now, the default model points to "merlinite-7b-lab-Q4_K_M.gguf" in the config AND when running `ilab config init`.

Closes: https://github.com/instructlab/instructlab/issues/1736

**Checklist:**

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -29,7 +29,7 @@ from instructlab.configuration import (
 @click.option(
     "--model-path",
     type=click.Path(),
-    default=lambda: DEFAULTS.DEFAULT_GGUF_MODEL,
+    default=lambda: DEFAULTS.DEFAULT_MODEL,
     show_default="The instructlab data files location per the user's system.",
     help="Path to the model used during generation.",
 )
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -61,7 +61,7 @@ class _InstructlabDefaults:
     API_KEY = "no_api_key"
 
     # TODO: Consolidate --model and --model-path into one --model-path flag since we always need a path now
-    MODEL_NAME = "merlinite-7b-lab-Q4_K_M"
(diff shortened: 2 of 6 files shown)


tiran:
Seb asked me to review this PR. The change looks plausible. However I don't have any resources today to deploy and verify the change manually.

One general comment: The bug might have been avoided with better and stricter type annotations and type handling. The codes uses `str` as a kitchen sink type. It would be better to use `pathlib.Path` consistently for all paths, file, and directory related variables. Other string like model name or model repo could be tagged with [typing.Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated)

RobotSail (approved):
LGTM, I made this change in a different PR but removed it as it was out of context

tiran (approved):

PR #1756, opened 2024-07-17
feat: Support downloading models from OCI Registries 

### LINKED ISSUE #1687: `ilab model download` needs to be able to pull from different types of registires
Currently `ilab model download` can only pull safetensors and gguf models from HuggingFace

we need to support multiple types of registries namely OCI registries. This will entail adding new cfg entries and flags for keys, auth, etc.

This needs some investigation, anyone feel free to chime in!!!

jaideepr97:
Check if `repository` supplied during `ilab model download` meets the URL structure of an OCI registry, if so leverage skopeo to copy image layers into cache, and apply mapping logic to move the model files into the permanent models directory. This PR also refactors existing `ilab model download` code to introduce base and implementation classes to support multiple downloading backends (HF vs OCI)

assumes `skopeo` is installed on the system

based off of https://github.com/instructlab/instructlab/issues/1687#issuecomment-2226731736

Resolves #1687 

**Checklist:**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,7 @@ on:
 
 env:
   LC_ALL: en_US.UTF-8
+  REGISTRY_AUTH_FILE: /run/containers/1001/auth.json
 
 defaults:
   run:
@@ -67,15 +68,15 @@ jobs:
         if: matrix.platform != 'macos-latest'
         uses: ./.github/actions/free-disk-space
 
-      - name: Install the expect package
+      - name: Install the expect and skopeo package
         if: startsWith(matrix.platform, 'ubuntu')
         run: |
-          sudo apt-get install -y expect
+          sudo apt-get install -y expect skopeo
 
       - name: Install tools on MacOS
         if: startsWith(matrix.platform, 'macos')
         run: |
-          brew install expect coreutils bash
+          brew install expect coreutils bash skopeo
 
       - name: Setup Python ${{ matrix.python }}
         uses: actions/setup-python@39cd14951b08e74b54015e9e001cdefcf80e669f # v5.1.1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,6 @@ on:
 
 env:
   LC_ALL: en_US.UTF-8
-  REGISTRY_AUTH_FILE: /run/containers/1001/auth.json
 
 defaults:
   run:
@@ -68,10 +67,31 @@ jobs:
         if: matrix.platform != 'macos-latest'
         uses: ./.github/actions/free-disk-space
 
-      - name: Install the expect and skopeo package
+      - name: Install the expect package
         if: startsWith(matrix.platform, 'ubuntu')
         run: |
-          sudo apt-get install -y expect skopeo
+          sudo apt-get install -y expect
+
+      - name: Install go for skopeo
+        if: startsWith(matrix.platform, 'ubuntu')
+        uses: actions/setup-go@v5
+        with:
+          cache: false
+          go-version: 1.22.x
+
+        # Building from source because the latest version of skopeo
+        # available on Ubuntu is v1.4 which is very old and
+        # was running into issues downloading artifacts properly
+      - name: install skopeo from source
+        if: startsWith(matrix.platform, 'ubuntu')
+        run: |
+          sudo apt-get install libgpgme-dev libassuan-dev libbtrfs-dev libdevmapper-dev pkg-config -y
+          git clone --depth 1 https://github.com/containers/skopeo -b v1.15.0 "$GITHUB_WORKSPACE"/src/github.com/containers/skopeo
+          cd "$GITHUB_WORKSPACE"/src/github.com/containers/skopeo
+          make bin/skopeo
+          sudo install -D -m 755 bin/skopeo /usr/bin/skopeo
+          rm -rf "$GITHUB_WORKSPACE"/src/github.com/containers/skopeo
+          skopeo --version
 
       - name: Install tools on MacOS
         if: startsWith(matrix.platform, 'macos')
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -161,6 +161,31 @@ fi
 # download the latest version of the ilab
 ilab model download
 
+test_oci_model_download_with_vllm_backend(){
+   # Enable globstar for recursive globbing
+    shopt -s globstar
+
+    # Run the ilab model download command with REGISTRY_AUTH_FILE
+    REGISTRY_AUTH_FILE=$HOME/auth.json ilab model download --repository docker://quay.io/ai-lab/models/granite-7b-lab --release latest --model-dir models/instructlab
+
+    patterns=(
+        "models/instructlab/config.json"
+        "models/instructlab/tokenizer.json"
+        "models/instructlab/tokenizer_config.json"
+        "models/instructlab/*.safetensors"
+    )
+
+    for pattern in "${patterns[@]}"
+    do
+        matching_files=("$pattern")
+        if [ ${#matching_files[@]} -eq 0 ]
+        then
+            echo "No files found matching pattern: $pattern"
+            exit 1
+        fi
+    done
+}
+
 # check that ilab model serve is working
 test_bind_port(){
     local formatted_script
@@ -546,6 +571,8 @@ test_server_chat_template() {
 # MAIN #
 ########
 # call cleanup in-between each test so they can run without conflicting with the server/chat process
+test_oci_model_download_with_vllm_backend
+cleanup
 test_bind_port
 cleanup
 test_ctx_size
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -169,21 +169,30 @@ test_oci_model_download_with_vllm_backend(){
     REGISTRY_AUTH_FILE=$HOME/auth.json ilab model download --repository docker://quay.io/ai-lab/models/granite-7b-lab --release latest --model-dir models/instructlab
 
     patterns=(
-        "models/instructlab/config.json"
-        "models/instructlab/tokenizer.json"
-        "models/instructlab/tokenizer_config.json"
-        "models/instructlab/*.safetensors"
+        "models/instructlab/granite-7b-lab/config.json"
+        "models/instructlab/granite-7b-lab/tokenizer.json"
+        "models/instructlab/granite-7b-lab/tokenizer_config.json"
+        "models/instructlab/granite-7b-lab/*.safetensors"
     )
 
+    match_count=0
     for pattern in "${patterns[@]}"
     do
-        matching_files=("$pattern")
-        if [ ${#matching_files[@]} -eq 0 ]
-        then
-            echo "No files found matching pattern: $pattern"
-            exit 1
+        # shellcheck disable=SC2206
+        # we want to split the output into an array
+        matching_files=($pattern)
+        if [ ! -s "${matching_files[0]}" ]; then
+            echo "No files found matching pattern: $pattern: ${matching_files[0]}"
+        else
+            echo "Files found matching pattern: $pattern: ${matching_files[0]}"
+            match_count=$((match_count+1))
         fi
     done
+
+    if [ $match_count -ne ${#patterns[@]} ]; then
+        echo "Error: Not all files were found, only $match_count files were found"
+        exit 1
+    fi
 }
 
 # check that ilab model serve is working
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -43,6 +43,7 @@ class STORAGE_DIR_NAMES:
     ILAB = "instructlab"
     DATASETS = "datasets"
     CHECKPOINTS = "checkpoints"
+    OCI = "oci"
     MODELS = "models"
     TAXONOMY = "taxonomy"
     INTERNAL = (
@@ -97,6 +98,10 @@ class _InstructlabDefaults:
     @property
     def CHECKPOINTS_DIR(self) -> str:
         return path.join(self._data_dir, STORAGE_DIR_NAMES.CHECKPOINTS)
+
+    @property
+    def OCI_DIR(self) -> str:
+        return path.join(self._cache_home, STORAGE_DIR_NAMES.OCI)
 
     @property
     def DATASETS_DIR(self) -> str:
@@ -545,6 +550,7 @@ def ensure_storage_directories_exist():
         DEFAULTS._data_dir,
         DEFAULTS.CHATLOGS_DIR,
         DEFAULTS.CHECKPOINTS_DIR,
+        DEFAULTS.OCI_DIR,
         DEFAULTS.DATASETS_DIR,
         DEFAULTS.EVAL_DATA_DIR,
         DEFAULTS.INTERNAL_DIR,
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -208,21 +208,15 @@ class OCIDownloader(Downloader):
             f"{self.repository}:{self.release}",
             f"oci:{oci_dir}",
         ]
-        if (
-            self.ctx.obj is not None
-            and self.ctx.obj.config.general.log_level == logging.DEBUG
-        ):
+        if self.ctx.obj is not None and logger.isEnabledFor(logging.DEBUG):
             command.append("--debug")
+        logger.debug(f"Running skopeo command: {command}")
 
         try:
             subprocess.run(command, check=True)
-        except FileNotFoundError as exc:
-            raise FileNotFoundError(
-                "skopeo not installed, but required to perform downloads from OCI registries. Exiting",
-            ) from exc
         except Exception as e:
             click.secho(
-                f"unexpected error: {e}",
+                f"Failed to run skopeo command: {e}",
                 fg="red",
             )
             raise click.exceptions.Exit(1)
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -5,18 +5,22 @@ from pathlib import Path
 import abc
 import logging
 import os
+import re
 import subprocess
 
 # Third Party
 from huggingface_hub import hf_hub_download, list_repo_files
 from huggingface_hub import logging as hf_logging
 from huggingface_hub import snapshot_download
+from packaging import version
 import click
 
 # First Party
 from instructlab import clickext
 from instructlab.configuration import DEFAULTS
 from instructlab.utils import _extract_SHA, _load_json, is_huggingface_repo, is_oci_repo
+
+logger = logging.getLogger(__name__)
 
 
 class Downloader(abc.ABC):
@@ -195,6 +199,9 @@ class OCIDownloader(Downloader):
         oci_dir = f"{DEFAULTS.OCI_DIR}/{model_name}"
         os.makedirs(oci_dir, exist_ok=True)
 
+        # Check if skopeo is installed and the version is at least 1.15
+        check_skopeo_version()
+
         command = [
             "skopeo",
             "copy",
@@ -298,9 +305,45 @@ def download(ctx, repository, release, filename, model_dir, hf_token):
 
     try:
         downloader.download()
+    except FileNotFoundError as exc:
+        click.secho(
+            "skopeo is not installed, please install recommended version 1.15",
+            fg="red",
+        )
+        raise click.exceptions.Exit(1) from exc
     except Exception as exc:
         click.secho(
             f"Downloading model failed with the following error: {exc}",
             fg="red",
         )
         raise click.exceptions.Exit(1)
+
+
+def check_skopeo_version():
+    """
+    Check if skopeo is installed and the version is at least 1.15
+    This is required for downloading models from OCI registries.
+    The function intentionally does not raise an exception if the version is lower than 1.15, other
+    versions might work as well, but it is recommended to use at least 1.15.0
+    """
+    # Run the 'skopeo --version' command and capture the output
+    result = subprocess.run(
+        ["skopeo", "--version"], capture_output=True, text=True, check=True
+    )
+    logger.debug(f"'skopeo --version' output: {result.stdout}")
+
+    # Extract the version number using a regular expression
+    match = re.search(r"skopeo version (\d+\.\d+\.\d+)", result.stdout)
+    if match:
+        installed_version = match.group(1)
+        logger.debug(f"detected skopeo version: {installed_version}")
+
+        # Compare the extracted version with the required version
+        if version.parse(installed_version) < version.parse("1.15.0"):
+            logger.error(
+                f"skopeo version {installed_version} is lower than 1.15. Consider upgrading. Downloading the model might fail."
+            )
+    else:
+        logger.error(
+            "Failed to determine skopeo version. Recommended version is 1.15. Downloading the model might fail."
+        )
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -1,10 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+from pathlib import Path
 import abc
-import json
+import logging
 import os
-import re
 import subprocess
 
 # Third Party
@@ -16,7 +16,7 @@ import click
 # First Party
 from instructlab import clickext
 from instructlab.configuration import DEFAULTS
-from instructlab.utils import is_huggingface_repo, is_oci_repo
+from instructlab.utils import _extract_SHA, _load_json, is_huggingface_repo, is_oci_repo
 
 
 class Downloader(abc.ABC):
@@ -38,7 +38,7 @@ class Downloader(abc.ABC):
 
 
 class HFDownloader(Downloader):
-    """Class to handle downloading safetensors and GGUF models from Huggingface"""
+    """Class to handle downloading safetensors and GGUF models from Hugging Face"""
 
     def __init__(
         self,
@@ -52,17 +52,16 @@ class HFDownloader(Downloader):
         super().__init__(
             repository=repository, release=release, download_dest=download_dest
         )
-        self.repository = repository
-        self.release = release
-        self.download_dest = download_dest
         self.filename = filename
         self.hf_token = hf_token
         self.ctx = ctx
 
     def download(self):
-        """Download the model(s) to train"""
+        """
+        Download specified model from Hugging Face
+        """
         click.echo(
-            f"Downloading model from huggingface: {self.repository}@{self.release} to {self.download_dest}..."
+            f"Downloading model from Hugging Face : {self.repository}@{self.release} to {self.download_dest}..."
         )
 
         if self.hf_token == "" and "instructlab" not in self.repository:
@@ -76,7 +75,7 @@ class HFDownloader(Downloader):
             if self.ctx.obj is not None:
                 hf_logging.set_verbosity(self.ctx.obj.config.general.log_level.upper())
             files = list_repo_files(repo_id=self.repository, token=self.hf_token)
-            if any(".safetensors" in string for string in files):
+            if any(".safetensors" in fname for fname in files):
                 self.download_safetensors()
             else:
                 self.download_gguf()
@@ -100,18 +99,18 @@ class HFDownloader(Downloader):
 
         except Exception as exc:
             click.secho(
-                f"Downloading GGUF model failed with the following HuggingFace Hub error: {exc}",
+                f"Downloading GGUF model failed with the following Hugging Face  Hub error: {exc}",
                 fg="red",
             )
             raise click.exceptions.Exit(1)
 
     def download_safetensors(self) -> None:
         try:
-            if not os.path.exists(os.path.join(self.download_dest, self.repository)):
-                os.makedirs(
-                    name=os.path.join(self.download_dest, self.repository),
-                    exist_ok=True,
-                )
+            os.makedirs(
+                name=os.path.join(self.download_dest, self.repository),
+                exist_ok=True,
+            )
+
             snapshot_download(
                 token=self.hf_token,
                 repo_id=self.repository,
@@ -120,7 +119,7 @@ class HFDownloader(Downloader):
             )
         except Exception as exc:
             click.secho(
-                f"Downloading safetensors model failed with the following HuggingFace Hub error: {exc}",
+                f"Downloading safetensors model failed with the following Hugging Face  Hub error: {exc}",
                 fg="red",
             )
             raise click.exceptions.Exit(1)
@@ -136,60 +135,53 @@ class OCIDownloader(Downloader):
         super().__init__(
             repository=repository, release=release, download_dest=download_dest
         )
-        self.repository = repository
-        self.release = release
-        self.download_dest = download_dest
         self.ctx = ctx
 
     def _build_oci_model_file_map(self, oci_model_path: str) -> dict:
         """
         Helper function to build a mapping between blob files and what they represent
+        Format for the index.json file can be found here: https://github.com/opencontainers/image-spec/blob/main/image-layout.md#indexjson-file
         """
         index_hash = ""
+        index_ref_path = f"{oci_model_path}/index.json"
         try:
-            with open(f"{oci_model_path}/index.json", mode="r", encoding="UTF-8") as f:
-                index_ref = json.load(f)
-            match = re.search("sha256:(.*)", index_ref["manifests"][0]["digest"])
+            index_ref = _load_json(Path(index_ref_path))
+            match = None
+            for manifest in index_ref["manifests"]:
+                if (
+                    manifest["mediaType"]
+                    == "application/vnd.oci.image.manifest.v1+json"
+                ):
+                    match = _extract_SHA(manifest["digest"])
 
             if match:
                 index_hash = match.group(1)
             else:
-                click.echo(f"could not find hash for index file at: {oci_model_path}")
-                raise click.exceptions.Exit(1)
-        except FileNotFoundError as exc:
-            raise ValueError(f"file not found: {oci_model_path}/index.json") from exc
-        except json.JSONDecodeError as exc:
-            raise ValueError(
-                f"could not read JSON file: {oci_model_path}/index.json"
-            ) from exc
+                raise ValueError(
+                    f"could not find hash for index file at: {oci_model_path}"
+                )
         except Exception as exc:
-            raise ValueError("unexpected error occurred: {e}") from exc
+            raise exc
 
+        blob_dir = f"{oci_model_path}/blobs/sha256"
         try:
-            with open(
-                f"{oci_model_path}/blobs/sha256/{index_hash}",
-                mode="r",
-                encoding="UTF-8",
-            ) as f:
-                index = json.load(f)
-        except FileNotFoundError as exc:
-            raise ValueError(f"file not found: {oci_model_path}/index.json") from exc
-        except json.JSONDecodeError as exc:
-            raise ValueError(
-                f"could not read JSON file: {oci_model_path}/index.json"
-            ) from exc
+            index = _load_json(Path(f"{blob_dir}/{index_hash}"))
         except Exception as exc:
-            raise ValueError("unexpected error occurred: {e}") from exc
+            raise exc
 
         title_ref = "org.opencontainers.image.title"
         oci_model_file_map = {}
+        try:
+            for layer in index["layers"]:
+                match = _extract_SHA(layer["digest"])
 
-        for layer in index["layers"]:
-            match = re.search("sha256:(.*)", layer["digest"])
-
-            if match:
-                blob_name = match.group(1)
-                oci_model_file_map[blob_name] = layer["annotations"][title_ref]
+                if match:
+                    blob_name = match.group(1)
+                    oci_model_file_map[blob_name] = layer["annotations"][title_ref]
+        except Exception as exc:
+            raise ValueError(
+                f"failed to build OCI model file mapping from: {blob_dir}/{index_hash}"
+            ) from exc
 
         return oci_model_file_map
 
@@ -198,8 +190,8 @@ class OCIDownloader(Downloader):
             f"Downloading model from OCI registry: {self.repository}@{self.release} to {self.download_dest}..."
         )
 
-        os.makedirs(self.download_dest, exist_ok=True)
         model_name = self.repository.split("/")[-1]
+        os.makedirs(os.path.join(self.download_dest, model_name), exist_ok=True)
         oci_dir = f"{DEFAULTS.OCI_DIR}/{model_name}"
         os.makedirs(oci_dir, exist_ok=True)
 
@@ -209,7 +201,10 @@ class OCIDownloader(Downloader):
             f"{self.repository}:{self.release}",
             f"oci:{oci_dir}",
         ]
-        if self.ctx.obj.config.general.log_level == "DEBUG":
+        if (
+            self.ctx.obj is not None
+            and self.ctx.obj.config.general.log_level == logging.DEBUG
+        ):
             command.append("--debug")
 
         try:
@@ -219,29 +214,29 @@ class OCIDownloader(Downloader):
                 "skopeo not installed, but required to perform downloads from OCI registries. Exiting",
             ) from exc
         except Exception as e:
-            raise ValueError(
-                f"CalledProcessError: command exited with non-zero code: {e}"
-            ) from e
+            click.secho(
+                f"unexpected error: {e}",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
 
         file_map = self._build_oci_model_file_map(oci_dir)
 
-        for _, _, files in os.walk(f"{oci_dir}/blobs/sha256/"):
+        blob_dir = f"{oci_dir}/blobs/sha256/"
+        for _, _, files in os.walk(blob_dir):
             for name in files:
                 if name not in file_map:
                     continue
                 dest = file_map[name]
-                if not os.path.exists(os.path.join(self.download_dest, model_name)):
-                    os.makedirs(
-                        os.path.join(self.download_dest, model_name), exist_ok=True
-                    )
+                dest_model_path = os.path.join(self.download_dest, model_name, dest)
                 # unlink any existing version of the file
-                if os.path.exists(os.path.join(self.download_dest, model_name, dest)):
-                    os.unlink(os.path.join(self.download_dest, model_name, dest))
+                if os.path.exists(dest_model_path):
+                    os.unlink(dest_model_path)
 
-                # create hard link to files in cache, to avoid redownloading if model has been downloaded before
-                os.link(
-                    os.path.join(f"{oci_dir}/blobs/sha256/", name),
-                    os.path.join(self.download_dest, model_name, dest),
+                # create symlink to files in cache, to avoid redownloading if model has been downloaded before
+                os.symlink(
+                    os.path.join(blob_dir, name),
+                    dest_model_path,
                 )
 
 
@@ -250,13 +245,13 @@ class OCIDownloader(Downloader):
     "--repository",
     default=DEFAULTS.MERLINITE_GGUF_REPO,  # TODO: add to config.yaml
     show_default=True,
-    help="HuggingFace or OCI repository of the model to download.",
+    help="Hugging Face  or OCI repository of the model to download.",
 )
 @click.option(
     "--release",
     default="main",  # TODO: add to config.yaml
     show_default=True,
-    help="The revision of the model to download - e.g. a branch, tag, or commit hash for Huggingface repositories and tag or commit has for OCI repositories.",
+    help="The revision of the model to download - e.g. a branch, tag, or commit hash for Hugging Face  repositories and tag or commit hash for OCI repositories.",
 )
 @click.option(
     "--filename",
@@ -296,7 +291,7 @@ def download(ctx, repository, release, filename, model_dir, hf_token):
         )
     else:
         click.secho(
-            f"repository {repository} matches neither Huggingface, nor OCI registry format. Please supply a valid repository",
+            f"repository {repository} matches neither Hugging Face nor OCI registry format. Please supply a valid repository",
             fg="red",
         )
         raise click.exceptions.Exit(1)
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -1,7 +1,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+import abc
+import json
 import os
+import re
+import subprocess
 
 # Third Party
 from huggingface_hub import hf_hub_download, list_repo_files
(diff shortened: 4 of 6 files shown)


src/instructlab/model/download.py
```diff
@@ -12,6 +16,197 @@
...
+
+        os.makedirs(self.download_dest, exist_ok=True)
+        model_name = self.repository.split("/")[-1]
+        oci_dir = f"{DEFAULTS.OCI_DIR}/{model_name}"
+        os.makedirs(oci_dir, exist_ok=True)
+
+        try:
+            subprocess.run(
+                [
+                    "skopeo",
```
leseb, line 184: A few thoughts:

* This expects that `skopeo` is present on the system. We should validate that upfront.
* Why Skopeo and not `crane`?
* Is skopeo taking care of the checksumming? Or is it done via a separate command?
* Should we support other tools?

src/instructlab/model/download.py
```diff
@@ -12,6 +16,199 @@
...
+            index_ref = json.load(f)
+        index_hash = re.search("sha256:(.*)", index_ref["manifests"][0]["digest"])
+
+        if index_hash is not None:
+            index_hash = index_hash.group(1)
+
+        with open(
+            f"{oci_model_path}/blobs/sha256/{index_hash}", mode="r", encoding="UTF-8"
+        ) as f:
+            index = json.load(f)
```
leseb, line 159: How do you handle loading failures?

jaideepr97: added some error handling logic, PTAL

src/instructlab/configuration.py
```diff
@@ -97,6 +98,10 @@ def _reset(self):
     def CHECKPOINTS_DIR(self) -> str:
         return path.join(self._cache_home, STORAGE_DIR_NAMES.CHECKPOINTS)
 
+    @property
```
cdoern, line 101: is this added to the list of dirs necessary upon `init`? I think there is a list somewhere

src/instructlab/model/download.py
```diff
@@ -12,6 +16,232 @@
...
+        try:
+            if self.ctx.obj is not None:
+                hf_logging.set_verbosity(self.ctx.obj.config.general.log_level.upper())
+            files = list_repo_files(repo_id=self.repository, token=self.hf_token)
+            if any(".safetensors" in string for string in files):
+                self.download_safetensors()
+            else:
+                self.download_gguf()
+
+        except Exception as exc:
```
cdoern, line 84: what kind of exceptions are there here? I am fine with this for now if its too hard to catch all different types

src/instructlab/model/download.py
```diff
@@ -12,6 +16,232 @@
...
+        try:
+            hf_hub_download(
+                token=self.hf_token,
+                repo_id=self.repository,
+                revision=self.release,
+                filename=self.filename,
+                local_dir=self.download_dest,
+            )
+
+        except Exception as exc:
```
cdoern, line 101: same here

src/instructlab/utils.py
```diff
@@ -541,3 +541,20 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+    # TO DO: flesh this out and make it a more robust check
+    oci_url_prefix = "docker://"
+    return repo_url.startswith(oci_url_prefix)
+
+
+def is_tool_present(name: str) -> bool:
+    # Standard
+    from shutil import which
+
+    return which(name) is not None
```
cdoern, line 560: I wonder how this will behave in CI and on systems with different permissions. Will it always pick up the proper bin?

src/instructlab/utils.py
```diff
@@ -541,3 +541,20 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
     return hostname, port
+
+
+def is_oci_repo(repo_url: str) -> bool:
+    """
+    Checks if a provided repository follows the OCI registry URL syntax
+    """
+
+    # TO DO: flesh this out and make it a more robust check
+    oci_url_prefix = "docker://"
```
cdoern, line 552: so there are a few different type of prefixes here. I think `docker://` is ok for a starter. I think `oci://` is also one though? unsure honestly

cdoern (commented):
This looks good so far. As for tests. I'd stick with e2e for this one.

The best way is probably to mock up a local registry? and then pass that in as the place to download from. Or point it to some registry we know for now will always be there.

unit will be hard but you can test the specific functions individually. You can modify the existing download tests to still work for the HF path though

src/instructlab/utils.py
jaideepr97: The assumption is that if `skopeo` is present it would be on the system path

src/instructlab/utils.py
jaideepr97: ack, I can add `oci://` as well

src/instructlab/configuration.py
jaideepr97: sounds like you're referring to https://github.com/instructlab/instructlab/blob/main/src/instructlab/configuration.py#L525

I can add it there, though I'm creating that directory anyway so it wouldn't matter 
I don't know if we want to stick to creating all directories at initialization or just create them as and when we need them in the code

src/instructlab/model/download.py
jaideepr97: unsure, I don't see `hf_hub_download` raising many exceptions 
https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/file_download.py#L1010

src/instructlab/model/download.py
jaideepr97: https://github.com/instructlab/instructlab/pull/1756/files/fb384f0445a591c65da5f66f1c931ffde6ac2d3c#r1683287534

src/instructlab/utils.py
n1hility: oci normally refers to oci directories vs registries, but I have some references to a different interpretation although from my understanding not widespread. Relatedly see https://github.com/containers/skopeo/issues/633

if we do support that convention we would need to remap it to a docker url scheme before passing off to skopeo

src/instructlab/model/download.py
```diff
@@ -12,6 +16,232 @@
...
+                    continue
+                dest = file_map[name]
+                if not os.path.exists(os.path.join(self.download_dest, model_name)):
+                    os.makedirs(
+                        os.path.join(self.download_dest, model_name), exist_ok=True
+                    )
+                    # create hard link to files in cache, to avoid redownloading if model has been downloaded before
+                    os.link(
+                        os.path.join(f"{oci_dir}/blobs/sha256/", name),
+                        os.path.join(self.download_dest, model_name, dest),
```
n1hility, line 243: If this is an update the destination might already exist. It should probably unlink before linking.

src/instructlab/utils.py
jaideepr97: thanks, jason
I'll probably leave it as is for now then

src/instructlab/model/download.py
jaideepr97: good catch! updated, thanks

src/instructlab/model/download.py
leseb: `crane` does not require a daemon either. Thanks for the clarification!

src/instructlab/utils.py
```diff
@@ -541,3 +541,20 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+    return repo_url.startswith(oci_url_prefix)
+
+
+def is_tool_present(name: str) -> bool:
```
leseb, line 556: You don't need a function, you can do this directly in an `if` statement. Thanks

src/instructlab/utils.py
```diff
@@ -541,3 +541,20 @@ def split_hostport(hostport: str) -> tuple[str, int]:
     if not hostname or not port:
         raise ValueError(f"Invalid host-port string: '{hostport}'")
     return hostname, port
+
+
+def is_oci_repo(repo_url: str) -> bool:
```
leseb, line 546: Given that check for a single string, I'd remove this function and do the check with an `if` statement.

src/instructlab/model/download.py
leseb: This code just moved so the line was already present.

src/instructlab/model/download.py
```diff
@@ -12,6 +16,236 @@
...
+                dest = file_map[name]
+                if not os.path.exists(os.path.join(self.download_dest, model_name)):
+                    os.makedirs(
+                        os.path.join(self.download_dest, model_name), exist_ok=True
+                    )
+                # unlink any existing version of the file
+                if os.path.exists(os.path.join(self.download_dest, model_name, dest)):
+                    os.unlink(os.path.join(self.download_dest, model_name, dest))
+
+                # create hard link to files in cache, to avoid redownloading if model has been downloaded before
```
leseb, line 244: Why a hard link and not a symlink?

src/instructlab/model/download.py
```diff
@@ -12,6 +16,236 @@
...
+        download_dest: str,
+    ) -> None:
+        super().__init__(
+            repository=repository, release=release, download_dest=download_dest
+        )
+        self.repository = repository
+        self.release = release
+        self.download_dest = download_dest
+
+    def _build_oci_model_file_map(self, oci_model_path: str) -> dict:
```
leseb, line 145: About this code, how do we ensure that the model does not change and this code doesn't break? What OCI releases a new spec/version and we lose compatibility?

Is skopeo, capable of doing this? Can skopeo give us the information we are looking for by reading the directory/image?

src/instructlab/utils.py
jaideepr97: is this not something that we may want to reuse in the future potentially?

src/instructlab/model/download.py
jaideepr97: I suppose there isn't a strong reason for that, maybe a symlink would work as well
the only thing is maybe in case we lose either the cache or the models_dir for whatever reason, we wouldn't have to redownlaod the models with a hard link but we would if it was a symlink

src/instructlab/utils.py
jaideepr97: I just made this a function because our oci check is very simple right now, but i thought we may need to make this more robust in the future, so we could develop this logic later and it would be more testable this way

src/instructlab/utils.py
leseb: There isn't much to re-use when the function is just line :)

jeremyeder:
I was able to download models using this branch.  Thanks!

```(venv) jeder@dev: /storage/tmp/instructlab $ ilab model download --repository docker://quay.io/ai-lab/models/granite-7b-lab --release latest
Downloading model from OCI registry: docker://quay.io/ai-lab/models/granite-7b-lab@latest to /home/jeder/.local/share/instructlab/models...
Getting image source signatures
Copying blob 88bc90d34dff done   |
Copying blob 5d96ed7903c0 done   |
Copying blob e8285413be69 done   |
Copying blob 59bbfb6b312b done   |
Copying blob 078e999c62c3 done   |
Copying blob f36673df9c7f done   |
Copying blob da5c4bd66dca done   |
Copying blob 9e556afd4421 done   |
Copying blob 6c507cde1b53 done   |
Copying blob afa302f67a27 done   |
Copying blob 72ce22c1313f done   |
Copying blob d16bf783cb66 done   |
Copying blob b4cbcd39f603 done   |
Copying config 44136fa355 done   |
Writing manifest to image destination```

File: src/instructlab/model/download.py (hunk 12 above)
Comment (leseb) [reply src/instructlab/model/download.py:145]: I spoke with @n1hility about that. We need to write down which version of the OCI spec was used to write this logic.

**Review by leseb — COMMENTED** [review]
I'm happy with the general state of the PR. The last thing I'd like to add (despite the existing comment) is a small functional test in `scripts/functional-tests.sh`. It's also almost time to squash your commits. Thanks!

**Comment by jaideepr97** [comment]
@leseb I was thinking about adding an e2e test in `scripts/basic-workflow-tests.sh` instead,  so that the completion of the e2e test would automatically validate that the model was downloaded correctly and in a usable state 
Lmk how that sounds to you

File: src/instructlab/utils.py (hunk 9 above)
Comment (jaideepr97) [reply src/instructlab/utils.py:556]: removed, thanks!

**Comment by leseb** [comment]
> @leseb I was thinking about adding an e2e test in `scripts/basic-workflow-tests.sh` instead, so that the completion of the e2e test would automatically validate that the model was downloaded correctly and in a usable state Lmk how that sounds to you

Does that mean not testing HF in e2e anymore? I believe the HF download is our main entry point so we should test it by default.

Any functional test is good! They both validate the model in different ways, and the `scripts/basic-workflow-tests.sh` is more advanced, takes more time to run, and is not free. IMO, if download + serve + chat is sufficient (it should be). I'm leaning towards `scripts/basic-workflow-tests.sh` simply because the GitHub runners are free. Thanks!

File: src/instructlab/model/download.py (hunk 12 above)
Comment (jaideepr97) [reply src/instructlab/model/download.py:146]: @leseb what would the best place to document that be? should it be in this PR?

**Comment by jaideepr97** [comment]
>Does that mean not testing HF in e2e anymore? I believe the HF download is our main entry point so we should test it by default.

I added a separate case, so both get tested 

>Any functional test is good! They both validate the model in different ways, and the scripts/basic-workflow-tests.sh is more advanced, takes more time to run, and is not free. IMO, if download + serve + chat is sufficient (it should be). I'm leaning towards scripts/basic-workflow-tests.sh simply because the GitHub runners are free. Thanks!

not sure I fully understood if you were for or against putting the test in `basic-worflow-tests.sh` ultimately. Could you clarify?

File: src/instructlab/model/download.py [hunk 13]
```diff
@@ -48,38 +283,28 @@
...
+    else:
+        downloader = HFDownloader(
+            repository=repository,
+            release=release,
+            download_dest=model_dir,
+            filename=filename,
+            hf_token=hf_token,
+            ctx=ctx,
+        )
+
```
leseb, line 302: What if the repo is not OCI and not HF? We need to handle that case, or is it handled somewhere I can't find it? We need to validate `repository` further. So you need if, elif, else here if not done.

```
@click.option(
    "--repository",
    default=DEFAULTS.MERLINITE_GGUF_REPO,  # TODO: add to config.yaml
    show_default=True,
    help="Hugging Face repository of the model to download.",
)
```

Also, the help section must be updated.

leseb:
> > Does that mean not testing HF in e2e anymore? I believe the HF download is our main entry point so we should test it by default.
> 
> I added a separate case, so both get tested

Can you clarify how both are being tested?

> 
> > Any functional test is good! They both validate the model in different ways, and the scripts/basic-workflow-tests.sh is more advanced, takes more time to run, and is not free. IMO, if download + serve + chat is sufficient (it should be). I'm leaning towards scripts/basic-workflow-tests.sh simply because the GitHub runners are free. Thanks!
> 
> not sure I fully understood if you were for or against putting the test in `basic-worflow-tests.sh` ultimately. Could you clarify?

Oops I made a typo sorry, I meant to say `scripts/functional-tests.sh` and not `scripts/basic-workflow-tests.sh`. I'm trying to think about the cost of the runners. 


In the end, let's not forget that the CI as a cost ($$$), and if we can reduce it, it's great.

src/instructlab/model/download.py
leseb: Doc string in code + commit message. Add as much details as you can, thanks 🙏🏻

src/instructlab/model/download.py
jaideepr97: added the check for repository type and updated the help msg

src/instructlab/model/download.py
```diff
@@ -48,38 +283,33 @@
...
+            repository=repository,
+            release=release,
+            download_dest=model_dir,
+            filename=filename,
+            hf_token=hf_token,
+            ctx=ctx,
+        )
+    else:
+        click.secho(
+            "repository supplied matches neither Huggingface, nor OCI registry format. Please supply a valid repository",
```
leseb, line 303: Please print `repository` when this message is shown.

src/instructlab/model/download.py
```diff
@@ -12,20 +17,250 @@
...
+        model_name = self.repository.split("/")[-1]
+        oci_dir = f"{DEFAULTS.OCI_DIR}/{model_name}"
+        os.makedirs(oci_dir, exist_ok=True)
+
+        if which("skopeo") is None:
+            click.secho(
+                "skopeo not detected, but required to perform downloads from OCI registries. Exiting",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
```
leseb, line 213: Sorry, but I just realized that instead of this block, we could simply catch the `FileNotFoundError` from the subprocess and handle it appropriately.

src/instructlab/model/download.py
```diff
@@ -12,20 +17,250 @@
...
+
+        try:
+            subprocess.run(
+                [
+                    "skopeo",
+                    "copy",
+                    f"{self.repository}:{self.release}",
+                    f"oci:{oci_dir}",
```
leseb, line 221: I think we should use Popen here so that we can:

* log stdout while the download is in progress
* print stderr/stdout when rc is not 0

This might look like:

```python
process = subprocess.Popen(
    [
        "skopeo",
        "copy",
        "docker://quay.io/ai-lab/models/granite-7b-lab:latest",
        f"oci:{oci_dir}",
    ],
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
    text=True,
)

# Print stdout line by line as it's being produced
while True:
    output = process.stdout.readline()
    if output == "" and process.poll() is not None:
        break
    if output:
        print(output.strip())

# Check if the process exited with an error
if process.returncode != 0:
    stderr = process.stderr.read()
    print(f"Error: {stderr}", file=sys.stderr)
```

src/instructlab/model/download.py
```diff
@@ -12,20 +17,250 @@
...
+                ],
+                check=True,
+            )
+        except Exception as e:
+            raise ValueError(
+                "CalledProcessError: command exited with non-zero code"
```
leseb, line 227: We need to print the error.

jaideepr97: fixed, thanks

src/instructlab/model/download.py
jaideepr97: fixed, thanks

src/instructlab/model/download.py
jaideepr97: added a note with version in both places

src/instructlab/model/download.py
jaideepr97: moved that error check msg to a fileNotFoundError handler

src/instructlab/model/download.py
```diff
@@ -12,20 +16,250 @@
...
+                    "skopeo",
+                    "copy",
+                    f"{self.repository}:{self.release}",
+                    f"oci:{oci_dir}",
+                ],
+                check=True,
+            )
+        except FileNotFoundError as e:
+            raise FileNotFoundError(
+                "skopeo not detected, but required to perform downloads from OCI registries. Exiting",
```
leseb, line 222:
```suggestion
                "skopeo not installed, but required to perform downloads from OCI registries. Exiting",
```

Also, if you don't e` please remove it.

src/instructlab/model/download.py
jaideepr97: possibly because using a PIPE breaks the in-place update that happens to the terminal o/p for the progress bar?

leseb: Oh that's right, weird, nevermind then, the current code is working better.

We still need to capture stdout/stderr on failure though. Where is it done?

jaideepr97: from https://docs.python.org/3.5/library/subprocess.html#subprocess.run
```
If check is true, and the process exits with a non-zero exit code, a [CalledProcessError](https://docs.python.org/3.5/library/subprocess.html#subprocess.CalledProcessError) exception will be raised.
```
that's what's being handled right now

leseb: It also says:

```
The CalledProcessError object will have the return code
    in the returncode attribute, and output & stderr attributes if those streams
    were captured.
```

Today are not captured as far as I can tell.

.github/workflows/test.yml
```diff
@@ -29,6 +29,7 @@ on:
 
 env:
   LC_ALL: en_US.UTF-8
+  REGISTRY_AUTH_FILE: /run/containers/1001/auth.json
```
leseb, line 32: Is this going to work? The func test has `REGISTRY_AUTH_FILE=$HOME/auth.json`

scripts/functional-tests.sh
```diff
@@ -159,6 +159,24 @@ fi
...
+
+    expect -c '
+        set timeout 120
+        spawn ilab chat
+        expect ">>>"
+        send "this is a test! what are you? do not exceed ten words in your reply.\r"
+        send "exit\r"
+        expect eof
+    '
+}
```
leseb, line 178: You can remove all that, as discussed offline vLLM is not supported in this test so we only want to validate that the command succeeded.

.github/workflows/test.yml
jaideepr97: no need for this anymore. so I will be removing it

leseb (commented):
I feel there's an issue with the current implementation, at least from a UX perspective. Currently, we use the `--repository` flag to point to the HF registry/domain/model and the `--release` flag to specify a release/tag/version of a model in that registry.

Now, we want to support a different registry format called OCI. The proposal is to re-use both `--repository` and `--release` flags for downloading a model from the OCI registry. The way we determine if the model is stored as OCI or in HF is by checking if the repository is prefixed with `docker://`.

I think it would be better to introduce a new flag, `--registry`, to represent the registry address. It would default to HF but would support OCI as well. This would provide us with a more consistent approach for both registries:

```
ilab model download --registry=quay.io --repository ai-lab/granite-7b-lab --release latest
```

In this scenario, the `docker://` URI would be added internally when the registry is `quay.io` or `docker.io`. This approach seems more consistent, user-friendly and offers a similar experience whether the user uses an HF or OCI registry.

Thoughts? @jaideepr97 @n1hility

n1hility:
> ```
> ilab model download --registry=quay.io --repository ai-lab/granite-7b-lab --release latest
> ```

I like the idea of supporting these flags. Although the problem with this approach is that user's can/will have their own registries, so we need a way to specify a protocol. We could add an additional flag:

 ilab model download --registry=quay.io --repository ai-lab/granite-7b-lab --release latest --protocol docker (maybe oci as well?)

Although this gets really long and not as user friendly/fast to type as, so IMO still worth having a URL convention:

ilab model download docker://ai-lab/granite-7b-lab:latest

Let's chat more. Great topic

src/instructlab/model/download.py
n1hility: The stdout / stderr are going to the ilab's stdout, stderr which you need for the tty progress bars capability (other than creating a pty would be a bit complex). You can interpret error codes 2 = not found and 1 some other error. A common approach would be "see above previous errors for failure details"

src/instructlab/model/download.py
jaideepr97: done, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,250 @@
...
+        os.makedirs(self.download_dest, exist_ok=True)
+        model_name = self.repository.split("/")[-1]
+        oci_dir = f"{DEFAULTS.OCI_DIR}/{model_name}"
+        os.makedirs(oci_dir, exist_ok=True)
+
+        try:
+            subprocess.run(
+                [
+                    "skopeo",
+                    "copy",
```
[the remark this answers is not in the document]
jaideepr97: added, thanks

.github/workflows/test.yml
jaideepr97: removed

scripts/functional-tests.sh
jaideepr97: removed

booxter:
Added milestone 0.18 since the referred issue has the same milestone. Please revert if this is wrong for some reason.

src/instructlab/model/download.py
```diff
@@ -12,20 +16,248 @@
...
+
+        try:
+            subprocess.run(command, check=True)
+        except FileNotFoundError as exc:
+            raise FileNotFoundError(
+                "skopeo not installed, but required to perform downloads from OCI registries. Exiting",
+            ) from exc
+        except Exception as e:
+            raise ValueError(
+                "CalledProcessError: command exited with non-zero code: {e}"
```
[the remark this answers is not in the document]
jaideepr97: fixed, thanks

jaideepr97:
>Added milestone 0.18 since the referred issue has the same milestone. Please revert if this is wrong for some reason.

AFAIK what we were following was that if an issue exists, the PR doesn't need to be in the milestone

tiran:
Have you considered to split the commands into `ilab model download hf` and `ilab model download oci`? Not all options apply to all types of registries.

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+                f"could not read JSON file: {oci_model_path}/index.json"
+            ) from exc
+        except Exception as exc:
+            raise ValueError("unexpected error occurred: {e}") from exc
+
+        title_ref = "org.opencontainers.image.title"
+        oci_model_file_map = {}
+
+        for layer in index["layers"]:
+            match = re.search("sha256:(.*)", layer["digest"])
```
booxter, line 188: nit: you could have a tiny helper function to extract sha from a layer, then use it here and above instead of implementing it twice.

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+            raise ValueError(
+                f"could not read JSON file: {oci_model_path}/index.json"
+            ) from exc
+        except Exception as exc:
+            raise ValueError("unexpected error occurred: {e}") from exc
+
+        title_ref = "org.opencontainers.image.title"
+        oci_model_file_map = {}
+
+        for layer in index["layers"]:
```
booxter, line 187: Above, you make an attempt to catch any exception from malformed `index.json` and convert it into a `ValueError` (except one case). But here, you assume the json conforms to standard (?). Should you also handle errors here? (at least with a bare `except Exception`?)

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+            ) from e
+
+        file_map = self._build_oci_model_file_map(oci_dir)
+
+        for _, _, files in os.walk(f"{oci_dir}/blobs/sha256/"):
+            for name in files:
+                if name not in file_map:
+                    continue
+                dest = file_map[name]
+                if not os.path.exists(os.path.join(self.download_dest, model_name)):
```
booxter, line 233: redundant since you  do `exist_ok=True`; Also, no need to attempt to create the same directory over and over on each file.

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+                    os.makedirs(
+                        os.path.join(self.download_dest, model_name), exist_ok=True
+                    )
+                # unlink any existing version of the file
+                if os.path.exists(os.path.join(self.download_dest, model_name, dest)):
```
booxter, line 238: nit: you construct the path name twice. calculate it once for both `exists` and `unlink`

src/instructlab/model/download.py
booxter: hardlinks won't work across file systems. Is it always the case that the cache and model_dir are on the same FS? I believe @tiran mentioned that he hosts the data dir on NFS while keeping cache local. I think it won't work then.

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+                    os.unlink(os.path.join(self.download_dest, model_name, dest))
+
+                # create hard link to files in cache, to avoid redownloading if model has been downloaded before
+                os.link(
+                    os.path.join(f"{oci_dir}/blobs/sha256/", name),
```
booxter, line 243: nitty-nit: consider introducing a variable to carry the sha-blob dir name (it's also used in line 228).

booxter (changes requested):
Thanks for the PR. I left all comments I have (major and tiny nits, plus some curious questions) in the review. Please take what makes sense and ignore the rest.

src/instructlab/model/download.py
jaideepr97: added, thanks

src/instructlab/model/download.py
jaideepr97: fixed, thanks

src/instructlab/model/download.py
jaideepr97: fixed, thanks

src/instructlab/model/download.py
jaideepr97: I'm all for helper functions, but whenever I've tried to introduce them in PRs in the past but I've gotten told to just do it in-place in reviews 🤷

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+
+    def _build_oci_model_file_map(self, oci_model_path: str) -> dict:
+        """
+        Helper function to build a mapping between blob files and what they represent
+        """
+        index_hash = ""
+        try:
+            with open(f"{oci_model_path}/index.json", mode="r", encoding="UTF-8") as f:
+                index_ref = json.load(f)
+            match = re.search("sha256:(.*)", index_ref["manifests"][0]["digest"])
```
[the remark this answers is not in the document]
jaideepr97: updated, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        except FileNotFoundError as exc:
+            raise ValueError(f"file not found: {oci_model_path}/index.json") from exc
+        except json.JSONDecodeError as exc:
+            raise ValueError(
+                f"could not read JSON file: {oci_model_path}/index.json"
+            ) from exc
+        except Exception as exc:
+            raise ValueError("unexpected error occurred: {e}") from exc
+
+        try:
```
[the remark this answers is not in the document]
jaideepr97: I agree, that would be ideal but https://github.com/instructlab/instructlab/pull/1756#discussion_r1688741027
I tried to add this exact function in an earlier PR

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+            if match:
+                index_hash = match.group(1)
+            else:
+                click.echo(f"could not find hash for index file at: {oci_model_path}")
+                raise click.exceptions.Exit(1)
+        except FileNotFoundError as exc:
+            raise ValueError(f"file not found: {oci_model_path}/index.json") from exc
+        except json.JSONDecodeError as exc:
+            raise ValueError(
+                f"could not read JSON file: {oci_model_path}/index.json"
```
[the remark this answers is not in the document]
jaideepr97: updated

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        self.download_dest = download_dest
+        self.ctx = ctx
+
+    def _build_oci_model_file_map(self, oci_model_path: str) -> dict:
+        """
+        Helper function to build a mapping between blob files and what they represent
+        """
+        index_hash = ""
+        try:
+            with open(f"{oci_model_path}/index.json", mode="r", encoding="UTF-8") as f:
```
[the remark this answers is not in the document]
jaideepr97: removed, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        oci_dir = f"{DEFAULTS.OCI_DIR}/{model_name}"
+        os.makedirs(oci_dir, exist_ok=True)
+
+        command = [
+            "skopeo",
+            "copy",
+            f"{self.repository}:{self.release}",
+            f"oci:{oci_dir}",
+        ]
+        if self.ctx.obj.config.general.log_level == "DEBUG":
```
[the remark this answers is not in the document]
jaideepr97: fixed, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+                """HF_TOKEN var needs to be set in your environment to download HF Model.
+                Alternatively, the token can be passed with --hf-token flag.
+                The HF Token is used to authenticate your identity to the Hugging Face Hub."""
+            )
+
+        try:
+            if self.ctx.obj is not None:
+                hf_logging.set_verbosity(self.ctx.obj.config.general.log_level.upper())
+            files = list_repo_files(repo_id=self.repository, token=self.hf_token)
+            if any(".safetensors" in string for string in files):
```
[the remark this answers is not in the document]
jaideepr97: fixed, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        )
+        self.repository = repository
+        self.release = release
+        self.download_dest = download_dest
+        self.filename = filename
+        self.hf_token = hf_token
+        self.ctx = ctx
+
+    def download(self):
+        """Download the model(s) to train"""
```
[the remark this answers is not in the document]
jaideepr97: fixed, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        click.echo(
+            f"Downloading model from huggingface: {self.repository}@{self.release} to {self.download_dest}..."
```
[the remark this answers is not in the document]
jaideepr97: fixed, thanks

src/instructlab/model/download.py
booxter: I don't know the context before, but here you also duplicate the same parsing logic in two places, so maybe it's an additional argument for having a helper doing the job once?

.github/workflows/test.yml
```diff
@@ -68,10 +68,28 @@ jobs:
...
+        uses: actions/setup-go@v5
+        with:
+          cache: false
+          go-version: 1.22.x
+
+      - name: install skopeo from source
+        if: startsWith(matrix.platform, 'ubuntu')
+        run: |
+          sudo apt-get install libgpgme-dev libassuan-dev libbtrfs-dev libdevmapper-dev pkg-config -y
+          git clone --depth 1 https://github.com/containers/skopeo -b v1.15.0 "$GITHUB_WORKSPACE"/src/github.com/containers/skopeo
```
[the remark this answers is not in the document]
n1hility: agree tag would be better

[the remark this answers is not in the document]
n1hility: oh wait that is a tag, its just creating a branch off of it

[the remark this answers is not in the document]
booxter: Ah right. I misinterpreted what it does. Thanks for double-checking.

n1hility:
> > Have you considered to split the commands into ilab model download hf and ilab model download oci? Not all options apply to all types of registries.
> 
> It might become hard to maintain first class subcommands for each new way to download that we decide to support as we go forward IMO I think having generic flags, with help messages that make their purpose clear under different circumstances could be more maintainable

One of the reasons I like the URI approach is that we get support for N transports for free. Really if you think about it downloading a model is really a network FS operation, download a directory. We could support https, nfs, rsync, git, smb, sftp oci/docker etc without having to change the API and all in a common user-understandable convention.

.github/workflows/test.yml
```diff
@@ -68,10 +68,28 @@ jobs:
...
+          sudo apt-get install -y expect
+
+      - name: Install go for skopeo
+        if: startsWith(matrix.platform, 'ubuntu')
+        uses: actions/setup-go@v5
+        with:
+          cache: false
+          go-version: 1.22.x
+
+      - name: install skopeo from source
```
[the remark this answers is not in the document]
jaideepr97: added a comment, it's because the latest version of skopeo available on ubuntu is 1.4 which is very old and was running into issues downloading the artifacts properly

.github/workflows/test.yml
```diff
@@ -67,15 +68,33 @@ jobs:
         if: matrix.platform != 'macos-latest'
         uses: ./.github/actions/free-disk-space
 
-      - name: Install the expect package
```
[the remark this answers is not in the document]
jaideepr97: fixed

src/instructlab/utils.py
```diff
@@ -541,3 +541,20 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+    """
+    Checks if a provided repository follows the OCI registry URL syntax
+    """
+
+    # TO DO: flesh this out and make it a more robust check
```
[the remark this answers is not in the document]
jaideepr97: fixed

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
     show_default=True,
-    help="Hugging Face repository of the model to download.",
+    help="HuggingFace or OCI repository of the model to download.",
 )
 @click.option(
     "--release",
     default="main",  # TODO: add to config.yaml
     show_default=True,
-    help="The git revision of the model to download - e.g. a branch, tag, or commit hash.",
+    help="The revision of the model to download - e.g. a branch, tag, or commit hash for Huggingface repositories and tag or commit has for OCI repositories.",
```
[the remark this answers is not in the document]
jaideepr97: fixed, thanks

src/instructlab/model/download.py
```diff
@@ -48,38 +279,33 @@
...
+            repository=repository,
+            release=release,
+            download_dest=model_dir,
+            filename=filename,
+            hf_token=hf_token,
+            ctx=ctx,
+        )
+    else:
+        click.secho(
+            f"repository {repository} matches neither Huggingface, nor OCI registry format. Please supply a valid repository",
```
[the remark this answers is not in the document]
jaideepr97: done

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        except Exception as exc:
+            click.secho(
+                f"Downloading GGUF model failed with the following HuggingFace Hub error: {exc}",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
+
+    def download_safetensors(self) -> None:
+        try:
+            if not os.path.exists(os.path.join(self.download_dest, self.repository)):
```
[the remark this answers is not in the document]
jaideepr97: removed, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        self.hf_token = hf_token
+        self.ctx = ctx
+
+    def download(self):
+        """Download the model(s) to train"""
+        click.echo(
+            f"Downloading model from huggingface: {self.repository}@{self.release} to {self.download_dest}..."
+        )
+
+        if self.hf_token == "" and "instructlab" not in self.repository:
```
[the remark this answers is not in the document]
jaideepr97: yeah, probably should be fixed in a follow up PR

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        release: str,
+        download_dest: str,
+        filename: str,
+        hf_token: str,
+        ctx,
+    ) -> None:
+        super().__init__(
+            repository=repository, release=release, download_dest=download_dest
+        )
+        self.repository = repository
```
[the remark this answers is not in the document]
jaideepr97: removed, thanks

src/instructlab/model/download.py
n1hility: I think it makes sense to switch this to fully qualified symlinks. Relative can sometimes cause issues with container volume mounting.

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+    """
+    Class to handle downloading safetensors models from OCI Registries
+    We are leveraging OCI v1.1 for this functionality
+    """
+
+    def __init__(self, repository: str, release: str, download_dest: str, ctx) -> None:
+        super().__init__(
+            repository=repository, release=release, download_dest=download_dest
+        )
+        self.repository = repository
```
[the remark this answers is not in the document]
jaideepr97: removed, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        self.release = release
+        self.download_dest = download_dest
+        self.ctx = ctx
+
+    def _build_oci_model_file_map(self, oci_model_path: str) -> dict:
+        """
+        Helper function to build a mapping between blob files and what they represent
```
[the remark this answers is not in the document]
jaideepr97: added, thanks

src/instructlab/model/download.py
jaideepr97: yeah
I'll add it again

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+
+        try:
+            with open(
+                f"{oci_model_path}/blobs/sha256/{index_hash}",
+                mode="r",
+                encoding="UTF-8",
+            ) as f:
+                index = json.load(f)
+        except FileNotFoundError as exc:
+            raise ValueError(f"file not found: {oci_model_path}/index.json") from exc
```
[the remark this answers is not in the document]
jaideepr97: fixed, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        try:
+            with open(f"{oci_model_path}/index.json", mode="r", encoding="UTF-8") as f:
+                index_ref = json.load(f)
+            match = re.search("sha256:(.*)", index_ref["manifests"][0]["digest"])
+
+            if match:
+                index_hash = match.group(1)
+            else:
+                click.echo(f"could not find hash for index file at: {oci_model_path}")
+                raise click.exceptions.Exit(1)
```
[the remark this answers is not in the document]
jaideepr97: done, thanks

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+                f"{oci_model_path}/blobs/sha256/{index_hash}",
+                mode="r",
+                encoding="UTF-8",
+            ) as f:
+                index = json.load(f)
+        except FileNotFoundError as exc:
+            raise ValueError(f"file not found: {oci_model_path}/index.json") from exc
+        except json.JSONDecodeError as exc:
+            raise ValueError(
+                f"could not read JSON file: {oci_model_path}/index.json"
```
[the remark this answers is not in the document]
jaideepr97: fixed, thanks

src/instructlab/model/download.py
jaideepr97: switched to symlinks

src/instructlab/model/download.py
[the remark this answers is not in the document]
jaideepr97: added a check, thanks

src/instructlab/model/download.py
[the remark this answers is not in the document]
jaideepr97: added

src/instructlab/model/download.py
jaideepr97: on second thought, i might leave this one as is - it seems like overkill for something that is just 2 lines anyway
I don't know that this is something we will have to do as often as loading a json file for eg to justify a helper function for it

jaideepr97: added

src/instructlab/model/download.py
```diff
@@ -12,20 +16,247 @@
...
+        if self.ctx.obj.config.general.log_level == "DEBUG":
+            command.append("--debug")
+
+        try:
+            subprocess.run(command, check=True)
+        except FileNotFoundError as exc:
+            raise FileNotFoundError(
+                "skopeo not installed, but required to perform downloads from OCI registries. Exiting",
+            ) from exc
+        except Exception as e:
```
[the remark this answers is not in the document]
jaideepr97: made it consistent with HF implementation
there's probably more work needed to be done to make exception handling a bit more consistent in the repo in general, but that can be taken up in a follow up PR

src/instructlab/model/download.py
jaideepr97: added

.github/workflows/test.yml
```diff
@@ -67,15 +68,33 @@ jobs:
...
+        uses: actions/setup-go@v5
+        with:
+          cache: false
+          go-version: 1.22.x
+
+      - name: install skopeo from source
+        if: startsWith(matrix.platform, 'ubuntu')
+        run: |
+          sudo apt-get install libgpgme-dev libassuan-dev libbtrfs-dev libdevmapper-dev pkg-config -y
+          git clone --depth 1 https://github.com/containers/skopeo -b v1.15.0 "$GITHUB_WORKSPACE"/src/github.com/containers/skopeo
```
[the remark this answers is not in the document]
jaideepr97: ack

src/instructlab/model/download.py
jaideepr97: added

n1hility (approved):

src/instructlab/utils.py
```diff
@@ -605,3 +605,39 @@ def ensure_legacy_dataset(
...
+            file_path,
+            encoding="UTF-8",
+        ) as f:
+            return json.load(f)
+    except FileNotFoundError as e:
+        raise ValueError(f"file not found: {file_path}") from e
+    except json.JSONDecodeError as e:
+        raise ValueError(f"could not read JSON file: {file_path}") from e
+    except Exception as e:
+        raise ValueError("unexpected error occurred:") from e
```
booxter, line 639: nit: remove `:`?

src/instructlab/utils.py
```diff
@@ -605,3 +605,39 @@ def ensure_legacy_dataset(
...
+
+
+def is_huggingface_repo(repo_name: str) -> bool:
+    # allow alphanumerics, underscores, hyphens and periods in huggingface repo names
+    # repo name should be of the format <owner>/<model>
+    pattern = r"^[\w.-]+\/[\w.-]+$"
+    return re.match(pattern, repo_name) is not None
+
+
+def _load_json(file_path: Path):
```
booxter, line 627: (No action required) These functions seem closely coupled to download module. (The sha one also to OCI implementation.) I'd put them back into `download.py`.

src/instructlab/utils.py
```diff
@@ -605,3 +605,39 @@ def ensure_legacy_dataset(
...
+            return json.load(f)
+    except FileNotFoundError as e:
+        raise ValueError(f"file not found: {file_path}") from e
+    except json.JSONDecodeError as e:
+        raise ValueError(f"could not read JSON file: {file_path}") from e
+    except Exception as e:
+        raise ValueError("unexpected error occurred:") from e
+
+
+def _extract_SHA(SHAstr: str):
```
booxter, line 642: nit: pep8 requires lower case names for variables: https://peps.python.org/pep-0008/#function-and-variable-names

src/instructlab/model/download.py
```diff
@@ -1,31 +1,258 @@
...
+        self.filename = filename
+        self.hf_token = hf_token
+        self.ctx = ctx
+
+    def download(self):
+        """
+        Download specified model from Hugging Face
+        """
+        click.echo(
+            f"Downloading model from Hugging Face : {self.repository}@{self.release} to {self.download_dest}..."
```
booxter, line 68: nit: remove space character before `:`.

src/instructlab/model/download.py
```diff
@@ -1,31 +1,258 @@
...
+                token=self.hf_token,
+                repo_id=self.repository,
+                revision=self.release,
+                filename=self.filename,
+                local_dir=self.download_dest,
+            )
+
+        except Exception as exc:
+            click.secho(
+                f"Downloading GGUF model failed with the following Hugging Face  Hub error: {exc}",
```
booxter, line 106: nit: double-space after `Face`.

src/instructlab/model/download.py
```diff
@@ -1,31 +1,258 @@
...
+            os.makedirs(
+                name=os.path.join(self.download_dest, self.repository),
+                exist_ok=True,
+            )
+
+            snapshot_download(
+                token=self.hf_token,
+                repo_id=self.repository,
+                revision=self.release,
+                local_dir=os.path.join(self.download_dest, self.repository),
```
booxter, line 122: nit: the local dir is calculated twice (here and in line 114). Could be done once.

src/instructlab/model/download.py
```diff
@@ -1,31 +1,258 @@
...
+        filename: str,
+        hf_token: str,
+        ctx,
+    ) -> None:
+        super().__init__(
+            repository=repository, release=release, download_dest=download_dest
+        )
+        self.filename = filename
+        self.hf_token = hf_token
+        self.ctx = ctx
```
booxter, line 61: nit: both implementations receive `ctx`. You could pass it to base class too and set it there to `self.ctx`.

src/instructlab/model/download.py
```diff
@@ -1,31 +1,258 @@
...
+
+            snapshot_download(
+                token=self.hf_token,
+                repo_id=self.repository,
+                revision=self.release,
+                local_dir=os.path.join(self.download_dest, self.repository),
+            )
+        except Exception as exc:
+            click.secho(
+                f"Downloading safetensors model failed with the following Hugging Face  Hub error: {exc}",
```
booxter, line 126: nit: ditto -> double space character

src/instructlab/model/download.py
```diff
@@ -1,31 +1,258 @@
...
+                ):
+                    match = _extract_SHA(manifest["digest"])
+
+            if match:
+                index_hash = match.group(1)
+            else:
+                raise ValueError(
+                    f"could not find hash for index file at: {oci_model_path}"
+                )
+        except Exception as exc:
```
booxter, line 167: This except block is redundant since it just re-raises the exception back. You can remove it and `try` and this will also un-indent the code.

---

That said, I wonder if instead of that, you'd want to reraise ValueError?

---

Same comments below (line 173).

src/instructlab/model/download.py
```diff
@@ -1,31 +1,258 @@
...
+        index_ref_path = f"{oci_model_path}/index.json"
+        try:
+            index_ref = _load_json(Path(index_ref_path))
+            match = None
+            for manifest in index_ref["manifests"]:
+                if (
+                    manifest["mediaType"]
+                    == "application/vnd.oci.image.manifest.v1+json"
+                ):
+                    match = _extract_SHA(manifest["digest"])
```
booxter, line 159: `break` immediately, otherwise it continues to the next `manifest`. (I don't think that's what you indend?)

src/instructlab/model/download.py
```diff
@@ -1,31 +1,258 @@
...
+
+                if match:
+                    blob_name = match.group(1)
+                    oci_model_file_map[blob_name] = layer["annotations"][title_ref]
+        except Exception as exc:
+            raise ValueError(
+                f"failed to build OCI model file mapping from: {blob_dir}/{index_hash}"
+            ) from exc
+
+        return oci_model_file_map
```
booxter, line 190: (Sorry I haven't noticed it before). So if e.g. no layer matched in line 180, then the returned dictionary will be empty. And then `download` will do nothing, silently returning (instead of e.g. logging an error.) Should we raise an error here, or in `download`, when the dictionary is empty?

src/instructlab/model/download.py
```diff
@@ -1,31 +1,258 @@
...
     show_default=True,
-    help="Hugging Face repository of the model to download.",
+    help="Hugging Face  or OCI repository of the model to download.",
 )
 @click.option(
     "--release",
     default="main",  # TODO: add to config.yaml
     show_default=True,
-    help="The git revision of the model to download - e.g. a branch, tag, or commit hash.",
+    help="The revision of the model to download - e.g. a branch, tag, or commit hash for Hugging Face  repositories and tag or commit hash for OCI repositories.",
```
booxter, line 255: nit: ditto - double space character

src/instructlab/model/download.py
```diff
@@ -48,38 +275,69 @@
...
-                token=hf_token,
-                repo_id=repository,
-                revision=release,
-                filename=filename,
-                local_dir=model_dir,
-            )
+        downloader.download()
+    except FileNotFoundError as exc:
+        click.secho(
+            "skopeo is not installed, please install recommended version 1.15",
```
booxter, line 304: This is a general function for any repo type (not just OCI). I think it's making an assumption that 1) OCI downloader will never raise the exception in any other scenario; 2) HF downloader will never raise it at all.

I think this `except` block belongs to where the `skopeo` presence is actually checked (inside `check_skopeo_version` or maybe inside the `OCIDownloader.download`).

booxter (approved):
I'm leaving some more comments. I think most can be ignored or taken as you see fit. There are two comments that I'd like to be addressed for sure, but these may go into a separate PR as a follow-up if you don't mind.

1. Move the click Exception raise for skopeo version from the general `download` function into `OCIDownloader`.
2. Raise an error on an empty blob map.

booxter:
To be clear: this can merge, my comments don't block the PR. @leseb also asked me to send the follow-up PR to fix the comments, I will do it after this PR merges.

leseb:
I cannot approve a PR that I participated in. So waiting for other maintainers to chime in :)

nathan-weinberg (approved):
Agreed with many of @booxter's nits but non-blocking

AaronJaegerVA:
@jaideepr97 I'm trying to use this functionality. Where should the model files in the container images be located in the filesystem?

jaideepr97:
>@jaideepr97 I'm trying to use this functionality. Where should the model files in the container images be located in the filesystem?

@AaronJaegerVA the model is meant to be an OCI artifact, not a container image

AaronJaegerVA:
> > @jaideepr97 I'm trying to use this functionality. Where should the model files in the container images be located in the filesystem?
> 
> @AaronJaegerVA the model is meant to be an OCI artifact, not a container image

Thanks for responding. Can you explain (or link to instructions) on how to create an acceptable OCI artifact?

AaronJaegerVA:
I found https://oras.land/ but it's completely unfamiliar. I wonder how do you create an OCI artifact from a directory worth of files cloned from huggingface?

AaronJaegerVA:
I was able to upload an artifact, but it wasn't recognized as a model. What kind of file(s) is it expecting?

AaronJaegerVA:
I tossed plan `.safetensors` files into my OCI repository, and tried `ilab model download ...` again. It looks better, but does not list any models. Any points @jaideepr97 ?

```
(app-root) ilab model download --repository docker://quay.example.com/vhaispjaegea/ibm-granite/granite-3.2-2b-instruct --release 0.0.9
INFO 2025-03-19 15:50:56,153 instructlab.model.download:192: Downloading model from OCI registry:
    Model: docker://quay.example.com/vhaispjaegea/ibm-granite/granite-3.2-2b-instruct@0.0.9
    Destination: /opt/app-root/src/.cache/instructlab/models
Copying blob 73833730ea9e skipped: already exists  
Copying blob 5720a4536051 skipped: already exists  
Copying config 44136fa355 done   | 
Writing manifest to image destination
INFO 2025-03-19 15:51:00,185 instructlab.model.download:288: 
ᕦ(òᴗóˇ)ᕤ docker://quay.example.com/vhaispjaegea/ibm-granite/granite-3.2-2b-instruct model download completed successfully! ᕦ(òᴗóˇ)ᕤ

INFO 2025-03-19 15:51:00,185 instructlab.model.download:302: Available models (`ilab model list`):
+------------+---------------+------+---------------+
| Model Name | Last Modified | Size | Absolute path |
+------------+---------------+------+---------------+
+------------+---------------+------+---------------+
(app-root) ls -l /opt/app-root/src/.cache/instructlab/models
total 0
drwxr-sr-x. 2 1001100000 1001100000 2 Mar 19 15:51 granite-3.2-2b-instruct
(app-root) ls -l /opt/app-root/src/.cache/instructlab/models/granite-3.2-2b-instruct/
total 1
lrwxrwxrwx. 1 1001100000 1001100000 142 Mar 19 15:51 model-00001-of-00002.safetensors -> /opt/app-root/src/.cache/instructlab/oci/granite-3.2-2b-instruct/blobs/sha256/5720a4536051b0cd90ccbd9b3d8a97283f9fb07a4c4066969f88b0d9d70e5872
lrwxrwxrwx. 1 1001100000 1001100000 142 Mar 19 15:51 model-00002-of-00002.safetensors -> /opt/app-root/src/.cache/instructlab/oci/granite-3.2-2b-instruct/blobs/sha256/73833730ea9eebdff5f79d6defd78a3cc05aeaa1d18fdd81d0db2f5a80ab5756
```

AaronJaegerVA:
I discovered by reviewing the code that for .safetensors models to be recognized, several .json files must also be present in the directory. These .json files are present in the huggingface repos for the models and include the tokenizer.json file.

PR #1758, opened 2024-07-17
Release strategy: alpha, beta, rc

tiran:
Proposal for using alpha, beta, and RCs.

**Checklist:**


docs/release-strategy.md
```diff
@@ -29,10 +29,16 @@ PRs and Issues associated with the next two milestones will be prioritized for r
...
 ## Git Branches and Tags
 
-Every `X.Y` release stream gets a new branch.
+Every `X.Y` release stream gets a new branch `release-X.Y`. The release branch is created with first beta, release candidate, or final `X.Y.0` release.
 
 Each release, `X.Y.Z`, exists as a tag named `vX.Y.Z`.
 
+Alpha releases are tagged as `vX.Y.Za1`, `vX.Y.Za2`, and so on. Alpha releases are created from the main branch, **not** from a `release-X.Y` branch.
+
+Beta releases are tagged as `vX.Y.Zb1`, `vX.Y.Zb2`, and so on from `release-X.Y` branch.
```
dhellmann, line 38: I've always used a tag as the base of a branch, too. So, tag a beta, make the release branch from the same point, then future tags for that release series are on the branch and clearly after only that initial branching version and not intermingled with pre-releases of other versions on the main branch.

docs/release-strategy.md
```diff
@@ -29,10 +29,16 @@ PRs and Issues associated with the next two milestones will be prioritized for r
 
 ## Git Branches and Tags
 
-Every `X.Y` release stream gets a new branch.
+Every `X.Y` release stream gets a new branch `release-X.Y`. The release branch is created with first beta, release candidate, or final `X.Y.0` release.
```
dhellmann, line 32: Do we need alphas, betas, and RCs? We want an alpha today to avoid breaking community users while we test some downstream build machinery. The pre-release type doesn't really matter for that, so we could just always use alphas. Maybe there's a reason to use different types upstream, though?

russellb (commented):
+1 to the clarification that we do alphas from `main`

docs/release-strategy.md
```diff
@@ -29,10 +29,16 @@ PRs and Issues associated with the next two milestones will be prioritized for r
...
 
 Each release, `X.Y.Z`, exists as a tag named `vX.Y.Z`.
 
+Alpha releases are tagged as `vX.Y.Za1`, `vX.Y.Za2`, and so on. Alpha releases are created from the main branch, **not** from a `release-X.Y` branch.
```
leseb, line 36:
```suggestion
Alpha releases are tagged as `vX.Y.Zalpha1`, `vX.Y.Za2`, and so on. Alpha releases are created from the main branch, **not** from a `release-X.Y` branch.
```

Going with fullname is clearer, it's cheap and users will immediately understand.

docs/release-strategy.md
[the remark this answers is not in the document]
tiran: good catch!

tiran: I would like to start using RCs from here on out with a couple of days of testing between rc1 and final.

- alphas are cut from the main branch and can have known bugs
- betas are cut from a release branch and should have stable API and requirements
- rcs should be stable. When everything goes right, then `1.2.3rc1` tag will become the `1.2.3` tag without any modifications.

docs/release-strategy.md
tiran: > I'm sorry, Dave. I'm afraid I can't do that.
> HAL 9000

We have to follow https://peps.python.org/pep-0440/#pre-releases

leseb: 🤷🏻‍♂️

docs/release-strategy.md
dhellmann: OK, that makes sense.

docs/release-strategy.md
```diff
@@ -6,12 +6,14 @@ This document discusses the release strategy and processes for the
...
 ## Versioning Scheme
 
-Releases use a `X.Y.Z` numbering scheme.
+Releases use a `X.Y.Z`, [PEP 440-compatible](https://peps.python.org/pep-0440/) numbering scheme.
 
 X-stream release are for major releases. At this stage in the project a major release has not been cut and we expect each release to be a new Y-stream.
 
 Z-stream releases are meant for critical bug and documentation fixes. Z-stream releases are cut as maintainers see fit.
 
+A final release may be preceded by a pre-release (alpha `X.Y.ZaN`, beta `X.Y.ZbN`, or release candidate `X.Y.ZrcN`). PyPI and pip automatically recognize them as pre-release. Pip ignores pre-releases unless the user explicitly requests a pre-release with `pip install --pre` or `pip install instructlab==0.16.0a1`.
```
nathan-weinberg, line 15:
```suggestion
A release may be preceded by a pre-release, e.g.
- Alpha (`X.Y.ZaN`)
- Beta (`X.Y.ZbN`)
- Release Candidate (`X.Y.ZrcN`)

PyPI and pip automatically recognize these types of tags as pre-release. Pip ignores pre-releases unless the user explicitly requests a pre-release with `pip install --pre` or `pip install instructlab==0.16.0a1`.
```

docs/release-strategy.md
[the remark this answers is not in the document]
nathan-weinberg: Yup, release branches are of the form `release-vX.Y`

docs/release-strategy.md
```diff
@@ -47,19 +55,38 @@ The Release Manager for each release is identified in the Description of the Mil
...
 
-### Y-Stream
+### Y-Stream alpha pre-release (`X.Y.0a1`)
+
+1. Determine a commit on the main branch that will serve as the basis for the next release - most of the time this should be the latest commit.
+1. Validate the release branch with an [E2E test](ci.md).
+1. Create a new release on GitHub targeting the release branch and using the latest Y-Stream tag as the previous release (e.g. `0.15.1` precedes `0.16.0a1`). If there have been other pre-release, then use the previous pre-release.
+1. Announce release via the following:
+    - The `#announce` channel on Slack
+    - The `announce` mailing list
```
nathan-weinberg, line 65: Do we want to do as broad of an announcement for pre-releases as we do regular releases?

docs/release-strategy.md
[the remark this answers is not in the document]
bjhargrave: This is the point I made in the CLI call this week. When we are ready to make an RC build, we need to make it from a release branch and not main. This then opens up main for work. I am not as concerned about beta releases from main but I am also ok with saying beta releases need to come from a release branch.

ktam3:
@tiran - could you take this out of draft pls

PR #1767, opened 2024-07-17
utils: Drop reading of taxonomy data

russellb:
Nothing uses the parsed taxonomy data that was returned by this code.
There is a proposal to change the taxonomy schema. Instead of fixing
this copy of this code, make it stop reading the contents aside from
doing automated validation.

Rename these functions to `validate_taxonomy` instead of
`read_taxonomy` since we only care about the validation part and not
returning any data from them.

If we look at changing this to use a new API, this helps clarify what
functionality we actually care about (validation, not the data).

Related to https://github.com/instructlab/sdg/issues/160

### DIFF
--- a/src/instructlab/taxonomy/diff.py
+++ b/src/instructlab/taxonomy/diff.py
@@ -48,7 +48,7 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
     """
     # pylint: disable=import-outside-toplevel
     # Local
-    from ..utils import get_taxonomy_diff, read_taxonomy
+    from ..utils import get_taxonomy_diff, validate_taxonomy
 
     if not taxonomy_base:
         taxonomy_base = ctx.obj.config.generate.taxonomy_base
@@ -71,7 +71,7 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
             for f in updated_taxonomy_files:
                 click.echo(f)
     try:
-        read_taxonomy(None, taxonomy_path, taxonomy_base, yaml_rules)
+        validate_taxonomy(None, taxonomy_path, taxonomy_base, yaml_rules)
     except (SystemExit, yaml.YAMLError) as exc:
         if not quiet:
             click.secho(
--- a/src/instructlab/taxonomy/diff.py
+++ b/src/instructlab/taxonomy/diff.py
@@ -48,7 +48,7 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
     """
     # pylint: disable=import-outside-toplevel
     # Local
-    from ..utils import get_taxonomy_diff, validate_taxonomy
+    from ..utils import TaxonomyReadingException, get_taxonomy_diff, validate_taxonomy
 
     if not taxonomy_base:
         taxonomy_base = ctx.obj.config.generate.taxonomy_base
@@ -62,7 +62,7 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
         else:  # taxonomy_path is dir
             try:
                 updated_taxonomy_files = get_taxonomy_diff(taxonomy_path, taxonomy_base)
-            except (SystemExit, GitError) as exc:
+            except (TaxonomyReadingException, GitError) as exc:
                 click.secho(
                     f"Reading taxonomy failed with the following error: {exc}",
                     fg="red",
@@ -72,7 +72,7 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
                 click.echo(f)
     try:
         validate_taxonomy(None, taxonomy_path, taxonomy_base, yaml_rules)
-    except (SystemExit, yaml.YAMLError) as exc:
+    except (TaxonomyReadingException, yaml.YAMLError) as exc:
         if not quiet:
             click.secho(
                 f"Reading taxonomy failed with the following error: {exc}",
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -152,7 +152,7 @@ def get_taxonomy_diff(repo="taxonomy", base="origin/main"):
         try:
             head_commit = repo.commit(base)
         except gitdb.exc.BadName as e:
-            raise SystemExit(
+            raise TaxonomyReadingException(
                 yaml.YAMLError(
                     f'Couldn\'t find the taxonomy git ref "{base}" from the current HEAD'
                 )
@@ -169,7 +169,7 @@ def get_taxonomy_diff(repo="taxonomy", base="origin/main"):
         try:
             current_commit = current_commit.parents[0]
         except IndexError as e:
-            raise SystemExit(
+            raise TaxonomyReadingException(
                 yaml.YAMLError(
                     f'Couldn\'t find the taxonomy base branch "{base}" from the current HEAD'
                 )
@@ -226,7 +226,7 @@ def get_documents(
 
             if file_contents:
                 return file_contents
-            raise SystemExit("Couldn't find knowledge documents")
+            raise TaxonomyReadingException("Couldn't find knowledge documents")
         except (OSError, exc.GitCommandError, FileNotFoundError) as e:
             raise e
 
@@ -460,7 +460,9 @@ def validate_taxonomy(_logger, taxonomy, taxonomy_base, yaml_rules):
                 f"{warnings} warnings (see above) due to taxonomy file not (fully) usable."
             )
         if errors:
-            raise SystemExit(yaml.YAMLError("Taxonomy file with errors! Exiting."))
+            raise TaxonomyReadingException(
+                yaml.YAMLError("Taxonomy file with errors! Exiting.")
+            )
     else:  # taxonomy is dir
         # Gather the new or changed YAMLs using git diff
         updated_taxonomy_files = get_taxonomy_diff(taxonomy, taxonomy_base)
@@ -480,7 +482,7 @@ def validate_taxonomy(_logger, taxonomy, taxonomy_base, yaml_rules):
                 f"{total_warnings} warnings (see above) due to taxonomy files that were not (fully) usable."
             )
         if total_errors:
-            raise SystemExit(
+            raise TaxonomyReadingException(
                 yaml.YAMLError(f"{total_errors} taxonomy files with errors! Exiting.")
             )
 
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -191,18 +191,18 @@ class SourceDict(TypedDict):
     patterns: List[str]
 
 
-def get_documents(
+def _validate_documents(
     source: SourceDict,
     skip_checkout: bool = False,
-) -> List[str]:
+) -> None:
     """
-    Retrieve the content of files from a Git repository.
+    Validate that we can retrieve the content of files from a Git repository.
 
     Args:
         source (dict): Source info containing repository URL, commit hash, and list of file patterns.
 
     Returns:
-         List[str]: List of document contents.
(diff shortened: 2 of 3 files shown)


src/instructlab/utils.py
```diff
@@ -438,49 +437,24 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
...
             errors += validation_errors
-            return None, warnings, errors
+            return warnings, errors
 
-        # get seed instruction data
-        tax_path = "->".join(taxonomy_path.parent.parts)
-        task_description = contents.get("task_description")
-        documents = contents.get("document")
-        if documents:
-            documents = get_documents(source=documents)
```
booxter, line 453: where is this function used now? Should we also kill it? (This will also open a path to clean up `git_clone_checkout`.)

booxter (changes requested):

src/instructlab/utils.py
bjhargrave: I think you still need to leave the reading of the documents repo. This is an important check for `taxonomy diff` since people often get this part of the yaml wrong and yaml schema validation cannot validate the URL, commit SHA, or patterns.

src/instructlab/taxonomy/diff.py
```diff
@@ -71,7 +71,7 @@ def diff(ctx, taxonomy_path, taxonomy_base, yaml_rules, quiet):
             for f in updated_taxonomy_files:
                 click.echo(f)
     try:
-        read_taxonomy(None, taxonomy_path, taxonomy_base, yaml_rules)
+        validate_taxonomy(None, taxonomy_path, taxonomy_base, yaml_rules)
     except (SystemExit, yaml.YAMLError) as exc:
```
booxter, line 75: (No action required) Insterestingly, we don't seem to catch `TaxonomyReadingException` that the function can raise.

src/instructlab/utils.py
booxter: If so, then the function should probably be also renamed into `validate_documents` (and changed so that it doesn't `raise SystemExit`but a more specific exception.

src/instructlab/taxonomy/diff.py
booxter: It's also unclear which exception the `SystemExit` catch is supposed to handle, since the one that (was) raised from `get_documents` is already converted into `TaxonomyReadingException`.

booxter (changes requested):

src/instructlab/utils.py
russellb: Great points. I've pushed changes to make better use of an already defined custom Exception type, as well as restoring the code to make sure we can load the knowledge document content, including renaming the function to make it more clear what we care about

russellb:
I expect we'll clean this up a good bit more as we get taxonomy parsing and validation code settled in https://github.com/instructlab/schema/pull/33

The immediate thing right now is dropping the assumptions that are incompatible with taxonomy schema v3, and hopefully help clarify what we need out of a taxonomy reading API here

src/instructlab/utils.py
```diff
@@ -215,18 +215,19 @@ def get_documents(
...
+            opened_files = False
             for pattern in file_patterns:
                 for file_path in glob.glob(os.path.join(repo.working_dir, pattern)):
                     if os.path.isfile(file_path) and file_path.endswith(".md"):
-                        with open(file_path, "r", encoding="utf-8") as file:
-                            file_contents.append(file.read())
+                        with open(file_path, "r", encoding="utf-8") as _file:
+                            # Success! we could open the file.
+                            # We don't actually care about the contents, though.
+                            opened_files = True
```
booxter, line 227: You could just return immediately (the effect would be the same). That said, I am not sure this is what you actually want to do here - the way it's written, it will *not* return an error even if all but one file are not accessible. You probably want to open each of the files.

src/instructlab/utils.py
```diff
@@ -215,18 +215,19 @@ def get_documents(
...
 
-            if file_contents:
-                return file_contents
-            raise SystemExit("Couldn't find knowledge documents")
+            if not opened_files:
+                raise TaxonomyReadingException("Couldn't find knowledge documents")
         except (OSError, exc.GitCommandError, FileNotFoundError) as e:
```
booxter, line 231: (No action required) This `except` seems to do nothing at all. We'll need to clean it up later.

src/instructlab/utils.py
```diff
@@ -215,18 +215,19 @@ def get_documents(
...
-            file_contents = []
 
             logger.debug("Processing files...")
+            opened_files = False
             for pattern in file_patterns:
                 for file_path in glob.glob(os.path.join(repo.working_dir, pattern)):
                     if os.path.isfile(file_path) and file_path.endswith(".md"):
-                        with open(file_path, "r", encoding="utf-8") as file:
-                            file_contents.append(file.read())
+                        with open(file_path, "r", encoding="utf-8") as _file:
```
booxter, line 224: since you don't use the file handler anywhere, you can just omit `as _file`.

src/instructlab/utils.py
```diff
@@ -191,12 +191,12 @@ class SourceDict(TypedDict):
     patterns: List[str]
 
 
-def get_documents(
+def _validate_documents(
     source: SourceDict,
     skip_checkout: bool = False,
-) -> List[str]:
+) -> None:
```
booxter, line 197: you will need to update the docstring in line 201 accordingly

src/instructlab/utils.py
```diff
@@ -438,55 +438,41 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
...
-        )
+        warnings, errors = validate_taxonomy_file(taxonomy, yaml_rules)
         if warnings:
             logger.warning(
                 f"{warnings} warnings (see above) due to taxonomy file not (fully) usable."
             )
         if errors:
-            raise SystemExit(yaml.YAMLError("Taxonomy file with errors! Exiting."))
+            raise TaxonomyReadingException(
+                yaml.YAMLError("Taxonomy file with errors! Exiting.")
```
booxter, line 474: "Exiting" here is probably wrong now since it's up to the caller to decide what to do with the exception.

src/instructlab/utils.py
```diff
@@ -498,20 +484,17 @@ def read_taxonomy(_logger, taxonomy, taxonomy_base, yaml_rules):
...
-            if data:
-                seed_instruction_data.extend(data)
         if total_warnings:
             logger.warning(
                 f"{total_warnings} warnings (see above) due to taxonomy files that were not (fully) usable."
             )
         if total_errors:
-            raise SystemExit(
+            raise TaxonomyReadingException(
                 yaml.YAMLError(f"{total_errors} taxonomy files with errors! Exiting.")
```
booxter, line 496: ditto

src/instructlab/utils.py
```diff
@@ -389,7 +389,7 @@ def read_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
                 f"{file_path_p} is not valid. The top-level element is not an object with key-value pairs."
             )
             errors += 1
-            return None, warnings, errors
+            return warnings, errors
```
booxter, line 392: (github interface is stupid so I leave the comment here) you missed the same change in line 386. (We should probably add a typing descriptor for the return value for the function; later.)

booxter (changes requested):

src/instructlab/utils.py
russellb: ah, nice.  changed!

src/instructlab/utils.py
russellb: fixed, thanks!

src/instructlab/utils.py
russellb: Yeah, I mostly just tried to keep the existing behavior. I agree that it seems pretty awkward.

Roughly it makes sure that at least one of the glob matters matches at least one valid file.

If a glob pattern matches a markdown file but we fail to open it for some reason, then it would fail. As you note below, I can't really think of a normal situation where that could happen.

src/instructlab/utils.py
russellb: yeah ... mostly tried to keep changes minimized to removing offending code that was blocking me. You're right that there's a lot of cleanup opportunity left.

src/instructlab/utils.py
russellb: Good catch!

src/instructlab/utils.py
russellb: done

src/instructlab/utils.py
booxter: I haven't noticed that the original version did the same wacky check. OK to leave it then.

booxter (approved):

src/instructlab/utils.py
```diff
@@ -215,18 +215,19 @@ def get_documents(
...
+            opened_files = False
             for pattern in file_patterns:
                 for file_path in glob.glob(os.path.join(repo.working_dir, pattern)):
                     if os.path.isfile(file_path) and file_path.endswith(".md"):
-                        with open(file_path, "r", encoding="utf-8") as file:
-                            file_contents.append(file.read())
+                        with open(file_path, "r", encoding="utf-8"):
+                            # Success! we could open the file.
+                            # We don't actually care about the contents, though.
+                            opened_files = True
```
bjhargrave, line 227: This check is slightly different. The prior code tests if there is any content in the aggregated files. If all the files are empty, then an exception is raised. Now we just check that the files can be opened. 


```suggestion
                        if os.path.isfile(file_path) and os.path.getsize(file_path) > 0:
    opened_files = True
                            # Success! The file in non-empty.
                            # We don't actually care about the contents, though.
                            opened_files = True
```

This avoids opening the file, which may be a time saver too.

booxter: Are you sure? I think `file.read() will return '' on empty files, so the code is equivalent - it will **not** raise when all files are empty.

src/instructlab/utils.py
```diff
@@ -215,18 +215,19 @@ def get_documents(
...
-            file_contents = []
 
             logger.debug("Processing files...")
+            opened_files = False
             for pattern in file_patterns:
                 for file_path in glob.glob(os.path.join(repo.working_dir, pattern)):
                     if os.path.isfile(file_path) and file_path.endswith(".md"):
-                        with open(file_path, "r", encoding="utf-8") as file:
-                            file_contents.append(file.read())
+                        with open(file_path, "r", encoding="utf-8"):
```
bjhargrave, line 224: One thing we no longer will catch, by not reading the file, is any UTF-8 encoding problems in the file. This error has been observed in contributions.

markmc: Which means that UTF-8 encoding issues would only be discovered when you run `ilab data generate` right?

I think that's fine. It's sort of surprising that we're doing any depth of validation in `ilab taxonomy diff` at all - given that it appears to just be a command that lists what files have changed

markmc:
`instructlab-sdg-v0.2.0` has been released with the knowledge v3 format

I order to merge support for using it in #1790, I'd like to get this merged ASAP

It looks like Russell has handled all the (great!) feedback. BJ is correct that we're getting a little less validation because we're not actually reading the content of knowledge documents, but see my point above - I think there's a limit to how much validation is reasonable in `ilab taxonomy diff`

Also, bear in mind that a new library API for reading taxonomy is probably the way forward - see instructlab/schema#33

markmc (approved):

src/instructlab/utils.py
nathan-weinberg: `ilab taxonomy diff` is a combination of historically what were separate `list` and `check` commands that were combined a few months back, for context

nathan-weinberg (approved):

Issue #1773, opened 2024-07-18
--max-ctx-size does not take effect when serve model with vllm


chudegao:
When specify --max-ctx-size in 'ilab serve --backend vllm ' command,  vllm still use default max seq length 32768 which may cause OOM.

```
(dlipy3) [root@lab instructlab]# ilab serve --model-path /root/train3 --backend vllm --max-ctx-size 2048
You are using an aliased command, this will be deprecated in a future release. Please consider using `ilab model serve` instead
Parameters:
     model_path: PosixPath('/root/train3') 	[type: pathlib.PosixPath, src: commandline]
        backend: 'vllm'   	[type: str, src: commandline]
   max_ctx_size: 2048     	[type: int, src: commandline]
     gpu_layers: -1       	[type: int, src: default]
    num_threads: None     	[type: None, src: default]
   model_family: None     	[type: None, src: default]
       log_file: None     	[type: None, src: default]
  chat_template: None     	[type: None, src: default]
DEBUG 2024-07-18 02:11:03,548 instructlab.model.backends.backends:198: Auto-detecting backend for model /root/train3
DEBUG 2024-07-18 02:11:03,548 instructlab.model.backends.backends:161: Model is a directory and system is Linux, using vllm backend.
DEBUG 2024-07-18 02:11:03,548 instructlab.model.backends.backends:206: Auto-detected backend: vllm
DEBUG 2024-07-18 02:11:03,548 instructlab.model.backends.backends:217: Validating 'vllm' backend
INFO 2024-07-18 02:11:03,548 instructlab.model.serve:136: Using model '/root/train3' with -1 gpu-layers and 2048 max context size.
INFO 2024-07-18 02:11:03,548 instructlab.model.serve:140: Serving model '/root/train3' with vllm
DEBUG 2024-07-18 02:11:03,549 instructlab.model.backends.vllm:184: vLLM serving command is: ['/opt/anaconda3/envs/dlipy3/bin/python', '-m', 'vllm.entrypoints.openai.api_server', '--model', '/root/train3', '--chat-template', '/tmp/tmp197so27u', '--distributed-executor-backend', 'mp', '--host', '10.244.128.61', '--port', '8003']
INFO 2024-07-18 02:11:03,549 instructlab.model.backends.vllm:196: vLLM starting up on pid 86818 at http://10.244.128.61:8003/v1
INFO 07-18 02:11:06 api_server.py:177] vLLM API server version 0.5.0.post1
[...]

PR #1775, opened 2024-07-18
chore: warn when vllm is used with unsupported arguments

leseb:
The serve cli has 3 options that won't do anything when the vllm backend is used, namely:

* gpu_layers
* num_threads
* max_ctx_size

If the user passes:

```
ilab model serve --backend --max-ctx-size 2048
```

A warning will be printed indicating the option is not supported:

```
WARNING 2024-07-18 10:38:21,649 instructlab.model.serve:203: Option
--max-ctx-size not supported by the backend.
```

Closes: https://github.com/instructlab/instructlab/issues/1773

**Checklist:**

### DIFF
--- a/src/instructlab/model/serve.py
+++ b/src/instructlab/model/serve.py
@@ -155,6 +155,9 @@ def serve(
             num_threads=num_threads,
         )
     elif backend == backends.VLLM:
+        # Warn if unsupported backend parameters are passed
+        warn_for_unsuported_backend_param(ctx)
+
         # Instantiate the vllm server
         if ctx.args:
             # extra click arguments after "--"
@@ -191,3 +194,13 @@ def serve(
     finally:
         backend_instance.shutdown()
         raise click.exceptions.Exit(0)
+
+
+def warn_for_unsuported_backend_param(
+    ctx: click.Context,
+):
+    for param in ["gpu_layers", "num_threads", "max_ctx_size"]:
+        if ctx.get_parameter_source(param) == click.core.ParameterSource.COMMANDLINE:
+            logger.warning(
+                f"Option '--{param.replace('_','-')}' not supported by the backend."
+            )
--- a/tests/test_lab_serve.py
+++ b/tests/test_lab_serve.py
@@ -1,13 +1,16 @@
 # Standard
 from unittest import mock
-from unittest.mock import patch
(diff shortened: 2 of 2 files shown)


leseb:
Further validation can be done by checking if the value in the config.yaml is different than the default and then warn again.

nathan-weinberg (approved):

src/instructlab/model/serve.py
```diff
@@ -191,3 +194,13 @@ def serve(
     finally:
         backend_instance.shutdown()
         raise click.exceptions.Exit(0)
+
+
+def warn_for_unsuported_backend_param(
```
russellb, line 199: non-blocking nit: would be nice if the function name, or at least docstring made it clear this was for backend == vllm

src/instructlab/model/serve.py
```diff
@@ -191,3 +194,13 @@ def serve(
...
+    ctx: click.Context,
+):
+    for param in ["gpu_layers", "num_threads", "max_ctx_size"]:
+        if ctx.get_parameter_source(param) == click.core.ParameterSource.COMMANDLINE:
```
russellb, line 203: why would you only log when it's used via command line?  is it not possible to set in the config file?

I know click supports setting flags via env vars too, but I don't think we have that turned on ...

leseb: It's simpler change to first test the CLI and I [mentioned](https://github.com/instructlab/instructlab/pull/1775#issuecomment-2235962348) that checking the config could be done as well. :)

russellb: You could just check if any values are different than the defaults and check all cases at once?

leseb: > You could just check if any values are different than the defaults and check all cases at once?

But that wouldn't solve the case where someone passes a flag pointing to the same value as the default one :/

russellb: I guess so. It seems like a pretty small change to add in any case.  I'm not blocking the PR on it, though.

leseb: Actually all gpu_layers, max_ctx_size are part of the llama_cpp section, so it's clear that they are not part of vllm.

leseb:
@Mergifyio rebase

Issue #1774, opened 2024-07-18
functional test creates a temporary directory for each run


leseb:
Since #1471, the directory structure has changed and the functional test was altered to setup the new directory and create a temporary directory for the test data.
While this makes sense when running on a fresh new runner, this makes the local test quite annoying since:

* a new temp is created
* a model is downloaded in this directory

Solution to accommodate local testing:

* allow to override `TEST_DIR` from `scripts/functional-tests.sh`
* remove or make the trap (`rm -rf "${TEST_DIR}"`) optional

tiran:
Fixing #1182 would also solve the problem. HF used to cache downloaded models in `~/.cache/huggingface` and then symlink the model files.

leseb:
> Fixing #1182 would also solve the problem. HF used to cache downloaded models in `~/.cache/huggingface` and then symlink the model files.

Looks like this won't be ready anytime soon :/

PR #1776, opened 2024-07-18
test: allow func test to use a specific temporary directory

leseb:
Since #1471, the directory structure has changed and the functional test was altered to setup the new directory and create a temporary directory for the test data. While this makes sense when running on a fresh new runner, this makes the local test quite annoying since:

a new temp is created
a model is downloaded in this directory

Now, for local users, the script can be run like this:

```
SKIP_TEST_DIR_CLEANUP=1 TEST_DIR=/tmp/foo ./scripts/functional-tests.sh
```

This assumes you control your test directory and its cleanup.

Closes: https://github.com/instructlab/instructlab/issues/1774

**Checklist:**

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: Apache-2.0
+#
+# If you are running locally and calling the script multiple times you may want to run like this:
+#
+# TEST_DIR=/tmp/foo ./scripts/functional-tests.sh
+#
+# As soon as TEST_DIR is set to any non-empty value, the test directory will NOT be removed
+# This assumes you control your test directory TEST_DIR and its cleanup
 
 set -ex
 
@@ -7,6 +14,14 @@ set -ex
 export SCRIPTDIR=$(dirname "$0")
 # build a prompt string that includes the time, source file, line number, and function name
 export PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+
+if [ -n "$TEST_DIR" ]; then
+    echo "TEST_DIR is set to $TEST_DIR by the caller, will not remove it"
+    TEST_DIR_SET_BY_CALLER=1
+fi
+
+# Support overriding the test directory for local testing otherwise creates a temporary directory
+TEST_DIR=${TEST_DIR:-$(mktemp -d)}
 
 export TEST_DIR
 export PACKAGE_NAME='instructlab'  # name we use of the top-level package directories for CLI data
@@ -49,7 +64,6 @@ function init_test_script() {
     local prev_home="${HOME}"
 
     printf 'initializing test script...\n'
-    TEST_DIR=$(mktemp -d)
     HOME="${TEST_DIR}"  # to ensure platformdirs uses the test directory for the duration of this script
     printf 'changing home directory to "%s", was: "%s"\n' "${HOME}" "${prev_home}"
 
@@ -126,7 +140,7 @@ cleanup() {
 }
 
 init_test_script
-trap 'cleanup "${?}"; rm -rf "${TEST_DIR}"' EXIT QUIT INT TERM
+trap 'cleanup "${?}"; test -z "${TEST_DIR_SET_BY_CALLER}" && rm -rf "${TEST_DIR}"' EXIT QUIT INT TERM
 
 rm -f "${ILAB_CONFIG_FILE}"
 
@@ -507,7 +521,7 @@ test_model_print(){
     fi
 
     # validate that we fail on invalid model
-    if ! expect -c '
+    if ! expect -d -c '
         set timeout 30
         spawn ilab model chat -m bar
         expect {
@@ -529,7 +543,7 @@ test_model_print(){
         }
     ' "$(basename "${expected_model_name}")"
     )
-    if ! expect -c"${expect_script}"; then
+    if ! expect -d -c"${expect_script}"; then
         echo "Error: expect test failed"
         exit 1
     fi
--- a/tox.ini
+++ b/tox.ini
@@ -41,6 +41,9 @@ basepython = {[testenv:py3]basepython}
 
 [testenv:py3-functional]
 basepython = {[testenv:py3]basepython}
+passenv =
+    {[testenv]passenv}
+    TEST_DIR
 
 # format, check, and linting targets don't build and install the project to
 # speed up testing.


scripts/functional-tests.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: Apache-2.0
+#
+# If you are running locally and calling the script multiple times you may want to run like this:
+#
+# CLEANUP_TEST_DIR=1 TEST_DIR=/tmp/foo ./scripts/functional-tests.sh
+#
+# This assumes you control your test diectory and its cleanup
```
nathan-weinberg, line 8:
```suggestion
# This assumes you control your test directory and its cleanup
```

nathan-weinberg (approved):

scripts/functional-tests.sh
```diff
@@ -124,7 +134,7 @@ cleanup() {
 }
 
 init_test_script
-trap 'cleanup "${?}"; rm -rf "${TEST_DIR}"' EXIT QUIT INT TERM
+trap 'cleanup "${?}"; test -z $CLEANUP_TEST_DIR && rm -rf "${TEST_DIR}"' EXIT QUIT INT TERM
```
booxter, line 137: `-z` means "the length of string is zero". This makes the test match even with `CLEANUP_TEST_DIR=0`, which I don't think is your intent. Suggest to use integer-ish comparison: `test x$CLEANUP_TEST_DIR = x1`.

scripts/functional-tests.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: Apache-2.0
+#
+# If you are running locally and calling the script multiple times you may want to run like this:
+#
+# CLEANUP_TEST_DIR=1 TEST_DIR=/tmp/foo ./scripts/functional-tests.sh
```
booxter, line 6: The script is being triggered through `tox` environment, which I believe filters out variables from external envvars, unless they are listed in the corresponding `tox.ini` section as `passenv`. You may want to update `tox.ini` so that `tox -e py311-functional` honors the change.

booxter, line 6: When would I want to *not* set `CLEANUP_TEST_DIR=1`? The directory name is uniquely generated on each script run. Leaving the downloaded files inside `$TEST_DIR` won't speed up the next invocation anyway. What am I missing? I question the need for the variable.

scripts/functional-tests.sh
```diff
@@ -8,6 +14,11 @@ export SCRIPTDIR=$(dirname "$0")
 # build a prompt string that includes the time, source file, line number, and function name
 export PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
+# Support overriding the test directory for local testing otherwise creates a temporary directory
+TEST_DIR=${TEST_DIR:-$(mktemp -d)}
```
booxter, line 18: There's another issue with the bare call to `mktemp` here - it will use `/tmp` (unless `$TMPDIR` is set to point to elsewhere.) And this directory is not meant for large files (e.g. on my vagrant box with Fedora 40, I get only 2GB available in `/tmp` - which makes the functional test runs fail with `No space left on device` because the model is 4GB+.

Quoting from [here](https://systemd.io/TEMPORARY_DIRECTORIES/):

```
/tmp/ should be used for smaller, size-bounded files only; /var/tmp/ should be used for everything else.
```

scripts/functional-tests.sh
```diff
@@ -8,6 +14,11 @@ export SCRIPTDIR=$(dirname "$0")
...
+# Support cleaning up the test directory once the script is done, defaults to True, set any value to disable
+export CLEANUP_TEST_DIR
```
booxter, line 20: Actually, the semantics and the variable name don't seem to match. The name would suggest that setting it as `CLEANUP_TEST_DIR=1` would *enable* the cleanup. But the implementation seems reversed: it will *disable* the cleanup when the variable is set. Consider renaming the variable to e.g. `SKIP_CLEANUP`.

---

You could also make the script automatically detect if `TEST_DIR` was set by the user - and disable the cleanup then without a separate flag variable. What's the case to pass `TEST_DIR` explicitly while expecting the directory to be removed at the end of the script? (If that's really needed, the caller can just `rm -rf $TEST_DIR` after the script returns.)

leseb: > Actually, the semantics and the variable name don't seem to match. The name would suggest that setting it as `CLEANUP_TEST_DIR=1` would _enable_ the cleanup. But the implementation seems reversed: it will _disable_ the cleanup when the variable is set. Consider renaming the variable to e.g. `SKIP_CLEANUP`.

Good point I'll will rename it.

> You could also make the script automatically detect if `TEST_DIR` was set by the user - and disable the cleanup then without a separate flag variable. What's the case to pass `TEST_DIR` explicitly while expecting the directory to be removed at the end of the script? (If that's really needed, the caller can just `rm -rf $TEST_DIR` after the script returns.)

Another good point is that cleanup should always happen to prevent the disk from filling up.
If we skip the cleanup and run the script multiple times, users will likely forget to do some cleanup, also passing a dedicated directory helps not having to download the model again.
I think having both gives us better control.

scripts/functional-tests.sh
leseb: That's not how I run the func test locally, I call the script directly, am I doing something wrong? :)

scripts/functional-tests.sh
leseb: Added in a separate commit, thanks!

leseb:
Ok now this is failing because of `/var/tmp` being too long:

```
tmp.gCUz5f/.local/share/instructlab/models/foo.gguf\r\n\u256d\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500 system \u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u256e\r\n\u2502 Welcome to InstructLab Chat w/                                               \u2502\r\n\u2502 \u001b[1m/VAR/TMP/INSTRUCTLAB-TEST-TMP.GCUZ5F/.LOCAL/SHARE/INSTRUCTLAB/MODELS/FOO.GGU\u001b[0m \u2502\r\n\u2502 \u001b[1mF\u001b[0m
```

scripts/functional-tests.sh
```diff
@@ -8,6 +15,11 @@ export SCRIPTDIR=$(dirname "$0")
 # build a prompt string that includes the time, source file, line number, and function name
 export PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
+# Support overriding the test directory for local testing otherwise creates a temporary directory
+TEST_DIR=${TEST_DIR:-$(mktemp -d --tmpdir=/var/tmp instructlab-test-tmp.XXXXXX)}
```
booxter, line 19: You should not assume `/var/tmp`, or `/tmp`, exists. (POSIX [leaves](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/xrat/V4_xbd_chap10.html) `/tmp` in the standard for historical reasons, but it's not advised to use it anyway.)

`mktemp` will already use `TMPDIR` to decide where to create a temporary file. We should allow the user to specify a different `TMPDIR` to use. They can already do it by `TMPDIR=/var/tmp ./scripts/functional-tests.sh` (but they cannot do it when they run the tests with `tox -e py311-functional` because `tox` strips off all envvars except those listed in `passenv` - hence my suggestion to allow the variable to be passed into the functional `tox` env.)

scripts/functional-tests.sh
```diff
@@ -8,6 +15,11 @@ export SCRIPTDIR=$(dirname "$0")
...
+# Support cleaning up the test directory once the script is done, defaults to True, set any value to disable
+export SKIP_TEST_DIR_CLEANUP
```
booxter, line 21: minor nit: strictly speaking, empty value is also "any value". This should probably say that any *non-empty* value should be set; also, `export`s aren't meaningful in this context, so if you'd like to keep it for documentation purposes, just do `SKIP_TEST_DIR_CLEANUP=`.

scripts/functional-tests.sh
booxter: And `TEST_DIR` and `SKIP_TEST_DIR_CLEANUP` should be added to `passenv` for the same reason.

booxter (changes requested):
I believe the functional test failure is because the chat UI truncates model paths that are too long, so "FOO.GGUF" is truncated to "FOO.GGU". This won't be an issue once https://github.com/instructlab/instructlab/pull/1802 is merged (thanks for this patch!) But in the meantime here, I don't think we actually need to change how `mktemp` picks the name for the automatically created directory (`instructlab-test-tmp.XXXXXX` tempate), nor to enforce a particular `--tmpdir` to use (users should be able to affect it by setting `TMPDIR`), so I assume we can revert the `mktemp` CLI change, and it should pass the suite again.

scripts/functional-tests.sh
```diff
@@ -124,7 +135,7 @@ cleanup() {
 }
 
 init_test_script
-trap 'cleanup "${?}"; rm -rf "${TEST_DIR}"' EXIT QUIT INT TERM
+trap 'cleanup "${?}"; test -z $SKIP_TEST_DIR_CLEANUP && rm -rf "${TEST_DIR}"' EXIT QUIT INT TERM
```
booxter, line 138: Still not sure we have to have a separate variable for `SKIP_TEST_DIR_CLEANUP`, or we could just detect that `TEST_DIR` was passed by the caller and avoid cleaning up the directory then.

leseb: Ok you sold me on this.

booxter (approved):
Thanks! I think `TMPDIR` should also be `passenv`'d but that's not the goal of this PR. Someone else can do it if needed.

tox.ini
```diff
@@ -39,6 +39,8 @@ basepython = {[testenv:py3]basepython}
 
 [testenv:py3-functional]
 basepython = {[testenv:py3]basepython}
+passenv =
+    TEST_DIR
```
tiran, line 43: I think this is the right syntax to pass the `passenv` settings from the parent section:
```suggestion
passenv =
    {[testenv]passenv}
    TEST_DIR
```

leseb: It's working for me without this.

tiran: Can you pass a modified CMAKE_ARGS without it?

leseb: Why would I do that? This is just for running the func test. Am I missing something?

scripts/functional-tests.sh
```diff
@@ -124,7 +138,7 @@ cleanup() {
 }
 
 init_test_script
-trap 'cleanup "${?}"; rm -rf "${TEST_DIR}"' EXIT QUIT INT TERM
+trap 'cleanup "${?}"; test -z $TEST_DIR_SET_BY_CALLER && rm -rf "${TEST_DIR}"' EXIT QUIT INT TERM
```
RobotSail, line 141: We should be using `"${TEST_DIR_SET_BY_CALLER}"` to avoid potential bugs when bash expands this variable.

```suggestion
trap 'cleanup "${?}"; test -z "${TEST_DIR_SET_BY_CALLER}" && rm -rf "${TEST_DIR}"' EXIT QUIT INT TERM
```

scripts/functional-tests.sh
```diff
@@ -8,6 +15,14 @@ export SCRIPTDIR=$(dirname "$0")
 # build a prompt string that includes the time, source file, line number, and function name
 export PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
+if [ -n "$TEST_DIR" ]; then
```
RobotSail, line 18: Please use double brackets when possible, this will help us reduce potential errors since pathname expansion or word splitting takes place between `[[` and  `]]`

```suggestion
if [[ -n "${TEST_DIR}" ]]; then
```

scripts/functional-tests.sh
booxter: I also thought so at first, but: Not really.

```
$ test -z && echo true
true
```

leseb: That's right, I forgot to quote it.

tox.ini
leseb: Added.

booxter: I think otherwise there's no way to run functional tests via `tox` with CUDA etc.

scripts/functional-tests.sh
leseb: Well I guess it does not hurt :)

booxter (approved):

scripts/functional-tests.sh
RobotSail: It's just something I feel we should adhere for consistency in order to prevent bugs. But you're welcome to leave this as is, since it's currently passing tests anyway.

RobotSail (approved):
LGTM

PR #1777, opened 2024-07-18
feat: add serve config for teacher model in `ilab data generate`

alimaredia:
Currently `ilab data generate` uses the configuration from the `serve` section of configuration file to serve the teacher model in the background. 

In the case where a user is using a specific model only as the teacher model, the user would have to edit the `serve` section of the config with the teacher model's settings.

Since the teacher model and it's configuration is static, and the serve section is dynamic, the teacher model and it's configuration should live in the `generate` section of the configuration. 

Giving the teacher model it's own configuration section also prevents the user from accidentally serving the teacher model, if they have not changed the `serve` section.

This PR is meant to be a starting point. Further work should evolve the configuration of the teacher model in `ilab data generate` to be similar to how the `ilab model evaluate` command configures the judge model for MT-Bench.

This PR also includes a commit that adds the `--enable-serving-output` flag to `ilab data generate`. This flag serves the same purpose as it does in `ilab model evaluate`.

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,11 @@
 * The `instructlab` package now uses optional dependencies for each supported hardware `cpu`,
   `cuda`, `hpu`, `mps`, and `rocm`. To install InstructLab for e.g. NVIDIA CUDA, use
   `pip install instructlab[cuda]`.
+* Add a `--enable-serving-output` flag for `ilab data generate`. This flag determines whether vLLM
+  will have its output suppressed when it serves the teacher model in the background.
+* The `generate` section of the config now has a `teacher` section. This section configures the
+  teacher model when it is automatically served in the background. This new section has the same
+  values as the `serve` section of the config.
 
 ### Breaking Changes
 
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -308,7 +308,7 @@ class _generate(BaseModel):
     chunk_word_count: PositiveInt = DEFAULTS.CHUNK_WORD_COUNT
     # DEPRECATED: see sdg_scale_factor instead
     # Left in place so that we can still detect and give a warning if its
-    # specified in an old configuraiton file.
+    # specified in an old configuration file.
     num_instructions: Optional[int] = Field(
         default=-1, deprecated="see 'sdg_scale_factor' instead", exclude=True
     )
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -265,13 +265,20 @@ class _serve(BaseModel):
     model_config = ConfigDict(extra="ignore", protected_namespaces=())
 
     # vllm configuration
-    vllm: _serve_vllm
+    vllm: _serve_vllm = _serve_vllm(
+        llm_family="",
+        vllm_args=[],
+    )
 
     # llama-cpp configuration
-    llama_cpp: _serve_llama_cpp
+    llama_cpp: _serve_llama_cpp = _serve_llama_cpp(
+        gpu_layers=-1,
+        max_ctx_size=4096,
+        llm_family="",
+    )
 
     # required fields
-    model_path: StrictStr
+    model_path: StrictStr = Field(default_factory=lambda: DEFAULTS.DEFAULT_MODEL)
     # additional fields with defaults
     host_port: StrictStr = "127.0.0.1:8000"
     chat_template: Optional[str] = None
@@ -296,7 +303,7 @@ class _generate(BaseModel):
     taxonomy_base: StrictStr
 
     # additional fields with defaults
-    teacher: _serve
+    teacher: _serve = Field(default_factory=_serve)
     num_cpus: PositiveInt = DEFAULTS.NUM_CPUS
     chunk_word_count: PositiveInt = DEFAULTS.CHUNK_WORD_COUNT
     # DEPRECATED: see sdg_scale_factor instead
@@ -406,18 +413,6 @@ def get_default_config() -> Config:
             model=DEFAULTS.DEFAULT_MODEL,
             taxonomy_path=DEFAULTS.TAXONOMY_DIR,
             taxonomy_base=DEFAULTS.TAXONOMY_BASE,
-            teacher=_serve(
-                model_path=DEFAULTS.DEFAULT_MODEL,
-                llama_cpp=_serve_llama_cpp(
-                    gpu_layers=-1,
-                    max_ctx_size=4096,
-                    llm_family="",
-                ),
-                vllm=_serve_vllm(
-                    llm_family="",
-                    vllm_args=[],
-                ),
-            ),
         ),
         serve=_serve(
             model_path=DEFAULTS.DEFAULT_MODEL,
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -242,32 +242,6 @@ class _chat(BaseModel):
     max_tokens: typing.Optional[int] = None
 
 
-class _generate(BaseModel):
-    """Class describing configuration of the generate sub-command."""
-
-    # model configuration
-    model_config = ConfigDict(extra="ignore")
-
-    # required fields
-    model: StrictStr
-    taxonomy_path: StrictStr
-    taxonomy_base: StrictStr
-
-    # additional fields with defaults
-    num_cpus: PositiveInt = DEFAULTS.NUM_CPUS
-    chunk_word_count: PositiveInt = DEFAULTS.CHUNK_WORD_COUNT
-    # DEPRECATED: see sdg_scale_factor instead
-    # Left in place so that we can still detect and give a warning if its
-    # specified in an old configuraiton file.
-    num_instructions: Optional[int] = Field(
-        default=-1, deprecated="see 'sdg_scale_factor' instead", exclude=True
-    )
-    sdg_scale_factor: Optional[PositiveInt] = DEFAULTS.SDG_SCALE_FACTOR
-    output_dir: StrictStr = Field(default_factory=lambda: DEFAULTS.DATASETS_DIR)
-    prompt_file: StrictStr = Field(default_factory=lambda: DEFAULTS.PROMPT_FILE)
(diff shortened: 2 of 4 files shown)


alimaredia:
Adding the `hold` label, which will be removed after I'm done testing this PR by hand. Reviews are still welcome.

leseb (commented):
The change looks ok to me but I'm missing some context. Can you explain how is configured the teacher model? Thanks

russellb:
I guess I just don’t feel this is necessary. Here’s how I think about it: something has to go in the serve config. What is the majority use case that is fair to expect should be in there? My thought: the teacher model config. 

An alternative answer to that could be “the target model to be trained, for use with ilab chat to test model behavior.”

Is the latter what you were thinking?

I’ll also say that I don’t like “general inference use cases” as an answer here because I don’t think using ilab makes much sense for that vs running vllm directly (via systemd). I’m not sure what value ilab provides in that case.

alimaredia:
@leseb The teacher model can be a separate model that specializes just generating the data from the taxonomy. One model you could use for example is Mixtral. This is similar to how we have a judge model just for MT-Bench in evaluation. The model is prometheus.

The teacher model could have it's own set of configuration, ex: `vllm_args`.

alimaredia:
> I guess I just don’t feel this is necessary. Here’s how I think about it: something has to go in the serve config. What is the majority use case that is fair to expect should be in there? My thought: the teacher model config.
> 
> An alternative answer to that could be “the target model to be trained, for use with ilab chat to test model behavior.”
> 
> Is the latter what you were thinking?
> 
> I’ll also say that I don’t like “general inference use cases” as an answer here because I don’t think using ilab makes much sense for that vs running vllm directly (via systemd). I’m not sure what value ilab provides in that case.

The alternative answer along with inference use cases is what I was thinking. Here's how I see the entire ilab workflow along with when models are served:

1. Download all the necessary models. These are the Base, Teacher, and Judge models.
2. Chat with the Base model on how much it knows the information you want it to know about. **Right now chat reads from the serve config if the model is started up in the background.**
3.  Make changes to the taxonomy, adding new skills and knowledge.
4. Run data generation. **Right now the model started up in the background also reads from the serve config.**
5. Train the model. Here MT-Bench is used in multiphase training. I think @JamesKunstle is calling the exact same functions in `model/evaluate.py` for the base and judge model and not reading anything from the serve config. 
6. Chat with the trained model to see what the model learned. As mentioned in step 2 **right now chat reads from the serve config if the model is started up in the background** but since it's derived from the base model, I think it's serving configuration should exactly be the same except for the model path.
7. Run evaluation. For MT-Bench the necessary base and judge model configuration happens within `evaluate.py`, not in the serve config.

As I mentioned above I think since the teacher model can be viewed as something static similar to the judge model, that's a good reason for it to have it's own config in the `generate` section.

@leseb @n1hility  do you have anything to add about the inference use case?

russellb:
OK - I buy that the base model is a better thing to assume stays in the base serve config.  Thanks.

src/instructlab/configuration.py
```diff
@@ -303,6 +283,27 @@ def api_base(self):
         return get_api_base(self.host_port)
 
 
+class _generate(BaseModel):
```
russellb, line 286: any reason you needed to move this? It makes it harder to see what actually changed.

If you feel it needs to move, I would generally recommend doing that type of refactoring in a dedicated commit that includes no functional changes. That way you can still see the real code changes in the commit series.

src/instructlab/configuration.py
```diff
@@ -303,6 +283,27 @@ def api_base(self):
...
+    # model configuration
+    model_config = ConfigDict(extra="ignore")
+
+    # required fields
+    model: StrictStr
+    taxonomy_path: StrictStr
+    taxonomy_base: StrictStr
+
+    # additional fields with defaults
+    teacher_serve_config: _serve
```
russellb, line 298: Can you please look into backwards compatibility? I haven't tested this, but I'm guessing that old configs will be broken by this?

My preference would be to use the default serve config if it is not specified in this section.

russellb (changes requested):

src/instructlab/configuration.py
alimaredia: I moved it below the `_serve` classes because those need to be declared first in order for `_generate` to have a member of that type. Without this move I get a `NameError` whenever `ilab` is invoked.

I wish I didn't have to move the class, but the only thing that has changed in the `_generate` class is just the addition of the `teacher_serve_config` variable.

src/instructlab/configuration.py
alimaredia: The addition of this in the _serve config would not make the `data generate` command able to work with older configs. Especially since the code change I made in `generate.py` now expects the teacher_serve_config.

Backwards compatibility within the older config wasn't something I considered when I made this PR simply because I thought there have been so many changes to the config that we're not expecting the older configs to be fowards compatible. 

It's for this reason I think https://github.com/instructlab/instructlab/pull/1749 should include a check that causes ilab to error out if a config doesn't have the `version` field.

What I could do is add a check in `generate.py` to see if the config has the `teacher_serve_config` section in the `generate` section and if it doesn't fall back to using the `serve` section of the config. How does that sound?

src/instructlab/configuration.py
booxter: I **think** you can reference via string, as described [here](https://docs.pydantic.dev/2.8/concepts/postponed_annotations/#self-referencing-or-recursive-models)

```
class Foo(BaseModel):
    bar: 'Bar'

class Bar(BaseModel):
    pass
```

src/instructlab/configuration.py
booxter, line 298: I think you could specify a default value for the field, and make all the nested models provide reasonable defaults; then you wouldn't need to pass the teacher argument below in `get_default_config`. To see what I mean, refer to: https://gist.github.com/booxter/e1332bbe3a5344ee217409d133a13776

I believe this will tackle the backwards compatibility concern by Russell too.

booxter (changes requested):

booxter (approved):

booxter:
Note: the last patch should be squashed before merging.

booxter (approved):

booxter:
@alimaredia looking at the checklist in the description, I think you may want to update changelog file to document the new section.

CHANGELOG.md
```diff
@@ -34,6 +34,11 @@
 * Default log format changes to include the logger name in the logs.
 * Introduces CUDA-specific requirements in a new file `requirements-cuda.txt`. To install these
   CUDA-enabled dependencies, simply run `pip install instructlab[cuda]`.
+* Add a `--enable-serving-output` flag for `ilab data generate`. This flag determines whether vLLM
+  will have its output supressed when it serves the teacher model in the background.
```
booxter, line 38: suppressed; the spellcheck job fails because of it

booxter:
@alimaredia do we still need the `hold` label here?

booxter (approved):

alimaredia:
@booxter Ran the branch by hand, and serving worked as expected. I've removed the hold label.

src/instructlab/configuration.py
russellb: That config versioning PR merged (#1749) -- are you planning to update this PR to bump the version in some way?

src/instructlab/configuration.py
```diff
@@ -310,6 +291,33 @@ def api_base(self):
...
+    taxonomy_path: StrictStr
+    taxonomy_base: StrictStr
+
+    # additional fields with defaults
+    teacher: _serve = Field(default_factory=_serve)
+    num_cpus: PositiveInt = DEFAULTS.NUM_CPUS
+    chunk_word_count: PositiveInt = DEFAULTS.CHUNK_WORD_COUNT
+    # DEPRECATED: see sdg_scale_factor instead
+    # Left in place so that we can still detect and give a warning if its
+    # specified in an old configuraiton file.
```
booxter, line 311: configuration

booxter (approved):
Thank you for quick respin.

booxter (approved):

alimaredia:
@russellb could you approve this PR so it merges?

src/instructlab/configuration.py
```diff
@@ -310,6 +291,33 @@ def api_base(self):
...
+    num_cpus: PositiveInt = DEFAULTS.NUM_CPUS
+    chunk_word_count: PositiveInt = DEFAULTS.CHUNK_WORD_COUNT
+    # DEPRECATED: see sdg_scale_factor instead
+    # Left in place so that we can still detect and give a warning if its
+    # specified in an old configuration file.
+    num_instructions: Optional[int] = Field(
+        default=-1, deprecated="see 'sdg_scale_factor' instead", exclude=True
+    )
+    sdg_scale_factor: Optional[PositiveInt] = DEFAULTS.SDG_SCALE_FACTOR
+    output_dir: StrictStr = Field(default_factory=lambda: DEFAULTS.DATASETS_DIR)
```
RobotSail, line 316: You can remove these `Field`s as they were previously used to prevent the config values from being cached before we ran tests. Nowadays, each test will reset the home directories to point to the [temporary path fixture](https://github.com/instructlab/instructlab/blob/c6adcb5ddd84d9358a2db56c5f30b237420f9b13/tests/conftest.py#L47).

```suggestion
    output_dir: StrictStr = DEFAULTS.DATASETS_DIR
```

RobotSail (commented):
I have one comment but once that's addressed we can merge this.

src/instructlab/configuration.py
alimaredia: @RobotSail The entire `_generate` section was moved in this PR, and one addition was made (the addition of the `teacher` member. 

The line you mentioned is the same as what's in `main` here (https://github.com/instructlab/instructlab/blob/main/src/instructlab/configuration.py#L266). To keep the scope of this PR focused I'd prefer to open up a follow up PR with that change, and possibly changes to any other variables in this section.

RobotSail: Makes sense, let's approve this then and then fix it later.

RobotSail (approved):
LGTM

khaledsulayman (approved):

PR #1779, opened 2024-07-18
tests: clarify the presence of some variables 

leseb:
At first glance, these variables might seem unnecessary given that the
same Python call is done later, and `mkdir -p` will create everything.
The idea is only to create the base directory and let `ilab` create the
subsequent folders.

**Checklist:**

Should go **AFTER** https://github.com/instructlab/instructlab/pull/1762

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -67,6 +67,9 @@ function init_test_script() {
     HOME="${TEST_DIR}"  # to ensure platformdirs uses the test directory for the duration of this script
     printf 'changing home directory to "%s", was: "%s"\n' "${HOME}" "${prev_home}"
 
+    # DO NOT REMOVE THE VARIABLE ASSIGNMENTS BELOW
+    # They might seem redundant given that `mkdir -p`` is called later, but they are necessary to
+    # ensure the directories are created by ilab CLI
     # get the directories from the platformdirs library
     CONFIG_DIR=$(python -c "import platformdirs; print(platformdirs.user_config_dir())")
     DATA_DIR=$(python -c "import platformdirs; print(platformdirs.user_data_dir())")


RobotSail (commented):
We actually shouldn't be doing this. The reason why I had created it this way originally was to ensure that at least the platform dies themselves exist, and that ilab can create the necessary subdirectories for itself correctly.

leseb:
> We actually shouldn't be doing this. The reason why I had created it this way originally was to ensure that at least the platform dirs themselves exist, and that ilab can create the necessary subdirectories for itself correctly.

Oh, so that's a way to execute `ensure_storage_directories_exist`? I can comment in the script on that so people don't trip over this again then.

RobotSail:
@leseb That will be really helpful, thank you. In the future I'll make sure that these non-obvious things are better documented with comments.

scripts/functional-tests.sh
```diff
@@ -67,6 +67,9 @@ function init_test_script() {
     HOME="${TEST_DIR}"  # to ensure platformdirs uses the test directory for the duration of this script
     printf 'changing home directory to "%s", was: "%s"\n' "${HOME}" "${prev_home}"
 
+    # DO NOT REMOVE THE VARIABLE ASSIGNMENTS BELOW
```
booxter, line 70: Sorry but: Which variables are we talking about here?.. There are several sections below. Does the comment apply to all of them?

RobotSail (approved):
Let's get this merged!!!

scripts/functional-tests.sh
RobotSail: Yes

booxter: OK. But isn't it *obvious* that they are needed? (They are used elsewhere in the script.) And how is `ilab` creating directories relevant here? I don't think `ilab` itself cares about any of the variables listed below.

I don't understand what this comment is trying to address / explain. Can you elaborate?

RobotSail: @booxter further down in the test script we are creating the directories for all of these variables. For those familiar with the init process, this may seem redundant as `ilab` has a function called `ensure_storage_directories_exist` where it recursively creates the storage directories that will be used in subsequent calls. But for the purpose of knowing how ilab works, we expect that these directories will already exist. So `ilab` is only responsible for those that it must create.

As I'm writing this out though, I'm realizing that we should probably also handle the case where these directories don't exist and `ilab` creates them manually, as this may be more realistic for some of our users.

nathan-weinberg (approved):

PR #1783, opened 2024-07-18
add -T option to e2e

cdoern:
-T enables the e2e to be run with JUST the full training library. When used the --legacy tests will be skipped.

**Checklist:**

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -19,6 +19,7 @@ DIFF_ARGS=("--taxonomy-path" "./taxonomy")
 TRAIN_ARGS=()
 GRANITE=0
 FULLTRAIN=0
+TRAIN_LIBRARY=0
 BACKEND="llama-cpp"
 HF_TOKEN=${HF_TOKEN:-}
 SDG_PIPELINE="simple"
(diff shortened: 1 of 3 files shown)


nathan-weinberg:
@cdoern can you please document the new flag in `docs/ci.md`? We have a running list of the flags there now

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -154,7 +154,7 @@ jobs:
 
           python3.11 -m pip show nvidia-nccl-cu12
           
-          ./scripts/basic-workflow-tests.sh -emf
+          ./scripts/basic-workflow-tests.sh -emfT
```
nathan-weinberg, line 157: Looking at the [matrix](https://github.com/instructlab/instructlab/blob/main/docs/ci.md/) it looks like full training was already working on this instance type and we need this to be added to the `x4` variant of this job?

@russellb can you confirm?

scripts/basic-workflow-tests.sh
```diff
@@ -277,10 +266,21 @@ test_train() {
             TRAIN_ARGS+=("--gguf-model-path" "${DATA_HOME}/instructlab/models/granite-7b-lab-Q4_K_M.gguf")
         fi
 
+        ilab model train "${TRAIN_ARGS[@]}"
+    else
+        # TODO Only cuda for now
+        TRAIN_ARGS+=("--legacy" "--device=cuda")
+        if [ "$FULLTRAIN" -eq 0 ]; then
+            TRAIN_ARGS+=("--4-bit-quant")
+        fi
```
nathan-weinberg, line 275: Looking at the `-f` flag I'm unclear as to how this and the new `-T` flag differ - is `-f` just using legacy training with safetensors?

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: Right.

My preference would be to leave this workflow alone and to add `-T` to `a10g-x4.yml`.

The change here is giving `-f` and `-T` which still has it running both training types. It would be better if this workflow only used full linux legacy training after this PR.

russellb: well actually `-T` overrides ``-f` it seems, but it still doesn't make sense to specify both.

and again, i'd leave `-f` here and add `-T` to the bigger one.

scripts/basic-workflow-tests.sh
russellb: * `-T` -- training lib
* `-f` -- `--legacy`
* neither option -- `--legacy --4-bit-quant`

.github/workflows/e2e-nvidia-a10g-x1.yml
cdoern: removed from a10x1, added to a10x4

russellb:
@cdoern This lgtm!

You just have one last little issue to fix.  The default e2e failed on:

```
./scripts/basic-workflow-tests.sh: line 260: TRAIN_LIBRARY: unbound variable
```

russellb (approved):
assuming the default e2e job failure is fixed now (looks like it is), then lgtm

Let's just run the `a10g-x4` job right after this merges to see how it looks

alinaryan (approved):

PR #1784, opened 2024-07-18
Remove the check for a valid pipeline alias

### LINKED ISSUE #1703: PR 1668: replace incomprehensible lambda with function
PR #1668 added a multi-line lambda callback that is hard to understand. I have been starring at the code for like 5 minutes and I'm still not sure why this works in the first place. And that's despite over 20 years of Python coding experience. 

```python
    callback=lambda ctx, param, value: value
    # check for directory instead of file
    if value in ["simple", "full"] or os.path.isdir(value)
    else ctx.fail(
        f'{value} is not a valid pipeline alias ("simple" or "full") or path to a directory including pipeline configuration files.'
    ),
```

Re-implement the callback with a helper function or a custom Click param type to make it more readable and testable.

**Additional context**

derekhiggins:
sdg now supports more then just the aliases defined by the sdg library. Allow all pipeline names so that the sgd library can assess if they are present.

Fixes #1703
Related instructlab/sdg#166

---
sdg PR that allows additional alias names https://github.com/instructlab/sdg/pull/166

### DIFF
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -2,7 +2,6 @@
 
 # Standard
 import logging
-import os
 
 # Third Party
 import click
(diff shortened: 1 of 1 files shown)


russellb (approved):
Validation will happen inside the sdg library still. This looks good to me.

It will also close https://github.com/instructlab/instructlab/issues/1703 (edit: oops you already said that)

nathan-weinberg (approved):

Issue #1707, opened 2024-07-13
`ensure_storage_directories_exist` assumes that user has not changed directories in `config.yaml`


tiran:
**Describe the bug**
`ensure_storage_directories_exist` creates internals directories for InstructLab. The function uses the values from the default, not the values from `config.yaml`. The implementation assumes that a user does not change any path in the config file. That is obviously a wrong assumption.

https://github.com/instructlab/instructlab/blob/90259f98157d2de1144fc9aa0b2c76bec7e4b37a/src/instructlab/configuration.py#L523-L544 

**Expected behavior**
The function `ensure_storage_directories_exist` should not exist in the first place. Instead the creation of directories should be delegated to code down the line. `ilab model chat` is a good example. It creates its chat logdirs in place:

https://github.com/instructlab/instructlab/blob/90259f98157d2de1144fc9aa0b2c76bec7e4b37a/src/instructlab/model/chat.py#L728-L733

**Additional context**
Problem was introduced in PR #1471

ktam3:
@RobotSail - PTAL, update on this?

nathan-weinberg:
Is this going to be worked on?

PR #1787, opened 2024-07-18
only create persistent storage files/dirs when needed

### LINKED ISSUE #1704: PR 1653: ilab creates and writes to file on every invocation
**Describe the bug**
PR #1653 added code, which causes `ilab` to always create or write to a file on every invocation of the CLI. Even `--help` on subcommands creates and writes to the file.

**To Reproduce**
```
$ strace -e openat ilab config --help 2>&1 | grep O_WRONLY
openat(AT_FDCWD, "/home/heimes/.local/share/instructlab/internal/train_configuration/additional/additional_args.yaml", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 3
```

**Expected behavior**
`ilab` should not create a file or write to files on startup.

**Additional context**
Suggestion:
- if the content of the JSON file is static, then the data should be included as a JSON file in the project as a project resource
- if the content is dynamic, then creation and write should be limited to execution of a subcommand. The content should be written to a temporary file or to the user cache directory. The user data directory is the wrong directory for the purpose.

**update**

Here is are better reproducers. 
```console
$ HOME=/dev/null ilab config --help
...
NotADirectoryError: [Errno 20] Not a directory: '/dev/null/.cache'
```

```console
$ rm -rf /tmp/instructlab/
$ HOME=/tmp/instructlab ilab config --help
$ if [ -d /tmp/instructlab/ ]; then echo "ERROR: --help created a directory"; fi
ERROR: --help created a directory
```

cdoern:
the additional_args file as well as many persistent storage dirs are re-created on each ilab invocation. Only create them if they do not exist.

resolves #1704

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -539,7 +539,8 @@ def ensure_storage_directories_exist():
     ]
 
     for dirpath in dirs_to_make:
-        os.makedirs(dirpath, exist_ok=True)
+        if not os.path.exists(dirpath):
+            os.makedirs(dirpath, exist_ok=True)
 
     additional_args_and_defaults = {
         "learning_rate": 2e-5,
@@ -557,8 +558,11 @@ def ensure_storage_directories_exist():
     }
 
     # create exper_args file for users to see/edit
-    with open(DEFAULTS.TRAIN_ADDITIONAL_OPTIONS_FILE, "w", encoding="utf-8") as outfile:
-        yaml.dump(additional_args_and_defaults, outfile)
+    if not os.path.isfile(DEFAULTS.TRAIN_ADDITIONAL_OPTIONS_FILE):
+        with open(
+            DEFAULTS.TRAIN_ADDITIONAL_OPTIONS_FILE, "w", encoding="utf-8"
+        ) as outfile:
+            yaml.dump(additional_args_and_defaults, outfile)
 
 
 class Lab:


tiran (dismissed):
Neither of the chances solve ticket #1704 . The first chance is just a less efficient code that results in the same outcome with more syscalls.

tiran:
Issue #1707 explains another aspect of the problematic design. The design flaw was introduced recently with #1471. It would be best to get rid of `ensure_storage_directories_exist` and create the directors in each CLI entry function. This change will also ensure that we take command line options into account.

cdoern:
ok @tiran I will go back to the drawing board and see if I can think of a better solution here! thanks for the feedback

RobotSail (approved):
LGTM

RobotSail:
@tiran I understand the concern around having too many syscalls to write to directories. However, the solution proposed in the aforementioned issue seems like a heavy lift, as we would need to wrap all of our writes & reads around a lazy write system. This PR seems like a decent compromise to at least reduce the number of write invocations.

As for the `ensure_storage_directories_exist` function - the way that `os.makedirs` is implemented has it so that the only redundant write call that happens is to `mkdir` itself - the rest of the path is recursively checked for its existence before being created. Not that this is a good design, but we know that the scope is at least limited to those dirs we've defined in the function. With this proposed change, we are reducing these `mkdir` invocations as well to only the times when they haven't been created.

alinaryan (approved):
Considering how close we are to GA, having read syscalls is better than having write syscalls. Can be further addressed in a follow up PR.

PR #1788, opened 2024-07-18
feat: add option for users to download default model during init

nathan-weinberg:
Currently, we ask users to specify a model path during init but we don't have an option to do an initial model download for first-time users like we do for taxonomy

**Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,20 +293,32 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 
 2. When prompted by the interface, press **Enter** to add a new default `config.yaml` file.
 
-3. When prompted, clone the `https://github.com/instructlab/taxonomy.git` repository into the current directory by typing **y**.
+3. When prompted, clone the `https://github.com/instructlab/taxonomy.git` repository into the storage directory by typing **y**.
 
    **Optional**: If you want to point to an existing local clone of the `taxonomy` repository, you can pass the path interactively or alternatively with the `--taxonomy-path` flag.
+
+4. When prompted, download the `merlinite-7b-lab-Q4_K_M.gguf` model into the storage directory by typing **y**.
+
+   > **NOTE** ⏳ This command can take few minutes or complete immediately depending on your internet connection or if the model is cached. If you have issues connecting to Hugging Face, refer to the [Hugging Face discussion forum](https://discuss.huggingface.co/) for more details.
 
    *Example output after initializing `ilab`*
 
    ```shell
    (venv) $ ilab config init
-   Welcome to InstructLab CLI. This guide will help you set up your environment.
+   Welcome to InstructLab CLI. This guide will help you to setup your environment.
    Please provide the following values to initiate the environment [press Enter for defaults]:
-   Path to taxonomy repo [taxonomy]: <ENTER>
-   `taxonomy` seems to not exists or is empty. Should I clone https://github.com/instructlab/taxonomy.git for you? [y/N]: y
+   Path to taxonomy repo [~/.local/share/instructlab/taxonomy]: 
+   `~/.local/share/instructlab/taxonomy` seems to not exist or is empty. Should I clone https://github.com/instructlab/taxonomy.git for you? [y/N]: y
    Cloning https://github.com/instructlab/taxonomy.git...
-   Generating `config.yaml` in the current directory...
+   Path to your model [~/.local/share/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf]: 
+   `~/.local/share/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf` seems to not exist. Should I download merlinite-7b-lab-Q4_K_M.gguf to ~/.local/share/instructlab/models for you? [y/N]: y
+   Downloading model from instructlab/merlinite-7b-lab-GGUF@main to ~/.local/share/instructlab/models...
+   Downloading 'merlinite-7b-lab-Q4_K_M.gguf' to '~/.local/share/instructlab/models/.huggingface/download/merlinite-7b-lab-Q4_K_M.gguf.9ca044d727db34750e1aeb04e3b18c3cf4a8c064a9ac96cf00448c506631d16c.incomplete'
+   INFO 2024-07-18 15:16:14,456 huggingface_hub.file_download:1877: Downloading 'merlinite-7b-lab-Q4_K_M.gguf' to '~/.local/share/instructlab/models/.huggingface/download/merlinite-7b-lab-Q4_K_M.gguf.9ca044d727db34750e1aeb04e3b18c3cf4a8c064a9ac96cf00448c506631d16c.incomplete'
+   merlinite-7b-lab-Q4_K_M.gguf: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 4.37G/4.37G [02:44<00:00, 26.6MB/s]
+   Download complete. Moving file to ~/.local/share/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
+   INFO 2024-07-18 15:18:59,059 huggingface_hub.file_download:1893: Download complete. Moving file to ~/.local/share/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
+   Generating `~/.config/instructlab/config.yaml`...
    Initialization completed successfully, you're ready to start using `ilab`. Enjoy!
    ```
 
@@ -314,7 +326,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 
 ### 📥 Download the model
 
-- Run the `ilab model download` command.
+- Run the `ilab model download` command if you did not already do it as part of the previous step.
 
   ```shell
   ilab model download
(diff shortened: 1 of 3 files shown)


nathan-weinberg:
@RobotSail any idea why I'm getting this unit test failure here? the model seems like it should be getting downloaded to the right location according to the logs

**local**
```bash
(venv) [nathan@nathan-redhat instructlab (auto-model)]$ ls /home/nathan/.local/share/instructlab/models
merlinite-7b-lab-Q4_K_M.gguf
```
**pytest**
```bash
  >       assert DEFAULTS.GGUF_MODEL_NAME in os.listdir(DEFAULTS._models_dir)
  E       AssertionError: assert 'merlinite-7b-lab-Q4_K_M.gguf' in []
  E        +  where 'merlinite-7b-lab-Q4_K_M.gguf' = <instructlab.configuration._InstructlabDefaults object at 0x7f0601eea950>.GGUF_MODEL_NAME
  E        +  and   [] = <built-in function listdir>('/tmp/pytest-of-runner/pytest-0/test_init_interactive_with_mod0/.local/share/instructlab/models')
  E        +    where <built-in function listdir> = os.listdir
  E        +    and   '/tmp/pytest-of-runner/pytest-0/test_init_interactive_with_mod0/.local/share/instructlab/models' = <instructlab.configuration._InstructlabDefaults object at 0x7f0601eea950>._models_dir
```

src/instructlab/config/init.py
```diff
@@ -125,6 +129,33 @@ def init(
...
             click.prompt("Path to your model", default=model_path)
         )
+
+    # check if model is present on disk or not
+    model_file = exists(model_path)
+    if model_file:
+        download_model_file = False
+    elif interactive:
+        download_model_file = click.confirm(
+            f"`{model_path}` seems to not exist. Should I download {DEFAULTS.GGUF_MODEL_NAME} to {DEFAULTS.MODELS_DIR} for you?"
```
leseb, line 139: Why hardcode the default and not use `model_path` and `repository`?

leseb (changes requested):

src/instructlab/config/init.py
nathan-weinberg: My thought was that a user could pass anything to those values - if something invalid is passed, we would just be issuing a download command that will fail anyway.

Since this is something that will only hit first-time users, it seemed appropriate to hold their hand a bit by just offering the defaults as a starting point.

That being said, I am not married to this approach, so I'd happily hear out other ideas!

src/instructlab/config/init.py
```diff
@@ -125,6 +129,33 @@ def init(
...
+
+    # download model file via 'ilab model download' if it needs to be downloaded
+    # pylint: disable=broad-exception-caught
+    if download_model_file:
+        try:
+            ctx.invoke(
+                download,
+                repository=DEFAULTS.MERLINITE_GGUF_REPO,
+                release="main",
+                filename=DEFAULTS.GGUF_MODEL_NAME,
```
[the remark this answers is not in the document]
nathan-weinberg: Same reasoning as above - for this block here I just wanted it to match the text earlier in the code so it was consistent.

src/instructlab/configuration.py
```diff
@@ -92,6 +92,7 @@ def _reset(self):
         self._cache_home = pd.user_cache_dir
         self._config_dir = pd.user_config_dir
         self._data_dir = pd.user_data_dir
+        self._models_dir = path.join(pd.user_data_dir, STORAGE_DIR_NAMES.MODELS)
```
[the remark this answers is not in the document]
nathan-weinberg: The comment above suggested to me that this was a safer approach for the respective unit test - you can see there I used it for checking the default model name was indeed added.

It may be unnecessary - @RobotSail perhaps you could weigh in here? If it's not beneficial I can definitely take a different approach.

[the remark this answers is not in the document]
RobotSail: @nathan-weinberg The `self._<name>_dir` attributes are meant for us to store the location of the data, config, and cache directories. We do it this way so that we can reset their values during testing with the `._reset` method. 

The attribute you're looking for already exists as `DEFAULTS.MODELS_DIR` in the global `DEFAULTS` object, so you shouldn't need to make this change.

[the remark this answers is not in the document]
nathan-weinberg: Thanks!

src/instructlab/config/init.py
leseb: It's odd that if a user requests a specific model and that model isn't available, ilab suggests downloading a different one. How are users supposed to obtain the model they requested in this case?

nathan-weinberg: We could do a scheme where we attempt to download the requested model to the default model path (i.e. use `repository` and not `model_path`) though this will fail the `init` if the model they specified isn't one that exists.

leseb: I think failing if the passed model is invalid makes sense. Thanks

nathan-weinberg: @leseb took another look at this - `repository` is just for the taxonomy repo, the only CLI arg for models is `model_path` currently - the only way I see us being able to use that to attempt a call to `ilab model download` is to try and truncate the path with something like `model_path.rsplit('/')` - I can go down this route but I feel that just using the default may be cleaner, wdyt?

leseb: Not sure I follow, why truncate?

nathan-weinberg: Not sure how else to massage the `model_path` input into something more useful than just recommending the default

README.md
```diff
@@ -296,28 +296,40 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
...
+   Downloading model from instructlab/merlinite-7b-lab-GGUF@main to ~/.local/share/instructlab/models...
+   Downloading 'merlinite-7b-lab-Q4_K_M.gguf' to '~/.local/share/instructlab/models/.huggingface/download/merlinite-7b-lab-Q4_K_M.gguf.9ca044d727db34750e1aeb04e3b18c3cf4a8c064a9ac96cf00448c506631d16c.incomplete'
+   INFO 2024-07-18 15:16:14,456 huggingface_hub.file_download:1877: Downloading 'merlinite-7b-lab-Q4_K_M.gguf' to '~/.local/share/instructlab/models/.huggingface/download/merlinite-7b-lab-Q4_K_M.gguf.9ca044d727db34750e1aeb04e3b18c3cf4a8c064a9ac96cf00448c506631d16c.incomplete'
+   merlinite-7b-lab-Q4_K_M.gguf: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 4.37G/4.37G [02:44<00:00, 26.6MB/s]
+   Download complete. Moving file to ~/.local/share/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
+   INFO 2024-07-18 15:18:59,059 huggingface_hub.file_download:1893: Download complete. Moving file to ~/.local/share/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
```
leseb, line 323: Frankly, I don't think this is adding much value and it will be different depending on the system, can we remove it? Thanks

src/instructlab/config/init.py
leseb: After revisiting this, I think we should try to download whatever the user has requested, even though it might be incorrect. For this you can use:

```python
if ctx.get_parameter_source("model_path") == click.core.ParameterSource.COMMANDLINE
```

If the `model_path` was not passed by the CLI we can keep what you have suggested here. Otherwise we continue to propose to download what the user passed.

How does that sound?

README.md
nathan-weinberg: This output is coming from `ilab model download` - I'm hesitant to change it in this one since it is modifying `ilab config init` - would it be fine to do this in a separate PR?

src/instructlab/config/init.py
nathan-weinberg: Sounds good to me!

Agu1l4r821 (approved):

src/instructlab/config/init.py
nathan-weinberg: @leseb getting back to this now - this still is going to require the `model_path` massaging - I'm going to push up a rebased commit to show what I mean - no rush on this since we are targeting for `0.19.0`

nathan-weinberg: Okay I'm not even sure it is possible with the massaging - to get everything we need from `model_path` we'd have to extrapolate 3 different `ilab model download` args from the single `model_path` arg
```python
            ctx.invoke(
                download,
                repository=DEFAULTS.MERLINITE_GGUF_REPO,
                release="main",
                filename=DEFAULTS.GGUF_MODEL_NAME,
                model_dir=DEFAULTS.MODELS_DIR,
            )
```
Everywhere I have a default specified

nathan-weinberg: I can still do it for `model_dir` but I don't see how to do the others

RobotSail (approved):
LGTM, this would be really nice to have. 

My only question is: 
Is there a way we can have it download all models, but maybe prompt the user if it will be a large one ahead of time?

For instance if we want to download mixtral 8x7B I would like to be prompted for it, but would also like to know that it's about to eat up 100GB of storage.

nathan-weinberg:
Will be superceded by changes in https://github.com/instructlab/instructlab/pull/1895

PR #1791, opened 2024-07-18
adds early-out for model type checking based on path type

JamesKunstle:
`is_model_safetensors` and `is_model_gguf` were raising exceptions rather than returning False for common file-based failure cases.

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -122,6 +122,19 @@ def is_model_safetensors(model_path: pathlib.Path) -> bool:
     Returns:
         bool: True if the model is a safetensors model, False otherwise.
     """
+
+    # guards against users passing str-type paths if not
+    # statically analyzing.
+    if isinstance(model_path, str):
+        logger.debug("'model_path' was passed as 'str'. Casting to pathlib.Path")
+        model_path = pathlib.Path(model_path)
+
+    if not model_path.is_dir():
+        logger.debug(
+            f"The path to the model {model_path.as_posix()} is not a directory, and therefore cannot be a model in .safetensors format.",
+        )
+        return False
+
     try:
         files = list(model_path.iterdir())
     except (FileNotFoundError, NotADirectoryError, PermissionError) as e:
@@ -167,6 +180,18 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
     # Third Party
     from gguf.constants import GGUF_MAGIC
 
+    # guards against users passing str-type paths if not
+    # statically analyzing.
+    if isinstance(model_path, str):
+        logger.debug("'model_path' was passed as 'str'. Casting to pathlib.Path")
+        model_path = pathlib.Path(model_path)
+
+    if not model_path.is_file():
+        logger.debug(
+            f"The path to the model {model_path.as_posix()} is not a file, and therefore cannot be a model in .gguf format.",
+        )
+        return False
+
     with open(model_path, "rb") as f:
         # Memory-map the file on the first 4 bytes (this is where the magic number is)
         mmapped_file = mmap.mmap(f.fileno(), length=4, access=mmap.ACCESS_READ)
@@ -194,7 +219,7 @@ def determine_backend(model_path: pathlib.Path) -> Tuple[str, str]:
                         - The reason why the backend was selected.
     """
     try:
-        if model_path.is_dir() and is_model_safetensors(model_path):
+        if is_model_safetensors(model_path):
             if sys.platform == "linux":
                 logger.debug(
                     f"Model is huggingface safetensors and system is Linux, using {VLLM} backend."


src/instructlab/model/backends/backends.py
```diff
@@ -136,7 +143,6 @@ def is_model_safetensors(model_path: pathlib.Path) -> bool:
     requires_files = {
         "config.json",
         "tokenizer.json",
-        "tokenizer.model",
```
JamesKunstle, line 139: removing this causes test to break because this is the only file that the test checks for.

src/instructlab/model/backends/backends.py
```diff
@@ -167,6 +174,13 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
     # Third Party
     from gguf.constants import GGUF_MAGIC
 
+    if not model_path.is_file():
+        logging.debug(
+            "The path to the model %s is not a file, and therefore cannot be a model in .gguf format.",
+            str(model_path),
```
leseb, line 180:
```suggestion
           model_path.as_posix(),
```

Feel free to do that instead, not a blocker.

src/instructlab/model/backends/backends.py
```diff
@@ -167,6 +174,13 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
     # Third Party
     from gguf.constants import GGUF_MAGIC
 
+    if not model_path.is_file():
+        logging.debug(
```
leseb, line 178:
```suggestion
        logger.debug(
```

src/instructlab/model/backends/backends.py
```diff
@@ -167,6 +174,13 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
...
+            "The path to the model %s is not a file, and therefore cannot be a model in .gguf format.",
```
leseb, line 179: Also the repo tends to do: `logger.debug(f"Message: {var}")` for formatting.

leseb (changes requested):

src/instructlab/model/backends/backends.py
```diff
@@ -122,6 +122,13 @@ def is_model_safetensors(model_path: pathlib.Path) -> bool:
     Returns:
         bool: True if the model is a safetensors model, False otherwise.
     """
+    if not model_path.is_dir():
+        logging.debug(
+            "The path to the model %s is not a directory, and therefore cannot be a model in .safetensors format.",
```
alimaredia, line 127: Ditto to what Seb said in https://github.com/instructlab/instructlab/pull/1791/files#r1683986353

jaideepr97 (commented):
+1 to seb and ali's comments

src/instructlab/model/backends/backends.py
JamesKunstle: fixed!

src/instructlab/model/backends/backends.py
JamesKunstle: fixed!

src/instructlab/model/backends/backends.py
JamesKunstle: ++

src/instructlab/model/backends/backends.py
JamesKunstle: Okay no problem, just changed over. In the future we may want to standardize on lazy logging instead- I looked into it a bit and it seems to be a bit faster.

nathan-weinberg:
Funny, these changes are very similar to something I was dealing with in https://github.com/instructlab/instructlab/pull/1795

src/instructlab/model/backends/backends.py
```diff
@@ -167,6 +173,12 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
     # Third Party
     from gguf.constants import GGUF_MAGIC
 
+    if not model_path.is_file():
```
nathan-weinberg, line 176: Similar story here to my comment above: https://www.freecodecamp.org/news/how-to-check-if-a-file-exists-in-python/

JamesKunstle:
@nathan-weinberg Very good points, I'll amend the code to be more inclusive

src/instructlab/model/backends/backends.py
```diff
@@ -122,6 +122,12 @@ def is_model_safetensors(model_path: pathlib.Path) -> bool:
     Returns:
         bool: True if the model is a safetensors model, False otherwise.
     """
+    if not model_path.is_dir():
```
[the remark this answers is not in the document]
JamesKunstle: @nathan-weinberg updated

src/instructlab/model/backends/backends.py
JamesKunstle: @nathan-weinberg updated

JamesKunstle:
@nathan-weinberg would you mind please dismissing requested changes- I made the requested change.

nathan-weinberg:
> @nathan-weinberg would you mind please dismissing requested changes- I made the requested change.

Those aren't mine, they're @leseb

src/instructlab/model/backends/backends.py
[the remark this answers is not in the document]
nathan-weinberg: Looks good, though this will throw an exception if that casting fails

src/instructlab/model/backends/backends.py
nathan-weinberg: Same comment as above

RobotSail (approved):
LGTM

JamesKunstle:
@nathan-weinberg He OK'ed dismissing if I updated the changes, I don't want to dismiss them myself as bad practice

src/instructlab/model/backends/backends.py
```diff
@@ -122,6 +122,19 @@ def is_model_safetensors(model_path: pathlib.Path) -> bool:
     Returns:
         bool: True if the model is a safetensors model, False otherwise.
     """
+
+    # guards against users passing str-type paths if not
+    # statically analyzing.
+    if isinstance(model_path, str):
```
booxter, line 128: I disagree with the premise of this change - assuming callers may pass something but a `Path`. Instead of adding ad-hoc conversion code like proposed here, the callers should be forced to pass the expected types, if not already. The latter can also be enforced by adding more typing checks up the call stack.

Otherwise, why not handling `model_path` being a `None`, or some other type? This is a slippery slope.

src/instructlab/model/backends/backends.py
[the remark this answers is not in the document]
booxter: Wouldn't `NotADirectoryError` handler catch the situation? Why do we need to handle this explicitly here, again?

booxter (changes requested):

src/instructlab/model/backends/backends.py
JamesKunstle: Definitely see your point. We can drop this if it's truly a bad route. It's a modest degree of leniency in this codebase because there's some heterogeneity on path handling- some modules might want to call this method that use str's representing paths. Since those representations are very close, I'm inclined to massage the input into a format that this function wants. Otherwise, if it's not something that's reasonably handle-able, it's rejected because it's `NotADirectory`.

booxter: Elsewhere, we were ripping off some of these "special handling for string paths", so I'd like to avoid this push-pull. The strategic direction I think is to convert all code that attempts to pass strings where Paths are expected to actually pass Paths. (Same for other types.) These are code bugs, so they should be fixed in-place, not worked around.

nathan-weinberg: I propose the following snippet
```python
if not isinstance(model_path, pathlib.Path):
    raise TypeError("'model_path' must be of type 'pathlib.Path')
```
This remains in the spirit of my comment, James's implementation, and Ihar's feedback.

src/instructlab/model/backends/backends.py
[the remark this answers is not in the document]
nathan-weinberg: With my suggestion above, we ensure a `Path` type is being passed, so agreed with @booxter here

src/instructlab/model/backends/backends.py
nathan-weinberg: Same code snippet suggestion here
```python
if not isinstance(model_path, pathlib.Path):
    raise TypeError("'model_path' must be of type 'pathlib.Path')
```

nathan-weinberg: Then below, can do something similar to the safetensors func
```python
try:
        # Memory-map the file on the first 4 bytes (this is where the magic number is)
        mmapped_file = mmap.mmap(f.fileno(), length=4, access=mmap.ACCESS_READ)
        # Read the first 4 bytes
        first_four_bytes = mmapped_file.read(4)
        # Convert the first four bytes to an integer
        first_four_bytes_int = int(struct.unpack("<I", first_four_bytes)[0])
        # Close the memory-mapped file
        mmapped_file.close()
        return first_four_bytes_int == GGUF_MAGIC
    except (FileNotFoundError, PermissionError) as e:
        logger.debug("Failed to read file: %s", e)
        return False
```
@booxter let me know your thoughts here as well ^^

src/instructlab/model/backends/backends.py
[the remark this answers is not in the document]
JamesKunstle: I think you're right @booxter. Explicitly checking if a path is a directory above is redundant since we catch an informative exception. I'll make a change reflective of that.

src/instructlab/model/backends/backends.py
JamesKunstle: I thought a bit about this offline- I think @booxter's route of trusting the function definition (that the input is a pathlib.Path) is the right way to go, otherwise we'd be writing input type-validation code for every function. We're using type annotations so anyone who passes a `str` to this method should be warned by their development environment that this isn't a good way to do things.

JamesKunstle:
@booxter's feedback is solid, doing what I'm suggesting here wouldn't really help. I'll close this PR and we can reopen if there are objections.

src/instructlab/model/backends/backends.py
```diff
@@ -167,6 +180,18 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
     # Third Party
     from gguf.constants import GGUF_MAGIC
 
+    # guards against users passing str-type paths if not
+    # statically analyzing.
+    if isinstance(model_path, str):
+        logger.debug("'model_path' was passed as 'str'. Casting to pathlib.Path")
+        model_path = pathlib.Path(model_path)
+
+    if not model_path.is_file():
```
booxter, line 189: @JamesKunstle I think this change may actually be useful, since right now if a directory is passed, the function will raise an exception (on `open`). The rest of changes here, as you confirmed, are probably not a good idea. (Thanks.)

PR #1793, opened 2024-07-18
removes tokenizer.model from required safetensors model files"

JamesKunstle:
`tokenizer.model` isn't required for a model directory to be a `.safetensors` model- this broke model loading for me because the checkpoints that we save don't have this file. This appears to be a common, but not a required file. (look at metaai/llama3-8b)

### DIFF
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -22,7 +22,7 @@ def create_safetensors_model_files(safetensors_model_path: pathlib.Path, valid: 
     test_json_dict = {"a": 1, "b": 2}
     json_object = json.dumps(test_json_dict, indent=4)
 
-    for file in ["config.json", "tokenizer.json", "tokenizer_config.json"]:
+    for file in ["tokenizer.json", "tokenizer_config.json"]:
         os.makedirs(os.path.dirname(safetensors_model_path / file), exist_ok=True)
         with open(safetensors_model_path / file, "a+", encoding="UTF-8") as f:
             f.write(json_object)
@@ -32,10 +32,8 @@ def create_safetensors_model_files(safetensors_model_path: pathlib.Path, valid: 
     ) as f:
         f.write("")
     if valid:
-        with open(
-            safetensors_model_path / "tokenizer.model", "a+", encoding="UTF-8"
-        ) as f:
-            f.write("")
+        with open(safetensors_model_path / "config.json", "a+", encoding="UTF-8") as f:
+            f.write(json_object)
 
 
 def test_free_port():
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -136,7 +136,6 @@ def is_model_safetensors(model_path: pathlib.Path) -> bool:
     requires_files = {
         "config.json",
         "tokenizer.json",
-        "tokenizer.model",
         "tokenizer_config.json",
     }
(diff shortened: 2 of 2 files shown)


tests/test_backends.py
```diff
@@ -33,7 +33,7 @@ def create_safetensors_model_files(safetensors_model_path: pathlib.Path, valid:
         f.write("")
     if valid:
         with open(
-            safetensors_model_path / "tokenizer.model", "a+", encoding="UTF-8"
```
JamesKunstle, line 36: needed to remove this from the test too haha

JamesKunstle:
Requires test fix that's updated in #1791

leseb (commented):
On paper, the change makes sense to me, please:

* look into the test failures
* update your commit message to match the PR description

Thanks!

jaideepr97:
@JamesKunstle LGTM as well, sorry for breaking this for you!
will approve after all CI checks pass

jaideepr97 (commented):
@JamesKunstle LGTM as well, sorry for breaking this for you!
will approve after all CI checks pass

tests/test_backends.py
```diff
@@ -22,7 +22,7 @@ def create_safetensors_model_files(safetensors_model_path: pathlib.Path, valid:
     test_json_dict = {"a": 1, "b": 2}
     json_object = json.dumps(test_json_dict, indent=4)
 
-    for file in ["config.json", "tokenizer.json", "tokenizer_config.json"]:
+    for file in ["tokenizer.json", "tokenizer_config.json"]:
```
alimaredia, line 25: Why did we remove `config.json` here? It still see it in `requires_files` in backends.py and it's still getting written in the test below.

jaideepr97: @alimaredia he just moved it a few lines down
what's happening here is that the test needs to verify both "valid" and "invalid" safetensors models. The way we're mocking an invalid model is for it to be missing the "config.json" file. so in this PR the config.json file only gets created for the "valid model" test case and gets skipped if valid is false

JamesKunstle: because that's the "is or isn't valid" test that this mock implements. If a "valid" model is written it'll include `config.json`, if an invalid model is written, it'll omit that file.

JamesKunstle:
> On paper, the change makes sense to me, please:
> 
> * look into the test failures
> * update your commit message to match the PR description
> 
> Thanks!

@leseb I updated the commit message- the current test failure is because I'm fixing the code and the test at once- how do we handle this failure state?

jaideepr97:
>the current test failure is because I'm fixing the code and the test at once- how do we handle this failure state?

@JamesKunstle not sure I follow, it's pretty standard to make code and test changes in the same PR, no? The CI should just run the test as is in your branch I think

JamesKunstle:
> > the current test failure is because I'm fixing the code and the test at once- how do we handle this failure state?
> 
> @JamesKunstle not sure I follow, it's pretty standard to make code and test changes in the same PR, no? The CI should just run the test as is in your branch

You're right, was typing too quickly. I think this actually breaks the test- will investigate.

JamesKunstle:
answer: test setup should've written json_object to config.json rather than empty string.

nathan-weinberg (approved):

jaideepr97 (approved):
LGTM

RobotSail (approved):
LGTM

PR #1795, opened 2024-07-18
fix: add checks to ensure evaluate model args are valid

nathan-weinberg:
While testing Eval, I noticed we are passing a GGUF file by default to the `evaluate` command but Evaluation won't fully support GGUF until https://github.com/instructlab/eval/issues/50 is complete

I filed https://github.com/instructlab/instructlab/issues/1792 to track this on the CLI side, but adding this minor check as well in the meantime

**Checklist:**

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -126,6 +126,15 @@ def get_evaluator(
                 fg="red",
             )
             raise click.exceptions.Exit(1)
+        # ensure user is passing full safetensors if they specify a local directory
+        # TODO: also allow GGUF once the following is resolved: https://github.com/instructlab/eval/issues/50
+        if os.path.isdir(model):
+            if not backends.is_model_safetensors(pathlib.Path(model)):
+                click.secho(
+                    "MMLU and MMLUBranch can currently only be used with safetensors",
+                    fg="red",
+                )
+                raise click.exceptions.Exit(1)
         if benchmark == Benchmark.MMLU:
             # Third Party
             from instructlab.eval.mmlu import MMLUEvaluator
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -87,6 +87,19 @@ def get_evaluator(
                 fg="red",
             )
             raise click.exceptions.Exit(1)
+        if os.path.exists(model):
+            model = pathlib.Path(model)
+            valid_model = False
+            if model.is_dir():
+                valid_model = backends.is_model_safetensors(model)
+            elif model.is_file():
+                valid_model = backends.is_model_gguf(model)
+            if not valid_model:
+                click.secho(
+                    "MTBench and MTBenchBranch need to be passed either a safetensors directory or a GGUF file",
+                    fg="red",
+                )
+                raise click.exceptions.Exit(1)
         if benchmark == Benchmark.MT_BENCH:
             # Third Party
             from instructlab.eval.mt_bench import MTBenchEvaluator
@@ -131,7 +144,7 @@ def get_evaluator(
         if os.path.isdir(model):
             if not backends.is_model_safetensors(pathlib.Path(model)):
                 click.secho(
-                    "MMLU and MMLUBranch can currently only be used with safetensors",
+                    "MMLU and MMLUBranch can currently only be used with a safetensors directory",
                     fg="red",
                 )
                 raise click.exceptions.Exit(1)
--- a/tests/test_lab_evaluate.py
+++ b/tests/test_lab_evaluate.py
@@ -496,7 +496,7 @@ def test_invalid_tasks_dir(cli_runner: CliRunner):
     assert result.exit_code != 0
 
 
-def test_invalid_mmlu_model(cli_runner: CliRunner, tmp_path):
+def test_invalid_model_path_mmlu(cli_runner: CliRunner, tmp_path):
     test_dir = tmp_path / "test"
     test_dir.mkdir()
     result = cli_runner.invoke(
@@ -512,7 +512,29 @@ def test_invalid_mmlu_model(cli_runner: CliRunner, tmp_path):
         ],
     )
     assert (
-        "MMLU and MMLUBranch can currently only be used with safetensors"
+        "MMLU and MMLUBranch can currently only be used with a safetensors directory"
+        in result.output
+    )
+    assert result.exit_code != 0
+
+
+def test_invalid_model_path_mt_bench(cli_runner: CliRunner, tmp_path):
+    test_dir = tmp_path / "test"
+    test_dir.mkdir()
+    result = cli_runner.invoke(
+        lab.ilab,
+        [
+            "--config=DEFAULT",
+            "model",
+            "evaluate",
+            "--benchmark",
+            "mt_bench",
+            "--model",
+            test_dir,
+        ],
+    )
+    assert (
+        "MTBench and MTBenchBranch need to be passed either a safetensors directory or a GGUF file"
         in result.output
     )
     assert result.exit_code != 0
--- a/tests/test_lab_evaluate.py
+++ b/tests/test_lab_evaluate.py
@@ -494,3 +494,25 @@ def test_invalid_tasks_dir(cli_runner: CliRunner):
     )
     assert "Tasks dir not found:" in result.output
     assert result.exit_code != 0
+
+
+def test_invalid_mmlu_model(cli_runner: CliRunner, tmp_path):
+    test_dir = tmp_path / "test"
+    test_dir.mkdir()
+    result = cli_runner.invoke(
+        lab.ilab,
+        [
+            "--config=DEFAULT",
+            "model",
+            "evaluate",
+            "--benchmark",
+            "mmlu",
+            "--model",
+            test_dir,
+        ],
+    )
+    assert (
+        "MMLU and MMLUBranch can currently only be used with safetensors"
+        in result.output
+    )
+    assert result.exit_code != 0


nathan-weinberg:
@leseb I noticed you introduced a function `is_model_gguf` for the backend work - do you think that's worth using here or this check is sufficient? only noticed the func after I'd already written this

leseb (changes requested):
See: https://github.com/instructlab/instructlab/pull/1795#issuecomment-2238580845

Also, can you add a test case for this? Thanks!

alimaredia:
The root of the issue is that MMLU and MMLU-Branch only work for .safetensors models right now. What we actually want to check is that the model is `.safetensors` and reject anything else. 

Seb and JD worked on adding a check for this here: https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/backends/backends.py#L115

src/instructlab/model/evaluate.py
```diff
@@ -465,6 +465,16 @@ def evaluate(
     tls_client_passwd,  # pylint: disable=unused-argument
     enable_serving_output,
 ):
+    # ensure user is passing full safetensors and not a GGUF if using MMLU or MMLUBranch
+    # https://github.com/instructlab/eval/issues/50
+    if model is not None:
+        if model.lower().endswith(".gguf") and benchmark in ["mmlu", "mmlu_branch"]:
```
booxter, line 471: (Looking at the other comment about only accepting .safetensors here, this may be not important anyway but:) please use symbols instead of bare strings:

```
class Benchmark(str, enum.Enum):
    MMLU = "mmlu"
    MMLU_BRANCH = "mmlu_branch"
```

nathan-weinberg: @booxter thanks for the tip - I've just pushed up a different implementation - would value your feedback on it!

alinaryan (approved):

booxter (approved):

src/instructlab/model/evaluate.py
```diff
@@ -60,6 +60,16 @@ def get_evaluator(
     otherwise raises an exception for the missing values
     """
 
+    # ensure user is passing full safetensors if they specify a local directory
+    # TODO: also allow GGUF once the following is resolved: https://github.com/instructlab/eval/issues/50
+    if model is not None and os.path.isdir(model):
```
russellb, line 65: So if someone has a path to a gguf file as opposed to its directory, this will be false. Is that what you want?

nathan-weinberg: Hmm, yeah I noted this in the bug https://github.com/instructlab/instructlab/issues/1792#issuecomment-2239811057 - I'm not sure if we _never_ are going to be able to use a specific file or if we're always going to require either a directory or HF repo. @alimaredia @alinaryan do either of you know?

nathan-weinberg: If files are verbotten, I think it'll be worth adding the following to this PR:
1. Additional check to ensure the we are not passing a local file
2. Additional unit test to ensure passing a file leads to an expected excepton

booxter: The `isdir` check is probably redundant btw, since I see the following inside `is_model_safetensors`:

```
    try:
        files = list(model_path.iterdir())
    except (FileNotFoundError, NotADirectoryError, PermissionError) as e:
        logger.debug("Failed to read directory: %s", e)
        return False
```

Note the `NotADirectoryError` is already handled there.

nathan-weinberg: The reason I added that there is because `model` in this case gets passed as a string, rather than a `pathlib.Path` type, so when you try to call `iterdir()` it throws an unhandled exception. 

This gets into the territory of what @JamesKunstle is doing in https://github.com/instructlab/instructlab/pull/1791 :smile:

src/instructlab/model/evaluate.py
```diff
@@ -60,6 +60,16 @@ def get_evaluator(
...
+        if not backends.is_model_safetensors(pathlib.Path(model)):
+            click.secho(
+                "Evaluation can currently only be used with safetensors",
```
danmcp, line 68: This is only true for MMLU right?

danmcp (changes requested):

src/instructlab/model/evaluate.py
booxter: That it's passed into is_model_safetensors as a string is a bug of the caller, and should be fixed in the caller. (Commented in the linked PR too.)

nathan-weinberg: That's what I'm doing here no? Thus the `isdir` check (though as @russellb points out we may want to abstract this to a directory or file path)

src/instructlab/model/evaluate.py
nathan-weinberg: Yes I think so - the check may make more sense within the MMLU benchmark param check section of the code - some discussion around this in the related bug as well: https://github.com/instructlab/instructlab/issues/1792

danmcp: +1 Having it inside the param checking block applying to mmlu and mmlu_branch above is probably best.  Until we can move serving out of lm_eval.

alimaredia: +1 to this, it seems like the only code changes should be made around line ~115 in main where the code paths that run MMLU and MMLU-Branch are.

alimaredia: and the log statement should reflect that MMLU and MMLU-Branch can currently only be run with safetensors.

src/instructlab/model/evaluate.py
nathan-weinberg: @danmcp can you confirm if there's any case where a singular file would be a valid input, or do we always need to be doing a local dir/HF repo?

src/instructlab/model/evaluate.py
nathan-weinberg: I will refactor to do this!

src/instructlab/model/evaluate.py
nathan-weinberg: if the former, I can change to `os.path.exists()`

booxter: Side note re: `if model is not None` part of the check: I don't think long term we should accept a None model here. See: https://github.com/instructlab/instructlab/pull/1810 At least as of right now, the code below in launch_server will raise TypeError if None ever sneaks past get_evaluator. I think it's better to just enforce it via click.option unless there's a use case for a missing --model?

nathan-weinberg: Thanks for opening that @booxter! AFAIK we don't have a case for a missing `--model` - could someone else from the Eval Squad confirm?

nathan-weinberg: The `if model is not None` was removed in my latest patchset, so shouldn't be any conflict there

booxter (approved):

jaideepr97:
@nathan-weinberg @alimaredia @alinaryan 
is the GGUF check being done because eval spins up a vLLM server? If so heads up that I have a PR to check for `.bin` files in addition to `.safetensors` files - presence of either would qualify the model as valid since `.bin` files can also be served via vLLM
this shouldn't break eval, and should allow us to also evaluate `.bin` models
https://github.com/instructlab/instructlab/pull/1809

leseb (approved):

src/instructlab/model/evaluate.py
russellb: From this discussion, it's not clear if `model` can ever be a path to a specific file. If so, this still seems insufficient.

nathan-weinberg: I've done some manual testing around this and made the following findings wrt the model arguments in additional to the discussion above:
1. MMLU/MMLUBranch require safetensors, either local or on HF, so this check should be sufficient for that
2. MTBench/MTBenchBranch must have a local path, either to a directory with safetensors or a GGUF file
I am going to update this PR to account for both, changing the original scope. Will update the PR title and commit messages to account for this.

nathan-weinberg:
> @nathan-weinberg @alimaredia @alinaryan @danmcp is the GGUF check being done because eval spins up a vLLM server? If so heads up that I have a PR to check for `.bin` files in addition to `.safetensors` files - presence of either would qualify the model as valid since `.bin` files can also be served via vLLM this shouldn't break eval, and should allow us to also evaluate `.bin` models #1809

For MTBench and MTBenchBranch, yes we are spinning up a vLLM server but there is also llama-cpp support so GGUF files can be used, though for most users using vLLM and safetensors makes the most sense here

nathan-weinberg:
@leseb @alinaryan @booxter I've done a bit of a refactor here - know you all already approved but wanted to give you a ping in case you had any comments on the new code

src/instructlab/model/evaluate.py
```diff
@@ -87,6 +87,19 @@ def get_evaluator(
                 fg="red",
             )
             raise click.exceptions.Exit(1)
+        if os.path.exists(model):
+            model = pathlib.Path(model)
+            valid = True
```
leseb, line 92: Why do you assume `True`? IMO this is uncommon. Also `valid` could be a bit clearer like `valid_model`?

leseb (commented):
Just nits.

src/instructlab/model/evaluate.py
nathan-weinberg: The value gets set either way so I didn't put a lot of thought into it - I will change this to `valid_model` and set it as `False` by default with a little extra handling

src/instructlab/model/evaluate.py
```diff
@@ -87,6 +87,19 @@ def get_evaluator(
...
+        if os.path.exists(model):
+            model = pathlib.Path(model)
+            valid_model = False
+            if model.is_dir():
+                valid_model = backends.is_model_safetensors(model)
+            elif model.is_file():
+                valid_model = backends.is_model_gguf(model)
+            if not valid_model:
+                click.secho(
+                    "MTBench and MTBenchBranch need to be passed either safetensors or a GGUF file",
```
booxter, line 99: nit: for a more clueless user (like me), it could be of help to explicitly say here (and below) in the error message that "safetensors" is supposed to be a directory (and not e.g. an actual .safetensors file.)

Something like: "MTBench and MTBenchBranch need to be passed either a safetensors directory or a GGUF file"

tests/test_lab_evaluate.py
```diff
@@ -494,3 +494,47 @@ def test_invalid_tasks_dir(cli_runner: CliRunner):
...
+    test_dir = tmp_path / "test"
+    test_dir.mkdir()
+    result = cli_runner.invoke(
+        lab.ilab,
+        [
+            "--config=DEFAULT",
+            "model",
+            "evaluate",
+            "--benchmark",
+            "mmlu",
```
booxter, line 509: you could probably expand both test cases with variants that would cover both `_branch` and non-`_branch` options. (e.g. with `pytest.mark.parametrize`).

booxter (approved):
It's fine; some ideas on how to improve if you'd like to.

src/instructlab/model/evaluate.py
nathan-weinberg: I like it! Will update with that text

tests/test_lab_evaluate.py
nathan-weinberg: I may do this in a followup if that's alright; want to try to get this in today and have some other items to do as well

tests/test_lab_evaluate.py
```diff
@@ -494,3 +494,47 @@ def test_invalid_tasks_dir(cli_runner: CliRunner):
...
+            "model",
+            "evaluate",
+            "--benchmark",
+            "mt_bench",
+            "--model",
+            test_dir,
+        ],
+    )
+    assert (
+        "MTBench and MTBenchBranch need to be passed either safetensors or a GGUF file"
```
booxter, line 537: think you'll have to update this too

booxter (changes requested):

tests/test_lab_evaluate.py
nathan-weinberg: whoops good catch, thank you!

booxter (approved):

src/instructlab/model/evaluate.py
```diff
@@ -87,6 +87,19 @@ def get_evaluator(
                 fg="red",
             )
             raise click.exceptions.Exit(1)
+        if os.path.exists(model):
+            model = pathlib.Path(model)
+            valid_model = False
```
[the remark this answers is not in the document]
booxter: No. Python will raise `NameError` if you never defined a variable before.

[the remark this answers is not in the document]
khaledsulayman: Ah good point. Still then maybe better to set it to None than False just for convention. But this is a minor nit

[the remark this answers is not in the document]
nathan-weinberg: Fine with either - @booxter do you have a preference? is one approach more pythonic than the other?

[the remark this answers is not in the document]
booxter: Honestly, I don't see why a `None` would be better here. Functionally it would behave the same way; but mixing types for a variable is not advised. (strict `bool` vs `bool | None`.) I like what you have already.

khaledsulayman (approved):

alinaryan (approved):

PR #1797, opened 2024-07-18
Enable SDG batching with vLLM

gabe-l-hart:
Epic: https://github.com/instructlab/sdg/issues/135
Requires: https://github.com/instructlab/sdg/pull/157

When using a serving backend that supports batches of requests (i.e. vLLM), the current SDG pipeline sends the entire dataset as a single large batch of requests to the OpenAI server. This may lead to some requests waiting too long for the response, resulting in timeout errors and potentially overloading the backend server with extremely large batches.

SDG now supports launching parallel tasks to send smaller batches in parallel to the OpenAPI server. This is controlled by the `batch_size` argument to `generate_data()`:

* `batch_size=None` - use the library's built-in default batch size ([currently 0](https://github.com/instructlab/sdg/blob/ca30d98bac7922a095bfb661cef1851ffc2df373/src/instructlab/sdg/generate_data.py#L311-L313), but [will change to 8](https://github.com/instructlab/sdg/issues/224) in a future release)
* `batch_size=0` - disable batching, just use one batch
* `batch_size=<int>` - use batches of the specified size

If the user sets the batch size to 8, the system will run a concurrent thread on each CPU on the system, each sending a batch of 8 prompts. So, on an 8 CPU system, this would result in a total of 64 requests processed simultaneously by the backend server.

This PR adds `--batch-size` to ilab data generate to allow overriding the default behaviour of a batch size of 8 with vLLM and zero with llama-cpp.

Supports https://github.com/instructlab/sdg/issues/135

### DIFF
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -145,6 +145,12 @@ logger = logging.getLogger(__name__)
     help="Data generation pipeline to use. Available: simple, full, or a valid path to a directory of pipeline workflow YAML files. Note that 'full' requires a larger teacher model, Mixtral-8x7b.",
 )
 @click.option(
+    "--batch-size",
+    type=click.IntRange(min=0),
+    default=None,
+    help="Number of elements to process in each batch through the SDG pipeline. Enabled by default for the vLLM serving backend, with a batch size of 8 chosen based on experiments to optimize for throughput. Use 0 to disable.",
+)
+@click.option(
     "--enable-serving-output",
     is_flag=True,
     help="Print serving engine logs.",
@@ -174,6 +180,7 @@ def generate(
     model_family,
     pipeline,
     enable_serving_output,
+    batch_size,
 ):
     """Generates synthetic data to enhance your example data"""
     # pylint: disable=import-outside-toplevel
@@ -192,12 +199,19 @@ def generate(
     if ctx.obj is not None:
         prompt_file_path = ctx.obj.config.generate.prompt_file
 
+    # If batch size is not set explicitly, default to 8
+    # Once https://github.com/instructlab/sdg/issues/224 is resolved we can
+    # pass batch_size=None to the library instead
+    if batch_size is None:
+        batch_size = 8
+
     backend_instance = None
     if endpoint_url:
         api_base = endpoint_url
     else:
         # First Party
         from instructlab.model.backends import backends
+        from instructlab.model.backends.llama_cpp import Server as llama_cpp_server
 
         ctx.obj.config.serve.llama_cpp.llm_family = model_family
         backend_instance = backends.select_backend(ctx.obj.config.generate.teacher)
@@ -210,6 +224,14 @@ def generate(
         except Exception as exc:
             click.secho(f"Failed to start server: {exc}", fg="red")
             raise click.exceptions.Exit(1)
+
+        # disable batching when running with the local llama.cpp server
+        if isinstance(backend_instance, llama_cpp_server):
+            if batch_size is not None:
+                logger.warning(
+                    "Disabling SDG batching - unsupported with llama.cpp serving"
+                )
+            batch_size = 0
     try:
         click.echo(
             f"Generating synthetic data using '{model}' model, taxonomy:'{taxonomy_path}' against {api_base} server"
@@ -236,6 +258,7 @@ def generate(
             tls_client_key=tls_client_key,
             tls_client_passwd=tls_client_passwd,
             pipeline=pipeline,
+            batch_size=batch_size,
         )
     except GenerateException as exc:
         click.secho(
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -142,7 +142,7 @@ logger = logging.getLogger(__name__)
     # Hidden until instructlab-sdg releases a version with multiple pipelines
     # For now only "simple" is supported in the latest release.
     hidden=True,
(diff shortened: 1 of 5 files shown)


src/instructlab/data/generate.py
```diff
@@ -195,6 +207,16 @@ def generate(
         except Exception as exc:
             click.secho(f"Failed to start server: {exc}", fg="red")
             raise click.exceptions.Exit(1)
+
+        # If batch size is not set explicitly, disable batching when running
+        # with the local llama.cpp server
```
derekhiggins, line 212: this isn't detecting llama_cpp vs vllm, AIUI the local backend_instance could be either and we only have a (known) issue using concurrency with llama_cpp.

gabe-l-hart: Ah! Thanks for clarifying. I'll refine it further.

src/instructlab/data/generate.py
```diff
@@ -143,7 +143,18 @@
...
-    help="Data generation pipeline to use. Available: simple, full, or a valid path to a directory of pipeline worlfow YAML files. Note that 'full' requires a larger teacher model, Mixtral-8x7b.",
+    help="Data generation pipeline to use. Available: simple, full, or a valid path to a directory of pipeline workflow YAML files. Note that 'full' requires a larger teacher model, Mixtral-8x7b.",
+)
+@click.option(
+    "--batch-size",
+    type=click.INT,
+    default=None,
+    callback=lambda ctx, param, value: value
+    if value is None or value >= 0
+    else ctx.fail("Batch size must be an integer >= 0"),
```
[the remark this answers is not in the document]
gabe-l-hart: Good catch. Fixed

src/instructlab/data/generate.py
gabe-l-hart: It now does an `isinstance` check on the backend type after `select_backend`. Thanks for pointing this out.

markmc:
CI issue:

```
INFO 2024-07-27 09:35:02,958 instructlab.sdg.pipeline:139: Running pipeline with multi-threaded batching. Using 4 workers for batches of size 8
```

src/instructlab/data/generate.py
```diff
@@ -210,6 +221,13 @@ def generate(
         except Exception as exc:
             click.secho(f"Failed to start server: {exc}", fg="red")
             raise click.exceptions.Exit(1)
+
+        # disable batching when running with the local llama.cpp server
+        if isinstance(backend_instance, llama_cpp_server):
+            batch_size = 0
+            logger.warning(
+                "Disabling SDG batching - unsupported with llama.cpp serving"
```
danmcp, line 229: Should we only log this message when the user requested a particular batch-size?

src/instructlab/data/generate.py
```diff
@@ -139,10 +139,13 @@
...
-    # For now only "simple" is supported in the latest release.
-    hidden=True,
-    help="Data generation pipeline to use. Available: simple, full, or a valid path to a directory of pipeline worlfow YAML files. Note that 'full' requires a larger teacher model, Mixtral-8x7b.",
+    help="Data generation pipeline to use. Available: simple, full, or a valid path to a directory of pipeline workflow YAML files. Note that 'full' requires a larger teacher model, Mixtral-8x7b.",
+)
+@click.option(
+    "--batch-size",
+    type=click.IntRange(min=0),
+    default=None,
+    help="Number of elements to process in each batch through the SDG pipeline. Enabled by default for the vLLM serving backend. Use 0 to disable.",
```
danmcp, line 148: Any way to give a tuning recommendation?

src/instructlab/data/generate.py
```diff
@@ -210,6 +220,16 @@ def generate(
         except Exception as exc:
             click.secho(f"Failed to start server: {exc}", fg="red")
             raise click.exceptions.Exit(1)
+
+        # If batch size is not set explicitly, disable batching when running
+        # with the local llama.cpp server
+        if isinstance(backend_instance, llama_cpp_server) and batch_size is None:
```
danmcp, line 226: Should this be:

```
if llama-cpp
  if batch_size is None
    batch_size=0
  else
    logger.warning(...)
```

src/instructlab/data/generate.py
markmc: From Shiv:

> we did some tests with the vllm endpoint and found that low batch size and increasing the number of workers worked best from a throughput perspective. So I’d recommend a batch size of 8 and setting num workers to num cpus.

Changing to:

> Enabled by default for the vLLM serving backend, with a batch size of 8 chosen based on experiments to optimize for throughput.

src/instructlab/data/generate.py
markmc: Yep, good point

src/instructlab/data/generate.py
markmc: Kind of - but I think we want to set `batch_size=0` in all cases for llama-cpp, even if the user specifies a batch size

markmc:
>  gabe-l-hart force-pushed the SDGBatchArguments branch from 7954527 to fcb84b3 
3 hours ago

@gabe-l-hart was that intentional? Sorry I didn't coordinate - just looking to get this merged quickly

gabe-l-hart:
@markmc Sorry if that wasn't wanted! I got the mergebot conflict notifuvation and wanted to clear the notification before diving into kid time. Feel free to force push back if needed.

n1hility (approved):

danmcp (approved):

markmc (approved):

PR #1804, opened 2024-07-19
docs: Reflect updates to training CI status

russellb:
- We now run legacy training only for `a10g-x1`.

- We run the training-lib in `a10g-x4`, including the use
  of the `full` SDG pipeline as input into training.


russellb:
Placing a `hold` while we want for the `a10g-x4` to finish some runs on main to make sure it's working as we expect.

cdoern (approved):

nathan-weinberg (approved):

PR #1805, opened 2024-07-19
ci: Run the a10g-x4 workflow twice per day

russellb:
This workflow tests the features that are the most difficult to test
in a personal development environment because they require the most
resources.

We're not ready to run it automatically on every commit for both
stability and cost reasons. Run it twice per day so we have a regular
stream of results based on what's in main.

It is also still possible to run it on demand on any PR or branch.

### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x4.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x4.yml
@@ -3,6 +3,8 @@
 name: E2E (NVIDIA A10G x4)
 
 on:
+  schedule:
+    - cron:  '0 8,20 * * *' # Runs at 8 AM and 8 PM UTC every day
   workflow_dispatch:
     inputs:
       pr_or_branch:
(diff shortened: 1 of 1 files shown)


russellb:
The hold is to merge #1804 first, which we will merge once we're happy with the current state of the job after some recent updates.

nathan-weinberg:
@russellb first off, love the badges :ok_hand: 

I am in favor of running these jobs periodically, but I'm not sure where the final part comes in when it comes to results reporting - if the jobs are failing, how will the Maintainer team be aware? Right now it seems the only way to know would be to check the results manually

russellb:
> @russellb first off, love the badges 👌
> 
> I am in favor of running these jobs periodically, but I'm not sure where the final part comes in when it comes to results reporting - if the jobs are failing, how will the Maintainer team be aware? Right now it seems the only way to know would be to check the results manually

I don't know of an easy way other than actively keeping an eye on it. I'll take a look at options, though I wouldn't block on it.

russellb:
> > @russellb first off, love the badges 👌
> > I am in favor of running these jobs periodically, but I'm not sure where the final part comes in when it comes to results reporting - if the jobs are failing, how will the Maintainer team be aware? Right now it seems the only way to know would be to check the results manually
> 
> I don't know of an easy way other than actively keeping an eye on it. I'll take a look at options, though I wouldn't block on it.

I filed #1806 to follow-up on notifications

.github/workflows/e2e-nvidia-a10g-x4.yml
```diff
@@ -3,6 +3,8 @@
 name: E2E (NVIDIA A10G x4)
 
 on:
+  schedule:
+    - cron:  '0 8,20 * * *' # Runs at 8 AM and 8 PM UTC every day
```
nathan-weinberg, line 7:
```suggestion
    - cron: '0 8,20 * * *' # Runs at 8 AM and 8 PM UTC every day
```

nathan-weinberg (approved):
one nit otherwise LGTM

cdoern (approved):

PR #1807, opened 2024-07-19
feat: automatically add upstream remote to instructlab clone

nathan-weinberg:
adds a remote 'upstream' as part of the setup process this allows users to pull directly from the upstream project on their clones without needing to manually add the remote

**Checklist:**

### DIFF
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -267,7 +267,7 @@ setup_rh_devenv() {
     local cloud_type=$1
     "${BASH_SOURCE[0]}" "$cloud_type" ssh -n "$INSTANCE_NAME" sudo dnf install git gcc make pip python3.11 python3.11-devel -y
     "${BASH_SOURCE[0]}" "$cloud_type" ssh -n "$INSTANCE_NAME" "sudo dnf install g++ -y || sudo dnf install gcc-c++"
-    "${BASH_SOURCE[0]}" "$cloud_type" ssh -n "$INSTANCE_NAME" "if [ ! -d instructlab.git ]; then git clone --bare https://github.com/instructlab/instructlab.git && git clone instructlab.git && pushd instructlab && git remote add syncrepo ../instructlab.git; fi"
+    "${BASH_SOURCE[0]}" "$cloud_type" ssh -n "$INSTANCE_NAME" "if [ ! -d instructlab.git ]; then git clone --bare https://github.com/instructlab/instructlab.git && git clone instructlab.git && pushd instructlab && git remote add syncrepo ../instructlab.git && git remote add upstream https://github.com/instructlab/instructlab.git; fi"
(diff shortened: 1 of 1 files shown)


scripts/infra/cloud-instance.sh
```diff
@@ -267,7 +267,7 @@ setup_rh_devenv() {
     local cloud_type=$1
     "${BASH_SOURCE[0]}" "$cloud_type" ssh -n "$INSTANCE_NAME" sudo dnf install git gcc make pip python3.11 python3.11-devel -y
     "${BASH_SOURCE[0]}" "$cloud_type" ssh -n "$INSTANCE_NAME" "sudo dnf install g++ -y || sudo dnf install gcc-c++"
-    "${BASH_SOURCE[0]}" "$cloud_type" ssh -n "$INSTANCE_NAME" "if [ ! -d instructlab.git ]; then git clone --bare https://github.com/instructlab/instructlab.git && git clone instructlab.git && pushd instructlab && git remote add syncrepo ../instructlab.git; fi"
+    "${BASH_SOURCE[0]}" "$cloud_type" ssh -n "$INSTANCE_NAME" "if [ ! -d instructlab.git ]; then git clone --bare https://github.com/instructlab/instructlab.git && git clone instructlab.git && pushd instructlab && git remote add syncrepo ../instructlab.git && git remote add upstream https://github.com/instructlab/instructlab.git; fi"
```
danmcp, line 270: This would be taking the opposite approach of using sync -c with your code on your laptop and using the instance as a sync target.  I don't guess I see any harm, but want to make sure the intended use was understandable.

nathan-weinberg: Ah I see - I have been using the `sync` feature, the reason I added this was I was on my instance and wanted to pull down the latest `main` directly to run a few things.

Any possible issues with this? Happy to close if it'll cause problems, but otherwise I think it's a nice-to-have.

danmcp: I don't think this will break any of the existing workflows.  It is a tiny bit weird the way it will be setup.  The bare repo's origin will be set to the same repo as the upstream of cloned repo.  Still nothing technically wrong with that AFAICT.

Also if we aren't going to doc it (which I kind of prefer at the moment) it will mostly go unnoticed.  My only fear would be someone would assume an intended use based on the upstream being there.  But if that's what they are looking for then so be it I guess.

danmcp (approved):

scripts/infra/cloud-instance.sh
danmcp: I don't think this will break any of the existing workflows. It is a tiny bit weird the way it will be setup. The bare repo's origin will be set to the same repo as the upstream of cloned repo. Still nothing technically wrong with that AFAICT.

Also if we aren't going to doc it (which I kind of prefer at the moment) it will mostly go unnoticed. My only fear would be someone would assume an intended use based on the upstream being there. But if that's what they are looking for then so be it I guess.

leseb (approved):

PR #1809, opened 2024-07-19
fix: check for .bin files during auto-detection of serving backend

jaideepr97:
Current check for auto-detection of backend requires a vLLM friendly model's directory to contain `*.safetensors` files to qualify as valid. However, we should also accept the presence of `*.bin` files in place of `*.safetensors` files as training outputs `.bin` checkpoints that can be served through vLLM

**Checklist:**

### DIFF
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -17,22 +17,22 @@ from instructlab.model.backends import backends
 from instructlab.model.backends.vllm import build_vllm_cmd
 
 
-# helper function to create dummy valid and invalid safetensor model directories
-def create_safetensors_model_files(safetensors_model_path: pathlib.Path, valid: bool):
+# helper function to create dummy valid and invalid safetensor or bin model directories
+def create_safetensors_or_bin_model_files(
+    model_path: pathlib.Path, model_file_type: str, valid: bool
+):
     test_json_dict = {"a": 1, "b": 2}
     json_object = json.dumps(test_json_dict, indent=4)
 
     for file in ["tokenizer.json", "tokenizer_config.json"]:
-        os.makedirs(os.path.dirname(safetensors_model_path / file), exist_ok=True)
-        with open(safetensors_model_path / file, "a+", encoding="UTF-8") as f:
+        os.makedirs(os.path.dirname(model_path / file), exist_ok=True)
+        with open(model_path / file, "a+", encoding="UTF-8") as f:
             f.write(json_object)
 
-    with open(
-        safetensors_model_path / "model.safetensors", "a+", encoding="UTF-8"
-    ) as f:
-        f.write("")
+    with open(model_path / f"model.{model_file_type}", "a+", encoding="UTF-8"):
+        pass
     if valid:
-        with open(safetensors_model_path / "config.json", "a+", encoding="UTF-8") as f:
+        with open(model_path / "config.json", "a+", encoding="UTF-8") as f:
             f.write(json_object)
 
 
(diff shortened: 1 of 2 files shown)


tests/test_backends.py
```diff
@@ -17,22 +17,30 @@
...
-        os.makedirs(os.path.dirname(safetensors_model_path / file), exist_ok=True)
-        with open(safetensors_model_path / file, "a+", encoding="UTF-8") as f:
+        os.makedirs(os.path.dirname(model_path / file), exist_ok=True)
+        with open(model_path / file, "a+", encoding="UTF-8") as f:
             f.write(json_object)
 
-    with open(
-        safetensors_model_path / "model.safetensors", "a+", encoding="UTF-8"
-    ) as f:
+    model_file = (
```
booxter, line 32:
```
with open(model_path / f"model.{model_file_type}", ...):
    pass
```

? (`pass` will be equivalent to `f.write("")` I think.)

booxter (changes requested):
Style change suggested; code-wise it's good I think.

Also, please re-wrap commit message lines to 50/72 rule. Thank you.

tests/test_backends.py
jaideepr97: what would the benefit of making that change be?

jaideepr97:
>Also, please re-wrap commit message lines to 50/72 rule. Thank you.

Interesting @booxter , wasn't familiar with this rule but it seems like we haven't really been following it in this repository 
I've mostly been asked to reuse the PR description as the commit msg

booxter:
@jaideepr97 Yes, commit messages in this repo are most often than not non-compliant with 50/72. I don't think the project ever made an explicit decision that the rule should be enforced, but it shouldn't stop us from applying the best practice. Core `git` tools like `git-log` expect that the rule is adhered. And while we can discuss if 72 or 100 limit is the best for the project, having SOME limit and not piling the commit message into a single title line shouldn't be controversial.

@russellb should the project formally document some rules for commit messages?

tests/test_backends.py
booxter: 2 lines vs 10 lines? The nested ternary `if-else` is also rather convoluted.

russellb:
> @jaideepr97 Yes, commit messages in this repo are most often than not non-compliant with 50/72. I don't think the project ever made an explicit decision that the rule should be enforced, but it shouldn't stop us from applying the best practice. Core `git` tools like `git-log` expect that the rule is adhered. And while we can discuss if 72 or 100 limit is the best for the project, having SOME limit and not piling the commit message into a single line shouldn't be controversial.
> 
> @russellb should the project formally document some rules for commit messages?

Yes.

I started https://github.com/instructlab/dev-docs/pull/110

haven't prioritized getting back to it to finish it up, though.

Maxusmusti (approved):
LGTM

tests/test_backends.py
RobotSail: Personally I'm not a fan of this approach, as it increases the valid values for file types from just `safetensors` and `bin` to allow all valid strings.

We should instead keep the values constricted to just `safetensors` and `bin` as those are types which we explicitly support, and instead throw a `ValueError` if the provided model type ever deviates.

In all likelihood, this function shouldn't be used outside of this codebase or even that much, so we shouldn't run into this error in the first place.

RobotSail: So we can just make the following change:
```suggestion
    if model_file_type not in ("safetensors", "bin"):
        raise ValueError(f"model file type is not supported: {model_file_type}")
    with open(os.path.join(model_path, f"model.{model_file_type}"), "a+", encoding="UTF-8"):
        pass
```

booxter: since this is test case, and if you'd like to raise an error, you can also `assert model_file_type in (...)`

booxter: (that said, I think there's a test variant here that actually passes "other".)

RobotSail (approved):
LGTM

leseb (approved):
@booxter do you have request changes left? Added the hold flag until your concerns are resolved. Thanks

tests/test_backends.py
jaideepr97: @booxter ah sorry - I totally misread what you were suggesting 
have updated the code to match that, thanks! 🙏 

@RobotSail this is just a test case so I think we do need to create valid and invalid scenarios to properly test the code

jaideepr97:
updated the commit to follow 50/72 rule @booxter 
sorry for the delay

Thanks for all the inputs, everyone!

booxter (approved):

PR #1810, opened 2024-07-19
evaluate: Demand --model argument through click library

booxter:
The code makes assumptions that model is a string (meaning, it was actually passed as --model argument). For example, it's passed into `launch_server` that demands a string (not a None), and that will raise a TypeError if None is passed.

The model being not None was (incompletely) validated inside `get_evaluator`, but not consistently. This patch just adds required=True to the click option, which allows to simplify the validation code, and avoid some potential typing issues.

**Checklist:**

- [X] Integration tests have been added, if necessary.


src/instructlab/model/evaluate.py
```diff
@@ -113,19 +111,13 @@ def get_evaluator(
 
     # ensure knowledge benchmarks have proper arguments if selected
     if benchmark in [Benchmark.MMLU, Benchmark.MMLU_BRANCH]:
-        required_args = [model, few_shots, batch_size]
-        required_arg_names = ["model"]
```
booxter, line 117: btw probably there's a bug here, see that few_shots and batch_size are NOT listed in required_arg_names. But I left it as-is.

tests/test_lab_evaluate.py
```diff
@@ -354,10 +354,7 @@ def test_no_model_mt_bench(cli_runner: CliRunner):
             "mt_bench",
         ],
     )
-    assert (
-        result.output
-        == "Benchmark mt_bench requires the following args to be set: ['model', 'judge-model']\n"
-    )
+    assert "Error: Missing option '--model'." in result.output
```
nathan-weinberg, line 357: Since the benchmark isn't relevant to the test anymore, worth renaming to just `test_no_model`?

src/instructlab/model/evaluate.py
```diff
@@ -335,6 +331,7 @@ def launch_server(
     "--model",
     type=click.STRING,
     cls=clickext.ConfigOption,
+    required=True,
```
danmcp, line 334: I thought the reason this wasn't required here was because it's in the configuration as well.

cdoern: required=True is the workflow we use for flags populated by the cfg bc they should and will always be there, right?

cdoern (commented):
this looks good and this should be how `judge_model` is handled as well I think I ran into this once during an e2e

nathan-weinberg:
@booxter we discussed this among some Eval Maintainers in a meeting yesterday - I think by-and-large we are agreed on the problem statement but we were less sure about requiring the flag as a solution - the reason for this is that there are some scenarios where you may not want to require the flag - for example, if you are working with a particular model you may want to set that model in the config for all command usages and not want to use the flag in `evaluate` - please let me know if that makes sense or if I've misunderstood anything about the PR!

PR #1812, opened 2024-07-20
refactor: Remove unnecessary `export`s from scripts

### LINKED ISSUE #1798: Remove `export` from test scripts
Currently in our test scripts, we make use of `export VAR` at the top level of each test script.

However, as pointed out in [this comment](https://github.com/instructlab/instructlab/pull/1762#discussion_r1683405990) and this [stackoverflow post](https://stackoverflow.com/a/1158231), the only reason to use `export` is when you want to make a variable available to subprocesses. 

Apart from a few variables such as `HOME`, we should avoid using this keyword when declaring new variables. 

Therefore, we should remove the usage of `export` in the existing test scripts that we currently have.

booxter:
`export` is used to pass a variable to a subprocess. These variables are not read by any subprocesses started by the scripts. No need to export them.

Resolves #1798

**Checklist:**

- [X] Integration tests have been added, if necessary.

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -35,14 +35,12 @@ BOLD='\033[1m'
 NC='\033[0m' # No Color
 
 SCRIPTDIR=$(dirname "$0")
-export E2E_TEST_DIR
+E2E_TEST_DIR=
 # E2E_TEST_DIR="$(pwd)/__e2e_test"
-export CONFIG_HOME
-export DATA_HOME
-export CACHE_HOME
-export CONFIG_HOME
-
-
+CONFIG_HOME=
+DATA_HOME=
+CACHE_HOME=
+CONFIG_HOME=
 
 ########################################
 # This function initializes the directory
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -11,9 +11,9 @@
 set -ex
 
 # shellcheck disable=SC2155
-export SCRIPTDIR=$(dirname "$0")
+SCRIPTDIR=$(dirname "$0")
 # build a prompt string that includes the time, source file, line number, and function name
-export PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
 if [ -n "$TEST_DIR" ]; then
     echo "TEST_DIR is set to $TEST_DIR by the caller, will not remove it"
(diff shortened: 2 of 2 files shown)


scripts/functional-tests.sh
```diff
@@ -4,25 +4,11 @@
 set -ex
 
 # shellcheck disable=SC2155
```
leseb, line 6: This can go away too now, rigth?

scripts/basic-workflow-tests.sh
```diff
@@ -31,11 +31,6 @@ BOLD='\033[1m'
 NC='\033[0m' # No Color
 
 SCRIPTDIR=$(dirname "$0")
-export E2E_TEST_DIR
```
RobotSail, line 34: Why are these variables being removed? We need them to exist for this end-to-end test.

scripts/functional-tests.sh
```diff
@@ -4,25 +4,11 @@
...
 # build a prompt string that includes the time, source file, line number, and function name
-export PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+PS4='+$(date +"%Y-%m-%d %T") ${BASH_VERSION}:${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
-export TEST_DIR
-export PACKAGE_NAME='instructlab'  # name we use of the top-level package directories for CLI data
-
-
-# get the directories from the platformdirs library
-export CONFIG_DIR
```
RobotSail, line 16: We can't remove these variables.

scripts/functional-tests.sh
```diff
@@ -4,25 +4,11 @@
...
-export DATA_DIR
-export CACHE_DIR
-
-
-# we define the path here to reference elsewhere, but the existence of this file
-# will be managed by the ilab CLI
-export ILAB_CONFIGDIR_LOCATION
```
RobotSail, line 23: We can't remove these variables.

RobotSail (changes requested):
We can't remove the variables entirely, we should just remove the `export` keyword on variables which aren't being used in subprocess calls. A number of these variables should still be declared and initialized so that their definitions can be easily found.

booxter:
> so that their definitions can be easily found

@RobotSail ack, we can leave them as `VAR=` for **documentation** purposes then. But AFAIU, technically the lines are not needed. (Because shell implicitly equates `echo $NON_EXISTING_ENVVAR` and `echo $ENVVAR_WITH_EMPTY_VALUE`.)

RobotSail:
@booxter It's important to keep these variables declared at the scope they are used at, both for documentation purposes as well as making it easier for other developers to work with these files. 

In general, our style should focus on the side of readability as opposed to just what is technically considered legal in BASH. This will ensure a more clear and consistent codebase overall.

I personally also make use of the BASH LSP extension in VSCode, so having these declared at the top-level is also very important to know where they're coming from:
https://marketplace.visualstudio.com/items?itemName=mads-hartmann.bash-ide-vscode

booxter:
Meh. Not worth it.

PR #1814, opened 2024-07-20
refactor: replace ternary if-else with or

booxter:


tiran (changes requested):
strong -1, the ternary operation was introduced in 2.4 for a good reason.

Your change also introduces changes for subtle bugs. For example `backend or cfg.backend` is not equivalent to `backend if backend is not None else cfg.backend`

RobotSail (changes requested):
We use the ternary operator since it makes the code readable and easy to understand what's actually happening under the hood. 

It's not a good idea to use the `or` operator for assigning variables as it evaluates falsy-ness which can lead to unexpected behavior.

booxter:
> It's not a good idea to use the or operator for assigning variables as it evaluates falsy-ness

I don't what it means. Python `or` returns one of the operands; it doesn't convert it to `bool`. Same as ternary operator.

RobotSail:
@booxter `or` evaluates a value as a boolean to understand whether it should be used for assignment or not. This can lead to unexpected behavior when the intended value evaluates to `False`. For example, if we have something like

```py
def get_temperature_or_default(temp: float | None) -> float:
    return temp or 0.1

config.temp = 0.0  # user wants there to be no variance in output by setting temp to 0.0
temp = get_temperature_or_default(config.temp)

# 0.0 is dropped because bool(0.0) == False
print(temp)
> 0.1
```

So to avoid these cases, we prefer to use the ternary operator so that we can be clear on the condition in which we take on a certain value.

booxter:
OK. This is the same point that tiran did above - that `x is None` != `x` in boolean context. We can definitely revert these back. There are cases that do `x if x else y` though - where ternary makes little sense.

RobotSail:
> There are cases that do `x if x else y` though - where ternary makes little sense.

I don't disagree, but I think in these cases we should be explicit about what the check is, e.g. `x is not None` unless in the cases where having them take on values such as `0`, `[]`, etc. in those cases as `False` is the intended behavior.

tiran:
Yes, there are cases where `egg or spam` trick is fine. In my professional experience of over 25 years of coding and over 20 years with Python, it is not worth to use special cases unless there is a benefit or technical reason for the special case. Code is more often read than written, and therefore should be self-documenting. Ternary operator is self-explaining and works reliable. I personally even go so far and prefer four-line if/else block over ternary in cases where it is more readable.

Instead of adding the `egg or spam` trick to our code base, we should rather remove any case of the trick and replace it with a ternary operator or if/else block. They are clearer, more explicit, and should be preferred.

booxter:
Clearly there are Opinions. :)

PR #1818, opened 2024-07-22
chore: update cloud provisioning

n1hility:
- Use precompiled nvidia driver approach to bring parity with bootc images
  + align driver versions and cuda libs.
- Update pip build process to be compliant with odh vllm
- Use constraints to ensure compatible dependency set
- Drop system pip install to avoid conflicts with venv approach
- Introduce new update command to manage updates, since drivers must be rebuilt
- Add ssh monitoring support for script initiated reboots

**Checklist:**

- [ x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
- ~~[ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.~~
- ~~[ ] Documentation has been updated, if necessary.~~
- ~~[ ] Unit tests have been added, if necessary.~~
- ~~[ ] Integration tests have been added, if necessary.~~

### DIFF
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -59,7 +59,7 @@ ec2__launch() {
         --security-group-ids "$EC2_SECURITY_GROUP_ID" \
         --subnet-id "$EC2_SUBNET_ID" \
         --key-name "$EC2_KEY_NAME" \
-        --block-device-mappings '{"DeviceName": "/dev/sda1","Ebs": {"VolumeSize": 200}}' \
+        --block-device-mappings '{"DeviceName": "/dev/sda1","Ebs": {"VolumeSize": 300}}' \
         --associate-public-ip-address \
         --tag-specifications "ResourceType=instance,Tags=[{Key=Name,Value=$INSTANCE_NAME}]" \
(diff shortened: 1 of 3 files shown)


scripts/infra/cloud-instance.sh
```diff
@@ -119,6 +119,22 @@ ec2__terminate() {
...
+        echo "Waiting..."; sleep 30
+        return
+    fi
+    while true; do
+        echo "Waiting for ssh..."
+        sleep 5
+        if nc -w 5 "$PUBLIC_DNS" 22; then
+            break
+        fi
+    done
```
leseb, line 135: Why not just run `ssh` like so? `ssh -o ConnectTimeout=5 user@address exit` This way no deps on nc

leseb (approved):

scripts/infra/cloud-instance.sh
n1hility: oh sure yeah, only reason I didn't do it like that is ssh includes the handshake time in that, and so you can get a spurious failure when everything is ok. But in the grand scheme of things another 5 seconds doesnt matter that much vs the dep. so Ill just switch it

PR #1825, opened 2024-07-22
chore(konflux): upstream e2e tests pipeline

lmilbaum:
This is an MVP Konflux pipeline to unblock invoking the upstream e2e tests on every konflux enabled PRs/MRs.

**Checklist:**

### DIFF
--- a/.tekton/integration-tests.yaml
+++ b/.tekton/integration-tests.yaml
@@ -1,0 +1,13 @@
+apiVersion: tekton.dev/v1
+kind: Pipeline
+metadata:
+  name: upstream integration tests
+spec:
+  description: >-
+    TBD
+  tasks:
+    - name: e2e-test
+      taskSpec:
+        steps:
+        - image: ubuntu
+          script: ./scripts/basic-workflow-tests.sh


russellb:
Can you explain a bit more what you have in mind for this? It looks like a WIP?

.tekton/integration-tests.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: tekton.dev/v1
+kind: Pipeline
+metadata:
+  name: upstream integration tests
+spec:
+  description: >-
+    TBD
```
nathan-weinberg, line 7: Could we add a description here so there's context for those that will come to see this file later?

lmilbaum: Sure. Can you help me with crafting this field content?

nathan-weinberg: What is the context?

lmilbaum: This pipeline is invoking the e2e tests. The description might include the e2e structure or/and what it is testing.

lmilbaum:
> > Can you explain a bit more what you have in mind for this? It looks like a WIP?
> 
> I gather that this is the start of some config for a different CI system. Some questions

Correct. It is intended for Konflux.

> 
> * What will this test that's different from what we test already?

This pipeline could be used on Konflux enabled repos

> * When will it run? on PRs? periodically?  something else?

It will run on downstream MRs

> * Where will results be published? (They need to be public for putting this config in the public repo to make sense)

At the moment, there are no plans to publish the results publicly.  We can plan it if that is a requirement.

> 
> I'd also like to see this documented in `docs/ci.md` to continue to capture details of our test coverage.

That could be done when this piece of code can be of value to the community

> 
> Thanks!

.tekton/integration-tests.yaml
nathan-weinberg: Okay - you would know better than I so if you can add something about that I can review it!

nathan-weinberg (approved):

nathan-weinberg:
@lmilbaum we are going to be revisiting a lot of the E2E testing as part of this epic: https://github.com/instructlab/instructlab/issues/2196

would not be opposed to revisiting this after some of that work has made it upstream!

PR #1676, opened 2024-07-11
Add remaining eval benchmarks to basic-workflow-tests

alinaryan:
This commit adds the remaining ilab model evaluate benchmarks to the basic-workflow-tests script including mmlu_branch, mt_bench, and mt_bench_branch. Also, includes downloading merlinite to be used
as a judge model in the 'branch' benchmarks.

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

Issue #1642, opened 2024-07-09
CI: `ilab model evaluate` in E2E runs are not incorporated into full ilab workflow


alimaredia:
**Describe the bug**

The ilab workflow tested in basic-workflow-tests.sh (https://github.com/instructlab/instructlab/blob/main/scripts/basic-workflow-tests.sh#L196) tests model evaluation, however evaluation is not being tested as it would be run in the full ilab workflow.

As development for training and SDG happen, the following should be incorporated into the model evaluation happening in the CI

1. The models these commands are testing should be the new model trained by `ilab model train`.
2. The Base model for the `mt_bench_branch` and `mmlu_branch` should be the model before training.
3. The taxonomy used for `mt_bench_branch` should be the taxonomy that was changed before the SDG step.
4. The mmlu_pr questions used in mmlu_branch are generated from the SDG step (ilab data generate).

nathan-weinberg:
Happy to help y'all with this if you want

nathan-weinberg:
@alimaredia @alinaryan @russellb this seems to be a duplicate of https://github.com/instructlab/instructlab/issues/1540 - am I wrong? If not we should close one or the other

PR #1833, opened 2024-07-22
Eval E2E test using trained model

cdoern:
enable eval on the a10x4, swap in trained model
    
 this enables eval to run using the last checkpoint from training. It also enables this workflow to run on the a10x4 runners

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -356,9 +356,10 @@ test_convert() {
 }
 
 test_evaluate() {
-    local MODEL_PATH="${CACHE_HOME}/instructlab/models/instructlab/granite-7b-lab"
     # Temporarily using merlinite as the base model to confirm the workflow executes correctly 
     local BASE_MODEL_PATH="${CACHE_HOME}/instructlab/models/instructlab/merlinite-7b-lab"
+    local MODEL_PATH=""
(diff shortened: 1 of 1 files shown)


cdoern:
this is built on top of #1676

mairin:
This is proposed as a blocker. What's the consequence on 0.18.0 if this doesn't make it in?

cdoern:
@mairin this is not really  a functional blocker as we have validated evaluations functionality manually. I'm fine with removing this from the blocker list!

mairin:
@cdoern OK cool, thanks! If this doesn't make it in we won't have eval tests of the trained model in CI, right?

nathan-weinberg:
@mairin that's correct - @cdoern i can help you try and get this over the line but we can prioritize appropriately (i.e. not hold a 0.18.0 release for this)

mairin:
This is not a blocker, it can make it into 0.18.0 if ready because it's CI and does not impact code.

alinaryan:
Blocked on testing. There is not enough capacity in the a10x4 ci test

nathan-weinberg:
> Blocked on testing. There is not enough capacity in the a10x4 ci test

Is it because of the size of the Prometheus model? We could use something like Merlinite instead since we're not as concerned with results here (if that's the issue)

ktam3:
Closing this in favor of #1642

PR #1839, opened 2024-07-23
ci: support vLLM installation in 4xA10 workflow

alimaredia:
This PR ensure vLLM can be installed correctly in the 4xA10 workflow. Separate PRs will ensure the `-v` flag to basic-workflows.sh is utilized.

### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x4.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x4.yml
@@ -32,7 +32,7 @@ jobs:
         with:
           mode: start
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
-          ec2-image-id: ami-00c51d9c1374eda97
+          ec2-image-id: ami-03f6686697972e40a
           ec2-instance-type: g5.12xlarge
           subnet-id: subnet-02d230cffd9385bd4
           security-group-id: sg-06300447c4a5fbef3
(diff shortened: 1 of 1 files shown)


alimaredia:
This clean run (https://github.com/russellb/ilab-runner/actions/runs/10058671526/job/27802382741) shows that changes to this workflow file result in a clean run.

The only difference between https://github.com/russellb/ilab-runner/blob/main/.github/workflows/e2e-nvidia-a10g-x4-amaredia.yml and what is in this PR is just ensuring the main branch of instructlab is used in the test not what's in the ilab-runner repo.

nathan-weinberg (approved):

.github/workflows/e2e-nvidia-a10g-x4.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         with:
           mode: start
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
-          ec2-image-id: ami-00c51d9c1374eda97
+          ec2-image-id: ami-03f6686697972e40a
```
booxter, line 35: I'm clueless about how AMIs are picked for CI, so the question is probably ignorant but: why is the AMI changed here and what does the change represent? Is the motivation to sync with the ilab-runner version regardless what is there, or does the change represent something material needed for the CI?

nathan-weinberg: My assumption is that this is related to @n1hility's change to the cloud config scripting here: https://github.com/instructlab/instructlab/pull/1818

n1hility: That is correct its a material change of need so that we have an aligned set of drivers that mirror the ai-lab bootc image and meet the deps of the vllm release.

n1hility (approved):

.github/workflows/e2e-nvidia-a10g-x4.yml
alimaredia: The AMI change here is relevant because it has all of the cuda drivers properly installed to build vLLM. This is happening when we run `pip install -r requirements-vllm-cuda.txt`. 

Without the AMI change, that `pip install` would error out.

booxter: Thank you for the explanation. You rock.

PR #1840, opened 2024-07-23
add `ilab model list`

cdoern:
`ilab model list` prints a table of all models in ~/.local/share/instructlab/models. Both GGUF and Safetensor models. This cmd also takes a `--list-checkpoints` flag that will expose the checkpoints generated by `ilab model train`. These live in ~/.cache/instructlab/checkpoints

also adds an e2e test checking functionality of gguf and safetensor models

an example output of this command looks like:

```
+------------------------------+---------------------+----------+
| Model Name                   | Last Modified       | Size     |
+------------------------------+---------------------+----------+
| merlinite-7b-lab-Q4_K_M.gguf | 2024-07-24 16:01:57 | 4.1 GiB  |
| instructlab/granite-7b-lab   | 2024-07-24 16:17:39 | 12.6 GiB |
+------------------------------+---------------------+----------+

```

**Checklist:**

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -148,6 +148,25 @@ test_download() {
     fi
 }
 
+test_list() {
+    task List the Downloaded Models
+
+    if [ "$GRANITE" -eq 1 ]; then
+        step Listing the granite GGUF model only
+        ilab model list | grep granite-7b-lab-Q4_K_M.gguf
+    elif [ "$MIXTRAL" -eq 1 ]; then
+        step Listing the mixtral model only
+        ilab model list | grep mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf
+    else
+        step Listing the merlinite GGUF model only
+        ilab model list | grep merlinite-7b-lab-Q4_K_M.gguf4
+    fi
+    
+    # regardless, download merl safetensors to test that capability
+    ilab model download --repository instructlab/merlinite-7b-lab
+    ilab model list | grep instructlab/merlinite-7b-lab
+}
+
 test_serve() {
     # Accepts an argument of the model, or default here
     if [ "$GRANITE" -eq 1 ]; then
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -167,20 +167,24 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
     # Third Party
     from gguf.constants import GGUF_MAGIC
 
-    with open(model_path, "rb") as f:
-        # Memory-map the file on the first 4 bytes (this is where the magic number is)
-        mmapped_file = mmap.mmap(f.fileno(), length=4, access=mmap.ACCESS_READ)
+    try:
+        with open(model_path, "rb") as f:
+            # Memory-map the file on the first 4 bytes (this is where the magic number is)
+            mmapped_file = mmap.mmap(f.fileno(), length=4, access=mmap.ACCESS_READ)
 
-        # Read the first 4 bytes
-        first_four_bytes = mmapped_file.read(4)
+            # Read the first 4 bytes
+            first_four_bytes = mmapped_file.read(4)
 
-        # Convert the first four bytes to an integer
-        first_four_bytes_int = int(struct.unpack("<I", first_four_bytes)[0])
+            # Convert the first four bytes to an integer
+            first_four_bytes_int = int(struct.unpack("<I", first_four_bytes)[0])
 
-        # Close the memory-mapped file
-        mmapped_file.close()
+            # Close the memory-mapped file
+            mmapped_file.close()
 
-        return first_four_bytes_int == GGUF_MAGIC
+            return first_four_bytes_int == GGUF_MAGIC
+    except IsADirectoryError as exc:
+        logger.debug(f"GGUF Path is a directory, returning {exc}")
+        return False
 
 
 def determine_backend(model_path: pathlib.Path) -> Tuple[str, str]:
--- a/src/instructlab/model/list.py
+++ b/src/instructlab/model/list.py
@@ -1,0 +1,151 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+from pathlib import Path
+import logging
+import os
+import time
+
+# Third Party
+import click
+
+# First Party
+from instructlab import clickext
+from instructlab.configuration import DEFAULTS
+from instructlab.model.backends.backends import is_model_gguf, is_model_safetensors
+
+logger = logging.getLogger(__name__)
+
+
+@click.command(name="list")
+@clickext.display_params
+@click.option(
+    "--model-dirs",
+    help="Base directories where models are stored.",
+    multiple=True,
+    default=[DEFAULTS.MODELS_DIR],
+    show_default=True,
+)
+@click.option(
+    "--list-checkpoints",
+    is_flag=True,
+)
+def model_list(model_dirs: list[str], list_checkpoints: bool):
+    """lists models"""
+    data = []
+    for directory in model_dirs:
+        for entry in Path(directory).iterdir():
+            # if file, just tally the size. This must be a GGUF.
+            if entry.is_file() and is_model_gguf(entry):
+                model, age, size = _analyze_gguf(entry)
+                data.append([model, age, size])
+            elif entry.is_dir():
+                for m in _analyze_dir(entry, list_checkpoints, directory):
+                    data.append(m)
+    print_table(["Model Name", "Last Modified", "Size"], data)
+
+
+# convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MiB, GiB)
+def convert_bytes_to_proper_mag(f_size: float):
+    magnitudes = ["KiB", "MiB", "GiB", "TiB"]
+    magnitude = "B"
+    for mag in magnitudes:
+        if f_size >= 1024:
+            magnitude = mag
+            f_size /= 1024
+        else:
+            return f_size, magnitude
+    return f_size, magnitude
+
+
+# print_table displays the models on the system in a nicely formatted table
+def print_table(headers: list[str], data: list[list[str]]):
+    column_widths = [
+        max(len(str(row[i])) for row in data + [headers]) for i in range(len(headers))
+    ]
+    # Print separator line between headers and data
+    horizontal_lines = ["-" * (width + 2) for width in column_widths]
+    joining_line = "+" + "+".join(horizontal_lines) + "+"
+    print(joining_line)
+    outputs = []
+    for header, width in zip(headers, column_widths, strict=False):
+        outputs.append(f" {header:{width}} ")
+    print("|" + "|".join(outputs) + "|")
+    print(joining_line)
+    for row in data:
+        outputs = []
+        for item, width in zip(row, column_widths, strict=False):
+            outputs.append(f" {item:{width}} ")
+        print("|" + "|".join(outputs) + "|")
+    print(joining_line)
+
+
+AnalyzeResultGGUF = list[str]
+AnalyzeResultDir = list[list[str]]
+
+
+def _analyze_gguf(entry: Path) -> AnalyzeResultGGUF:
+    # stat the gguf, add it to the table
+    stat = Path(entry.absolute()).stat(follow_symlinks=False)
+    f_size = stat.st_size
+    f_size, magnitude = convert_bytes_to_proper_mag(f_size)
+    # add to table
+    modification_time = os.path.getmtime(entry.absolute())
+    modification_time_readable = time.strftime(
+        "%Y-%m-%d %H:%M:%S", time.localtime(modification_time)
+    )
+    return [entry.name, modification_time_readable, f"{f_size:.1f} {magnitude}"]
+
+
+def _analyze_dir(
+    entry: Path, list_checkpoints: bool, directory: str
+) -> AnalyzeResultDir:
+    actual_model_name = ""
+    all_files_sizes = 0
+    add_model = False
+    models = []
+    # walk entire dir.
+    for root, _, files in os.walk(entry.as_posix()):
+        normalized_path = os.path.normpath(root)
+        # Split the path into its components
+        parts = normalized_path.split(os.sep)
+        # Get the last two parts and join them back into a path
+        last_two_parts = os.path.join(parts[-2], parts[-1])
+        # if this is a dir it could be:
+        # top level repo dir `instructlab/`
+        # top level model dir `instructlab/granite-7b-lab`
+        # checkpoint top level dir `step-19`
+        # any lower level dir: `instructlab/granite-7b-lab/.huggingface/download.....`
+        # so, check if model is valid Safetensor, GGUF, or list it regardless w/ `--list-checkpoints`
+        # if --list-checkpoints is specified, we will list all checkpoints in the checkpoints dir regardless of the validity
+        if is_model_safetensors(Path(normalized_path)) or is_model_gguf(
+            Path(normalized_path)
+        ):
+            actual_model_name = last_two_parts
+            all_files_sizes = 0
+            add_model = True
+        else:
+            if list_checkpoints and directory is DEFAULTS.CHECKPOINTS_DIR:
+                logging.debug("Including model regardless of model validity")
+            else:
+                continue
+        for f in files:
+            # stat each file in the dir, add the size in Bytes, then convert to proper magnitude
+            full_file = os.path.join(root, f)
+            stat = Path(full_file).stat(follow_symlinks=False)
+            all_files_sizes += stat.st_size
+        all_files_sizes, magnitude = convert_bytes_to_proper_mag(all_files_sizes)
+        if add_model:
+            # add to table
+            modification_time = os.path.getmtime(entry.absolute())
+            modification_time_readable = time.strftime(
+                "%Y-%m-%d %H:%M:%S", time.localtime(modification_time)
+            )
+            models.append(
+                [
+                    actual_model_name,
+                    modification_time_readable,
+                    f"{all_files_sizes:.1f} {magnitude}",
+                ]
+            )
+    return models
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -156,3 +156,4 @@ vllm
 wikisql
 xcode
 xlarge
+Safetensor
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@
    `ilab config init` and can be used to specify alternative configurations for any command,
    ensuring that defaults such as taxonomy repositories and base models are honored from the global
    config.
+* `ilab model list`: a new command which lists all GGUF and Safetensor Models on the system.
 
 ### Breaking Changes
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,7 @@ issues = "https://github.com/instructlab/instructlab/issues"
 "serve" = "instructlab.model.serve:serve"
 "test" = "instructlab.model.test:test"
 "train" = "instructlab.model.train:train"
+"list" = "instructlab.model.list:model_list"
 
 [project.entry-points."instructlab.command.system"]
 "info" = "instructlab.system.info:info"
@@ -77,6 +78,7 @@ issues = "https://github.com/instructlab/instructlab/issues"
 "sysinfo" = "instructlab.system.info:info"
 "test" = "instructlab.model.test:test"
 "train" = "instructlab.model.train:train"
+"list" = "instructlab.model.list:model_list"
 
 [tool.setuptools_scm]
 version_file = "src/instructlab/_version.py"
--- a/src/instructlab/model/model.py
+++ b/src/instructlab/model/model.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# pylint: disable=redefined-builtin
 # Third Party
 import click
 
--- a/tests/test_lab.py
+++ b/tests/test_lab.py
@@ -90,6 +90,7 @@ subcommands: list[Command] = [
     Command(("model", "serve")),
     Command(("model", "test")),
     Command(("model", "train")),
+    Command(("model", "list")),
     Command(("data",), needs_config=False),
     Command(("data", "generate")),
     Command(("system",), needs_config=False),
@@ -110,6 +111,7 @@ aliases = [
     "diff",
     "generate",
     "sysinfo",
+    "list",
 ]
 
 


scripts/basic-workflow-tests.sh
```diff
@@ -148,6 +148,26 @@ test_download() {
...
+        ilab model list | grep mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf
+    else
+        step Listing the merlinite GGUF model only
+        ilab model list | grep merlinite-7b-lab-Q4_K_M.gguf4
+    fi
+    
+    # regardless, download merl safetensors to test that capability
+
+    ilab model download --repository instructlab/merlinite-7b-lab
+    ilab model list | grep instructlab/merlinite-7b-lab
```
nathan-weinberg, line 168:
```suggestion
    # irregardless, download merlinte safetensors to test that capability
    ilab model download --repository instructlab/merlinite-7b-lab
    ilab model list | grep instructlab/merlinite-7b-lab
```

src/instructlab/model/list.py
```diff
@@ -0,0 +1,110 @@
...
+    is_flag=True,
+    show_default=True,
+)
+def list(model_dirs: str, list_checkpoints: bool):
+    """lists models"""
+    t = PrettyTable(["Model Name", "Model Age", "Model Size"], hrules=ALL)
+    for dir in model_dirs:
+        for entry in Path(dir).iterdir():
+            # if file, just tally the size. This must be a GGUF.
+            if entry.is_file() and ".gguf" in entry.name:
```
nathan-weinberg, line 45: Think we should use `is_model_gguf` here: https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/backends/backends.py#L159

src/instructlab/model/list.py
```diff
@@ -0,0 +1,110 @@
...
+    show_default=True,
+)
+@click.option(
+    "--list-checkpoints",
+    type=click.BOOL,
+    default=False,
+    is_flag=True,
+    show_default=True,
+)
+def list(model_dirs: str, list_checkpoints: bool):
```
nathan-weinberg, line 39: The type hint suggest this should be passed as a string, but the default val appears to be a list?

scripts/basic-workflow-tests.sh
cdoern: keeping regardless lol

requirements.txt
```diff
@@ -45,3 +45,4 @@ wandb>=0.16.4
 # it is for CLI usage ONLY
 yamllint>=1.35.1
 referencing>=0.35.0
+prettytable>=3.10.0
```
RobotSail, line 48: What is the impact of adding this dependency - is this something that we've cleared?

src/instructlab/model/list.py
```diff
@@ -0,0 +1,111 @@
...
+            if entry.is_file() and is_model_gguf(entry):
+                # stat the gguf, add it to the table
+                stat = Path(entry.absolute()).stat(follow_symlinks=False)
+                f_size = stat.st_size
+                f_size, magnitude = convert_bytes_to_proper_mag(f_size)
+                # add to table
+                age = pandas.to_timedelta(time.time_ns() - stat.st_ctime_ns)
+                t.add_row([entry.name, age, f"{f_size} {magnitude}"])
+            elif entry.is_dir():
+                actual_model_name = ""
```
RobotSail, line 55: It seems like there's a lot happening in this `elif` branch and it's leading to a lot of visual clutter. I recommend moving this out to a secondary function since it contains a nested for loop w/ a try/catch.

src/instructlab/model/list.py
```diff
@@ -0,0 +1,111 @@
...
+                    # any lower level dir: `instructlab/granite-7b-lab/.huggingface/download.....`
+                    # so, query HF and see if its a model. For NOW only list models with HF's `ModelCard`.
+                    # if --list-checkpoints is specified, we will list all checkpoints in the checkpoints dir regardless of the ModelCard
+                    try:
+                        ModelCard.load(last_two_parts)
+                        actual_model_name = last_two_parts
+                        all_files_sizes = 0
+                        add_model = True
+                    except (ValueError, requests_exceptions.HTTPError):
+                        if list_checkpoints and dir is DEFAULTS.CHECKPOINTS_DIR:
```
RobotSail, line 78: You don't have to change this but here's a tip I like to share since it leads to cleaner branching logic:
```py
# invert the condition
if not list_checkpoints or dir is not DEFAULTS.CHECKPOINTS_DIR:
    continue

# then you can continue here in the main
logging.debug("Including model regardless of ModelCard validity")
```

RobotSail (commented):
A few things should be addressed but otherwise LGTM

src/instructlab/model/list.py
```diff
@@ -0,0 +1,132 @@
...
+    multiple=True,
+    default=[DEFAULTS.MODELS_DIR],
+    show_default=True,
+)
+@click.option(
+    "--list-checkpoints",
+    type=click.BOOL,
+    default=False,
+    is_flag=True,
+    show_default=True,
```
tiran, line 37: Type and default should not be needed for a flag
```suggestion
    is_flag=True,
```

src/instructlab/model/list.py
```diff
@@ -0,0 +1,132 @@
...
+                # add to table
+                age = pandas.to_timedelta(time.time_ns() - stat.st_ctime_ns)
+                data.append([entry.name, str(age), f"{f_size} {magnitude}"])
+            elif entry.is_dir():
+                actual_model_name = ""
+                all_files_sizes = 0
+                add_model = False
+                # walk entire dir.
+                for root, _, files in os.walk(entry.as_posix()):
+                    normalized_path = os.path.normpath(root)
```
tiran, line 59: pathlib objects have a `walk` method and `resolve` method. The following change should simplify your code. Then you can replace `parts = normalized_path.split(os.sep)` with `root.parts`.

```suggestion
                for root, _, files in entry.resolve().walk():
```

src/instructlab/model/list.py
tiran: That's an excellent suggestion. I also recommend to move if/elif block into a separate private function that returns name, age, and size. It may even a good idea to move each analyze code into its own private function, too. That would make it easier to write unit tests for each implementation.

```python
AnalyzeResult = tuple[str, str, str]
def _analyze_gguf(entry: pathlib.Path) -> AnalyzeResult:
    ...
```

src/instructlab/model/list.py
```diff
@@ -0,0 +1,132 @@
...
+    for directory in model_dirs:
+        for entry in Path(directory).iterdir():
+            # if file, just tally the size. This must be a GGUF.
+            if entry.is_file() and is_model_gguf(entry):
+                # stat the gguf, add it to the table
+                stat = Path(entry.absolute()).stat(follow_symlinks=False)
+                f_size = stat.st_size
+                f_size, magnitude = convert_bytes_to_proper_mag(f_size)
+                # add to table
+                age = pandas.to_timedelta(time.time_ns() - stat.st_ctime_ns)
```
tiran, line 51: ctime is probably the wrong thing. It's the time stamp of the last status change of the file's metadata (owner, permission, ...). mtime is the last file data modification time.

tiran (dismissed):

src/instructlab/model/list.py
cdoern: ah cool, thanks!

src/instructlab/model/list.py
cdoern: thanks for the tip! will change accordingly

cdoern: getting `AttributeError: 'PosixPath' object has no attribute 'walk'`

tiran: argh, my bad. The method was added in 3.12. We have to support 3.10 and 3.11.

src/instructlab/model/list.py
```diff
@@ -0,0 +1,139 @@
...
+        last_two_parts = os.path.join(parts[-2], parts[-1])
+        # if this is a dir it could be:
+        # top level repo dir `instructlab/`
+        # top level model dir `instructlab/granite-7b-lab`
+        # checkpoint top level dir `step-19`
+        # any lower level dir: `instructlab/granite-7b-lab/.huggingface/download.....`
+        # so, query HF and see if its a model. For NOW only list models with HF's `ModelCard`.
+        # if --list-checkpoints is specified, we will list all checkpoints in the checkpoints dir regardless of the ModelCard
+        try:
+            ModelCard.load(last_two_parts)
```
jaideepr97, line 118: what if we just reused the `is_model_safetensors()` check to only list out valid models? 
If we must use the model card method maybe that should be added to the `is_model_safetensors()` check so we have one comprehensive check for how we decide something is a valid model

jaideepr97 (commented):
Thanks for the PR @cdoern 
Please update the changelog as well since this is new functionality

src/instructlab/model/list.py
RobotSail: Wouldn't we also want to include llama-cpp models for the time being as well?

jaideepr97: @RobotSail there's a separate check for gguf models, this block only checks against model paths that are directories

cdoern: switched it

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+            return f_size, magnitude
+    return f_size, magnitude
+
+
+# print_table displays the models on the system in a nicely formatted table
+def print_table(headers: list[str], data: list[list[str]]):
+    column_widths = [
+        max(len(str(row[i])) for row in data + [headers]) for i in range(len(headers))
+    ]
+    for i, header in enumerate(headers):
```
RobotSail, line 67: The way you're printing the formatted headers is very smart. You can also optimize this loop even further by using `zip` to go through both iterators:


```suggestion
    for header, width in zip(headers, column_widths):
        print(f"{item:width}", end=" |)
    print()
```

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+        if f_size >= 1024:
+            magnitude = mag
+            f_size /= 1024
+        else:
+            return f_size, magnitude
+    return f_size, magnitude
+
+
+# print_table displays the models on the system in a nicely formatted table
+def print_table(headers: list[str], data: list[list[str]]):
```
RobotSail, line 63: I'm not sure how to suggest this big of a change and don't feel like making a separate PR for this, but I've fixed up how we're printing this so now it's printing in the format we originally wanted from PrettyTable:

```
+--------------------------------+---------------------------+-----------------------+
| Model Name                     | Model Age                 | Model Size            |
+--------------------------------+---------------------------+-----------------------+
| merlinite-7b-lab-Q4_K_M-2.gguf | 0 days 00:35:09.434472546 | 4.068468689918518 GB  |
| instructlab/granite-7b-lab     | 0 days 00:32:32.872044662 | 12.553834554739296 GB |
| merlinite-7b-lab-Q4_K_M.gguf   | 0 days 00:40:45.984725562 | 4.068468689918518 GB  |
+--------------------------------+---------------------------+-----------------------+
```


```suggestion
# print_table displays the models on the system in a nicely formatted table
def print_table(headers: list[str], data: list[list[str]]):
    column_widths = [
        max(len(str(row[i])) for row in data + [headers]) for i in range(len(headers))
    ]

    # Print separator line between headers and data
    horizontal_lines = ["-" * (width + 2) for width in column_widths]
    joining_line = "+" + "+".join(horizontal_lines) + "+"

    print(joining_line)

    outputs = []
    for header, width in zip(headers, column_widths, strict=False):
        outputs.append(f" {header:{width}} ")
    print("|" + "|".join(outputs) + "|")

    print(joining_line)

    for row in data:
        outputs = []
        for item, width in zip(row, column_widths, strict=False):
            outputs.append(f" {item:{width}} ")
        print("|" + "|".join(outputs) + "|")

    print(joining_line)```

**Review by RobotSail — APPROVED** [review]
Really awesome work on this PR Charlie, this will be a great add. 

The only thing we should do before merging is fixing the `Model Age` field and `Model Size` field so that:
- Model Size is rounded to 2 decimal points (currently we're printing full precision)
- Model Age should be renamed to "Downloaded At" because that's what it seems to be representing right now
- For the values of model age, we should change it in one of the two following ways:
  - Print out the most significant unit like how podman does, e.g.: `2 seconds ago`, `5 minutes ago`, `1 hour ago`, etc.
  - OR just print out a simple timestamp in a format like mm:dd:yy, hh:mm 

I also provided a snippet in the review to polish up the output a little bit to resemble PrettyTable. 

But I don't want to hold this PR up too much, so we should get whatever the simplest fixes are in and merge this. 

Thank you for putting this together, this is great work!

File: src/instructlab/model/list.py [hunk 13]
```diff
@@ -0,0 +1,138 @@
...
+                data.append([model, age, size])
+            elif entry.is_dir():
+                for m in _analyze_dir(entry, list_checkpoints, directory):
+                    data.append(m)
+    print_table(["Model Name", "Model Age", "Model Size"], data)
+
+
+# convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MB, GB)
+def convert_bytes_to_proper_mag(f_size: float):
+    magnitudes = ["KB", "MB", "GB"]
```
leseb, line 51:
```suggestion
    magnitudes = ["KiB", "MiB", "GiB", "TiB"]
```

Since you are using 1024 you are not printing Gygabyte but Gigibyte and so on. Let's add TiB too.

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def _analyze_gguf(entry: Path) -> AnalyzeResultGGUF:
+    # stat the gguf, add it to the table
+    stat = Path(entry.absolute()).stat(follow_symlinks=False)
+    f_size = stat.st_size
+    f_size, magnitude = convert_bytes_to_proper_mag(f_size)
+    # add to table
+    age = pandas.to_timedelta(time.time_ns() - stat.st_mtime_ns)
+    return [entry.name, str(age), f"{f_size} {magnitude}"]
```
leseb, line 91:
```suggestion
    return [entry.name, str(age), f"{f_size:.1f} {magnitude}"]
```

You don't need full precision, printing `4.1 GiB` is more than enough.

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+AnalyzeResultDir = list[list[str]]
+
+
+def _analyze_gguf(entry: Path) -> AnalyzeResultGGUF:
+    # stat the gguf, add it to the table
+    stat = Path(entry.absolute()).stat(follow_symlinks=False)
+    f_size = stat.st_size
+    f_size, magnitude = convert_bytes_to_proper_mag(f_size)
+    # add to table
+    age = pandas.to_timedelta(time.time_ns() - stat.st_mtime_ns)
```
leseb, line 90: Why do we need a nanosecond precision here? This seems quite overkill.

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+
+# Standard
+from pathlib import Path
+import logging
+import os
+import time
+
+# Third Party
+import click
+import pandas
```
leseb, line 11: Panda is not a small library, it took a few seconds for me to load it. Are we ok adding that up to the CLI?

leseb (changes requested):

src/instructlab/model/backends/backends.py
```diff
@@ -167,20 +167,24 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
...
 
-        # Convert the first four bytes to an integer
-        first_four_bytes_int = int(struct.unpack("<I", first_four_bytes)[0])
+            # Convert the first four bytes to an integer
+            first_four_bytes_int = int(struct.unpack("<I", first_four_bytes)[0])
 
-        # Close the memory-mapped file
-        mmapped_file.close()
+            # Close the memory-mapped file
+            mmapped_file.close()
```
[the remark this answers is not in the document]
jaideepr97: for me as well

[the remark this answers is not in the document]
cdoern: spacing is what It should be I believe. I have noticed this recently with GH reviews representing changes like this.

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+        # Get the last two parts and join them back into a path
+        last_two_parts = os.path.join(parts[-2], parts[-1])
+        # if this is a dir it could be:
+        # top level repo dir `instructlab/`
+        # top level model dir `instructlab/granite-7b-lab`
+        # checkpoint top level dir `step-19`
+        # any lower level dir: `instructlab/granite-7b-lab/.huggingface/download.....`
+        # so, check if model is valid Safetensor, GGUF, or list it regardless w/ `--list-checkpoints`
+        # if --list-checkpoints is specified, we will list all checkpoints in the checkpoints dir regardless of the ModelCard
+        if is_model_safetensors(Path(normalized_path)) or is_model_gguf(
```
jaideepr97, line 115: please update comment to remove ModelCard reference on 114
Also wondering if we need the `is_model_gguf` check again here? seems redundant

src/instructlab/model/list.py
cdoern: switched, also removed pandas

src/instructlab/model/list.py
cdoern: removed, sorry!

src/instructlab/model/list.py
cdoern: we need gguf check probably, there could be nested dirs.

src/instructlab/model/list.py
cdoern: thanks, incorporated

cdoern: this is REALLY nice

src/instructlab/model/backends/backends.py
```diff
@@ -167,20 +167,24 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
...
-        # Memory-map the file on the first 4 bytes (this is where the magic number is)
-        mmapped_file = mmap.mmap(f.fileno(), length=4, access=mmap.ACCESS_READ)
+    try:
+        with open(model_path, "rb") as f:
+            # Memory-map the file on the first 4 bytes (this is where the magic number is)
+            mmapped_file = mmap.mmap(f.fileno(), length=4, access=mmap.ACCESS_READ)
 
-        # Read the first 4 bytes
-        first_four_bytes = mmapped_file.read(4)
+            # Read the first 4 bytes
```
leseb, line 175: Can you make sure this is not a space/tab issue and stick with spaces? Thanks

src/instructlab/model/list.py
```diff
@@ -0,0 +1,151 @@
...
+            if entry.is_file() and is_model_gguf(entry):
+                model, age, size = _analyze_gguf(entry)
+                data.append([model, age, size])
+            elif entry.is_dir():
+                for m in _analyze_dir(entry, list_checkpoints, directory):
+                    data.append(m)
+    print_table(["Model Name", "Last Modified", "Size"], data)
+
+
+# convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MB, GB)
```
leseb, line 48:
```suggestion
# convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MiB, GiB)
```

src/instructlab/model/backends/backends.py
cdoern: when I edit the file, changing to spaces results in no changes to the file to commit, same goes for tabs, unsure how to fix how this is rendering

cdoern:
just fixed the MiB and GiB comment and pushed it up

cdoern:
*shakes fist at changelog*

src/instructlab/model/list.py
```diff
@@ -0,0 +1,151 @@
...
+    f_size, magnitude = convert_bytes_to_proper_mag(f_size)
+    # add to table
+    modification_time = os.path.getmtime(entry.absolute())
+    modification_time_readable = time.strftime(
+        "%Y-%m-%d %H:%M:%S", time.localtime(modification_time)
+    )
+    return [entry.name, modification_time_readable, f"{f_size:.1f} {magnitude}"]
+
+
+def _analyze_dir(
```
leseb, line 100: do we follow symlinks?

cdoern: nope `stat = Path(full_file).stat(follow_symlinks=False)`

I think by default `os.walk` doesn't either?

leseb: Aren't we using symlinks from cache to model dir? @RobotSail do you remember? Thanks

jaideepr97: OCI downloaded models are also symlinked from cache to model dir 
do we need to follow them to get metadata about the file, or is that stuff carried over in the symlink?

cdoern: cc @mbestavros

cdoern: speaking to oleg and mark now, they will follow up here

mbestavros: @RobotSail and I hacked a bit on model paths last week - IIRC the model loaders ran into issues when we provided symlinks to them. I'd verify that but that would be the strongest reason I think

cdoern: I think we should get this in as is, and follow up with enhancements IMO

booxter:
@cdoern I didn't realize that this work is also included in another PR and left a bunch of comments that apply here [elsewhere](https://github.com/instructlab/instructlab/pull/1861#pullrequestreview-2199749611). Would you mind checking the comments there so that I don't have to copy them verbatim? Thanks!

cdoern:
> @cdoern I didn't realize that this work is also included in another PR and left a bunch of comments that apply here [elsewhere](https://github.com/instructlab/instructlab/pull/1861#pullrequestreview-2199749611). Would you mind checking the comments there so that I don't have to copy them verbatim? Thanks!

The review on the other PR is very eleventh hour in the sense that this PR has two approvals and is about to merge. I think if anything it would make sense to follow up in a different PR and/or include these changes in @RobotSail's `ilab data list` PR, though it is a separate cmd. 

I do really appreciate the thoroughness of the review, and if you want can make an immediate follow up. We would really like to get this functionality into the beta release today

booxter:
@cdoern I can send a follow-up, sure.

scripts/basic-workflow-tests.sh
```diff
@@ -148,6 +148,25 @@ test_download() {
...
 }
 
+test_list() {
+    task List the Downloaded Models
+
+    if [ "$GRANITE" -eq 1 ]; then
+        step Listing the granite GGUF model only
+        ilab model list | grep granite-7b-lab-Q4_K_M.gguf
+    elif [ "$MIXTRAL" -eq 1 ]; then
+        step Listing the mixtral model only
```
nathan-weinberg, line 158:
```suggestion
        step Listing the mixtral GGUF model only
```

scripts/basic-workflow-tests.sh
```diff
@@ -148,6 +148,25 @@ test_download() {
...
+        ilab model list | grep mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf
+    else
+        step Listing the merlinite GGUF model only
+        ilab model list | grep merlinite-7b-lab-Q4_K_M.gguf4
+    fi
+    
+    # regardless, download merl safetensors to test that capability
```
nathan-weinberg, line 165: from my last review
```suggestion
    # regardless, download merlinite safetensors to test that capability
```

src/instructlab/model/backends/backends.py
```diff
@@ -167,20 +167,24 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
...
 
-        # Close the memory-mapped file
-        mmapped_file.close()
+            # Close the memory-mapped file
+            mmapped_file.close()
 
-        return first_four_bytes_int == GGUF_MAGIC
+            return first_four_bytes_int == GGUF_MAGIC
+    except IsADirectoryError as exc:
+        logger.debug(f"GGUF Path is a directory, returning {exc}")
```
nathan-weinberg, line 186:
```suggestion
        logger.debug(f"'model_path' is a directory, returning {exc}")
```

nathan-weinberg (approved):
Couple nits but otherwise LGTM

mairin:
@tiran Just a note I gave Charlie the ok to remove request for review; the requests were addressed as asked and a while had passed and I felt we had to get into today's beta because it was blocking another PR. My apologies if I gave bad guidance on this.

PR #1841, opened 2024-07-23
ci: update basic-workflow-tests.sh to work with vllm

alimaredia:
**Checklist:**

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euf
+set -xeuf
 
 # This is a basic workflow test of the ilab CLI.
 #
@@ -25,6 +25,7 @@ HF_TOKEN=${HF_TOKEN:-}
 SDG_PIPELINE="simple"
 SKIP_TRAIN=${SKIP_TRAIN:-0}
 EVAL=0
+MIXTRAL_GPTQ_MODEL="TheBloke/Mixtral-8x7b-Instruct-v0.1-GPTQ"
 
 export GREP_COLORS='mt=1;33'
 BOLD='\033[1m'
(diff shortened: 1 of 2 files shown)


alimaredia:
Blocked by: https://github.com/instructlab/instructlab/pull/1839

scripts/basic-workflow-tests.sh
```diff
@@ -120,8 +121,8 @@ test_init() {
         ilab config init --non-interactive
     fi
     step Checking config.yaml
-    if [ "${MIXTRAL}" -eq 1 ]; then
-        sed -i -e 's|merlinite.*|mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf|' "${CONFIG_HOME}/instructlab/config.yaml"
+    if [ "${MIXTRAL}" -eq 1 ] && [ "${BACKEND}" == "vllm" ]; then
```
booxter, line 124: `set_defaults` seems to guarantee that it's never `!= "vllm"`. Wonder if we have to check it everywhere below.

scripts/basic-workflow-tests.sh
```diff
@@ -120,8 +121,8 @@ test_init() {
...
+        sed -i -e 's|merlinite.*|${MIXTRAL_GPTQ_MODEL}|' "${CONFIG_HOME}/instructlab/config.yaml"
```
booxter, line 125: This won't work: `''` won't expand the variable

booxter (changes requested):

booxter (approved):
LGTM. Some of `if MIXTRAL && vllm` checks could probably be just `if MIXTRAL` because `set_defaults` guarantee that it's never one but not another.

alimaredia:
@booxter When I started this line of work I had vllm and non-vllm cases when flags for granite, mixtral, and neither were passed in. When Jason and I talked yesterday we decided that this was overkill.

The reason I left the `if Mixtral && vllm` checks were just so that it was obvious to someone new to the script that the two are intertwined. That being said I removed them, since they were somewhat of a relic of past work I'd done.

booxter (commented):
Code-wise, looks correct. Which CI job utilizes the workflow?

leseb (approved):

leseb:
I tried to run the CI job manually but it failed [here](https://github.com/instructlab/instructlab/actions/runs/10074501141/job/27850757333)

Issue #1800, opened 2024-07-19
[regression] ilab no longer supports multiple work profiles and config files


tiran:
**Describe the bug**
The ticket is formulated like an enhancement, but I'm flagging this as a bug, because it is a regression from 0.17 and earlier versions. Multiple work profiles was a super useful feature for testing and presenting InstructLab to customers and users.

As a user, I want to use multiple `config.yaml` and work profiles, so I can several independent work profiles for InstructLab. This enables me to have several independent configurations, checkouts of taxonomy, training data, and models on my machine at the same time.

Until PR #1471 landed, this was very easy. The config file and all other files (models, taxonomy, training data) were references in relation to the current working directory. To create or use a new, independent work profile, I merely had to `cd` into another directory. With the new storage system, the feature is gone.

- `ilab config init` has no option to set a different config file. It is always creating its config file in an hard-coded path. `ilab --config=myconfig.yaml config init` does not change the path to default conf

- On Linux, it is possible to hack around the problem by setting three env vars. But that's awkward and does not work on macOS or Windows. `platformdirs` does not read `os.environ` for any platform except for Linux. Even on Linux it is problematic, because the `XDG` env vars affect other program. That's why users don't want to set the env vars in their current shell. There are other problems, e.g. taxonomy checkout uses `git` command and the `git` command reads its configuration from XDG directories.

   ```
   XDG_CACHE_HOME=$HOME/instructlab/profile1/cache \
   XDG_CONFIG_HOME=$HOME/instructlab/profile1/config \
   XDG_DATA_HOME=$HOME/instructlab/profile1/data \
   XDG_STATE_HOME=$HOME/instructlab/profile1/state \
   ilab config init
   ```

**To Reproduce**
See above

**Expected behavior**
InstructLab should reintroduce a way to have separate configuration and work profiles. Switching profiles should be easy, e.g. with a command or env var.

**Device Info (please complete the following information):**
 - 0.18.0a3.dev10

jaideepr97:
@cdoern 
will your work in #1399 also include creating separate workspace dirs for different profiles? if so that might address this issue 
also related: https://github.com/instructlab/instructlab/issues/729

ktam3:
@cdoern - check this

PR #1842, opened 2024-07-23
Add support for an `ilab` global config

cdoern:
When the env var `ILAB_GLOBAL_CONFIG` is set, that config is used as the
template to render the ~/.config/instructlab/config.yaml user space
config. The interactive mode is skipped when calling `ilab config init`.
This works at the `ilab` main entrypoint, this can be used to specify a
different configuration file too when running any commands.

This ensures defaults like taxonomy repos, base models, etc, are honored
from the global config.

Co-authored-by: Sébastien Han <seb@redhat.com>
Signed-off-by: Charlie Doern <cdoern@redhat.com>

**Checklist:**

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -3,6 +3,8 @@
 # Standard
 from os import listdir
 from os.path import dirname, exists
+import pathlib
+import typing
 
 # Third Party
 from git import GitError, Repo
@@ -12,6 +14,7 @@ import click
 from instructlab import clickext, utils
 from instructlab.configuration import (
     DEFAULTS,
+    Config,
     ensure_storage_directories_exist,
     get_default_config,
     read_train_profile,
@@ -61,7 +64,9 @@ from instructlab.configuration import (
 )
 @click.option("--train-profile", type=click.Path(), default=None)
 @clickext.display_params
+@click.pass_context
 def init(
+    ctx,
     interactive,
     model_path: str,
     taxonomy_path,
@@ -72,6 +77,70 @@ def init(
 ):
     """Initializes environment for InstructLab"""
     ensure_storage_directories_exist()
+    # If the caller used `ilab init` then the parent command is `ilab`, so we need to get the
+    # parameter source from the parent command.
+    if ctx.parent.command.name == "ilab":
+        param_source = ctx.parent.get_parameter_source("config_file")
+    # if the caller used `ilab config init` then the parent command is `config`, so we need to get
+    # the parameter source from the parent of the parent command.
+    else:
+        param_source = ctx.parent.parent.get_parameter_source("config_file")
+
+    if param_source == click.core.ParameterSource.ENVIRONMENT:
+        model_path, taxonomy_path, taxonomy_base, cfg = get_params_from_env(ctx.obj)
+    else:
+        try:
+            model_path, taxonomy_path, cfg = get_params_default(
+                interactive,
+                repository,
+                min_taxonomy,
+                model_path,
+                taxonomy_path,
+            )
+        except click.exceptions.Exit as e:
+            ctx.exit(e.exit_code)
+
+    click.echo(f"Generating `{DEFAULTS.CONFIG_FILE}`...")
+    if train_profile is not None:
+        cfg.train = read_train_profile(train_profile)
+
+    cfg.chat.model = model_path
+    cfg.generate.model = model_path
+    cfg.serve.model_path = model_path
+    cfg.generate.taxonomy_path = taxonomy_path
+    cfg.generate.taxonomy_base = taxonomy_base
+    cfg.evaluate.model = model_path
+    cfg.evaluate.mt_bench_branch.taxonomy_path = taxonomy_path
+    write_config(cfg)
+
+    click.secho(
+        "Initialization completed successfully, you're ready to start using `ilab`. Enjoy!",
+        fg="green",
+    )
+
+
+def get_params_from_env(
+    obj: typing.Optional[typing.Any],
+) -> typing.Tuple[str, str, str, Config]:
+    if obj is None or not hasattr(obj, "config"):
+        raise ValueError("obj must not be None and must have a 'config' attribute")
+
+    return (
+        obj.config.serve.model_path,
+        obj.config.generate.taxonomy_path,
+        obj.config.generate.taxonomy_base,
+        obj.config,
+    )
+
+
+def get_params_default(
+    interactive: bool,
+    repository: str,
+    min_taxonomy: bool,
+    model_path: str,
+    taxonomy_path: pathlib.Path,
+) -> typing.Tuple[str, pathlib.Path, Config]:
+    cfg = get_default_config()
     clone_taxonomy_repo = True
     if interactive:
         if exists(DEFAULTS.CONFIG_FILE):
@@ -79,7 +148,7 @@ def init(
                 f"Found {DEFAULTS.CONFIG_FILE} in the current directory, do you still want to continue?"
             )
             if not overwrite:
-                return
+                raise click.exceptions.Exit(0)
         click.echo(
             "Welcome to InstructLab CLI. This guide will help you to setup your environment."
         )
@@ -125,20 +194,5 @@ def init(
         model_path = utils.expand_path(
             click.prompt("Path to your model", default=model_path)
         )
-    click.echo(f"Generating `{DEFAULTS.CONFIG_FILE}`...")
-    cfg = get_default_config()
-    if train_profile is not None:
-        cfg.train = read_train_profile(train_profile)
-    cfg.chat.model = model_path
-    cfg.generate.model = model_path
-    cfg.serve.model_path = model_path
-    cfg.generate.taxonomy_path = taxonomy_path
-    cfg.generate.taxonomy_base = taxonomy_base
-    cfg.evaluate.model = model_path
-    cfg.evaluate.mt_bench_branch.taxonomy_path = taxonomy_path
-    write_config(cfg)
 
-    click.secho(
-        "Initialization completed successfully, you're ready to start using `ilab`. Enjoy!",
-        fg="green",
-    )
+    return model_path, taxonomy_path, cfg
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -62,6 +62,11 @@ class _InstructlabDefaults:
     # define static defaults up here
     API_KEY = "no_api_key"
 
+    # ILAB_GLOBAL_CONFIG is the environment variable that can be used to override the default config
+    # file. When set, the CLI will use the file specified in the environment variable as a sample to
+    # generate the default config file.
+    ILAB_GLOBAL_CONFIG = "ILAB_GLOBAL_CONFIG"
+
     # TODO: Consolidate --model and --model-path into one --model-path flag since we always need a path now
     MODEL_NAME_OLD = "merlinite-7b-lab-Q4_K_M"
     MERLINITE_GGUF_REPO = "instructlab/merlinite-7b-lab-GGUF"
--- a/tests/test_lab_config.py
+++ b/tests/test_lab_config.py
@@ -1,7 +1,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# Standard
+import logging
+import pathlib
+
 # Third Party
 from click.testing import CliRunner
+import pytest
 import yaml
 
 # First Party
@@ -16,3 +21,40 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
     assert parsed
 
     assert configuration.Config(**parsed)
+
+
+@pytest.mark.parametrize(
+    "command",
+    [
+        (["config", "init"]),
+        (
+            ["init"]
+        ),  # TODO: remove this test once the deprecated alias 'ilab init' is removed
+    ],
+)
+def test_ilab_config_init_with_env_var_config(
+    cli_runner: CliRunner, tmp_path: pathlib.Path, command: list
+) -> None:
+    # Common setup code
+    cfg = configuration.get_default_config()
+    assert cfg.general.log_level == logging.getLevelName(logging.INFO)
+    cfg.general.log_level = logging.getLevelName(logging.DEBUG)
+    cfg_file = tmp_path / "config-gold.yaml"
+    with cfg_file.open("w") as f:
+        yaml.dump(cfg.model_dump(), f)
+
+    # Invoke the CLI command
+    result = cli_runner.invoke(
+        lab.ilab, command, env={"ILAB_GLOBAL_CONFIG": cfg_file.as_posix()}
+    )
+    assert result.exit_code == 0, result.stdout
+
+    # Load and check the generated config file
+    config_path = pathlib.Path(configuration.DEFAULTS.CONFIG_FILE)
+    assert config_path.exists()
+    with config_path.open(encoding="utf-8") as f:
+        parsed = yaml.safe_load(f)
+    assert parsed
+    assert configuration.Config(**parsed).general.log_level == logging.getLevelName(
+        logging.DEBUG
+    )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,12 @@
 * The `generate` section of the config now has a `teacher` section. This section configures the
   teacher model when it is automatically served in the background. This new section has the same
   values as the `serve` section of the config.
+* Support for `ILAB_GLOBAL_CONFIG` environment variable: When set, this environment variable
+   specifies a global configuration file that serves as the template for the
+   `~/.config/instructlab/config.yaml` user space config. This bypasses the interactive mode in
+   `ilab config init` and can be used to specify alternative configurations for any command,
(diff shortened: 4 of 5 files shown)


leseb:
I don't think this can work until https://github.com/instructlab/instructlab/issues/1800 is resolved.

src/instructlab/config/init.py
```diff
@@ -72,61 +75,71 @@ def init(
...
-    if interactive:
-        if exists(DEFAULTS.CONFIG_FILE):
-            overwrite = click.confirm(
-                f"Found {DEFAULTS.CONFIG_FILE} in the current directory, do you still want to continue?"
+    cfg = None
+    if os.environ.get(DEFAULTS.ILAB_GLOBAL_CONFIG) is not None:
+        model_path = ctx.obj.config.serve.model_path
+        taxonomy_path = ctx.obj.config.generate.taxonomy_path
+        taxonomy_base = ctx.obj.config.generate.taxonomy_base
+        cfg = ctx.obj.config
```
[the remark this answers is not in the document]
leseb: It's working for me, I'm going to push an update to this branch now.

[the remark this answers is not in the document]
jaideepr97: are these the only fields we care about copying over? It's still not working for me but that might be because I'm editing fields like `num_cpu` since my undersanding is we want to take the entire config as is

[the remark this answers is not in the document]
jaideepr97: nvm, I suppose the entire golden config should be getting stored in the `ctx` obj so I'm not sure why my changes aren't getting reflected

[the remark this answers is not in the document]
leseb: `ILAB_GLOBAL_CONFIG=config-leseb.yaml ilab config init` works for me, for now I just edited`general.log_level` and it's reflected in the generated config file.

I'm writing test!

src/instructlab/config/init.py
```diff
@@ -72,61 +74,80 @@ def init(
...
-            "Please provide the following values to initiate the "
-            "environment [press Enter for defaults]:"
-        )
-        taxonomy_path = utils.expand_path(
-            click.prompt("Path to taxonomy repo", default=taxonomy_path)
-        )
+    cfg = None
+    # If the caller used `ilab init` then the parent command is `ilab`, so we need to get the
+    # parameter source from the parent command. If the caller used `ilab config init` then the
+    if ctx.parent.command.name == "ilab":
```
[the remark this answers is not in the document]
leseb: correct!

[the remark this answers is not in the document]
cdoern: also I think after speaking w/ @mairin we want running something like `ilab model chat` to just _work_ if you have the `ILAB_GLOBAL_CONFIG` env var set for those types of users. So technically we will tell them to run `ilab config init` but if they skip it, it will pull from the golden cfg? 

This may be confusing though

RobotSail (approved):
LGTM

jaideepr97 (approved):

booxter:
I don't know if there was a design decision already made  somewhere to implement it in this particular way. (If so, it would help to see a link to where it was discussed.) So the comment may or may not be helpful. Thanks for consideration anyway.

---

A common "golden config" approach used by traditional Unix tools and elsewhere is:

1. put a "golden" config somewhere in `/usr/share/`, e.g. installed with a package;
2. let the system admin to adjust it in `/etc/` by **overriding** options from the golden config defaults; (meaning: listing changes only)
3. let user to adjust it in `$HOME/.local/config/` by **overriding** options from the both the golden and system-admin configs. (also listing changes only).
4. finally, apply the in-app (embedded) defaults if no prior config overrode them.

This enables a hierarchy of decisions (from packager to system admin to user). This also allows to rely on defaults defined in the golden config as well as in the app.

---

AFAIU here the approach is to have the "golden" config **copied** into `$HOME`, meaning that any changes to system-level / package defaults won't be reflected in the user config. This may an issue in the future when a packager would like to propagate a better setting to users automatically.

I wonder if we could **merge** several configs in runtime instead of copying it once on `init`.

---

Let me know what you think.

bjhargrave:
> 3\. let user to adjust it in `$HOME/.local/config/`

Do you mean `$XDG_CONFIG_HOME` (defaults to `$HOME/.config/`) ?

booxter:
@bjhargrave Yes I meant the XDG thingy. 👍

jaideepr97:
This change could probably also use a changelog update

src/instructlab/config/init.py
```diff
@@ -72,61 +74,80 @@ def init(
...
-    elif interactive:
-        clone_taxonomy_repo = click.confirm(
-            f"`{taxonomy_path}` seems to not exist or is empty. Should I clone {repository} for you?"
-        )
+    if param_source == click.core.ParameterSource.ENVIRONMENT:
+        model_path = ctx.obj.config.serve.model_path
+        taxonomy_path = ctx.obj.config.generate.taxonomy_path
+        taxonomy_base = ctx.obj.config.generate.taxonomy_base
+        cfg = ctx.obj.config
+    else:
```
booxter, line 92: (No action required) This function becomes a bit too long to my taste. I'd suggest to split out the paths for golden config env (lines 88-91) and "default" (lines 93-147) into separate functions, and call them here. You'd have:

```
if param_source == ...:
    model_path, taxonomy_path, taxonomy_base, cfg = get_params_from_env(cfg)
else:
    model_path, taxonomy_path, taxonomy_base, cfg = get_params_default(cfg, ...)

...

if cfg is None:
    ...
```

It would be somewhat easier to consume the code then.

src/instructlab/config/init.py
```diff
@@ -72,61 +74,80 @@ def init(
...
-        clone_taxonomy_repo = False
-    elif interactive:
-        clone_taxonomy_repo = click.confirm(
-            f"`{taxonomy_path}` seems to not exist or is empty. Should I clone {repository} for you?"
-        )
+    if param_source == click.core.ParameterSource.ENVIRONMENT:
+        model_path = ctx.obj.config.serve.model_path
+        taxonomy_path = ctx.obj.config.generate.taxonomy_path
+        taxonomy_base = ctx.obj.config.generate.taxonomy_base
+        cfg = ctx.obj.config
```
booxter, line 91: I [think](https://github.com/instructlab/instructlab/pull/1847#discussion_r1690148608) this is never a `None`, so you could probably move `cfg = get_default_config()` from line 150 under the `else` branch for "default" (non-golden-env) case.

src/instructlab/configuration.py
```diff
@@ -609,7 +614,7 @@ def init(
     config_obj: Config
     error_msg: str | None = None
     ensure_storage_directories_exist()
-    if config_file == "DEFAULT":
+    if config_file == "DEFAULT" or config_file is None:
```
booxter, line 617: when is `config_file is None`? This looks like it comes from the click option that has the default value set to a string:

```
@click.option(
    "--config",
    "config_file",
    type=click.Path(),
    default=cfg.DEFAULTS.CONFIG_FILE,
    envvar=cfg.DEFAULTS.ILAB_GLOBAL_CONFIG,
    show_default=True,
    help="Path to a configuration file.",
)
```

I reverted the change to see which test case fails, but none does. Is this needed?

tests/test_lab_config.py
```diff
@@ -16,3 +20,93 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
 
     assert configuration.Config(**parsed)
+
+
+def test_ilab_config_init_with_env_var_config(
+    cli_runner: CliRunner, tmp_path: pathlib.Path
+) -> None:
+    # create a dummy config file
+    cfg = configuration.get_default_config()
+    assert cfg.general.log_level == "INFO"
```
booxter, line 30: nit: There are `logging` module symbols for all levels that you could use here instead of strings.

tests/test_lab_config.py
```diff
@@ -16,3 +20,93 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+    cfg.general.log_level = "DEBUG"
+    cfg_file = tmp_path / "config.yaml"
+    with cfg_file.open("w") as f:
+        yaml.dump(cfg.model_dump(), f)
+
+    result = cli_runner.invoke(
+        lab.ilab, ["config", "init"], env={"ILAB_GLOBAL_CONFIG": cfg_file.as_posix()}
+    )
+    assert result.exit_code == 0, result.stderr
```
booxter, line 39: I think our CliRunner does not capture stderr (mix_stderr=True by default). (I think this should be changed, but that's beyond the scope for this PR.)

But I think you actually care to see the stdout if it fails?

tests/test_lab_config.py
```diff
@@ -16,3 +20,93 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+
+    # the following feels ugly, if someone has a better way to extract the path from the output
```
booxter, line 41: I probably miss something, but why do you even need to extract the path from the output? Can't you just read the file from the CONFIG_DIR?

```
extracted_path = pathlib.Path(configuration.DEFAULTS.CONFIG_FILE)
```

This whole block seems unnecessary.

tests/test_lab_config.py
```diff
@@ -16,3 +20,93 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
     assert configuration.Config(**parsed)
+
+
+def test_ilab_config_init_with_env_var_config(
+    cli_runner: CliRunner, tmp_path: pathlib.Path
+) -> None:
+    # create a dummy config file
+    cfg = configuration.get_default_config()
+    assert cfg.general.log_level == "INFO"
+    cfg.general.log_level = "DEBUG"
```
booxter, line 31: (No action required) This test assumes that the default is NOT "debug". Could we pick some other setting in the config that surely won't ever have our custom value as the default? I dunno, set `chat.max_tokens` to `123456`. (This will also allow us to NOT import `logging` to access the `DEBUG` symbol.)

tests/test_lab_config.py
```diff
@@ -16,3 +20,93 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+    assert (
+        extracted is not None
+    ), "No path containing 'config.yaml' was found in matches"
+
+    # load the generated config file
+    extracted_path = pathlib.Path(extracted)
+    assert extracted_path.exists()
+    with extracted_path.open(encoding="utf-8") as f:
+        parsed = yaml.safe_load(f)
+        assert parsed
```
booxter, line 60: nitty-nit: unindent since you no longer need to hold the file descriptor after line 59.

tests/test_lab_config.py
```diff
@@ -16,3 +20,93 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+
+        # check that the log level was set to DEBUG
+        assert configuration.Config(**parsed).general.log_level == "DEBUG"
+
+    # remove the file
```
booxter, line 65: I assume that `tmp_path` fixture will remove the directory?

tests/test_lab_config.py
```diff
@@ -16,3 +20,93 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+    # create a dummy config file
+    cfg = configuration.get_default_config()
+    assert cfg.general.log_level == "INFO"
+    cfg.general.log_level = "DEBUG"
+    cfg_file = tmp_path / "config.yaml"
+    with cfg_file.open("w") as f:
+        yaml.dump(cfg.model_dump(), f)
+
+    result = cli_runner.invoke(
+        lab.ilab, ["init"], env={"ILAB_GLOBAL_CONFIG": cfg_file.as_posix()}
```
booxter, line 82: so what's the difference between the two cases? This line only? Can we make these into two variants of the same base test case? This is very redundant.

booxter (changes requested):
The test cases need some love. Otherwise, the PR does what it promises to do and can probably go forward. Though I think it's not the best approach to manage config files overall (see the other comment). (That said, an alternative can probably be adopted later with relatively minor pain regardless whether this PR merges or not.)

To reiterate, once test cases are cleaned up, I am ok with merging it.

tests/test_lab_config.py
booxter: Pff. Nevermind, I just noticed the `assert INFO` above. Then at least the test case won't silently become no-op if we ever change the default. But the other argument stands I think.

src/instructlab/config/init.py
```diff
@@ -72,61 +74,80 @@ def init(
...
-        click.echo(
-            "Please provide the following values to initiate the "
-            "environment [press Enter for defaults]:"
-        )
-        taxonomy_path = utils.expand_path(
-            click.prompt("Path to taxonomy repo", default=taxonomy_path)
-        )
+    cfg = None
+    # If the caller used `ilab init` then the parent command is `ilab`, so we need to get the
+    # parameter source from the parent command. If the caller used `ilab config init` then the
```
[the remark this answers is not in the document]
leseb: Sorry this belongs below actually. Thanks.

src/instructlab/configuration.py
leseb: It was None because at one point the default got removed. Good catch, thanks.

tests/test_lab_config.py
leseb: I forgot that the file gets deleted after the third run or something.

booxter:
@leseb I will report a RFE issue later to revisit configuration procedure (if it's not already reported somewhere - will check). This PR can stand on its own as a short term solution.

leseb:
> @leseb I will report a RFE issue later to revisit configuration procedure (if it's not already reported somewhere - will check). This PR can stand on its own as a short term solution.

Thanks, I'll remove the 'hold' label and let the bot merge it then.

src/instructlab/config/init.py
```diff
@@ -72,14 +77,83 @@ def init(
...
+    click.secho(
+        "Initialization completed successfully, you're ready to start using `ilab`. Enjoy!",
+        fg="green",
+    )
+
+
+def get_params_from_env(
+    obj: typing.Optional[typing.Any],
+) -> typing.Tuple[str, str, str, Config]:
+    if obj is None or not hasattr(obj, "config"):
```
booxter, line 126: methinks it's always non-None and it always has config?

src/instructlab/config/init.py
```diff
@@ -72,14 +77,83 @@ def init(
...
         if exists(DEFAULTS.CONFIG_FILE):
             overwrite = click.confirm(
                 f"Found {DEFAULTS.CONFIG_FILE} in the current directory, do you still want to continue?"
             )
             if not overwrite:
-                return
+                return (
+                    "",
+                    pathlib.Path("."),
+                    get_default_config(),
```
booxter, line 155: You could return `cfg` since you already have it. That said, I think you could raise an exception and catch it outside as an indication of exit state, instead of reusing return value for indication about this except-ional situation.

tests/test_lab_config.py
```diff
@@ -16,3 +20,44 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+    assert config_path.exists()
+    with config_path.open(encoding="utf-8") as f:
+        parsed = yaml.safe_load(f)
+    assert parsed
+    assert configuration.Config(**parsed).general.log_level == logging.getLevelName(
+        logging.DEBUG
+    )
+
+
+def test_ilab_config_init_with_env_var_config(
```
booxter, line 53: pytest parametrize could probably be used here.

booxter (commented):
Minor things that shouldn't block the PR.

src/instructlab/config/init.py
booxter: Is it because the `typing.Optional` used for the parameter?

src/instructlab/config/init.py
leseb: Addressed, both.

tests/test_lab_config.py
leseb: Done.

src/instructlab/config/init.py
leseb: Yes. I think the error was:

```
src/instructlab/config/init.py:125: error: Item "None" of "Any | None" has no attribute "config"
```

tests/test_lab_config.py
```diff
@@ -16,3 +21,46 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+        (["config", "init"]),
+        (
+            ["init"]
+        ),  # TODO: remove this test once the deprecated alias 'ilab init' is removed
+    ],
+)
+def test_ilab_config_init_with_env_var_config(
+    cli_runner: CliRunner, tmp_path: pathlib.Path, command
+) -> None:
+    setup_and_invoke_test(cli_runner, tmp_path, command)
```
booxter, line 66: nit: at this point, having a separate `setup_and_invoke_test` function is redundant and you could merge the body here.

booxter (approved):

tests/test_lab_config.py
leseb: argh, right!

booxter (approved):
Piece of cake. I will check the mypy thingy after this merges. Kudos!

booxter (approved):

PR #1843, opened 2024-07-23
Add autodetection and vllm arg adjustment for bitsandbytes quantization

Maxusmusti:
When loading a bitsandbytes quantized model with vllm, we need to ensure that the `--quantization` and `--load-format` args are appropriately set to `bitsandbytes`, otherwise the model cannot be loaded. This PR introduces the logic to check whether a model is vllm quantized and adjust the `vllm_args` accordingly.

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/vllm.py
+++ b/src/instructlab/model/backends/vllm.py
@@ -285,12 +285,15 @@ def build_vllm_cmd(
     # Auto-detect whether model is quantized w/ bitsandbytes and add potentially missing args
     quant_arg_present = contains_argument("--quantization", vllm_args)
     load_arg_present = contains_argument("--load-format", vllm_args)
-    if not quant_arg_present or not load_arg_present:
+    eager_arg_present = contains_argument("--enforce-eager", vllm_args)
+    if not (quant_arg_present and load_arg_present and eager_arg_present):
         if is_bnb_quantized(model_path):
             if not quant_arg_present:
                 vllm_cmd.extend(["--quantization", "bitsandbytes"])
             if not load_arg_present:
                 vllm_cmd.extend(["--load-format", "bitsandbytes"])
+            if not eager_arg_present:
+                vllm_cmd.append("--enforce-eager")
 
     # Force multiprocessing for distributed serving, vLLM will try "ray" if it's installed but we do
     # not support it (yet?). We don't install Ray but we might end up running on systems that have it,
--- a/src/instructlab/model/backends/vllm.py
+++ b/src/instructlab/model/backends/vllm.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+import json
 import logging
 import os
 import pathlib
@@ -202,6 +203,29 @@ def run_vllm(
     return vllm_process, tmp_files
 
 
+def is_bnb_quantized(model_path: pathlib.Path) -> bool:
+    """
+    Check if provided model has quantization config with bitsandbytes specified.
+
+    Args:
+        model_path    (Path):         The path to the model files
+
+    Returns:
+        bool: Whether or not the model has been quantized via bitsandbytes
+    """
+    config_json = model_path / "config.json"
+    if not model_path.is_dir() or not config_json.is_file():
+        return False
+
+    with config_json.open(encoding="utf-8") as f:
+        config = json.load(f)
+
+    return (
+        "quantization_config" in config
+        and config["quantization_config"].get("quant_method", "") == "bitsandbytes"
+    )
+
+
 def build_vllm_cmd(
     host: str,
     port: int,
@@ -258,6 +282,13 @@ def build_vllm_cmd(
             path = pathlib.Path(chat_template)
             verify_template_exists(path)
             vllm_cmd.extend(["--chat-template", chat_template])
+
+    # Auto-detect whether model is quantized w/ bitsandbytes and add potentially missing args
+    if is_bnb_quantized(model_path):
+        if not contains_argument("--quantization", vllm_args):
+            vllm_cmd.extend(["--quantization", "bitsandbytes"])
+        if not contains_argument("--load-format", vllm_args):
+            vllm_cmd.extend(["--load-format", "bitsandbytes"])
 
     # Force multiprocessing for distributed serving, vLLM will try "ray" if it's installed but we do
     # not support it (yet?). We don't install Ray but we might end up running on systems that have it,
--- a/src/instructlab/model/backends/vllm.py
+++ b/src/instructlab/model/backends/vllm.py
@@ -220,9 +220,8 @@ def is_bnb_quantized(model_path: pathlib.Path) -> bool:
     with config_json.open(encoding="utf-8") as f:
         config = json.load(f)
 
-    return (
-        "quantization_config" in config
-        and config["quantization_config"].get("quant_method", "") == "bitsandbytes"
+    return bool(
+        config.get("quantization_config", {}).get("quant_method", "") == "bitsandbytes"
     )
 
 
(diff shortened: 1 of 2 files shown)


src/instructlab/model/backends/vllm.py
```diff
@@ -202,6 +203,25 @@ def run_vllm(
...
+
+    Returns:
+        bool: Whether or not the model has been quantized via bitsandbytes
+    """
+    with open(os.path.join(os.fspath(model_path), "config.json"), "r") as config_json:
+        config = json.load(config_json)
+
+    return (
+        "quantization_config" in config
+        and config["quantization_config"]["quant_method"] == "bitsandbytes"
```
RobotSail, line 221: Change this line to use `.get` in case it's not there for some reason:
```suggestion
        and config["quantization_config"].get("quant_method", "") == "bitsandbytes"
```

RobotSail (approved):
LGTM

RobotSail:
Tested this locally both with the `ilab model serve` as well as directly running vLLM serve and it worked in both instances. The checkpoints used were the quantized LoRA checkpoints that we generated via the training library using bitsandbytes quantization.

RobotSail:
To test this PR locally, you must update the `requirements-vllm-cuda.txt` file to contain the following:

```

vllm @git+https://github.com/bd-iaas-us/vllm@nf5 ; sys_platform == 'linux' and platform_machine == 'x86_64'
```

And rebuild it by running the following command:
```
CUDACXX=/usr/local/cuda-12/bin/nvcc  DCMAKE_CUDA_ARCHITECTURES="native" FORCE_CMAKE=1 pip install -r requirements-vllm-cuda.txt
```

RobotSail:
For this to work out-of-the-box, we're currently relying on this PR to be merged into upstream vLLM: https://github.com/vllm-project/vllm/pull/5753

src/instructlab/model/backends/vllm.py
```diff
@@ -202,6 +203,30 @@ def run_vllm(
...
+    """
+    Check if provided model has quantization config with bitsandbytes specified.
+
+    Args:
+        model_path    (Path):         The path to the model files
+
+    Returns:
+        bool: Whether or not the model has been quantized via bitsandbytes
+    """
+    if not model_path.is_dir() or "config.json" not in os.listdir(model_path):
```
tiran, line 216:
```suggestion
    config_json = model_path / "config.json"
    if not model_path.is_dir() or not config_json.is_file():
```

src/instructlab/model/backends/vllm.py
```diff
@@ -202,6 +203,30 @@ def run_vllm(
...
+        return False
+
+    with open(
+        os.path.join(os.fspath(model_path), "config.json"), "r", encoding="utf-8"
+    ) as config_json:
+        config = json.load(config_json)
```
tiran, line 222:
```suggestion
    with config_json.open(encoding="utf-8") as f:
        config = json.load(f)
```

tiran (dismissed):
Use the force of pathlib, Luke!

leseb (changes requested):
Please add a new unit test for `build_vllm_cmd` to capture this scenario. Thanks

src/instructlab/model/backends/vllm.py
```diff
@@ -259,6 +283,13 @@ def build_vllm_cmd(
             verify_template_exists(path)
             vllm_cmd.extend(["--chat-template", chat_template])
 
+    # Auto-detect whether model is quantized w/ bitsandbytes and add potentially missing args
+    if is_bnb_quantized(model_path):
+        if not contains_argument("--quantization", vllm_args):
+            vllm_cmd.extend(["--quantization", "bitsandbytes"])
+        if not contains_argument("--load-format", vllm_args):
+            vllm_cmd.extend(["--load-format", "bitsandbytes"])
+
```
leseb, line 292: I think these should be called before `is_bnb_quantized` so that we don't call `is_bnb_quantized` (and avoid loading the json etc) if the arguments already contain `"--quantization"` and `"--load-format"`.

src/instructlab/model/backends/vllm.py
```diff
@@ -202,6 +203,29 @@ def run_vllm(
...
+    if not model_path.is_dir() or not config_json.is_file():
+        return False
+
+    with config_json.open(encoding="utf-8") as f:
+        config = json.load(f)
+
+    return (
+        "quantization_config" in config
+        and config["quantization_config"].get("quant_method", "") == "bitsandbytes"
+    )
```
leseb, line 226:
```suggestion
    return config.get("quantization_config", {}).get("quant_method", "") == "bitsandbytes"
```

leseb (changes requested):

leseb (approved):

n1hility (approved):

src/instructlab/model/backends/vllm.py
```diff
@@ -259,6 +282,19 @@ def build_vllm_cmd(
             verify_template_exists(path)
             vllm_cmd.extend(["--chat-template", chat_template])
 
+    # Auto-detect whether model is quantized w/ bitsandbytes and add potentially missing args
+    quant_arg_present = contains_argument("--quantization", vllm_args)
+    load_arg_present = contains_argument("--load-format", vllm_args)
+    eager_arg_present = contains_argument("--enforce-eager", vllm_args)
```
leseb, line 288: Does this have any relationship with `--quantization` and `--load-format` or is it independent? If not, I'd extract it from the condition. Regardless I'd also add a comment on what the flag does and why it is needed.

Thanks!

leseb (commented):
Small clarification nit, thanks

leseb:
Just hold until this https://github.com/instructlab/instructlab/pull/1843#discussion_r1692587957 is resolved. Thanks

src/instructlab/model/backends/vllm.py
Maxusmusti: In this case they are related. The context is that currently in vllm, there is a bug with 4-bit bnb models (specifically with cuda graphs) that causes generation output to be much worse. To get around this, the solution is to include the `--enforce-eager` flag, which will restore our generation quality. We only want to auto-enable this flag for the 4-bit bnb models, though, as in every other case it provides no benefit, and can actually be a bit slower.

tests/test_backends.py
```diff
@@ -21,7 +21,11 @@
...
 ):
-    test_json_dict = {"a": 1, "b": 2}
+    test_json_dict = {
+        "a": 1,
+        "b": 2,
+        "quantization_config": {"quant_method": "bitsandbytes"},
+    }
     json_object = json.dumps(test_json_dict, indent=4)
 
     for file in ["tokenizer.json", "tokenizer_config.json"]:
```
booxter, line 31: should this change also go into tokenizer config though? (I know it's just tests, but still...)

tests/test_backends.py
```diff
@@ -21,7 +21,11 @@
 def create_safetensors_or_bin_model_files(
     model_path: pathlib.Path, model_file_type: str, valid: bool
 ):
-    test_json_dict = {"a": 1, "b": 2}
```
booxter, line 24: (Observation) I wonder why these tests have to write this dict into all files. Does `config.json` need any of it, except `{}`?

tests/test_backends.py
```diff
@@ -21,7 +21,11 @@
...
+    test_json_dict = {
+        "a": 1,
+        "b": 2,
+        "quantization_config": {"quant_method": "bitsandbytes"},
```
booxter, line 27: should we have a variant of the same test case where the config is NOT set?

src/instructlab/model/backends/vllm.py
```diff
@@ -259,6 +282,20 @@ def build_vllm_cmd(
...
+    if not (quant_arg_present and load_arg_present and eager_arg_present):
+        if is_bnb_quantized(model_path):
+            if not quant_arg_present:
+                vllm_cmd.extend(["--quantization", "bitsandbytes"])
+            if not load_arg_present:
+                vllm_cmd.extend(["--load-format", "bitsandbytes"])
+            # Currently needed to retain generation quality w/ 4-bit bnb + vLLM (bypass graph bug)
```
booxter, line 295: Is there a bug to refer to in the comment? Please add where one can find more details.

tests/test_backends.py
```diff
@@ -283,3 +287,38 @@ def test_build_vllm_cmd_with_args_provided(tmp_path: pathlib.Path):
...
+        "bitsandbytes",
+        "--load-format",
+        "bitsandbytes",
+        "--enforce-eager",
+        "--distributed-executor-backend",
+        "mp",
+    ]
+    create_safetensors_or_bin_model_files(model_path, "safetensors", True)
+    cmd, _ = build_vllm_cmd(
+        host, port, model_family, model_path, str(chat_template), vllm_args
```
booxter, line 322: (No action required) You could pass `auto` for template and avoid creating the file in the test case code.

tests/test_backends.py
```diff
@@ -283,3 +287,38 @@ def test_build_vllm_cmd_with_args_provided(tmp_path: pathlib.Path):
...
+    )
+    assert cmd == expected_cmd
```
booxter, line 324: This is a bit fragile: every time some other argument will be added to the default command line, this test will have to be updated. Consider instead confirming that the exact three options you test here are present (and not that the whole command line matches.)

booxter (changes requested):

booxter (approved):
Actually, probably I don't have anything too important to block this. But I'd like the author to send a follow-up if they decide to merge as-is.

Maxusmusti:
Thanks @booxter, I've created an issue to address the primary concern of test fragility and organization, feel free to add any additional details I may have left out: https://github.com/instructlab/instructlab/issues/1922

booxter:
Thanks @Maxusmusti

PR #1845, opened 2024-07-23
Support the pretraining samples in the legacy dataset conversion utility

RobotSail:
PR #1644 introduced a dataset conversion utility which provided the ability for datasets
in the HuggingFaces messages format to be converted into the legacy format used by Linux train.

However; SDG is now planning on providing a pretraining dataset which is represented as
phase007. This new dataset will usually only contain only a single message which can be identified
by the `pretraining` role inside the `messages` list. E.g.:

```json
{
	"messages": [
		{
			"role": "pretraining",
			"content": "<|user|>Why did the chicken cross the road?<|assistant|>To get to the other side!"
		}
	]
}
```

This PR provides support for this new dataset format by identifying if a dataset contains the
`pretraining` sample and then parsing it out to map into the legacy sample format:
```json
{
	"user": "Why did the chicken cross the road?",
	"assistant": "To get to the other side!",
	"system": ""
}
```

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,9 @@
 * `ilab model list`: a new command which lists all GGUF and Safetensor Models on the system.
 * `ilab data list`: a new command which lists the generated datasets in the user's datasets
   directory.
+* Legacy Linux training now supports the new messages format. When a dataset is provided in the
+  HuggingFace messages format, `ilab` will automatically convert it back into the legacy format.
+* Legacy Linux training is now compatible with the phase07 pretraining format.
 
 ### Breaking Changes
 
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -42,6 +42,16 @@ rules:
 """
 
 
+class SDGTokens:
+    """
+    Provides constant definitions for the tokens used by SDG.
+    """
+
+    USER = "<|user|>"
+    ASSISTANT = "<|assistant|>"
+    PRETRAINING = "<|pretraining|>"
+
+
 class TaxonomyReadingException(Exception):
     """An exception raised during reading of the taxonomy."""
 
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -44,6 +44,9 @@
 * The `generate` section of the config now has a `teacher` section. This section configures the
   teacher model when it is automatically served in the background. This new section has the same
   values as the `serve` section of the config.
+* Legacy Linux training now supports the new messages format. When a dataset is provided in the
+  HuggingFace messages format, `ilab` will automatically convert it back into the legacy format.
+* Legacy Linux training is now compatible with the phase007 pretraining format.
```
leseb, line 49: phase007? Where can I read more about that? :)

leseb (commented):
Do we have existing tests that validate this? Do they need to be expanded? Thanks!

RobotSail:
> Do we have existing tests that validate this? Do they need to be expanded? Thanks!

I'll add tests for this.

CHANGELOG.md
RobotSail: We don't have any documentation up for phase007 yet as I believe this is still something being worked on by SDG and training. But you can read about how it's used by the full training library in https://github.com/instructlab/training/pull/137

RobotSail:
@leseb I've updated this PR to include a more comprehensive test suite. I've also moved some of the old tests that were trying to test the dataset conversion in the linux test suite into the `test_utils.py` file. It turns out that the way we were mocking the linux train function meant that it received no coverage, so these tests weren't actually doing anything.

alimaredia (approved):

tests/test_utils.py
```diff
@@ -29,6 +31,157 @@ def test_validate_documents(self, git_clone_checkout):
...
+
+    def test_convert_to_legacy_from_messages(self):
+        messages: typing.List[MessageSample] = [
+            {
+                "messages": [
+                    {
+                        "role": "system",
+                        "content": "You are a friendly assistant",
+                    },
+                    {"role": "user", "content": "Who is pickle rick?"},
```
leseb, line 120: Love it!

leseb (approved):

src/instructlab/utils.py
```diff
@@ -556,15 +566,89 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+    Given a pretraining message sample which contains the tokens '<|user|>' and '<|assistant|>',
+    return the strings corresponding to the user & assistant inputs.
+
+    This function assumes that a given message begins with the '<|user|>' token
+    and contains '<|assistant|>' somewhere later.
+
+    Raises ValueError if neither the '<|user|>' or '<|assistant|>' tokens exist
+    within the given sample.
+    """
+    if SDGTokens.ASSISTANT not in pretraining:
```
booxter, line 591: nit:

```
for token in (ASSISTANT, USER):
    if token not in pretraining:
         raise ...
```

src/instructlab/utils.py
```diff
@@ -556,15 +566,89 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+        raise ValueError(
+            f"pretraining sample doesn't contain the '{SDGTokens.ASSISTANT}' token: {pretraining}"
+        )
+    if SDGTokens.USER not in pretraining:
+        raise ValueError(
+            f"pretraining sample doesn't contain the '{SDGTokens.USER}' token: {pretraining}"
+        )
+    user, assistant = pretraining.split(SDGTokens.ASSISTANT)
```
booxter, line 599: Question: Is it ever possible that the `pretraining` contains multiple tokens? Wonder if this should be handled somehow. (`split` will then return 3+-tuple, which will fail to unpack the return value here.)

src/instructlab/utils.py
```diff
@@ -556,15 +566,89 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+    sample = ds[0]
+    return any(m["role"] == "pretraining" for m in sample["messages"])
+
+
+def get_user_assistant_from_pretraining(pretraining: str) -> Tuple[str, str]:
+    """
+    Given a pretraining message sample which contains the tokens '<|user|>' and '<|assistant|>',
+    return the strings corresponding to the user & assistant inputs.
+
+    This function assumes that a given message begins with the '<|user|>' token
```
booxter, line 585: Is this enforced somewhere in the format definition? Or is just implementation detail that may change later? (Could you please share a link to where the format is defined? Maybe add a link to the comment?)

src/instructlab/utils.py
```diff
@@ -556,15 +566,89 @@ def split_hostport(hostport: str) -> tuple[str, int]:
...
+                if msg["role"] == "pretraining"
+            ),
+            None,
+        )
+        # not sure if 'system' will exist in phase007 or not - leaving here just in case
+        system = next(
+            (msg["content"] for msg in sample["messages"] if msg["role"] == "system"),
+            "",
+        )
+        if not pretraining:
```
booxter, line 630: nit: I think you could raise it before `system` is calculated.

src/instructlab/utils.py
```diff
@@ -601,6 +701,10 @@ def ensure_legacy_dataset(
     Given a dataset that's either in the HF messages format or the legacy ilab train format,
     ensure that the returned dataset is always in the legacy ilab train format.
     """
+    if len(dataset) == 0:
```
booxter, line 704: nit:

```
if not dataset:
    return []
```

tests/test_utils.py
```diff
@@ -29,6 +31,157 @@ def test_validate_documents(self, git_clone_checkout):
...
+                        "content": content,
+                    },
+                ],
+                "group": " test",
+                "dataset": "test-dataset",
+                "metadata": "test",
+            }
+        ]
+        if exception:
+            with pytest.raises(ValueError, match=match):
```
booxter, line 89: nit: `raises(exception, ...)`?

tests/test_lab_train.py
```diff
@@ -210,86 +209,6 @@ def test_invalid_data_dir(self, cli_runner: CliRunner):
             assert "Could not read from data directory" in result.output
             assert result.exit_code == 1
 
-    @patch("instructlab.utils.is_macos_with_m_chip", return_value=False)
```
booxter, line 213: Nice!

booxter (approved):
Assuming this reflects the format (I can't be sure since it's not referred from the patch it seems), the code does what it should do. I also love how testing is switched to target conversion functions instead of running through `click` runner.

src/instructlab/utils.py
RobotSail: That's a bit cleaner, I'll do this instead.

src/instructlab/utils.py
RobotSail: It's not necessarily enforced, but there's no logical reason for anything to exist before `<|user|>` since we're providing that as a separate message.

We can change this to instead extract all of the text between `<|user|>` and `<|assistant|>` rather than assuming that `<|user|>` is the very first token of the string.

src/instructlab/utils.py
RobotSail: For now no, but in the future if this is something we decide to add then we'll have to support it. I'll add a comment to the docstring indicating that the assumption is a single instance of each, and that user precedes assistant.

src/instructlab/utils.py
RobotSail: Good point.

src/instructlab/utils.py
RobotSail: I'll make this change.

tests/test_utils.py
RobotSail: PyTest doesn't like this unfortunately. It requires you to provide the exact exception or else it'll throw some nasty errors.

RobotSail: MyPy* not PyTest

booxter (approved):

danmcp:
@mergifyio rebase

danmcp:
@mergifyio rebase

PR #1849, opened 2024-07-24
Follow-up to OCI download PR merged

booxter:
This addresses comments from https://github.com/instructlab/instructlab/pull/1756 ; plus it optimizes the way symlinks are created, a bit.

### DIFF
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -227,21 +227,17 @@ class OCIDownloader(Downloader):
             raise click.exceptions.Exit(1)
 
         blob_dir = f"{oci_dir}/blobs/sha256/"
-        for _, _, files in os.walk(blob_dir):
-            for name in files:
-                if name not in file_map:
-                    continue
-                dest = file_map[name]
-                dest_model_path = os.path.join(self.download_dest, model_name, dest)
-                # unlink any existing version of the file
-                if os.path.exists(dest_model_path):
-                    os.unlink(dest_model_path)
+        for name, dest in file_map.items():
+            dest_model_path = os.path.join(self.download_dest, model_name, dest)
+            # unlink any existing version of the file
+            if os.path.exists(dest_model_path):
+                os.unlink(dest_model_path)
 
-                # create symlink to files in cache, to avoid redownloading if model has been downloaded before
-                os.symlink(
-                    os.path.join(blob_dir, name),
-                    dest_model_path,
-                )
+            # create symlink to files in cache to avoid redownloading if model has been downloaded before
+            os.symlink(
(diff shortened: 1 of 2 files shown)


src/instructlab/model/download.py
```diff
@@ -135,11 +135,21 @@ class OCIDownloader(Downloader):
     We are leveraging OCI v1.1 for this functionality
     """
 
-    def __init__(self, repository: str, release: str, download_dest: str, ctx) -> None:
-        super().__init__(
-            repository=repository, release=release, download_dest=download_dest
-        )
-        self.ctx = ctx
+    @staticmethod
+    def _load_json(file_path: Path):
```
jaideepr97, line 139: @booxter couldn't this generic function to read json files be used elsewhere in the project as well? That was the main reason I'd put it in the `utils.py` file

src/instructlab/model/download.py
```diff
@@ -149,35 +159,35 @@ def _build_oci_model_file_map(self, oci_model_path: str) -> dict:
...
-                    match = _extract_SHA(manifest["digest"])
+                    match = self._extract_SHA(manifest["digest"])
+                    break
 
             if match:
                 index_hash = match.group(1)
             else:
                 raise ValueError(
-                    f"could not find hash for index file at: {oci_model_path}"
+                    f"failed to find image hash in the index file: {oci_model_path}"
```
jaideepr97, line 176: This is more for my learning, but is it correct to refer to this file as the "image"? It references one of the blobs that represents the manifest that contains the mapping between other blobs and their respective filenames

jaideepr97 (commented):
Thanks for the follow up PR @booxter 
Just a couple things from me

src/instructlab/model/download.py
booxter: That may be true (though it's not used anywhere else right now). Then the name of the function should not be private (remove `_`)?

src/instructlab/model/download.py
booxter: Oh that may well be that I screwed this one up. The intent here was to say it's not a hash *for* but *in* the index file. The `image` part is wrong - I will drop it.

src/instructlab/model/download.py
jaideepr97: That would be my preference anyway
I can think of a couple other places we could use this function

booxter: I will return the load_json one back; and will leave extract_sha in place since it's very specific to OCI.

jaideepr97 (approved):
LGTM thanks @booxter

nathan-weinberg (approved):

PR #1852, opened 2024-07-24
introduce backends.common to reduce circular dependencies

makelinux:
The backends module and modules llama_cpp and vllm have circular dependency, which is anti-pattern.

The backends module, along with the llama_cpp and vllm modules, have a circular dependencies, which is considered an anti-pattern."
    
This PR is intentionally kept small to facilitate an easier code review. The plan is to relocate the relevant code from `backends` to `common` in a subsequent PR, eliminating the circular dependency entirely.

Current state:
```mermaid
graph LR;
  backends --> llama_cpp
  llama_cpp --> backends
  backends --> vllm
  vllm --> backends
```
Desired state:

```mermaid
graph LR;
  backends --> llama_cpp
  llama_cpp -->common
  backends --> common
  backends --> vllm
  vllm --> common
```

References:
1. https://en.wikipedia.org/wiki/Circular_dependency
2. https://en.wikipedia.org/wiki/Acyclic_dependencies_principle

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -31,26 +31,11 @@ from ...client import check_api_base
 from ...configuration import _serve as serve_config
 from ...configuration import get_api_base, get_model_family
 from ...utils import split_hostport
+from .common import CHAT_TEMPLATE_AUTO, LLAMA_CPP, VLLM, templates
 
 logger = logging.getLogger(__name__)
 
-LLAMA_CPP = "llama-cpp"
-VLLM = "vllm"
 SUPPORTED_BACKENDS = frozenset({LLAMA_CPP, VLLM})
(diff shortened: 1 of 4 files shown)


khaledsulayman:
@makelinux could you link the follow-up PR if it's ready? Might be better to merge these 2 in succession.

leseb (approved):

makelinux:
> @makelinux could you link the follow-up PR if it's ready? Might be better to merge these 2 in succession.
@khaledsulayman,

Unfortunately, it's not practical to submit several commits or pull requests simultaneously. Combining multiple changes in a single pull request increases the risk of merge conflicts and may require multiple rebases. Creating a pull request based on another is also challenging, if not impossible. If you have any suggestions or know of a workaround, please let me know.

RobotSail (approved):
LGTM

markstur (approved):
lgtm
Thanks!

Issue #1851, opened 2024-07-24
Lock down major.minor versions in requirements.txt


n1hility:
In the run up to 0.18.0 we should lock down all of our deps to minor versions in our requirements file to prevent surprise updates coming in (e.g. like torch 2.4).

bjhargrave:
See also https://github.com/instructlab/instructlab/pull/1425

n1hility:
I'm closing this in favor of using constraints per the proposal in #122

nathan-weinberg:
@n1hility think you wanted to link https://github.com/instructlab/dev-docs/pull/122

PR #1853, opened 2024-07-24
deps: add version ceiling for all dependencies

nathan-weinberg:
Related to https://github.com/instructlab/instructlab/issues/1851

This first PR adds a ceiling to all `ilab` dependencies

My approach was this:
1. Next Y version up from the currently defined _minimum_ version was made the ceiling
2. Two dependencies - `tokenizers` and `transformers` did not work with this approach, as seen below
```bash
ERROR: Cannot install instructlab and instructlab==0.17.1.dev609 because these package versions have conflicting dependencies.

The conflict is caused by:
    instructlab 0.17.1.dev609 depends on tokenizers<0.12.0 and >=0.11.1
    transformers 4.41.2 depends on tokenizers<0.20 and >=0.19
```
For these, I added a ceiling that is the next Y version up from what is currently installed on `main` as of the time of this writing, as you can see below
```bash
tokenizers                0.19.1
transformers           4.43.2
```

One open question @n1hility and I were discussing is if we should take approach 2 for _all_ dependencies, i.e. update the ceiling to be the Y version above what is currently installed on `main` for _all_ dependencies. I did notice when diff-ing the two we now have some older versions of packages after this change. Happy to update the PR with that or another approach but opening this as a starting point.

**Checklist:**

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,12 +3,12 @@
 -r requirements.txt
 
 pre-commit>=3.0.4,<4.0
-pydeps>=1.12.12,<2
+pydeps>=1.12.12,<2.0
 pylint>=2.16.2,<4.0
 pylint-pydantic
-pyspelling>=2.0
+pyspelling>=2.0,<2.1
 pytest
 pytest-asyncio
 pytest-cov[toml]
 pytest-html
-tox>=4.4.2,<5
+tox>=4.4.2,<5.0.0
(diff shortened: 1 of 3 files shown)


n1hility:
@tiran PTAL Also any other suggestions/ideas you have to lock things down short of pinning everything before release would be appreciated.

tiran:
This approach will make our downstream work harder. We will no longer be able to pull in security and bug fix updates if necessary. It also does not solve the actual problem, because the change only adds version ranges to leave packages. You have to constraint every package in the entire package set to a fixed version, then change the constraint when necessary.

nathan-weinberg:
After some additional discussion we are going to take a different approach, thanks all for the feedback!

PR #1854, opened 2024-07-24
mergify: add mergify autolabeling for dependency PRs

nathan-weinberg:
**Checklist:**

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -220,3 +220,14 @@ pull_request_rules:
     label:
       remove:
         - one-approval
+
+- name: label-dependencies
+  description: Automatically apply dependencies label
+  conditions:
+    - or:
+      - files~=requirements.*\.txt$
+      - files=requirements/**/*
+  actions:
+    label:
+      add:
+        - dependencies


booxter (approved):

RobotSail (approved):
LGTM

leseb (approved):

.github/mergify.yml
```diff
@@ -220,3 +220,14 @@ pull_request_rules:
     label:
       remove:
         - one-approval
+
+- name: label-dependencies
+  description: Automatically apply dependencies label
+  conditions:
+    - or:
+      - files~=requirements.*\.txt$
+      - files=requirements/**/*
```
bjhargrave, line 229: This is a glob, not a regex.

bjhargrave (changes requested):

.github/mergify.yml
nathan-weinberg: We're using something like this here: https://github.com/instructlab/instructlab/blob/main/.github/mergify.yml#L98

Mergify claims this is valid config - but if it doesn't work as expected, it needs to be fixed everywhere in this file

bjhargrave: That is wrong too. According to https://docs.mergify.com/configuration/conditions/#operators-list

`files=` is an exact match test, not a glob or regex. So `files=requirements/**/*` will never match anything.

`files~=` is a regex (not a glob). 

I do not see any case where globs are supported.

bjhargrave: > Can you make a followup PR for both?

See https://github.com/instructlab/instructlab/pull/1941

PR #1855, opened 2024-07-24
tests: Don't access result.stderr from cliRunner

booxter:
Our cliRunner mixes stderr and stdout (mix_stderr=True is the default for the click runner.) So, accessing result.stderr won't work.


RobotSail (approved):
LGTM, thank you for catching this

leseb (approved):

PR #1857, opened 2024-07-25
ci: ensure $FULLTRAIN for 4xA10 workflow

alimaredia:
The 4xA10 workflow is intended to mirror the full ilab workflow. If `FULLTRAIN`  is set to `0` the final ilab chat and serve aren't run (https://github.com/instructlab/instructlab/blob/main/scripts/basic-workflow-tests.sh#L304). If the evaluation flag was specified in the future it would not be run as well.

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x4.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x4.yml
@@ -145,7 +145,7 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
           . venv/bin/activate
-          ./scripts/basic-workflow-tests.sh -mFMT
+          ./scripts/basic-workflow-tests.sh -mfFMT
 
       - name: Add comment to PR if the workflow failed
         if: failure() && steps.check_pr.outputs.is_pr == 'true'


RobotSail (approved):
LGTM, I don't see why not.

khaledsulayman (approved):

alimaredia:
@RobotSail or anyone else that's knowledgable, if we enable FULLTRAIN and the TRAIN_LIBRARY we won't hit the same issues as we see for the A10x1 instance with only FULLTRAIN right? 

This is a recent run I did of that
https://github.com/russellb/ilab-runner/actions/runs/10085943339/job/27887699505

RobotSail:
@alimaredia Fulltrain works but only for the LoRA case since we aren't doing full model parallelism with DeepSpeed currently.

RobotSail:
We would need to update how checkpointing works in the training library so we can split up the model and be able to run a full fine-tune.

alimaredia:
Test to ensure the two training flags can go together is here: https://github.com/russellb/ilab-runner/actions/runs/10087870830/job/27892725725

alimaredia:
@RobotSail looks like that last log failed during the training step. It only errored on serving because there was no model for `ilab model serve` to output.

RobotSail:
@alimaredia I made a comment on that - you'll need to update the `--model-path` when serving so that it's pointing to the correct directory. We use temp directories in e2e while the config has it hard-coded to a specific dir. Therefore we need to just pass the correct path.

nathan-weinberg:
@alimaredia where is this at?

nathan-weinberg:
Superceded by https://github.com/instructlab/instructlab/pull/2291

PR #1859, opened 2024-07-25
feat: display default dirs in `ilab system info`

### LINKED ISSUE #1858: Display default directories with ilab sysinfo

With the introduction of persistent storage, it's become easier for `ilab` to manage storage and objects internally.
Unfortunately, the same cannot be said for the end user since it's become harder to understand where stuff lives.

We should either add a new command such as `ilab system storage` or `ilab system directories` that displays the
active directories being used by ilab after the config is loaded.

Alternatively, we could include this information in the `ilab system info` command under a `defaults` field
which simply prints out the **DEFAULT** directories used by ilab. These directories may differ from what's in the
actual config, but that's what `ilab config show` is for - since that renders the actively loaded config.

RobotSail:
With the introduction of persistent storage, it has become easier for ilab to manage
data, but harder for the user to. This commit adds 3 fields to `ilab system info` which
display where the default cache, data, and config directories are.

Resolves #1858 

<!-- Thank you for contributing to InstructLab! -->

<!-- STEPS TO FOLLOW:
  1. Add a description of the changes (frequently the same as the commit description)
  2. Enter the issue number next to "Resolves #" below (if there is no tracking issue resolved, **remove that section**)
  3. Follow the steps in the checklist below, starting with the **Commit Message Formatting**.
-->

<!-- Uncomment this section with the issue number if an issue is being resolved
Resolves #
--->

**Checklist:**

This is how the output of `ilab system info` now looks with this PR:

```
(venv) osilkin:instructlab/ (list-default-dirs✗) $ ilab system info                                                                                                                                                                                                                                                          [0:25:47]
sys.version: 3.11.8 (main, Jul  3 2024, 13:50:00) [Clang 15.0.0 (clang-1500.3.9.4)]
sys.platform: darwin
os.name: posix
platform.release: 23.5.0
platform.machine: arm64
instructlab.version: 0.18.0a7.dev4
instructlab-dolomite.version: 0.0.1
instructlab-eval.version: 0.1.0
instructlab-quantize.version: 0.1.0
instructlab-schema.version: 0.3.1
instructlab-sdg.version: 0.2.0
instructlab-training.version: 0.3.0
torch.version: 2.3.1
torch.version.cuda: None
torch.version.hip: None
torch.cuda.available: False
llama_cpp_python.version: 0.2.79
defaults.cache_dir: /Users/osilkin/Library/Caches/instructlab
defaults.data_dir: /Users/osilkin/Library/Application Support/instructlab
defaults.config_home: /Users/osilkin/Library/Caches/instructlab
```

### DIFF
--- a/src/instructlab/system/info.py
+++ b/src/instructlab/system/info.py
@@ -11,6 +11,7 @@ import click
 
 # First Party
 from instructlab import clickext
+from instructlab.configuration import DEFAULTS
 
 
 def _platform_info() -> typing.Dict[str, typing.Any]:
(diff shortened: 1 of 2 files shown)


src/instructlab/system/info.py
```diff
@@ -130,6 +131,33 @@ def _instructlab_info():
     return {f"{name}.version": ver for name, ver in pkgs}
 
 
+def _default_paths() -> dict[str, str]:
+    """Default directories used by ilab"""
+    # d = {
```
RobotSail, line 136: I kept this in the commit because the other idea I had was enumerating the defaults, but this seems way too messy and cluttered. 

Maybe there's a better place where these can go? `ilab config show` currently renders the *active* paths used by ilab, but not necessarily the defaults.

tiran (changes requested):
I would keep the output of `ilab system info` to a useful minimum. It's a debug feature for humans and for bug reporting. If you display too much, people won't find the useful bits and pieces. If you want to display all paths, then please add a new command like `ilab system paths` or `ilab config paths`. The three paths are useful to show.

The `defaults` prefix feels liek it is leaking internal implementation details to the user. You are printing the config directory of instructlab. The key should be something like `instructlab.config_directory` or `instructlab.paths.config_directory`

- Remove the commented out section
- Move the values below `_instructlab_info()`
- Change them to consistent names. Some variables have `+dir` and others `_home` as suffix.

RobotSail:
@tiran Your approach makes much more sense. I'll update this PR to do that instead.

tiran:
> @tiran Your approach makes much more sense. I'll update this PR to do that instead.

Let's make it two PRs. Update this PR to land the `ilab system info` improvement and make a new one to list paths.

RobotSail:
@tiran Do we still want to make this change to `ilab system info` though? Based on your response it sounds like a separate debug command of some kind may make more sense. I completely agree that putting paths in `ilab system info` will create a messy output.

ktam3:
@RobotSail - could you please rebase?

RobotSail:
@ktam3 I'm just gonna close it

PR #1861, opened 2024-07-25
feat: add `ilab data list`

RobotSail:
- **add `ilab model list`**
- **feat: introduce `ilab data list` to list datasets**

This PR follows the work done by @cdoern in PR #1840. 

Specifically, this PR now enables the user to list out available datasets
using a new command: `ilab data list`. This way, when the user generates
data, they can access it much more easily.

Here's an example output from the new command:

```
+------------------------------------------------------------+---------------------+-----------+
| Dataset                                                    | Created At          | File size |
+------------------------------------------------------------+---------------------+-----------+
| train_merlinite-7b-lab-Q4_K_M_2024-07-25T00_47_21.jsonl    | 2024-07-25 00:49:37 | 16.94 KB  |
| test_merlinite-7b-lab-Q4_K_M_2024-07-25T00_50_57.jsonl     | 2024-07-25 00:50:57 | 2.84 KB   |
| messages_merlinite-7b-lab-Q4_K_M_2024-07-25T00_47_21.jsonl | 2024-07-25 00:49:37 | 19.17 KB  |
| test_merlinite-7b-lab-Q4_K_M_2024-07-25T00_47_21.jsonl     | 2024-07-25 00:47:21 | 2.84 KB   |
+------------------------------------------------------------+---------------------+-----------+
```

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,8 @@
    ensuring that defaults such as taxonomy repositories and base models are honored from the global
    config.
 * `ilab model list`: a new command which lists all GGUF and Safetensor Models on the system.
+* `ilab data list`: a new command which lists the generated datasets in the user's datasets
+  directory.
 
 ### Breaking Changes
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,7 @@ issues = "https://github.com/instructlab/instructlab/issues"
 
 [project.entry-points."instructlab.command.data"]
 "generate" = "instructlab.data.generate:generate"
+"list" = "instructlab.data.list:list_data"
 
 [project.entry-points."instructlab.command.model"]
 "chat" = "instructlab.model.chat:chat"
--- a/src/instructlab/data/list.py
+++ b/src/instructlab/data/list.py
@@ -1,0 +1,77 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+from pathlib import Path
+from typing import List, TypedDict
+import logging
+import time
+
+# Third Party
+import click
+
+# First Party
+from instructlab import clickext
+from instructlab.configuration import DEFAULTS
+from instructlab.utils import convert_bytes_to_proper_mag, print_table
+
+logger = logging.getLogger(__name__)
+
+
+class DatasetListing(TypedDict):
+    """
+    Represents a single dataset listing.
+    """
+
+    filename: str
+    created_at: str
+    file_size: str
+
+
+@click.command(name="list")
+@clickext.display_params
+@click.option(
+    "--dataset-dirs",
+    help="Base directories where datasets are stored.",
+    multiple=True,
+    default=[DEFAULTS.DATASETS_DIR],
+    show_default=True,
+)
+@click.pass_context
+def list_data(ctx, dataset_dirs: list[str]):
+    """lists datasets"""
+    data: List[List[str]] = []
+    for directory in dataset_dirs:
+        dirpath = Path(directory)
+        directories: List[Path] = [dirpath]
+        top_level_dir = dirpath
+        if not dirpath.exists():
+            ctx.fail(f"directory does not exist: {dirpath}")
+        while len(directories) > 0:
+            current_dir = directories.pop()
+            for entry in current_dir.iterdir():
+                if entry.is_dir():
+                    directories.append(entry)
+                    continue
+                if entry.suffix != ".jsonl":
+                    continue
+
+                stat = entry.stat()
+                created_at = time.strftime(
+                    "%Y-%m-%d %H:%M:%S", time.localtime(stat.st_ctime)
+                )
+                fsize, mag = convert_bytes_to_proper_mag(stat.st_size)
+                formatted_size = f"{fsize:.2f} {mag}"
+
+                # capture the entries as 'messages.jsonl', 'dir/dataset2.jsonl'
+                # with respect to the top-level directory
+                relative_name = str(entry.relative_to(top_level_dir))
+                data.append(
+                    [
+                        relative_name,
+                        created_at,
+                        formatted_size,
+                    ]
+                )
+
+    headers = ["Dataset", "Created At", "File size"]
+    print_table(headers, data)
--- a/src/instructlab/model/list.py
+++ b/src/instructlab/model/list.py
@@ -13,6 +13,7 @@ import click
 from instructlab import clickext
 from instructlab.configuration import DEFAULTS
 from instructlab.model.backends.backends import is_model_gguf, is_model_safetensors
+from instructlab.utils import convert_bytes_to_proper_mag, print_table
 
 logger = logging.getLogger(__name__)
 
@@ -46,38 +47,6 @@ def model_list(model_dirs: list[str], list_checkpoints: bool):
 
 
 # convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MiB, GiB)
-def convert_bytes_to_proper_mag(f_size: float):
-    magnitudes = ["KiB", "MiB", "GiB", "TiB"]
-    magnitude = "B"
-    for mag in magnitudes:
-        if f_size >= 1024:
-            magnitude = mag
-            f_size /= 1024
-        else:
-            return f_size, magnitude
-    return f_size, magnitude
-
-
-# print_table displays the models on the system in a nicely formatted table
-def print_table(headers: list[str], data: list[list[str]]):
-    column_widths = [
-        max(len(str(row[i])) for row in data + [headers]) for i in range(len(headers))
-    ]
-    # Print separator line between headers and data
-    horizontal_lines = ["-" * (width + 2) for width in column_widths]
-    joining_line = "+" + "+".join(horizontal_lines) + "+"
-    print(joining_line)
-    outputs = []
-    for header, width in zip(headers, column_widths, strict=False):
-        outputs.append(f" {header:{width}} ")
-    print("|" + "|".join(outputs) + "|")
-    print(joining_line)
-    for row in data:
-        outputs = []
-        for item, width in zip(row, column_widths, strict=False):
-            outputs.append(f" {item:{width}} ")
-        print("|" + "|".join(outputs) + "|")
-    print(joining_line)
 
 
 AnalyzeResultGGUF = list[str]
@@ -88,13 +57,13 @@ def _analyze_gguf(entry: Path) -> AnalyzeResultGGUF:
     # stat the gguf, add it to the table
     stat = Path(entry.absolute()).stat(follow_symlinks=False)
     f_size = stat.st_size
-    f_size, magnitude = convert_bytes_to_proper_mag(f_size)
+    adjusted_size, magnitude = convert_bytes_to_proper_mag(f_size)
     # add to table
     modification_time = os.path.getmtime(entry.absolute())
     modification_time_readable = time.strftime(
         "%Y-%m-%d %H:%M:%S", time.localtime(modification_time)
     )
-    return [entry.name, modification_time_readable, f"{f_size:.1f} {magnitude}"]
+    return [entry.name, modification_time_readable, f"{adjusted_size:.1f} {magnitude}"]
 
 
 def _analyze_dir(
@@ -134,7 +103,7 @@ def _analyze_dir(
             full_file = os.path.join(root, f)
             stat = Path(full_file).stat(follow_symlinks=False)
             all_files_sizes += stat.st_size
-        all_files_sizes, magnitude = convert_bytes_to_proper_mag(all_files_sizes)
+        adjusted_all_sizes, magnitude = convert_bytes_to_proper_mag(all_files_sizes)
         if add_model:
             # add to table
             modification_time = os.path.getmtime(entry.absolute())
@@ -145,7 +114,7 @@ def _analyze_dir(
                 [
                     actual_model_name,
                     modification_time_readable,
-                    f"{all_files_sizes:.1f} {magnitude}",
+                    f"{adjusted_all_sizes:.1f} {magnitude}",
                 ]
             )
     return models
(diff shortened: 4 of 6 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,9 @@
 * The `generate` section of the config now has a `teacher` section. This section configures the
   teacher model when it is automatically served in the background. This new section has the same
   values as the `serve` section of the config.
+* `ilab model list*: a new command which lists all GGUF and Safetensor Models on the system.
+* `ilab data list`: a new command which lists out the generate datasets in the user's datasets
```
jaideepr97, line 51:
```suggestion
* `ilab data list`: a new command which lists out the generated datasets in the user's datasets
```

src/instructlab/data/list.py
```diff
@@ -0,0 +1,101 @@
...
+    magnitudes = ["KB", "MB", "GB"]
+    magnitude = "B"
+    adjusted_fsize = float(f_size)
+    for mag in magnitudes:
+        if adjusted_fsize >= 1024:
+            magnitude = mag
+            adjusted_fsize /= 1024
+        else:
+            return adjusted_fsize, magnitude
+    return adjusted_fsize, magnitude
```
jaideepr97, line 40: can we just make this a general utility function since it will be getting used more than once?

src/instructlab/data/list.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+# print_table displays the datasets on the system in a nicely formatted table
+def print_table(headers: list[str], data: list[list[str]]):
```
jaideepr97, line 44: same with this function, better not to duplicate if we can avoid it

src/instructlab/data/list.py
```diff
@@ -0,0 +1,101 @@
...
+    help="Base directories where datasets are stored.",
+    multiple=True,
+    default=[DEFAULTS.DATASETS_DIR],
+    show_default=True,
+)
+def list_data(dataset_dirs: list[str]):
+    """lists datasets"""
+    entries: List[DatasetListing] = []
+    for directory in dataset_dirs:
+        for entry in Path(directory).iterdir():
```
jaideepr97, line 84: This is my ignorance, but what happens if someone passes a mutli-level nested directory? Does `iterdir()` recursively get to the bottom of the tree and only process files? should we be verifying that the entry is of type `.jsonl` here?

jaideepr97 (commented):
Thanks for the PR @RobotSail 
I'm not sure I see why @cdoern's changes are included here - unless we make some of these functions general utility functions that are used by both `model list` and `data list` I don't see much dependency between the 2 
could we remove all the `model list` related code from this PR?

RobotSail:
@jaideepr97  I'm basing this on top of his work to avoid unnecessary duplication, I expect his PR to be merged first as well at which point we can rebase onto main.

src/instructlab/data/list.py
RobotSail: Yes but I'm keeping it here for now to avoid a messy rebase because otherwise it's going to exist in the `model/list.py` file.

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+            if list_checkpoints and directory is DEFAULTS.CHECKPOINTS_DIR:
+                logging.debug("Including model regardless of model validity")
+            else:
+                continue
+        for f in files:
+            # stat each file in the dir, add the size in Bytes, then convert to proper magnitude
+            full_file = os.path.join(root, f)
+            stat = Path(full_file).stat(follow_symlinks=False)
+            all_files_sizes += stat.st_size
+        all_files_sizes, magnitude = convert_bytes_to_proper_mag(all_files_sizes)
```
booxter, line 131: (nit) this seems unused if `add_model` is `False`. Should this move under the `if`? (Why do we even run the code above if it's not `add_model`? Seems like a waste of CPU? What am I missing?

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def _analyze_gguf(entry: Path) -> AnalyzeResultGGUF:
+    # stat the gguf, add it to the table
+    stat = Path(entry.absolute()).stat(follow_symlinks=False)
+    f_size = stat.st_size
+    f_size, magnitude = convert_bytes_to_proper_mag(f_size)
+    # add to table
+    age = pandas.to_timedelta(time.time_ns() - stat.st_mtime_ns)
+    return [entry.name, str(age), f"{f_size} {magnitude}"]
```
booxter, line 91: I think this risks producing very long size strings. Consider limiting the precision to e.g. 2 fractional digits.

```
f"{2./3}"
'0.6666666666666666'
```

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+AnalyzeResultGGUF = list[str]
+AnalyzeResultDir = list[list[str]]
+
+
+def _analyze_gguf(entry: Path) -> AnalyzeResultGGUF:
+    # stat the gguf, add it to the table
+    stat = Path(entry.absolute()).stat(follow_symlinks=False)
+    f_size = stat.st_size
+    f_size, magnitude = convert_bytes_to_proper_mag(f_size)
+    # add to table
```
booxter, line 89: (nit) this comment is misplaced (this function doesn't add anything to a table) and not very helpful here. Consider removing it.

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+# convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MB, GB)
+def convert_bytes_to_proper_mag(f_size: float):
+    magnitudes = ["KB", "MB", "GB"]
+    magnitude = "B"
+    for mag in magnitudes:
+        if f_size >= 1024:
+            magnitude = mag
+            f_size /= 1024
+        else:
+            return f_size, magnitude
```
booxter, line 58: nit: `break` and let the last `return` do the job

src/instructlab/model/list.py
```diff
@@ -0,0 +1,138 @@
...
+            if entry.is_file() and is_model_gguf(entry):
+                model, age, size = _analyze_gguf(entry)
+                data.append([model, age, size])
+            elif entry.is_dir():
+                for m in _analyze_dir(entry, list_checkpoints, directory):
+                    data.append(m)
+    print_table(["Model Name", "Model Age", "Model Size"], data)
+
+
+# convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MB, GB)
```
booxter, line 49: Remove `dir/file` from the description, this is really irrelevant for the function what it takes. It just takes bytes and spills out magnitude.

---

Also maybe convert to docstring?

src/instructlab/data/list.py
booxter: I don't understand which messy rebase you are talking about...

src/instructlab/data/list.py
booxter, line 84: you should handle the case when the directory doesn't exist. Otherwise `.iterdir` will raise.

src/instructlab/data/list.py
```diff
@@ -0,0 +1,101 @@
...
+            stat = entry.stat()
+            created_at = time.strftime(
+                "%Y-%m-%d %H:%M:%S", time.localtime(stat.st_ctime)
+            )
+            fsize, mag = convert_bytes_to_proper_mag(stat.st_size)
+            formatted_size = f"{fsize:.2f} {mag}"
```
booxter, line 90: The "bytes-to-string-that-represents-human-readable-size" is useful in both `data` and `model` paths. It should be a util function.

src/instructlab/data/list.py
booxter: I think we should filter our anything that is not `isfile`.

src/instructlab/data/list.py
```diff
@@ -0,0 +1,101 @@
...
+            entries.append(
```
booxter, line 91: nit: you construct `entries` list, then verbatim convert it into `data`. Why not just construct `data` here?

booxter (commented):
`model list` code is very hard to understand (esp. the directory part.) Perhaps we can simplify it a bit with the small improvements I shared, it will become easier to consume.

There's some code duplication that I think would be nice to address before merging.

There are some checks missing (e.g. for passed names actually representing directories.) that may otherwise make `ilab` crash with tracebacks; but nothing major.

---

But the code seems to work. So I'm not blocking.

booxter:
Gosh, only now I realized that this includes code from another PR and most comments belong elsewhere. Oh well! Sorry for that, I will ask elsewhere to pick up most of comments I left here.

src/instructlab/data/list.py
RobotSail: Good point

leseb:
What's the rationale for proposing this as a blocker?

RobotSail:
@leseb This is a UX blocker - we cannot ship something that has all of the data files in some arbitrary directory on the user's system without providing an easy way for them to see where these things exist.

src/instructlab/data/list.py
booxter: This is proposed as part of https://github.com/instructlab/instructlab/pull/1885

CHANGELOG.md
RobotSail: Thanks for catching this!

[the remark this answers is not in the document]
RobotSail: Thanks!

CHANGELOG.md
```diff
@@ -47,6 +47,9 @@
...
+  directory
```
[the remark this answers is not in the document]
RobotSail: Good catch!

booxter:
FYI I removed all comments for the code that came from `model list` PR. I left only those that are relevant here.

src/instructlab/data/list.py
RobotSail: Will we ever have a case where the SDG dataset is outputted into a directory?

booxter: Depends on how you'd like to handle a case where something that is not a last level directory was passed. If you are ok with reporting directories in the output (probably with a minuscule size - since we don't drill down like we do in `_analyze_dir` in `model list` implementation), then this is enough.

---

If, as @jaideepr97 seems to suggest, we want to drill down, then `iterdir()` is not the right way. Instead, `os.walk` should be used.

src/instructlab/data/list.py
RobotSail: Good point.

src/instructlab/data/list.py
RobotSail: I see what you mean. The reason why we should avoid this for the time being is that it introduces confusion in the implementation (rightfully so) which shouldn't be a priority for 0.18.0.
Specifically, by drilling through subdirectories, we would end up printing them in the format of `foo/bar/dataset.jsonl`. This is misleading since it's printed as a relative path, and so the user may end up referencing it as such. However; the default is to look at the user's persisent storage to extract these datasets. In this way, we can at least provide the guidance that the datasets are relative to where they are stored in local storage. Otherwise, it will greatly convulate the work we need to do for documentation and presentation.

pyproject.toml
```diff
@@ -49,6 +49,7 @@ issues = "https://github.com/instructlab/instructlab/issues"
 
 [project.entry-points."instructlab.command.data"]
 "generate" = "instructlab.data.generate:generate"
+"list" = "instructlab.data.list:list_data"
```
booxter, line 52: (No action required) nit: model list has `model_list` function, but here you have `list_data`. Maybe change to `data_list` for consistency?

src/instructlab/data/list.py
```diff
@@ -0,0 +1,67 @@
...
+    show_default=True,
+)
+@click.pass_context
+def list_data(ctx, dataset_dirs: list[str]):
+    """lists datasets"""
+    data: List[List[str]] = []
+    for directory in dataset_dirs:
+        dirpath = Path(directory)
+        if not dirpath.exists():
+            ctx.fail(f"directory does not exist: {dirpath}")
```
booxter, line 47: This is one approach. An alternative would be to skip to the next directory. (As implemented here: https://github.com/instructlab/instructlab/pull/1885) Whatever is done, both should be synchronized for consistency.

src/instructlab/data/list.py
```diff
@@ -0,0 +1,67 @@
...
+        for entry in Path(directory).iterdir():
```
booxter, line 48: nit: you already have Path (as `dirpath`), no need to instantiate it again.

booxter (approved):
Thanks. We should sync behaviour of `model list` and `data list` when a directory doesn't exist (fail completely or skip to next)

jaideepr97 (approved):
LGTM @RobotSail 
please fix lint failure

RobotSail:
@booxter I've updated it now so the behavior is displaying all files that end in `.jsonl` recursively and relative to the dataset directory. For instance, here's an example output:
```
(venv) osilkin:instructlab/ (list-datasets✗) $ ilab data list                                                                                                                                             [16:46:00]
+-------------------------------------------------------------+---------------------+-----------+
| Dataset                                                     | Created At          | File size |
+-------------------------------------------------------------+---------------------+-----------+
| train_merlinite-7b-lab-Q4_K_M_2024-07-25T00_47_21.jsonl     | 2024-07-25 00:49:37 | 16.94 KB  |
| test/test_merlinite-7b-lab-Q4_K_M_2024-07-25T00_50_57.jsonl | 2024-07-26 12:43:58 | 2.84 KB   |
+-------------------------------------------------------------+---------------------+-----------+
```

pyproject.toml
RobotSail: Sure, I'll make this change.

src/instructlab/data/list.py
RobotSail: I like this approach more, but I want to stay consistent with what's in main right now.

src/instructlab/data/list.py
RobotSail: Good catch, I'll remove this.

cdoern (approved):
lgtm. I like that you moved some of the functions I had into utils.py, makes more sense for them to be there. I appreciate your hard work on this!

PR #1863, opened 2024-07-25
Add pip constraint files from Fromager

tiran:
Add constraints files for CPU and CUDA based on Fromager's dataset for 0.18.0a5. The constraints have been cleaned up from problematic or buggy entries.

Use constraints in tox tests and e2e tests.`

Related to #1851

**Checklist:**


Makefile
```diff
@@ -229,3 +229,19 @@ actionlint: ## Lint GitHub Action workflows
 		exit 1 ; \
 	fi
 	$(CMD_PREFIX) actionlint -color
+
+# generate constraints.txt from fromager constraints file
+# remove all build-only dependencies
+# remoev all meson packages
+# remove prometheus-fastapi-instrumentator and its starlette (bad starlette version)
+# remove instructlab packages
+requirements/constraints-cuda.txt: fromager.txt
```
bjhargrave, line 238: Is `fromager.txt` file meant to be included in this PR?

tiran: No, the file is coming from an internal system. I added the code to Makefile so I can re-generate the file. I can remove if that's too confusing.

bjhargrave: It seems there should be someway for anyone to rebuild the constraints files. But it seems this currently depends upon some private data. This does not seem sustainable from a community point-of-view.

tiran: That is possible with `pip freeze`, but you have to install the entire environment first. Other options are `pip-compile` and `pip-compile-multi`, but they don't work for our packages, yet. `flash-attn` and some other packages have some hacks that break the tools.

PR #1867, opened 2024-07-25
Update README.md Instructions Steps

brianbier:
I've updated the readme to help folks who are setting up instructlabs and run into the following error:

`zsh: no matches found: instructlab[mps]`

This error occurs when try to execute the following command:

`pip install instructlab[mps]`

For some strange reason the terminal doesn't know what to do with the install and throw and error but if you wrap the package to install in quotes it works with out any issues

Solution:

`pip install 'instructlab[mps]'`

**Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
       python3 -m venv --upgrade-deps venv
       source venv/bin/activate
       pip cache remove llama_cpp_python
-      pip install instructlab[cpu] \
+      pip install 'instructlab[cpu]' \
          --extra-index-url=https://download.pytorch.org/whl/cpu \
          -C cmake.args="-DLLAMA_NATIVE=off"
       ```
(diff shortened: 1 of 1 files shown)


booxter (commented):
`zsh` probably attempts to interpret the string as a pattern. Wonder if we should use a single quote string to avoid any kind of parsing any shell may attempt?

booxter:
Yeah I think this is the explanation: https://kinopyo.com/en/blog/escape-square-bracket-by-default-in-zsh

I suggest we just put the strings in single-quotes to be sure. Thoughts?

brianbier:
After Reading that blog it makes more sense to be in single quotes. I tested both and the command work but i've updated it to single.

leseb (approved):

leseb (dismissed):
Please squash your commits. Thanks

brianbier:
> Please squash your commits. Thanks

Any particular reason?

brianbier:
Ready 👍

booxter (approved):
Thank you!

booxter:
(The commit message could be truncated a bit though. Seems redundant. Could be smth like:)

```
docs: Fix `pip install` instructions for zsh

Signed-Off-By: ...
```

booxter (approved):
Thank you for updating the commit message!

brianbier:
@leseb any other thoughts?

bjhargrave (approved):

markstur (approved):

Issue #1866, opened 2024-07-25
Model serving with vLLM seems to be broken on Mac


jaideepr97:
**Describe the bug**

model serving with vLLM seems to be broken when supplied with a safetensors model directory even if supplying `--backend vllm` explicitly. Here is the observed output:
```
(venv) ➜  instructlab ilab model serve --backend vllm --model-path /Users/jrao/Library/Caches/instructlab/models/instructlab/merlinite-7b-lab 
Failed to determine backend: Cannot determine which backend to use: 
```

**To Reproduce**
1. Run the above command 

**Expected behavior**

vLLM server should start up without issue 

**Screenshots**

**Device Info (please complete the following information):**
 - Hardware Specs: Apple M3 Pro, 36GB memory 
 - OS Version: macOS Sonoma 14.5
 - Python Version: 3.11.9
 - InstructLab Version: 0.18.0a8.dev8

**Additional context**

jaideepr97:
closing as I was reminded that vLLM does not support mac at all

leseb:
That doesn't change the fact that the exception message is not fw'ed correctly/

PR #1868, opened 2024-07-25
fix: remove useless try-expect

leseb:
We don't expect any `ValueError` coming from `is_model_safetensors()`.
So the try-expect is not needed.

Closes: https://github.com/instructlab/instructlab/issues/1866
Co-authored-by: Ihar Hrachyshka <ihrachys@redhat.com>

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -193,23 +193,20 @@ def determine_backend(model_path: pathlib.Path) -> Tuple[str, str]:
                         - The backend to use.
                         - The reason why the backend was selected.
     """
-    try:
-        if model_path.is_dir() and is_model_safetensors(model_path):
-            if sys.platform == "linux":
-                logger.debug(
-                    f"Model is huggingface safetensors and system is Linux, using {VLLM} backend."
-                )
-                return (
-                    VLLM,
-                    "model path is a directory containing huggingface safetensors files and running on Linux.",
-                )
-            raise ValueError(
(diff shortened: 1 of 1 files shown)


booxter (approved):

src/instructlab/model/backends/backends.py
```diff
@@ -209,7 +209,7 @@ def determine_backend(model_path: pathlib.Path) -> Tuple[str, str]:
                 "If you want to run the model on a different system (e.g. macOS), please use a GGUF file."
             )
     except Exception as e:
-        raise ValueError from e
+        raise ValueError(f"{e}") from e
```
booxter, line 212: this ValueError should probably have some value-add message if we raise it at all. This is a bit duplicative. Maybe "Unexpected error happened" or smth. And to avoid duplication, this `try` block should probably look like:

```
try:
    ...
except ValueError:
    raise
except Exception as e:
    raise ValueError(<message>> from e
```

---

That said, I don't understand which other exceptions but `ValueError` we expect here at all (I checked the code in `is_model_safetensors` and I don't see where it would raise anything - assuming code is correctly written.) so maybe this `try-expect` is worthless in the first place?

leseb: You're absolutely right, the try-expect is not needed.

booxter (approved):

jaideepr97 (approved):

nathan-weinberg (approved):

PR #1869, opened 2024-07-25
Fix --help return code when config is missing (Was: Don't expect Lab or Lab.config to be a None)

booxter:
These cleanups were identified in https://github.com/instructlab/instructlab/pull/1842 and https://github.com/instructlab/instructlab/pull/1847#discussion_r1690148608

It also fixes scenario where --help was returning 2 or 0 depending on whether config file was present.

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -1417,7 +1417,6 @@ class Lab:
 def init(
     ctx: click.Context, config_file: str | os.PathLike[str], debug_level: int = 0
 ) -> None:
-    config_obj: Config
     error_msg: str | None = None
     if config_file == "DEFAULT":
         # if user passed --config DEFAULT we should ensure all proper dirs are created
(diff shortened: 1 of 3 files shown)


src/instructlab/configuration.py
```diff
@@ -637,33 +635,25 @@ def init(
             "file using `--config=<path>`."
         )
 
-    # special case: --help should always work
```
leseb, line 640: Why is the `--help` case gone?

booxter: Because `config_obj` is never `None`, so this is dead code.

And also (with config file removed):

```
.tox/py311-unit/bin/ilab --help

Options:
                 /home/vagrant/.config/instructlab/config.yaml]

Commands:

Aliases:
  chat      model chat
  convert   model convert
  diff      taxonomy diff
  download  model download
  evaluate  model evaluate
  generate  data generate
  init      config init
  serve     model serve
  sysinfo   system info
  test      model test
  train     model train
.tox/py311-unit/bin/ilab config --help
Usage: ilab config [OPTIONS] COMMAND [ARGS]...

  Command Group for Interacting with the Config of InstructLab.

Options:
  --help  Show this message and exit.

Commands:
  init  Initializes environment for InstructLab
  show  Displays the current config as YAML
```

I think `--help` is implemented by `click` itself?

booxter: Wonder if you are onto something here though. Maybe this should check `error_msg` instead of `config_obj`?

Maybe this is the case it's trying to handle:

```
.tox/py311-unit/bin/ilab config show --help
Usage: ilab config [OPTIONS] COMMAND [ARGS]...
Try 'ilab config --help' for help.

Error: `/home/vagrant/.config/instructlab/config.yaml` does not exist or is not a readable file.
Please run `ilab config init` or point to a valid configuration file using `--config=<path>`.
```

booxter: With `config_obj is None and` part removed, we can get:

```
.tox/py311-unit/bin/ilab config show --help
Usage: ilab config show [OPTIONS]

  Displays the current config as YAML

Options:
  --help  Show this message and exit.
```

booxter: I've reverted (partially) this change + added a test update that covers the --help case (it was broken before).

booxter:
Checked that --help no longer returns different values depending on config file existence. Abandoning for this reason.

PR #1880, opened 2024-07-25
Dont read default config in tests

booxter:
The tests were failing depending on whether the config file existed in the usual location.

### DIFF
--- a/tests/test_lab.py
+++ b/tests/test_lab.py
@@ -189,7 +189,7 @@ def test_ilab_missing_config(command: Command, cli_runner: CliRunner) -> None:
     result = cli_runner.invoke(lab.ilab, cmd)
 
     if command.needs_config:
-        assert result.exit_code == 2, result
+        assert result.exit_code == 2, result.stdout
         assert "does not exist or is not a readable file" in result.stdout
     else:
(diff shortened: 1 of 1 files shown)


tests/test_lab.py
```diff
@@ -180,10 +180,16 @@ def test_ilab_commands_tested():
 def test_ilab_missing_config(command: Command, cli_runner: CliRunner) -> None:
     cmd = command.get_args(default_config=False)
     assert "--config" not in cmd
+
+    # make sure cli doesn't attempt to read config from the default location
+    # (which may actually exist)
+    if command.needs_config:
+        cmd = ["--config", "./non-existing.yaml"] + cmd
```
booxter, line 187: I put this here because for some reason, `ilab --config ./xxx` doesn't behave the same way as `ilab`, (though one could expect it to, I think.) Perhaps that's a bug in CLI definition, but I am not sure. Maybe @tiran will have an idea.

tiran: You were on the right track! PR #1887 solves the problem in a generic way and also ensures that we do not introduce the same problem ever again.

booxter:
Better approach: https://github.com/instructlab/instructlab/pull/1887

PR #1882, opened 2024-07-25
fix: have `chat` and `generate` use the model being actively served

RobotSail:
Presently, the `ilab model chat` and `ilab data generate` models require
the exact path to the model being actively served. This is a bad UX
since the path had to already have been specified correctly once
when it was initially being provided to `ilab model serve`.
Furthermore, this makes it very easy to get wrong and introduce new
bugs since we are having to propogate the model actively being used in serve.
This commit fixes that by automatically detecting the model ID and using that
for the model. This should also be fine because it is not expected that
the user will be serving more than one model at a time anyway.

**Checklist:**

### DIFF
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -8,6 +8,7 @@ import click
 
 # First Party
 from instructlab import clickext
+from instructlab import client as ilabclient
 from instructlab.configuration import DEFAULTS
 
 # Local
@@ -210,6 +211,27 @@ def generate(
         except Exception as exc:
             click.secho(f"Failed to start server: {exc}", fg="red")
             raise click.exceptions.Exit(1)
+
+    # figure out what model is running
+    try:
+        models = ilabclient.list_models(
+            api_base=api_base, http_client=http_client(ctx.params)
+        )
+
+        model = models.data[0].id if models is not None else None
+        logger.debug(f"Using model from server {model}")
+    except ilabclient.ClientException as exc:
+        click.secho(
+            f"Failed to list models from {api_base}. Please check the API key and endpoint.",
+            fg="red",
+        )
+        # Right now is_temp_server() does not check if a subprocessed vllm is up
+        # shut it down just in case an exception is raised in the try
+        # TODO: revise is_temp_server to check if a vllm server is running
+        if backend_instance is not None:
+            backend_instance.shutdown()
+        raise click.exceptions.Exit(1) from exc
+
     try:
         click.echo(
             f"Generating synthetic data using '{model}' model, taxonomy:'{taxonomy_path}' against {api_base} server"
--- a/src/instructlab/model/chat.py
+++ b/src/instructlab/model/chat.py
@@ -195,56 +195,27 @@ def chat(
     # in server mode (`ilab model serve` is running somewhere, or we are talking to another
     # OpenAI compatible endpoint).
     if not is_temp_server_running():
-        # Try to get the model name right if we know we're talking to a local `ilab model serve`.
-        #
-        # If the model from the CLI and the one in the config are the same, use the one from the
-        # server if they are different else let's use what the user provided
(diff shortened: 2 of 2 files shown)


src/instructlab/model/chat.py
```diff
@@ -195,56 +195,27 @@ def chat(
...
-        # from the config is the same as the default value, then the user didn't provide a value
-        # we then compare it with the value from the server to see if it's different
-        if (
-            # We need to get the base name of the model because the model path is a full path and
-            # the once from the config is just the model name
-            os.path.basename(model) == cfg.DEFAULTS.GGUF_MODEL_NAME
-            and os.path.basename(ctx.obj.config.chat.model)
-            == cfg.DEFAULTS.GGUF_MODEL_NAME
-            and api_base == ctx.obj.config.serve.api_base()
-        ):
```
leseb, line 215: What's the reason to stop honoring the `-m` flag here? This looks like a regression.

src/instructlab/data/generate.py
```diff
@@ -210,6 +211,27 @@ def generate(
         except Exception as exc:
             click.secho(f"Failed to start server: {exc}", fg="red")
             raise click.exceptions.Exit(1)
+
+    # figure out what model is running
```
leseb, line 215: I believe the logic should be identical to what chat was doing before this patch.

leseb (changes requested):

src/instructlab/data/generate.py
```diff
@@ -210,6 +211,27 @@ def generate(
...
+    try:
+        models = ilabclient.list_models(
+            api_base=api_base, http_client=http_client(ctx.params)
+        )
+
+        model = models.data[0].id if models is not None else None
```
booxter, line 221: so you now always override what is passed via --model arg? Do we need the argument then?

src/instructlab/data/generate.py
```diff
@@ -210,6 +211,27 @@ def generate(
...
+        logger.debug(f"Using model from server {model}")
+    except ilabclient.ClientException as exc:
+        click.secho(
+            f"Failed to list models from {api_base}. Please check the API key and endpoint.",
+            fg="red",
+        )
+        # Right now is_temp_server() does not check if a subprocessed vllm is up
```
booxter, line 228: you mean `is_temp_server_running`?

src/instructlab/data/generate.py
booxter, line 221: is it always correct to take the first model in the list? I see the commit message claims:

```
it is not expected that
the user will be serving more than one model at a time anyway
```

But is it always true? Can't we `generate` against a different server, not the one running as `serve`? AFAIU the `api_base` can point to any `OpenAI` compatible endpoint.

booxter, line 221: nit: This should probably gracefully handle a degenerate case where `data` is empty (no models are served).

booxter, line 221: Is `None` a good default for `model` though? I suspect the generation will crash if you don't pass anything. Should you handle it gracefully by raising an error?

src/instructlab/data/generate.py
```diff
@@ -210,6 +211,27 @@ def generate(
...
+        # shut it down just in case an exception is raised in the try
+        # TODO: revise is_temp_server to check if a vllm server is running
+        if backend_instance is not None:
```
booxter, line 231: This cleanup is already done by the `finally` section at the very end of this function. Just move this `try-except` block under the `try` started below in line 235.

booxter:
@RobotSail this is set as a blocker. Would you mind respinning it to resolve git conflicts and comments?

ktam3:
Per convo with Oleg, closing this PR in favor of

https://github.com/instructlab/instructlab/pull/1992
https://github.com/instructlab/instructlab/pull/1990

PR #1885, opened 2024-07-25
refactor: Misc changes noticed and proposed in `model list` PR

booxter:
This is a refactoring follow-up from  https://github.com/instructlab/instructlab/pull/1840

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -167,24 +167,23 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
     # Third Party
     from gguf.constants import GGUF_MAGIC
 
-    try:
-        with open(model_path, "rb") as f:
-            # Memory-map the file on the first 4 bytes (this is where the magic number is)
-            mmapped_file = mmap.mmap(f.fileno(), length=4, access=mmap.ACCESS_READ)
-
-            # Read the first 4 bytes
-            first_four_bytes = mmapped_file.read(4)
-
-            # Convert the first four bytes to an integer
-            first_four_bytes_int = int(struct.unpack("<I", first_four_bytes)[0])
-
-            # Close the memory-mapped file
-            mmapped_file.close()
-
-            return first_four_bytes_int == GGUF_MAGIC
-    except IsADirectoryError as exc:
-        logger.debug(f"GGUF Path is a directory, returning {exc}")
+    if not model_path.is_file():
         return False
+
+    with model_path.open("rb") as f:
+        # Memory-map the file on the first 4 bytes (this is where the magic number is)
+        mmapped_file = mmap.mmap(f.fileno(), length=4, access=mmap.ACCESS_READ)
+
+        # Read the first 4 bytes
+        first_four_bytes = mmapped_file.read(4)
+
+        # Close the memory-mapped file
+        mmapped_file.close()
+
+    # Convert the first four bytes to an integer
+    first_four_bytes_int = int(struct.unpack("<I", first_four_bytes)[0])
+
+    return first_four_bytes_int == GGUF_MAGIC
 
 
 def determine_backend(model_path: pathlib.Path) -> Tuple[str, str]:
--- a/src/instructlab/model/list.py
+++ b/src/instructlab/model/list.py
@@ -2,6 +2,7 @@
 
 # Standard
 from pathlib import Path
+from typing import Generator, Tuple
 import logging
 import os
 import time
@@ -15,6 +16,9 @@ from instructlab.configuration import DEFAULTS
 from instructlab.model.backends.backends import is_model_gguf, is_model_safetensors
 
 logger = logging.getLogger(__name__)
+
+ModelInfo = Tuple[str, str, str]
+TableRow = ModelInfo
 
 
 @click.command(name="list")
@@ -34,32 +38,35 @@ def model_list(model_dirs: list[str], list_checkpoints: bool):
     """lists models"""
     data = []
     for directory in model_dirs:
-        for entry in Path(directory).iterdir():
-            # if file, just tally the size. This must be a GGUF.
-            if entry.is_file() and is_model_gguf(entry):
-                model, age, size = _analyze_gguf(entry)
-                data.append([model, age, size])
+        d = Path(directory)
+        if not d.is_dir():
+            continue
+        for entry in d.iterdir():
+            if is_model_gguf(entry):
+                data.append(_analyze_gguf(entry))
             elif entry.is_dir():
-                for m in _analyze_dir(entry, list_checkpoints, directory):
-                    data.append(m)
-    print_table(["Model Name", "Last Modified", "Size"], data)
+                data += list(_analyze_dir(entry, list_checkpoints, directory))
+    print_table(("Model Name", "Last Modified", "Size"), data)
 
 
-# convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MiB, GiB)
-def convert_bytes_to_proper_mag(f_size: float):
-    magnitudes = ["KiB", "MiB", "GiB", "TiB"]
+def convert_bytes_to_human_readable_size(n_bytes: int) -> str:
+    """
+    Convert bytes size to a human readable string with the appropriate magnitude. (MiB, GiB, ...)
+    """
     magnitude = "B"
-    for mag in magnitudes:
-        if f_size >= 1024:
-            magnitude = mag
-            f_size /= 1024
-        else:
-            return f_size, magnitude
-    return f_size, magnitude
+    f_bytes = float(n_bytes)
+    for mag in ("KiB", "MiB", "GiB", "TiB"):
+        if f_bytes < 1024:
+            break
+        magnitude = mag
+        f_bytes /= 1024
+    return f"{f_bytes:.2f} {magnitude}"
 
 
-# print_table displays the models on the system in a nicely formatted table
-def print_table(headers: list[str], data: list[list[str]]):
+def print_table(headers: TableRow, data: list[TableRow]):
+    """
+    Displays rows of data in a nicely formatted table.
+    """
     column_widths = [
         max(len(str(row[i])) for row in data + [headers]) for i in range(len(headers))
     ]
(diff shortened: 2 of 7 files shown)


src/instructlab/model/list.py
```diff
@@ -34,32 +38,35 @@ def model_list(model_dirs: list[str], list_checkpoints: bool):
...
-
-# print_table displays the models on the system in a nicely formatted table
-def print_table(headers: list[str], data: list[list[str]]):
+    f_bytes = float(n_bytes)
+    for mag in ("KiB", "MiB", "GiB", "TiB"):
+        if f_bytes < 1024:
+            break
+        magnitude = mag
+        f_bytes /= 1024
+    return f"{.2} {magnitude}"
```
leseb, line 63:
```suggestion
    return f"{.1} {magnitude}"
```

non-blocking: I think a single digit after the decimal point is sufficient.

src/instructlab/model/backends/backends.py
```diff
@@ -167,24 +166,16 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
     # Third Party
     from gguf.constants import GGUF_MAGIC
 
-    try:
-        with open(model_path, "rb") as f:
-            # Memory-map the file on the first 4 bytes (this is where the magic number is)
-            mmapped_file = mmap.mmap(f.fileno(), length=4, access=mmap.ACCESS_READ)
```
leseb, line 173: I'm missing the power of mmap already 😢

src/instructlab/model/list.py
```diff
@@ -34,32 +38,35 @@ def model_list(model_dirs: list[str], list_checkpoints: bool):
...
-                    data.append(m)
-    print_table(["Model Name", "Last Modified", "Size"], data)
+                data += list(_analyze_dir(entry, list_checkpoints, directory))
+    print_table(("Model Name", "Last Modified", "Size"), data)
 
 
-# convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MiB, GiB)
-def convert_bytes_to_proper_mag(f_size: float):
-    magnitudes = ["KiB", "MiB", "GiB", "TiB"]
+def get_bytes_size_string(n_bytes: int) -> str:
```
leseb, line 52: I think the newly proposed name is less clear. How about `convert_bytes_to_human_readable_size`? Something along those lines might be more descriptive.

leseb (changes requested):

src/instructlab/model/list.py
```diff
@@ -80,72 +87,36 @@ def print_table(headers: list[str], data: list[list[str]]):
...
-    add_model = False
-    models = []
-    # walk entire dir.
-    for root, _, files in os.walk(entry.as_posix()):
-        normalized_path = os.path.normpath(root)
-        # Split the path into its components
-        parts = normalized_path.split(os.sep)
-        # Get the last two parts and join them back into a path
-        last_two_parts = os.path.join(parts[-2], parts[-1])
-        # if this is a dir it could be:
```
cdoern, line 114: can we please preserve these comments, they are useful

src/instructlab/model/list.py
```diff
@@ -34,32 +38,35 @@ def model_list(model_dirs: list[str], list_checkpoints: bool):
     """lists models"""
     data = []
     for directory in model_dirs:
-        for entry in Path(directory).iterdir():
-            # if file, just tally the size. This must be a GGUF.
```
cdoern, line 38: same here

src/instructlab/model/list.py
RobotSail: I disagree, we should keep it to 2 digits - this is also how podman does it.

src/instructlab/model/list.py
booxter: OK I will rewrite them a bit then.

src/instructlab/model/backends/backends.py
```diff
@@ -167,24 +166,16 @@ def is_model_gguf(model_path: pathlib.Path) -> bool:
...
-
-            # Read the first 4 bytes
-            first_four_bytes = mmapped_file.read(4)
+    if not model_path.is_file():
+        return False
 
-            # Convert the first four bytes to an integer
-            first_four_bytes_int = int(struct.unpack("<I", first_four_bytes)[0])
+    with model_path.open("rb") as f:
```
RobotSail, line 172: Should we specify the encoding here?

src/instructlab/model/list.py
booxter: I hope we agree that both `.1` and `.2` are better than unlimited (which is the current status). I picked `.2` because that's what @RobotSail used in his `data list` PR. I think it's a reasonable value. I am not a UX designer to reasonably argue for one or another. I'll keep as-is but feel free to push a follow-up changing it to something else.

src/instructlab/model/list.py
booxter: @cdoern the code now is:

```
if is_gguf(...):
    data.append(analyze_gguf(...))
```

The comment doesn't apply anymore, I think. Whatever I would write now, it seems repetitive. Do you have an idea what I should capture here now to clarify the code? (Perhaps I don't see what needs clarification in this section. Please elaborate.)

src/instructlab/model/list.py
```diff
@@ -80,72 +87,36 @@ def print_table(headers: list[str], data: list[list[str]]):
...
-                    f"{all_files_sizes:.1f} {magnitude}",
-                ]
-            )
-    return models
+) -> Generator[ModelInfo, None, None]:
+    list_all = list_checkpoints and directory == DEFAULTS.CHECKPOINTS_DIR
+    # TODO: switch to Path.walk after python3.12 bump
+    for root_, _, files in os.walk(entry.as_posix()):
+        root = Path(root_)
+        if not (list_all or is_model_safetensors(root)):
```
booxter, line 110: methinks now I should have is_model_gguf here back. Thought it's useless but I misread the intent. The intent I think is when pointing at the top level repo dir

booxter: Hm, maybe not. `root` is NEVER a file here. So it will never be a gguf. Perhaps the intent (but not implementation) of `model list` was to drill down the directory and report on each found gguf file? But then one would have to check is_model_* on `files`, not on `root` (as done in the original code and with the refactor).

@cdoern what was the intent of `is_model_gguf` check here?

src/instructlab/model/backends/backends.py
booxter: Binary read ("b") means it [won't](https://docs.python.org/3/library/functions.html#open) be interpreted.

```
As mentioned in the Overview, Python distinguishes between binary and text I/O. Files opened in binary mode (including 'b' in the mode argument) return contents as bytes objects without any decoding. In text mode (the default, or when 't' is included in the mode argument), the contents of the file are returned as str, the bytes having been first decoded using a platform-dependent encoding or using the specified encoding if given.
```

src/instructlab/model/list.py
booxter: I've updated as per my understanding of the code, but feel free to correct me. I think either `gguf` cases are not applicable, or the code has a bug. You decide what the intent was.

src/instructlab/model/list.py
```diff
@@ -34,32 +38,35 @@ def model_list(model_dirs: list[str], list_checkpoints: bool):
...
-
-# print_table displays the models on the system in a nicely formatted table
-def print_table(headers: list[str], data: list[list[str]]):
+    f_bytes = float(n_bytes)
+    for mag in ("KiB", "MiB", "GiB", "TiB"):
+        if f_bytes < 1024:
+            break
+        magnitude = mag
+        f_bytes /= 1024
+    return f"{.2} {magnitude}"
```
booxter, line 63: I think I may have messed here. Shouldn't it refer to `f_bytes` in the f-string? (Stunned since I think it worked before? Will have to check.)

src/instructlab/model/list.py
```diff
@@ -34,32 +38,35 @@ def model_list(model_dirs: list[str], list_checkpoints: bool):
...
-                    data.append(m)
-    print_table(["Model Name", "Last Modified", "Size"], data)
+                data += list(_analyze_dir(entry, list_checkpoints, directory))
+    print_table(("Model Name", "Last Modified", "Size"), data)
 
 
-# convert_bytes_to_proper_mag takes a dir/file size in Bytes and converts it to the proper Magnitude (MiB, GiB)
-def convert_bytes_to_proper_mag(f_size: float):
-    magnitudes = ["KiB", "MiB", "GiB", "TiB"]
+def convert_bytes_to_human_readable_size(n_bytes: int) -> str:
```
RobotSail, line 52: We shouldn't do this because it couples the implementation of this function too tightly to where it gets used. This will make things more complicated down the road. We should instead just return a float along with the magnitude representation, and then let the caller do whatever with it as they like. 
The other problem here is that it assumes what the display decides to round the number to. What if we want to do 2 floating points in one place, but actually check the whole value somewhere else? Now we need to add in an additional `precision` arg, etc.

RobotSail (commented):
Thank you for this PR, I appreciate you taking the time to clean up our codebase. Apart from a few changes, I think this will be useful to have. The only concern I have here is that we're fixing a lot of things all at once instead of targeting specific things that we want to fix.

While I personally prefer to fix things in one go, I think the workflow for this project mandates that things are fixed in atomic PRs. I recommend splitting this PR up into smaller ones that are more specific about what the intention is. If you do that, I think we could get your fixes in relatively quickly.

src/instructlab/model/list.py
booxter: What I'm trying to avoid here is that this code in this module duplicates this conversion and calculation. The part that produces `Tuple[float, str]` may be spun out into a separate function for others' use (I assume you'd like to see it in `utils.py`?)

booxter:
The original PR merged in a single piece. I think it's applicable to merge the follow-up PR that addresses its **ignored** comments in a single piece too. (The only piece here that was **not** pointed out in the original PR is the `mmap` removal. I see why one may argue to split this change out, _kinda_.)

Otherwise, the precedent is set that it becomes harder to clean tech debt up than producing it. (Juggling a dozen of small PRs has its cost.)

@cdoern please check the PR again. I'm specifically looking for confirmation that `is_model_gguf` check was not needed (or - alternatively - that the original PR has a bug?) Thanks.

booxter:
FYI I split out the mmap patch here: https://github.com/instructlab/instructlab/pull/1893 since it was not directly related to the PR this is a follow-up for.

src/instructlab/model/list.py
booxter: I think this PR may actually fix the gguf handling: https://github.com/instructlab/instructlab/pull/1895 So I guess it was intended, but broken before?

src/instructlab/model/list.py
booxter: @jaideepr97 FYI I think you are fixing this bug mentioned above in your PR

booxter:
The code base moved so far that rebasing this PR would be like starting from scratch. We will clean some of this stuff up later if we touch the relevant code again.

PR #1889, opened 2024-07-26
Add the checkpoint-dir arg

derekhiggins:


markmc:
Couple of todos here:

1. Require `instructlab-sdg>=0.2.2` for this feature. See https://github.com/instructlab/sdg/releases/tag/v0.2.2
2. I'm not sure this needs to be a parameter - why not just generate these checkpoints by default (if batching is enabled) in the same directory (or a subdir) as the generated samples?

markmc:
from @shivchander 

> +1 on this,  I believe this is how it is currently setup on the research fork. I dont see a need to expose checkpoint-dir explicitly, we already let users configure output_dir, i think having it as a subdir under this works

markmc:
Replaced by https://github.com/instructlab/instructlab/pull/1915

PR #1897, opened 2024-07-29
RHELAI-625: set the teacher.model_path under generate

eranco74:
This should fix 'ilab data generate' failing on not finding /root/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf (default model) in case it was initialized with another model

**Issue resolved by this Pull Request:**
Resolves #RHELAI-625

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [x] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [x] Documentation has been updated, if necessary.
- [x] Unit tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #1901, opened 2024-07-29
set teacher model in generate to the value passed via `--model`

cdoern:
`ilab data generate` ignores `--model` and instead uses the teacher model in the cfg. honor `--model` as an override

**Checklist:**

### DIFF
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -19,8 +19,10 @@ logger = logging.getLogger(__name__)
 @click.command()
 @click.option(
     "--model",
-    default=lambda: DEFAULTS.DEFAULT_MODEL,
-    show_default="The default model used by the instructlab system, located in the data directory.",
+    "model_path",
+    cls=clickext.ConfigOption,
+    config_sections="teacher",
+    show_default=False,
     help="Name of the model used during generation.",
 )
 @click.option(
@@ -156,7 +158,7 @@ logger = logging.getLogger(__name__)
 @clickext.display_params
 def generate(
     ctx,
-    model,
+    model_path,
     num_cpus,
     num_instructions,
     sdg_scale_factor,
@@ -210,8 +212,11 @@ def generate(
         from instructlab.model.backends import backends
         from instructlab.model.backends.llama_cpp import Server as llama_cpp_server
 
+        # TODO (cdoern): we really should not edit the cfg object
         ctx.obj.config.serve.llama_cpp.llm_family = model_family
-        backend_instance = backends.select_backend(ctx.obj.config.generate.teacher)
+        backend_instance = backends.select_backend(
+            cfg=ctx.obj.config.generate.teacher, model_path=model_path
+        )
 
         try:
             # Run the backend server
@@ -231,14 +236,14 @@ def generate(
             batch_size = 0
     try:
         click.echo(
-            f"Generating synthetic data using '{model}' model, taxonomy:'{taxonomy_path}' against {api_base} server"
+            f"Generating synthetic data using '{model_path}' model, taxonomy:'{taxonomy_path}' against {api_base} server"
         )
         generate_data(
             logger=logging.getLogger("instructlab.sdg"),  # TODO: remove
             api_base=api_base,
             api_key=api_key,
             model_family=model_family,
-            model_name=model,
+            model_name=model_path,
             num_cpus=num_cpus,
             num_instructions_to_generate=sdg_scale_factor,
             taxonomy=taxonomy_path,


src/instructlab/data/generate.py
```diff
@@ -210,7 +210,9 @@ def generate(
         from instructlab.model.backends import backends
         from instructlab.model.backends.llama_cpp import Server as llama_cpp_server
 
+        # TODO (cdoern): we really should not edit the cfg object
```
booxter, line 213: ok then why not just pass the `model` into `select_backend`?

cdoern: `select_backend` should be considered relatively frozen ATM. I don't really want to alter functionality those close to a release which many components depend on

booxter: I mean... you pass the value into the function as an argument, no? see line 216. why not just pass it without setting on `ctx.obj`?

booxter: Disregard, I misread that `select_backend` doesn't accept just the model, but the whole config section.

booxter: But I think you are breaking customizing the model via config file:


```
$ cat ~/.config/instructlab/config.yaml | grep -A 8 teacher
  teacher:
    backend: null
    chat_template: null
    host_port: 127.0.0.1:8000
    llama_cpp:
      gpu_layers: -1
      llm_family: ''
      max_ctx_size: 4096
    model_path: INVALID_MODEL
$ .tox/py311-unit/bin/ilab data generate
INFO 2024-07-29 15:58:21,301 numexpr.utils:161: NumExpr defaulting to 4 threads.
INFO 2024-07-29 15:58:21,741 datasets:58: PyTorch version 2.3.1+cpu available.
Model used: /home/vagrant/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf

```

RobotSail (approved):
LGTM

src/instructlab/data/generate.py
booxter: (that is with:)

```
diff --git a/src/instructlab/data/generate.py b/src/instructlab/data/generate.py
index 1da2aba..29b35ec 100644
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -213,6 +213,7 @@ def generate(
         # TODO (cdoern): we really should not edit the cfg object
         ctx.obj.config.serve.llama_cpp.llm_family = model_family
         ctx.obj.config.generate.teacher.model_path = model
+        print("Model used:", ctx.obj.config.generate.teacher.model_path)
         backend_instance = backends.select_backend(ctx.obj.config.generate.teacher)

         try:

```

booxter (changes requested):
This breaks custom teacher model via config. This should be fixed. This also shows that there are no tests for this scenario. (They should be added; also this --model fix should get its own test scenario.)

cdoern:
ok @booxter this now reads `--model` via the `ConfigOption` class, I pass the var directly to the `select_backends` method rather than changing the ctx.cfg

src/instructlab/data/generate.py
```diff
@@ -210,8 +210,11 @@ def generate(
...
         ctx.obj.config.serve.llama_cpp.llm_family = model_family
-        backend_instance = backends.select_backend(ctx.obj.config.generate.teacher)
+        backend_instance = backends.select_backend(
+            cfg=ctx.obj.config.serve, model_path=model
+        )
```
jaideepr97, line 217: @cdoern should we add a check to make sure `model != default` before essentially overriding what is supplied in the ctx? To make sure we dont replace a custom mdoel with the default as @booxter pointed out

jaideepr97: nvm, sounds like this comment might be obsolete now

src/instructlab/data/generate.py
```diff
@@ -210,8 +210,11 @@ def generate(
         from instructlab.model.backends import backends
         from instructlab.model.backends.llama_cpp import Server as llama_cpp_server
 
+        # TODO (cdoern): we really should not edit the cfg object
         ctx.obj.config.serve.llama_cpp.llm_family = model_family
-        backend_instance = backends.select_backend(ctx.obj.config.generate.teacher)
+        backend_instance = backends.select_backend(
+            cfg=ctx.obj.config.serve, model_path=model
```
booxter, line 216: so why do we now ignore `generate.teacher` section? It seems like it no longer used anywhere.

```
$ git grep '\.teacher'
tests/test_config.py:        assert cfg.generate.teacher.model_path == default_model
tests/test_config.py:        assert cfg.generate.teacher.llama_cpp is not None
tests/test_config.py:        assert cfg.generate.teacher.llama_cpp.gpu_layers == -1
tests/test_config.py:        assert cfg.generate.teacher.llama_cpp.max_ctx_size == 4096
tests/test_config.py:        assert cfg.generate.teacher.llama_cpp.llm_family == ""
tests/test_config.py:        assert cfg.generate.teacher.vllm is not None
tests/test_config.py:        assert cfg.generate.teacher.vllm.vllm_args == []
tests/test_config.py:        assert cfg.generate.teacher.host_port == "127.0.0.1:8000"
tests/test_config.py:        assert cfg.generate.teacher.backend is None
```

booxter (changes requested):
It does seem to switch from .generate.teacher to .serve section. I don't think it should be done though. Why?

---

Also, tests should be added for scenarios that this PR may affect.

nathan-weinberg:
Is this a dup of https://github.com/instructlab/instructlab/pull/1897?

cdoern:
nope. #1897 seems to be funneling the `cfg` option in properly in init.py @nathan-weinberg this PR deals with `--model` being ignored. I'll rename

alimaredia:
@cdoern Great catch on the flag not being useful anymore Have you read through the discussion in https://github.com/instructlab/instructlab/pull/1777 for why that teacher section was introduced in the config in first place?

What if for the time being we removed the `--model` flag entirely? Users can still use whatever model they want by running `ilab serve` separately and passing the endpoint of served there to `--endpoint`. Using a custom endpoint overrides the teacher background serving paths. 

If we stick with the `--model` and had it override the generate.teacher.model_path like you have in the first iteration of the PR, we don't have flags to override the other serving options for it (ex: vllm_args) for it.

cdoern:
@alimaredia we should not remove `--model`. Being able to serve -> generate in one command is pretty nice UX as compared to the alternative.

The fix here is pretty simple now that we have figured out the root cause. I'd prefer not to bikeshed too far given that this is fixing currently broken functionality

alinaryan (approved):
This LGTM given how close we are to a release. IMO, removing --model entirely could be revisited after GA.

cdoern:
I am kind of anti adding tests to this for two reasons

1. folks need to start running the e2e workflow this afternoon using `--model`
2. we currently don't have `test_lab_generate.py` so this PR would need to introduce a new test suite for a relatively small change

if people feel really strongly about adding tests I can do it, but this is pretty trivial

booxter (approved):
Code is fine now, and I am ok with merging it.

---

I still would love to see tests being added as the default behavior for changes in the project. (Unless coverage is already present of course.)

alimaredia:
This change seems fine to me. It fixes a broken flag, and changes the default value of the correct flag to look in the correct place which is the generate.teacher.model_path.

That being said, right now `--model` is the only flag that changes anything in the teacher config. Follow up work should be done to either add more flags to configure the teacher model, or possibly remove them. We might also consider renaming `--model` to `--teacher`.

Issue #1899, opened 2024-07-29
ilab taxonomy diff should account for .yaml and .yml


alinaryan:
in ilab taxonomy diff, both .yml and .yaml should be accounted for

nathan-weinberg:
@bjhargrave just want to confirm that `.yml` is supported in the taxonomy?

nathan-weinberg:
@cdoern @alinaryan if that's the case this may not be a bug, but perhaps I can add a clearer message?

cdoern:
it is still a bug given that the error for this is pretty unclear

bjhargrave:
There have been past issues on this topic which have been closed. Also the new taxonomy reading api in the schema repo instructlab/schema#33 should have a better error message for this.

tiran:
Previous discussion: #1306

alinaryan:
Thanks, will close this one

Issue #1306, opened 2024-06-07
Support .yml extension in addition to .yaml for qna files


gfbarros:
**Is your feature request related to a problem? Please describe.**
The fairly common .yml extension for yaml files is quietly ignored when handling new taxonomy qna files. 

**Describe the solution you'd like**
Allow using .yml files for taxonomy.

bjhargrave:
The taxonomy repo requires `qna.yaml`. I think the backend does also. So any change to allow `qna.yml` would need to be properly coordinated through all the stakeholder projects.

gfbarros:
I checked the taxonomy and backed repo files for mentions of yaml and didn't see this. Could you show me where I missed it?

bjhargrave:
> Taxonomy skill files must be a valid [YAML](https://yaml.org/) file named qna.yaml.

https://github.com/instructlab/taxonomy/blob/main/README.md#getting-started-with-skill-contributions



https://github.com/instructlab/taxonomy/blob/main/.github/scripts/check-yaml.py#L129-L133

are some examples.

tiran:
I would prefer if we would stick to a consistent `.yaml` and not add support for `.yml` extension. A second extension adds a very minor UX improvement. Instead of support `.yml`, how about `ilab diff` checks for `.yml`, informs the user to rename the file, and fails?

markstur:
it's debatable because both .yml and .yaml are commonly used, but allowing both can result in a messy taxonomy with both qna.yml and qna.yaml (even in the same dir).  So I felt from the beginning it's just best to allow one and throw an error on the other.  I think it is just a warning.  Is the warning clear enough to make it an easy fix?

Otherwise open to debate about allowing both but inconsistency is bad

leseb:
I think the project should stay consistent with the extensions we support, so I'd prefer to stick with .yaml. For what it's worth, the [YAML project](https://github.com/search?q=org%3Ayaml+.yaml+language%3AYAML&type=code&l=YAML) itself uses .yaml, so let's follow the standard.

leseb:
It looks like the consensus is to not go with another extension and stay we `.yaml`, however @tiran's [proposal](https://github.com/instructlab/instructlab/issues/1306#issuecomment-2158123655) is a good middle-ground.

@gfbarros would you like to take a stab at it? Thanks!

gfbarros:
Yes, I'll take a stab at it later this week. thx.

PR #1905, opened 2024-07-29
refactor: strictly require `qna.yaml` filename for `ilab taxonomy diff`

nathan-weinberg:
As discussed in https://github.com/instructlab/instructlab/issues/1899 and https://github.com/instructlab/instructlab/issues/1306 there is currently no indication to CLI users that `.yml` files are not supported for the taxonomy. This PR adds a warning log line to the CLI output to communicate that any files in the taxonomy not named `qna.yaml` are invalid and won't be checked as part of `ilab taxonomy diff`.

Example output:
```bash
(venv) [ec2-user@ip-10-0-7-10 instructlab]$ ilab taxonomy diff
WARNING 2024-07-30 19:18:27,919 instructlab.utils:174: Found a 'qna.yml' file: compositional_skills/writing/freeform/foo-lang/qna.yml: taxonomy files must be named 'qna.yaml'. File will not be checked.
WARNING 2024-07-30 19:18:27,937 instructlab.utils:174: Found a 'qna.yml' file: compositional_skills/writing/freeform/foo-lang/qna.yml: taxonomy files must be named 'qna.yaml'. File will not be checked.
Taxonomy in /home/ec2-user/.local/share/instructlab/taxonomy is valid :)
```

**Checklist:**

### DIFF
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -416,6 +416,7 @@ def validate_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
         )
         warnings += 1
         return warnings, errors
+    # update path to point directly at file if it is not already
     for i in range(len(file_path_p.parts) - 1, -1, -1):
         if file_path_p.parts[i] in TAXONOMY_FOLDERS:
             taxonomy_path = Path(*file_path_p.parts[i:])
@@ -437,11 +438,13 @@ def validate_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
             errors += 1
             return warnings, errors
 
-        # do general YAML linting if specified
+        # do general YAML linting
         version = get_version(contents)
         if version > 1:  # no linting for version 1 yaml
-            if yaml_rules is not None:
-                if os.path.isfile(yaml_rules):
+            if yaml_rules is not None:  # user attempted to pass custom yaml rules file
+                if os.path.isfile(
+                    yaml_rules
+                ):  # custom rules file was found, use specified config
                     logger.debug(f"Using YAML rules from {yaml_rules}")
                     yamllint_cmd = [
                         "yamllint",
@@ -452,7 +455,7 @@ def validate_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
                         str(file_path_p),
                         "-s",
                     ]
-                else:
+                else:  # custom rules file was not found, use default config
                     logger.debug(f"Cannot find {yaml_rules}. Using default rules.")
                     yamllint_cmd = [
                         "yamllint",
@@ -463,7 +466,7 @@ def validate_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
                         str(file_path_p),
                         "-s",
                     ]
-            else:
+            else:  # no custom rules file was specified, use default config
                 yamllint_cmd = [
                     "yamllint",
                     "-f",
@@ -473,6 +476,7 @@ def validate_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
                     str(file_path_p),
                     "-s",
                 ]
+            # run YAML linter via yamllint subprocess
             try:
                 subprocess.check_output(yamllint_cmd, text=True)
             except subprocess.CalledProcessError as e:
@@ -486,6 +490,7 @@ def validate_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
                 logger.error("\n".join(lint_messages))
                 return warnings, errors
 
+        # validate file matches InstructLab Taxonomy Schema
         validation_errors = validate_yaml(contents, taxonomy_path)
         if validation_errors:
             errors += validation_errors
@@ -508,7 +513,7 @@ def validate_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
 
 
 # TODO: remove `_logger` parameter after instructlab.sdg is fixed.
-def validate_taxonomy(_logger, taxonomy, taxonomy_base, yaml_rules):
+def validate_taxonomy(_logger, taxonomy, taxonomy_base, yaml_rules) -> None:
     if os.path.isfile(taxonomy):
         warnings, errors = validate_taxonomy_file(taxonomy, yaml_rules)
         if warnings:
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -166,14 +166,18 @@ TAXONOMY_FOLDERS: List[str] = ["compositional_skills", "knowledge"]
 """Taxonomy folders which are also the schema names"""
 
 
-def istaxonomyfile(fn):
+def is_taxonomy_file(fn: str):
     path = Path(fn)
+    if path.suffix == ".yml" and path.parts[0] in TAXONOMY_FOLDERS:
+        logger.warning(
+            f"Found a '.yml' file: {path}: taxonomy files must have a '.yaml' extension. File will not be checked."
+        )
     return path.suffix == ".yaml" and path.parts[0] in TAXONOMY_FOLDERS
 
 
 def get_taxonomy_diff(repo="taxonomy", base="origin/main"):
     repo = git.Repo(repo)
-    untracked_files = [u for u in repo.untracked_files if istaxonomyfile(u)]
+    untracked_files = [u for u in repo.untracked_files if is_taxonomy_file(u)]
 
     branches = [b.name for b in repo.branches]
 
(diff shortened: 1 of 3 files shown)


src/instructlab/utils.py
```diff
@@ -164,14 +164,18 @@ def lab_check_callback(*args, **kwargs):
 """Taxonomy folders which are also the schema names"""
 
 
-def istaxonomyfile(fn):
+def is_taxonomy_file(fn: str):
```
bjhargrave, line 167: Only `qna.yaml` is a valid taxonomy yaml. This code currently passes `foo.yaml` which then blows up later. We should probably warn if any `*.ya?ml` other than `qna.yaml` is found.

src/instructlab/utils.py
```diff
@@ -164,14 +164,18 @@ def lab_check_callback(*args, **kwargs):
...
     path = Path(fn)
+    if path.suffix == ".yml" and path.parts[0] in TAXONOMY_FOLDERS:
+        logger.warning(
+            f"Found a '.yml' file: {path}: taxonomy files must have a '.yaml' extension. File will not be checked."
+        )
     return path.suffix == ".yaml" and path.parts[0] in TAXONOMY_FOLDERS
```
booxter, line 173: should we just log when this returns False?.. (Plus adding the validation for `qna.yaml` full name, as @bjhargrave suggested above.)

booxter (commented):
This will work, but probably can be improved somewhat.

src/instructlab/utils.py
bjhargrave: It can return false for attribution.txt which should not log a message.

src/instructlab/utils.py
cdoern: in my testing all file names work, not just qna.yaml. Is there somewhere in the code that only qna.yaml is accepted?

src/instructlab/utils.py
nathan-weinberg: Right, and `attribution.txt` is a valid file - this particular function is only used to determine whether or not a file pointer is externally recognizable as a possible taxonomy entry - additional functions beyond this do checks of
1. Byte content (attempt loading and check if the content is not empty and matches the generic `Mapping` type)
2. Generic YAML linting 
3. Specific InstructLab Schema Validation

nathan-weinberg: I added a commit to explain this more in-line in the code

src/instructlab/utils.py
```diff
@@ -166,14 +166,18 @@ def lab_check_callback(*args, **kwargs):
 """Taxonomy folders which are also the schema names"""
 
 
-def istaxonomyfile(fn):
+def is_taxonomy_file(fn: str):
     path = Path(fn)
+    if path.suffix == ".yml" and path.parts[0] in TAXONOMY_FOLDERS:
```
booxter, line 171: nit: you check that the file is in taxonomy folder twice now. We could:

```
if path.parts[0] not in TAXONOMY_FOLDERS:
    return False
if path.suffix == ".yml":
    # only log for .yml; other extensions are fine
    logger.warning(...)
    return False
if path.suffix == ".yaml":
    return True
```

booxter (approved):
Thank you. This can merge I think.

src/instructlab/utils.py
nathan-weinberg: Ooh I like this, thanks!

src/instructlab/utils.py
bjhargrave: It is quite heavily documented in the taxonomy repo (and enforced by the CI) that the 
taxonomy file is named qna.yaml.

> Taxonomy skill files must be a valid [YAML](https://yaml.org/) file named qna.yaml.

https://github.com/instructlab/taxonomy/blob/2e445afe323db0bcaf4d57268a3fb7279e3c3b69/scripts/check-yaml.py#L126-L130

 All doc examples in this repo use qna.yaml.

src/instructlab/utils.py
nathan-weinberg: Did something _slightly_ different, lmk what you think :+1:

src/instructlab/utils.py
bjhargrave: We also will need to allow md files in the folder as someone could refer to a prior commit in their taxonomy repo for knowledge documents. Since the folder can contain other files than the qna, this is why the qna.yaml file name is important.

src/instructlab/utils.py
bjhargrave: I think we still need to confirm the file name is qna.yaml not just any file with a yaml suffix. I have already encountered a taxonomy folder with a rules.yaml file with the yamllint config to be used. This upset ilab as it assumed it was a qna.yaml file.

src/instructlab/utils.py
```diff
@@ -166,14 +166,21 @@ def lab_check_callback(*args, **kwargs):
...
 
-def istaxonomyfile(fn):
+def is_taxonomy_file(fn: str) -> bool:
     path = Path(fn)
-    return path.suffix == ".yaml" and path.parts[0] in TAXONOMY_FOLDERS
+    if path.parts[0] not in TAXONOMY_FOLDERS:
+        return False
+    if path.suffix == ".yml":
+        logger.warning(
+            f"Found a '.yml' file: {path}: taxonomy files must have a '.yaml' extension. File will not be checked."
```
bjhargrave, line 175:
```suggestion
    if path.name == "qna.yml":
        logger.warning(
            f"Found a 'qna.yml' file: {path}: taxonomy files must be named 'qna.yaml'. File will not be checked."
```

src/instructlab/utils.py
```diff
@@ -166,14 +166,21 @@ def lab_check_callback(*args, **kwargs):
...
+        )
+        return False
+    return path.suffix == ".yaml"
```
bjhargrave, line 178:
```suggestion
    return path.name == "qna.yaml"
```

bjhargrave (changes requested):
We need to require the file name to be qna.yaml and warn when qna.yml is used. Other files in the folder must be ignored.

src/instructlab/utils.py
booxter: Wasn't the plan here before this suggested change to log for **any** .yml file, not just qna.yml?

bjhargrave: But there can be other files in the folder which are irrelevant for ilab. We want to alert to the user to the improper .yml extension on qna.yml. But for any other name, we can just ignore the file.

src/instructlab/utils.py
mairin: it's not a hard requirement from my pov that it be named qna.yaml, but i am comfortable enforcing that as the name required if we are not confident the codebase can handle another name.

nathan-weinberg: It is an upstream requirement for the community Taxonomy, but we are still determining if this is the case for downstream implementations. Let's please have this conversation in a single place instead of spreading it around multiple points in the PR.

nathan-weinberg: @mairin beat me to it

bjhargrave: It is the enforced experience in the taxonomy repo that the file is named `qna.yaml` and nothing else. This is the only file which is validated by the taxonomy repo CI process. I think it is important for consistency and to allow other files to exist in a taxonomy folder which can be politely ignored by ilab.

bjhargrave: It seems odd (to me) that ilab would demand the extension be .yaml yet allow the file name to be anything. This would allow for someone to do qna1.yaml and qna2.yaml in the same taxonomy folder. What would that mean? Would we support multiple qna files in the same folder?

nathan-weinberg: Okay the point has been made (in several places :wink:) I will make the change when I have time to do so, by EOD

bcrochet (approved):
/lgtm

bjhargrave (approved):

booxter (approved):

src/instructlab/utils.py
```diff
@@ -433,11 +441,13 @@ def validate_taxonomy_file(file_path: str, yaml_rules: Optional[str] = None):
...
             return warnings, errors
 
-        # do general YAML linting if specified
+        # do general YAML linting
         version = get_version(contents)
         if version > 1:  # no linting for version 1 yaml
-            if yaml_rules is not None:
-                if os.path.isfile(yaml_rules):
+            if yaml_rules is not None:  # user attempted to pass custom yaml rules file
+                if os.path.isfile(
```
cdoern, line 448: should we fail is yaml rules is not none and this isfile check is false?

cdoern (commented):
one comment, looks good though

src/instructlab/utils.py
nathan-weinberg: Hmmm I see your point - we have the debug statement below, but maybe we could make it into a warning instead?

nathan-weinberg: Come to think of it, the `debug` statement when the file _is_ found could maybe be `info`?

bjhargrave: This code will all be removed and replaced by calling the new taxonomy parse API in instructlab-schema. So I would not worry too much about this here.

cdoern (approved):

PR #1913, opened 2024-07-29
fix: allow chat to resolve the local model being served

RobotSail:
When `ilab model chat` is executed, it tries to find the default model being served
and use that when available. However; this behavior was not properly updated when
extending support for serving models via vLLM. As a result, the current code assumes that
the default model being served is the one which is specified as a default in the config.
Since users may instead download and serve a different model path, ilab model chat will
fail to detect and select this model when running ilab model chat.

This commit updates this behavior by introducing a default endpoint that ilab model serve
will commonly serve on (http://127.0.0.1:8000/v1), and checks to see if there is a
model being served at that endpoint when no other endpoint is specified. If the check
fails, then ilab model chat will proceed with the current behavior of seeking the default
model specified in the config for serve. If the check succeeds, then ilab model chat
simply proceeds using this endpoint for the remainder of the code.

**Checklist:**

### DIFF
--- a/src/instructlab/model/chat.py
+++ b/src/instructlab/model/chat.py
@@ -26,6 +26,7 @@ import openai
 from instructlab import clickext
 from instructlab import client as ilabclient
 from instructlab import configuration as cfg
+from instructlab.utils import HttpClientParams
 
 # Local
 from ..utils import get_sysprompt, http_client
@@ -60,6 +61,24 @@ CONTEXTS = {
 PROMPT_HISTORY_FILEPATH = os.path.expanduser("~/.local/chat-cli.history")
 
 PROMPT_PREFIX = ">>> "
+
+DEFAULT_ENDPOINT = cfg.get_api_base(cfg._serve().host_port)
+
+
+def is_openai_server_and_serving_model(
+    endpoint: str, api_key: str, http_params: HttpClientParams
+) -> bool:
+    """
+    Given an endpoint, returns whether or not the server is OpenAI-compatible
+    and is actively serving at least one model.
+    """
+    try:
+        models = ilabclient.list_models(
+            endpoint, api_key=api_key, http_client=http_client(http_params)
+        )
+        return len(models.data) > 0
+    except ilabclient.ClientException:
+        return False
 
 
 @click.command()
@@ -113,6 +132,7 @@ PROMPT_PREFIX = ">>> "
     "--endpoint-url",
     type=click.STRING,
     help="Custom URL endpoint for OpenAI-compatible API. Defaults to the `ilab model serve` endpoint.",
+    default=DEFAULT_ENDPOINT,
 )
 @click.option(
     "--api-key",
@@ -173,9 +193,20 @@ def chat(
     # First Party
     from instructlab.model.backends.llama_cpp import is_temp_server_running
 
-    # TODO: this whole code block is replicated in generate.py. Refactor to a common function.
     backend_instance = None
-    if endpoint_url:
+    if endpoint_url != DEFAULT_ENDPOINT or (
+        endpoint_url == DEFAULT_ENDPOINT
+        and is_openai_server_and_serving_model(
+            endpoint_url,
+            api_key,
+            {
+                "tls_client_cert": tls_client_cert,
+                "tls_client_key": tls_client_key,
+                "tls_client_passwd": tls_client_passwd,
+                "tls_insecure": tls_insecure,
+            },
+        )
+    ):
         api_base = endpoint_url
     else:
         # First Party
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -335,7 +335,15 @@ def launch_server(
     try:
         # http_client is handling tls params
         api_base = backend_instance.run_detached(
-            http_client(ctx.params), background=not enable_serving_output
+            http_client(
+                {
+                    "tls_client_cert": ctx.params["tls_client_cert"],
+                    "tls_client_key": ctx.params["tls_client_key"],
+                    "tls_client_passwd": ctx.params["tls_client_passwd"],
+                    "tls_insecure": ctx.params["tls_insecure"],
+                }
+            ),
+            background=not enable_serving_output,
         )
     except Exception as exc:
         click.secho(f"Failed to start server: {exc}", fg="red")
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -76,6 +76,17 @@ class LegacyMessageSample(TypedDict):
     assistant: str
 
 
+class HttpClientParams(TypedDict):
+    """
+    Types the parameters used when initializing the HTTP client.
+    """
+
+    tls_client_cert: str | None
+    tls_client_key: str | None
+    tls_client_passwd: str | None
+    tls_insecure: bool
+
+
 def macos_requirement(echo_func, exit_exception):
     """Adds a check for MacOS before running a method.
 
@@ -532,7 +543,7 @@ def get_ssl_cert_config(tls_client_cert, tls_client_key, tls_client_passwd):
         return tls_client_cert, tls_client_key, tls_client_passwd
 
 
-def http_client(params):
+def http_client(params: HttpClientParams):
     return httpx.Client(
         cert=get_ssl_cert_config(
             params.get("tls_client_cert", None),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,8 @@
 
 * `ilab config`: Fixed a bug where `ilab` didn't recognize `train.lora_quantize_dtype: null` as a valid
    value.
+* `ilab model chat`: Fixed an issue where the default served model couldn't be resolved when running
+   model besides the default `merlinite-7b-lab-Q4_K_M.gguf`.
 
 ## v0.17
 
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -170,10 +170,10 @@ def generate(
     yaml_rules,
     chunk_word_count,
     server_ctx_size,
-    tls_insecure,
-    tls_client_cert,
-    tls_client_key,
-    tls_client_passwd,
+    tls_insecure: bool,
+    tls_client_cert: str | None,
+    tls_client_key: str | None,
+    tls_client_passwd: str | None,
     model_family,
     pipeline,
     enable_serving_output,
@@ -216,7 +216,15 @@ def generate(
         try:
             # Run the backend server
             api_base = backend_instance.run_detached(
-                http_client(ctx.params), background=not enable_serving_output
+                http_client(
+                    {
+                        "tls_client_cert": tls_client_cert,
+                        "tls_client_key": tls_client_key,
+                        "tls_client_passwd": tls_client_passwd,
+                        "tls_insecure": tls_insecure,
+                    }
+                ),
+                background=not enable_serving_output,
             )
         except Exception as exc:
             click.secho(f"Failed to start server: {exc}", fg="red")


src/instructlab/model/chat.py
```diff
@@ -61,6 +62,38 @@
 
 PROMPT_PREFIX = ">>> "
 
+DEFAULT_ENDPOINT = "http://127.0.0.1:8000/v1"
+
+
+def is_openai_server_and_serving_model(endpoint: str) -> bool:
```
cdoern, line 68: I think this logic exists somewhere else? I added code a while back to ensure you could only chat w/ a model being served

cdoern: yep: https://github.com/instructlab/instructlab/blob/632df7b83b87060b38eb978e075a12ce03c01446/src/instructlab/model/chat.py#L703-L710

OpenAI compatible servers have a .list() cmd that returns all models in the server

RobotSail: Oh good call, I'll update it to use that instead so we aren't duplicating unnecessary code.

RobotSail: Updated.

src/instructlab/model/chat.py
```diff
@@ -61,6 +62,30 @@
...
+    a model.
+    """
+    client = OpenAI(
+        base_url=endpoint,
+        api_key=api_key,
+        timeout=cfg.DEFAULTS.CONNECTION_TIMEOUT,
+        http_client=http_client(http_params),
+    )
+    try:
+        # ensure the model specified exists on the server. with backends like vllm, this is crucial.
```
booxter, line 83: `the model specified` where is it specified? I don't see where you'd pass an argument to the function. Looks like this function just checks that SOME models are served by the endpoint.

---

Can you please also elaborate in the comment why this is "crucial" for vllm and not some other endpoint?

src/instructlab/model/chat.py
```diff
@@ -61,6 +62,30 @@
...
 PROMPT_PREFIX = ">>> "
 
+DEFAULT_ENDPOINT = "http://127.0.0.1:8000/v1"
+
+
+def is_openai_server_and_serving_model(
+    endpoint: str, api_key: str, http_params: HttpClientParams
+) -> bool:
+    """
+    Given an endpoint, returns whether or not the given response is expected
```
booxter, line 72: `given response` which response is it?

src/instructlab/model/chat.py
```diff
@@ -61,6 +62,30 @@
...
+    client = OpenAI(
+        base_url=endpoint,
+        api_key=api_key,
+        timeout=cfg.DEFAULTS.CONNECTION_TIMEOUT,
+        http_client=http_client(http_params),
+    )
+    try:
+        # ensure the model specified exists on the server. with backends like vllm, this is crucial.
+        model_list = client.models.list().data
+        return len(model_list) > 0
```
booxter, line 85: nit: `return bool(model_list)` since it doesn't seem important which length it has as long as it's not empty.

src/instructlab/model/chat.py
```diff
@@ -61,6 +62,30 @@
...
+
+def is_openai_server_and_serving_model(
+    endpoint: str, api_key: str, http_params: HttpClientParams
+) -> bool:
+    """
+    Given an endpoint, returns whether or not the given response is expected
+    from an OpenAI-compatible server, and whether or not it's actively serving
+    a model.
+    """
+    client = OpenAI(
```
booxter, line 76: `chat_cli` already instantiates the client and fetches the models. I think the code should be split into a function and reused. And/or there's `list_models` in `utils.py` that can be of use here instead.

src/instructlab/model/chat.py
```diff
@@ -61,6 +62,30 @@
 
 PROMPT_PREFIX = ">>> "
 
+DEFAULT_ENDPOINT = "http://127.0.0.1:8000/v1"
```
booxter, line 65: There's already code and constants that can be used to produce the string (`get_api_base`, `host_port` default value in `_serve` class.) These should be reused.

src/instructlab/model/chat.py
```diff
@@ -175,7 +201,19 @@ def chat(
 
     # TODO: this whole code block is replicated in generate.py. Refactor to a common function.
```
booxter, line 202: This comment is suspicious. Should we also update `generate` code to do the same - resolve the local endpoint if it's running?

booxter (commented):
Should the test suite be updated to cover this case? (Serve vllm on default endpoint, then, then validate that `chat` connects correctly / ignores the config setting.)

src/instructlab/model/chat.py
RobotSail: While that would functionally work, we should keep it as is since it's descriptive about what the exact condition is.

src/instructlab/model/chat.py
RobotSail: The comment was just copied over from an earlier section in the code. I believe the issue for `generate` should be partially solved in PR https://github.com/instructlab/instructlab/pull/1901. We should keep the scope of this PR to just chat for now.

src/instructlab/model/chat.py
RobotSail: Good call, I'll update this to use that instead.

src/instructlab/model/chat.py
RobotSail: The response from the endpoint.

nathan-weinberg: When you refer to "the given response" I read that as if it's another arg - is it `http_params`?

RobotSail: No - those are just values that we need to be able to make the request using the OpenAI client.

nathan-weinberg:
Not an SME so only left a couple comments but also added @alimaredia who knows much more about vLLM than myself

src/instructlab/model/chat.py
RobotSail: @nathan-weinberg This is just saying that we received a response consistent with that of an OpenAI-compatible server. Consider the case when we have another REST server at the URL but is completely unrelated - the response we get wouldn't be expected.

nathan-weinberg: gotcha @RobotSail - might i suggest the following `Given an server endpoint, returns whether or not the server is OpenAI-compatable`

@booxter wdyt?

RobotSail: Yes Nathan, that wording is much better. Lets definitely change it since the way I wrote it isn't very clear.

src/instructlab/model/chat.py
RobotSail: Ignore my previous comment... this isn't even a line that gets touched in this PR. Lets please limit the review comments to what is being changed.

src/instructlab/model/chat.py
```diff
@@ -61,6 +62,28 @@
 
 PROMPT_PREFIX = ">>> "
 
+DEFAULT_ENDPOINT = cfg.get_api_base(cfg._serve.host_port)
+
+
+def is_openai_server_and_serving_model(
```
alimaredia, line 68: Where is ` is_openai_server_and_serving_model()` getting called? I don't see another call for it in the PR

src/instructlab/model/chat.py
RobotSail: It's also worth mentioning that we're only querying a local server for this, so if the server exists then the overhead here would be very minimal. It shouldn't impact the user experience in any way.

src/instructlab/model/chat.py
RobotSail: Ruh roh - looks like I may have accidentally deleted that change from this PR.

src/instructlab/model/chat.py
booxter: I am not suggesting you to pass the result of the function called elsewhere. I suggest you reuse the code. It is never simpler to maintain the same code in multiple places. You have a function in `utils.py` that will return what you want. Why not just using it? I am confused. Just call it here and then return `len(...) > 0` as you do.

src/instructlab/model/chat.py
booxter: It's not about particular lines. At the very least, this PR affects the comment since now it's invalid (because the PR modifies the code - it's no longer correct that the block replicates anything. It doesn't anymore.)

You can remove the comment (since it's invalid), or modify the `generate.py` to keep the comment valid.

src/instructlab/model/chat.py
RobotSail: I see - I misread the original comment as suggesting that we reuse the output from the list models function. I will update this to reuse the suggested function.

src/instructlab/model/chat.py
booxter: @nathan-weinberg this is definitely better. The function also checks that models are served though. Maybe it should have been also captured. (That said, the name of the function is so clear that I don't know if there's much of value in having any docstring...)

src/instructlab/model/chat.py
```diff
@@ -61,6 +62,27 @@
 
 PROMPT_PREFIX = ">>> "
 
+DEFAULT_ENDPOINT = cfg.get_api_base(cfg._serve().host_port)
```
booxter, line 65: (No action required) You could also move the default value for `host_port` into a variable, then you wouldn't need to call `_serve()` here.

src/instructlab/utils.py
```diff
@@ -76,6 +76,17 @@ class LegacyMessageSample(TypedDict):
     assistant: str
 
 
+class HttpClientParams(TypedDict):
```
booxter, line 79: Love it btw.

booxter (commented):
Just waiting for the next revision with `list_models` reused and `generate.py` comment clarified / removed. Otherwise, this is ready to merge I think.

src/instructlab/model/evaluate.py
```diff
@@ -335,7 +335,15 @@ def launch_server(
     try:
         # http_client is handling tls params
         api_base = backend_instance.run_detached(
-            http_client(ctx.params), background=not enable_serving_output
+            http_client(
```
alimaredia, line 338: What's the reason this change was included but not one for `run_detached()` in data/generate.py?

RobotSail: MyPy didn't complain about this change needing to be made there, but we should definitely update it there as well.

src/instructlab/model/chat.py
```diff
@@ -61,6 +62,24 @@
...
+    """
+    Given an endpoint, returns whether or not the server is OpenAI-compatible
+    and is actively serving at least one model.
+    """
+    try:
+        models = ilabclient.list_models(
+            endpoint, api_key=api_key, http_client=http_client(http_params)
+        )
+        return len(models) > 0
+    except openai.OpenAIError:
```
booxter, line 80: `list_models` raises `ClientException`, not `OpenAIError`.

booxter (changes requested):

src/instructlab/model/chat.py
RobotSail: Good catch, thanks!

src/instructlab/model/chat.py
RobotSail: We can probably put it in the `DEFAULTS` object. But let's not do that in this PR, we should really just get this merged so we can resolve the original issue we wanted to fix - which is that `ilab model chat` doesn't automatically connect to the running server when launched with the default values.

src/instructlab/model/chat.py
RobotSail: That's a very good point, it sounds like the best way forward will be to remove the comment.

booxter (approved):
Thank you.

alimaredia (approved):

booxter (approved):

jaideepr97 (approved):
approving this - I think its a neat change 
that being said I personally think its a little confusing that we rely on the config for some things but circumvent it for other things, and the line is not always super clear to me

PR #1917, opened 2024-07-29
LoRA Adjustments

### LINKED ISSUE #1906: is_padding_free needs to be false when running lora
when is_padding_free is true, train is not able to find the lora layers and tries to parse through keys and see if ending matches what needs to be saved. 
is_padding_free needs to be false when running lora or don’t use is_padding_free alongside lora

cdoern:
add the ability to turn LoRA off. Currently all of the LoRA config fields are mandatory and have strict defaults meaning LoRA is always on when called via the CLI.
Also make it so that if LoRA is used with --is-padding-free=True, we fail. In testing these two do not work together

resolves #1906

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -497,7 +497,7 @@ class _train(BaseModel):
 
     deepspeed_cpu_offload_optimizer: bool
 
-    lora_rank: int
+    lora_rank: int | None = None
     lora_quantize_dtype: str | None
 
     is_padding_free: bool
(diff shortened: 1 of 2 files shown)


src/instructlab/configuration.py
```diff
@@ -380,8 +380,8 @@ class _train(BaseModel):
 
     deepspeed_cpu_offload_optimizer: bool
 
-    lora_rank: int
-    lora_quantize_dtype: str
+    lora_rank: Optional[int] = None
```
RobotSail, line 383: You'll want to use `int | None` here - Pydantic doesn't seem to interpret `typing.Optional` properly:

```suggestion
    lora_rank: int | None = None
```

alinaryan:
@cdoern can you grant me write access to your repo? I'm trying to fix the lint failures/take this over

src/instructlab/configuration.py
alinaryan: The default `None` is not set in the config properly if set this way, but since this is more readable I kept this suggestion with a small change to be able to set a default.

src/instructlab/configuration.py
```diff
@@ -691,13 +691,30 @@ def map_train_to_library(params):
...
-        rank=params["lora_rank"],
-        alpha=params["lora_alpha"],
-        dropout=params["lora_dropout"],
-        target_modules=params["lora_target_modules"],
-        quantize_data_type=params["lora_quantize_dtype"],
-    )
+    lora_args = LoraOptions()
+    lora = False
+    if params["lora_rank"] is not None:
+        if not lora:
```
booxter, line 697: nit: just set `lora = True`, no need to check it's `False` first.

booxter (approved):
Ways to improve the PR: update CHANGELOG about no longer requiring the `--lora-*` options; re-wrap commit messages to 50/72 rule; add a "unit" test case for the conflicting scenario.

RobotSail (approved):
LGTM

alinaryan:
This merged before it was ready, created followup in #2008

Issue #1904, opened 2024-07-29
Add additional GPUs flag to ilab model serve


alinaryan:
An extra `-- --tensor-parallel-size n` (n being the num of gpus) is needed in the `ilab model serve` command when serving the model. This should be doc'ed in the readme. Also, we would like to point out that everything after that flag is expected to be a vLLM command. 

In 0.19.0, the vLLM args should become flags for `ilab model serve`.

mairin:
What is the outcome if this doesn't make it to 0.18.0? Trying to understand whether or not this is a true blocker.

alimaredia:
@mairin This issue is actually 2 things.
1. This is a UX improvment. It's to give the user the option to specify `--gpus` for `ilab model serve`. Right now to specify that they'd need to edit the config file, and change the `vllm_args` section to include a specific vLLM flag. 

Charlie opened up https://github.com/instructlab/instructlab/pull/1931 to fix that issue. I narrowed down the scope for his PR in https://github.com/instructlab/instructlab/pull/1974. I think we can get 1974 in today. 

I think providing just the flag for ilab serve should be a blocker. What Charlie had done by a gpu value into the serving config has bigger scope and I think we could move to the next release.
 
2. Upstream Docs for item 1.

CC: @alinaryan @cdoern

PR #1931, opened 2024-07-30
graduate --tensor-parallel-size

cdoern:
the vllm_args format: `-- --<flag_name>` hides commonly changed serve options behind an extra step. The commonly changed options also are not represented in our help text. Graduate --tensor-parallel-size out of the hidden flags, rename it to `--gpus` and use the value of the flag unless `--tensor-parallel-size` is already set in the cfg.

**Checklist:**


src/instructlab/model/serve.py
```diff
@@ -121,6 +130,14 @@ def serve(
     from instructlab.model.backends import llama_cpp, vllm
 
     host, port = utils.split_hostport(ctx.obj.config.serve.host_port)
+
+    if tensor_parallel_size is not None:
+        if "--tensor-parallel-size" in ctx.args:
+            logger.info(
+                "'--gpus' flag used alongside '--tensor-parallel-size' in the vllm_args section of the config. Using the value found in the config"
```
bcrochet, line 137: CLI args should always override a config file. Or am I misunderstanding the log message?

bcrochet (commented):
Looking for a clarification.

src/instructlab/model/serve.py
```diff
@@ -94,6 +94,14 @@ def signal_handler(
         "Automatically detected based on the model file properties.\n"
     ),
 )
+@click.option(
+    "--gpus",
```
booxter, line 98: so it's `--gpus` in CLI but `tensor_parallel_size` in config? This is confusing. What's the rationale to have different names for these?

alimaredia:
@cdoern I'm going to wait on this PR to merge until I work on the docs for it here: https://github.com/instructlab/instructlab/issues/1904

src/instructlab/configuration.py
```diff
@@ -256,6 +256,7 @@ class _serve_vllm(BaseModel):
     """Class describing configuration of vllm serving backend."""
 
     llm_family: str = ""
+    tensor_parallel_size: Optional[int] = None
```
alimaredia, line 259: This should be gpus just like it is in the SDG config.

alimaredia:
I opened up https://github.com/instructlab/instructlab/pull/1974 to at least add the flag to `ilab model serve` for `0.18.0`. I see this PR as the beginning of figuring out how we set gpus in the config, and how all of the commands that use vLLM serving will use a consistent interface. That should be in scope for `0.19.0`.

Wdyt @cdoern

cdoern:
I disagree @alimaredia I think this one makes more sense given that most flags correlate to config entities. I will change the name of this one to gpus

alimaredia:
The changes you made look great. One thing that should be changed is the generate section still has two `gpus` in it. 
```
generate:
  chunk_word_count: 1000
  gpus: null
  model: /home/ali/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
  ...
  teacher:
    backend: null
   ...
    vllm:
      gpus: null
      llm_family: ''
      max_startup_attempts: 300
      vllm_args: []
```

Here's a patch for that should address it:
```
diff --git a/src/instructlab/configuration.py b/src/instructlab/configuration.py
index dc25d3f..8d57eb6 100644
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -338,7 +338,6 @@ class _generate(BaseModel):
     output_dir: StrictStr = Field(default_factory=lambda: DEFAULTS.DATASETS_DIR)
     prompt_file: StrictStr = Field(default_factory=lambda: DEFAULTS.PROMPT_FILE)
     seed_file: StrictStr = Field(default_factory=lambda: DEFAULTS.SEED_FILE)
-    gpus: Optional[int] = None
 
 
 class _mmlu(BaseModel):
diff --git a/src/instructlab/data/generate.py b/src/instructlab/data/generate.py
index 92ce7c2..a324201 100644
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -161,6 +161,7 @@ logger = logging.getLogger(__name__)
     "--gpus",
     type=click.IntRange(min=0),
     cls=clickext.ConfigOption,
+    config_sections="teacher.vllm",
     help="Number of GPUs to run generation on",
 )
 @click.pass_context
diff --git a/tests/testdata/default_config.yaml b/tests/testdata/default_config.yaml
index ae2bbca..131cdd9 100644
--- a/tests/testdata/default_config.yaml
+++ b/tests/testdata/default_config.yaml
@@ -29,7 +29,6 @@ general:
   log_level: INFO
 generate:
   chunk_word_count: 1000
-  gpus: null
   model: /cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
   num_cpus: 10
   output_dir: /data/instructlab/datasets
```

src/instructlab/model/serve.py
```diff
@@ -121,6 +129,14 @@ def serve(
     from instructlab.model.backends import llama_cpp, vllm
 
     host, port = utils.split_hostport(ctx.obj.config.serve.host_port)
+
+    if gpus is not None:
```
alimaredia, line 133: The logic I had https://github.com/instructlab/instructlab/pull/1974/files seems tighter.

There are two collisions you need to resolve/add logging for:
1) If tensor parallel size is set in vllm_args in the config. My PR proposes using the flag and throw a warning about it.
2) If tensor parallel size is set in ctx.args. This PR and mine propsed use ctx.args.

It looks like you're just missing 1) in this PR.

My PR also moves this logic down into the vLLM section and uses `contains_argument()` as Dan suggested.

src/instructlab/configuration.py
```diff
@@ -266,6 +266,7 @@ class _serve_vllm(BaseModel):
 
     llm_family: str = ""
     max_startup_attempts: int | None = None
+    gpus: Optional[int] = None
```
danmcp, line 269: We should probably change the evaluate default to None as well.  It is currently set to default to 1, so it will override serves config by default.

danmcp:
@cdoern has this been manually tested with serve and generate?

alimaredia:
@danmcp I'm taking over this PR from Charlie. Testing it manually right now.

src/instructlab/model/serve.py
```diff
@@ -94,6 +94,13 @@ def signal_handler(
         "Automatically detected based on the model file properties.\n"
     ),
 )
+@click.option(
+    "--gpus",
+    type=click.IntRange(min=0),
```
booxter, line 99: what's the meaning of `0` in this case? should it start from `1`? A brief look at `vllm` code suggests it won't work. (The "world_size" will end up as 0?)

alimaredia:
Manual testing was successful. It showed multiple gpus running when I specified `--gpus` for serve and generate, and if `-- --tensor-parallel-size` is specified for `ilab model serve --gpus` the `tensor-parallel-size` flag overrides gpus.

src/instructlab/configuration.py
```diff
@@ -367,7 +367,7 @@ class _evaluate(BaseModel):
     base_model: str
     branch: Optional[str] = None
     base_branch: Optional[str] = None
-    gpus: Optional[int] = 1
+    gpus: Optional[int] = None
```
danmcp, line 370: I think None needs to be allowed here as well: https://github.com/instructlab/instructlab/blame/main/src/instructlab/model/evaluate.py#L279

src/instructlab/model/serve.py
```diff
@@ -155,9 +164,29 @@ def serve(
...
+
+            if contains_argument(
+                "--tensor-parallel-size", ctx.obj.config.serve.vllm.vllm_args
+            ):
+                logger.info(
+                    "'--gpus' flag used alongside '--tensor-parallel-size' in the vllm_args section of the config file. Using value of the --gpus flag."
+                )
+
         # Instantiate the vllm server
         if ctx.args:
```
booxter, line 191: Does this line mean that when `--gpus` is passed, then arguments from `vllm_args` config section are ignored? (Even the arguments that have nothing to do with tensor size?) Is it intended here?

tests/test_config.py
```diff
@@ -233,6 +231,7 @@ def test_config_modified_settings(self, tmp_path_home):
     max_ctx_size: 2048
     llm_family: ''
   vllm:
+    gpus: 8
```
booxter, line 234: This is ok, but the test is a bit limited, since it only checks that `vllm_args` on config object are set. But then nothing checks that they are propagated down to vllm backend. I think it may be a problem (see my other comment in `serve.py`.)

src/instructlab/model/serve.py
```diff
@@ -155,9 +164,29 @@ def serve(
             num_threads=num_threads,
         )
     elif backend == backends.VLLM:
+        # First Party
+        from instructlab.model.backends.vllm import contains_argument
+
         # Warn if unsupported backend parameters are passed
         warn_for_unsuported_backend_param(ctx)
 
+        if gpus:
```
booxter, line 173: (Suggestion) This code is a bit convoluted. It would be helpful to see some test cases that demonstrate how this works. Ideally, something that would allow us to assert on `vllm_args` state past line 197. (May require some code extraction into a separate testable function.)

src/instructlab/model/serve.py
danmcp: I don't think `if ctx.args` has anything to do with --gpus directly.  This logic is replacing everything in vllm_args in config with what's after the --

I didn't actually realize it was doing this vs extending vllm_args.  Guessing that's the correct practice based on the author but should validate.

src/instructlab/model/serve.py
danmcp: That's a good question I don't think has been asked yet.  This was recently changed from INT to IntRange(min=0) in multiple places.  You could argue 0 should imply CPU only.  But not saying it will currently work.  Probably should handle as a separate issue.

booxter: The current implementation will do nothing, I think (`if args` will result in `False`.) But this may change. Wonder if we need to allow it in the first place.

src/instructlab/model/serve.py
booxter: Consider this: `ilab model serve --gpus=4`; also, in config file, `vllm_args` are set to `--random-option-to-vllm`. In this case,

1. `ctx.args` will start as empty list (because no `--` was passed.)
2. Then we `.extend` it in line 177.
3. Now, line 191 triggers, which will make the code use `--gpus=4` for `vllm_args`.
4. Line 195 is no longer triggered -> config file setting is ignored.

This feels wrong? We should probably merge the two sources of arguments (`--gpus=4` and the config `vllm_args`), not ignore the latter. I think this `if-else` here was designed so that when you pass `-- --more --options`, it allows you to override what is in the config file. But this is different?

danmcp: Ah, I see what you saw in the first place now!  Great catch!  I did not fully appreciate what was getting extended above.  I have the same instinct as you of asking why we are not updating vllm_args instead.

booxter: I'd really love to see some targeted tests for this piece of code that would demonstrate all concerning permutations:

1. defaults.
2. gpus and no vllm-args.
3. vllm-args and no gpus.
4. gpus and vllm-args with just the tensor-parallel-size.
5. gpus and vllm-args with tensor-parallel-size only.
6. gpus and vllm-args with tensor-parallel-size AND some other options.
7. gpus and vllm-args with some other options but not tensor-parallel-size.

Perhaps there's more I missed.

alimaredia: Dan is right. ctx.args is overwriting vllm_args for the server, instead of extending them. I would have expected extending `serve.vllm.vllm_args` to be the intended behavior. 

That doesn't have to do with the --gpus option though. Maybe we open up a separate PR for that?

alimaredia: @booxter you're right. Given the current logic If the user specifies `--gpus` and ctx.args is empty then ctx.args is extended to have `--tensor-parallel-size` and `str(gpus)`.

Since `ctx.args` is not empty anymore, it gets assigned to `vllm_args` and that is what's passed into `vllm.Server()`, not `serve.vllm.vllm_args`. 

I'm going to make some changes to this PR and report back.

danmcp:
@alimaredia Need to regenerate the default config for the unit tests

src/instructlab/model/evaluate.py
```diff
@@ -301,6 +301,7 @@ def launch_server(
             logger.warning(
                 f"Based on your hardware configuration, when using vLLM, we recommend setting max-workers between {recommended_min_workers} and {recommended_max_workers} for optimal performance"
             )
+        gpus = gpus or eval_serve.vllm.gpus
```
booxter, line 304: Shouldn't this be achieved with:

```
    cls=clickext.ConfigOption,
    config_sections="vllm",
```

...in the config option definition? Like it's done for the `serve` `--gpus` CLI option.

src/instructlab/model/evaluate.py
```diff
@@ -301,6 +301,7 @@ def launch_server(
...
         if gpus:
             # Warn when overriding from vllm_args
             tps_prefix = "--tensor-parallel-size"
```
booxter, line 307: Ouch. So we already have some code that attempts to do tackle much the same thing, just for `evaluate`?.. Ideally, we would have a **single** implementation for both code paths. They don't seem like they should be different in how they handle `--gpus`...

src/instructlab/model/serve.py
alimaredia: The intended precedence for vllm argument setting I have after updating this PR is:
`ctx.args` > `--flag` > `serve.vllm.vllm_args`

where:
`ctx.args` is anything passed after `ilab model serve --`
`--flag` is  `--gpus` or any future flag for `ilab model serve` that would be an argument to vllm under the hood.
`serve.vllm.vllm_args` is what's in the config file.

@danmcp @booxter what do you think?

booxter (commented):
I think the `evaluate` defaulting logic for the `--gpus` argument should use `clickext`, as for `serve` variant of the same.

---

Also, I still think that we should have this code spun out into a function, then test cases added to cover all permutations possible. And it's a risk to carry two parallel and nearly identical implementations of the same logic for `evaluate` and `serve`, going forward. This separated function could be reused for both commands, I think.

src/instructlab/model/serve.py
booxter: I've checked the latest version, and I think the precedence is correct.

src/instructlab/model/evaluate.py
alimaredia: You called it. When you realize `ilab data generate` also has a `--gpus` flag you realize that there scope of this PR could get quite big with a sizable refactor.

That was why I opened up the very small https://github.com/instructlab/instructlab/pull/1974/files to land the feature in `0.18.0`. I think as long as some unit tests are added, this PR is fine as is with an issue that needs to be created to properly scope out how to share code between the three commands.

src/instructlab/model/evaluate.py
alimaredia: The `evaluate` section of the config does not have a `vllm` section. See the comment below on how we need to go to the drawing board to figure out what the common interface under the hood/in the config should be across commands for `--gpus`.

src/instructlab/model/evaluate.py
booxter: @alimaredia refactor can be left for later. Do you mind creating an issue for this? We can get back to it post GA.

src/instructlab/model/evaluate.py
booxter: Oh gosh!

booxter (approved):
@alimaredia thanks for clarifications inline.

The code is fine. I am ok to merge without any additional changes; but we should at least report issues to track down deficiencies identified, namely:

1. missing test coverage.
2. code reuse for the code that reconciles gpus and vvlm_args.
3. extend evaluate interface with `--gpus` / vllm_args.

If you decide to report these, please post links in comments. Thank you.

src/instructlab/model/serve.py
danmcp: I've added a bunch of tests.  Hopefully they are a good start.

src/instructlab/model/serve.py
danmcp: The tests do a decent job of validating this precedence.  The only part which is a bit strange is gpus > tps in the config.  But tps > gpus in the command.  As long as we continue to output from the command which arg is winning, I think it's workable as is.

danmcp: > I'd really love to see some targeted tests for this piece of code that would demonstrate all concerning permutations:
> 
> 1. defaults.
I think this is covered
> 2. gpus and no vllm-args.
Covered
> 3. vllm-args and no gpus.
Covered
> 4. gpus and vllm-args with just the tensor-parallel-size.
Covered
> 5. gpus and vllm-args with tensor-parallel-size only.
Not sure how this is different than 4
> 6. gpus and vllm-args with tensor-parallel-size AND some other options.
Covered with ctx tps but not config.
> 7. gpus and vllm-args with some other options but not tensor-parallel-size.
Not yet covered
> 
> Perhaps there's more I missed.
I did cover several other combinations between the 4 places the value can be set.  If there are other important ones, we can either add them in this PR or open issues/PRs.

booxter (approved):
@danmcp you are my hero of the day. 🥳

tests/test_lab_serve.py
```diff
@@ -109,6 +122,295 @@ def test_chat_manual(mock_popen, mock_determine, _, cli_runner: CliRunner):
...
+        [
+            f"--config={CFG_FILE_NAME}",
+            "model",
+            "serve",
+            "--model-path=foo",
+            "--gpus",
+            "4",
+        ],
+    )
+    assert args[15] == "--tensor-parallel-size"
```
booxter, line 383: (no action needed) these asserts are a bit brittle, but I think we'll deal with them when we refactor the code that calculates cmd out into a function that we would be able to target properly.

danmcp: Agree on the brittleness.  Checking the length at least makes sure we are targeting the effective args.

danmcp: It's still not great but at least made it a little less brittle by using len(args) to calc the index.

tests/test_lab_serve.py
```diff
@@ -51,6 +54,21 @@ def assert_template(args, expect_chat, path_chat, chat_value):
         assert template == chat_value
 
 
+def assert_tps(args, tps):
+    assert args[len(args) - 2] == "--tensor-parallel-size"
```
booxter, line 58: nit: this is the same as `args[-2]`

booxter (approved):

tests/test_lab_serve.py
danmcp: doh, thanks!

booxter (approved):
Thanks for respinning @danmcp Obviously, these were just nits and I would be happy to merge without them addressed, but I appreciate your energy in polishing code. 👍

tests/test_config.py
danmcp: Resolving based on new tests

danmcp (approved):

src/instructlab/model/serve.py
```diff
@@ -155,17 +164,37 @@ def serve(
...
+            if contains_argument(
+                "--tensor-parallel-size", ctx.obj.config.serve.vllm.vllm_args
+            ):
+                logger.info(
+                    "'--gpus' flag used alongside '--tensor-parallel-size' in the vllm_args section of the config file. Using value of the --gpus flag."
+                )
```
khaledsulayman, line 179: I slightly prefer failing and explicitly telling the user to use 1 or the other but I won't block on this

khaledsulayman (approved):
left a nit but I'll file a separate issue for it if we want to get this in ASAP
LGTM!

PR #1940, opened 2024-07-31
Add missing ilab model list help screen description

alinaryan:
**Checklist:**

### DIFF
--- a/src/instructlab/model/list.py
+++ b/src/instructlab/model/list.py
@@ -29,6 +29,7 @@ logger = logging.getLogger(__name__)
 )
 @click.option(
     "--list-checkpoints",
+    help="Also list checkpoints [in addition to existing models].",
     is_flag=True,
 )
 def model_list(model_dirs: list[str], list_checkpoints: bool):


alinaryan:
New output:

```
Usage: ilab model list [OPTIONS]

  Lists models

Options:
  --model-dirs TEXT   Base directories where models are stored.  [default:
                      /home/alina/.cache/instructlab/models]
  --list-checkpoints  List the trained model checkpoints.
  --help              Show this message and exit.
```

src/instructlab/model/list.py
```diff
@@ -29,6 +29,7 @@
 )
 @click.option(
     "--list-checkpoints",
+    help="List the trained model checkpoints.",
```
RobotSail, line 32: You should probably reword this to "Also include trained model checkpoints" or something.

RobotSail (approved):
LGTM, just please change the wording a little bit - the reason is that the current wording makes it seem like it's listing the trained checkpoints _instead_ of the existing models.

src/instructlab/model/list.py
alinaryan: So it's listing both trained model checkpoints and existing models?

RobotSail: Yeah it's not a binary flag, it's mainly a fallback case we check if the given path isn't a safetensors directory or gguf model:

https://github.com/instructlab/instructlab/blob/f0590b710689d0ae0051e744b47f2517b563a828/src/instructlab/model/list.py#L97

alinaryan: How is this?

RobotSail: I think the way you've written it is good for a command description, but we'll probably want to change it to "include the trained checkpoints in the output", since we're extending the existing functionality.

danmcp:
@mergifyio rebase

booxter (approved):

src/instructlab/model/list.py
```diff
@@ -29,6 +29,7 @@
 )
 @click.option(
     "--list-checkpoints",
+    help="List the trained model checkpoints and all existing models.",
```
booxter, line 32: since it's a description of the effect of this particular option, perhaps you could say smth like "Also list checkpoints [in addition to existing models]."

alinaryan: Changed it to this

RobotSail (approved):
LGTM

PR #1941, opened 2024-07-31
mergify config: Fix and update for valid regex and consistency

bjhargrave:
This fix updates the `files` attributes to use valid regex expressions (`~=` operator) when needed and exact match (`=` operator) otherwise. Globs are not supported and are replaced by regex. We use `^` and `$` to properly anchor the regex as needed. Leading and trailing `.*` are removed as they are not necessary.

For consistency across the multiple workflow conditions, we move the `check-success` condition to the top. We also remove unneeded quote characters as they were inconsistently used.

The `files` conditions are updated to properly match the globs used in the corresponding workflows. References to no-longer-present files are removed.

A `check-success` condition for the `spellcheck` job is added as it was missing. An old action for release-v0.14 is removed.

**Checklist:**

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -7,38 +7,42 @@ pull_request_rules:
     - "#changes-requested-reviews-by=0"
     - or:
       - base=main
-      - base~=release-.*
+      - base~=^release-
     - label!=hold
     - label!=do-not-merge
     - label!=needs-rebase
     - check-success=DCO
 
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):
This is great - thanks @bjhargrave!

booxter (approved):
This makes sense based on what I gather after reading the docs [here](https://docs.mergify.com/configuration/conditions/#operators-list). The regexes are python, not shell globs.

PR #1949, opened 2024-08-01
Model family vllm data generate & `get_model_family()` fixes

alimaredia:
1st commit:

```
set teacher.vllm.llm_family from model_family flag.
This was only being set for the llama_cpp backend.
```
2nd commit:
    
    This commit changes the following with
    regard to get_model_family in configuration.py:
    
    1. rename "forced" to "family" in the function
     to make it more readable.
    
    2. get_model_family() was always guessing the
    model_family even if a valid family was input.
    Now if a valid family is input, it returns that.
    
    3. add debug logging in backend.py to see the
    result of get_model_family()
    
    4. add a case in test_get_model_family() for when
    model_family == "" which is it's default value
    in the serve config and in other places in the
    code base.

For anyone tracing through the code, here are the calls that lead to `get_model_family()`
```
1. data/generate.py -- model_family is assigned to teacher.vllm.llm_family
2. data/generate.py -- select_backend() is called with the teacher cfg
3. model/backends/backend.py -- a vllm_server is initialized with cfg.vllm.llm_family as the model_family
4. model/backends/vllm.py -- run_detached() calls run_vllm() with self.model_family
5. model/backends/vllm.py -- in run_vllm() model_family is passed into build_vllm_cmd()
6. model/backends/vllm.py -- in format_template(), model_family is passed into get_model_template()
7. model/backends/backend.py -- in get_model_template(), model_family is passed into get_model_family() which is in configuration.py
```

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -534,12 +534,15 @@ def get_api_base(host_port: str) -> str:
     return f"http://{host_port}/v1"
 
 
-def get_model_family(forced, model_path):
-    forced = MODEL_FAMILY_MAPPINGS.get(forced, forced)
-    if forced and forced.lower() not in MODEL_FAMILIES:
-        raise ConfigException(f"Unknown model family: {forced}")
+def get_model_family(family, model_path):
+    family = MODEL_FAMILY_MAPPINGS.get(family, family)
+    if family:
+        if family.lower() not in MODEL_FAMILIES:
+            raise ConfigException(f"Unknown model family: {family}")
 
-    # Try to guess the model family based on the model's filename
+        return family.lower()
+
+    # If family is not set try to guess the model family based on the model's filename
     guess = match(r"^\w*", path.basename(model_path)).group(0).lower()
     guess = MODEL_FAMILY_MAPPINGS.get(guess, guess)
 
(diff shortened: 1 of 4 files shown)


booxter (changes requested):
This breaks config file setting for the family.

src/instructlab/configuration.py
```diff
@@ -534,12 +534,15 @@ def get_api_base(host_port: str) -> str:
...
-    if forced and forced.lower() not in MODEL_FAMILIES:
-        raise ConfigException(f"Unknown model family: {forced}")
-
-    # Try to guess the model family based on the model's filename
+def get_model_family(family, model_path):
+    family = MODEL_FAMILY_MAPPINGS.get(family, family)
+    if family:
+        if family.lower() in MODEL_FAMILIES:
+            return family.lower()
+        else:
```
booxter, line 542: nit: I think this `else` is useless

booxter (approved):
Your commit message could probably mention that it also fixes behavior for llama case.

booxter (approved):

danmcp (approved):

cdoern (approved):

PR #1954, opened 2024-08-01
fix: log error if user supplies invalid repository for OCI download

jaideepr97:
Raise an error if user tries to specify version/tag within the supplied `--repository` instead of specifying the tag/version via `--release` when downloading models from OCI registries 

**Checklist:**

### DIFF
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -191,6 +191,15 @@ class OCIDownloader(Downloader):
             f"Downloading model from OCI registry: {self.repository}@{self.release} to {self.download_dest}..."
         )
 
+        # raise an exception if user specified tag/SHA embedded in repository instead of specifying --release
+        match = re.search(r"^(?:[^:]*:){2}(.*)$", self.repository)
+        if match:
+            click.secho(
+                f"Invalid repository supplied: Please specify tag/version '{match.group(1)}' via --release",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
+
         model_name = self.repository.split("/")[-1]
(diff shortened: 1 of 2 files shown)


src/instructlab/model/download.py
```diff
@@ -191,6 +191,15 @@ def download(self):
             f"Downloading model from OCI registry: {self.repository}@{self.release} to {self.download_dest}..."
         )
 
+        # raise an exception if user specified tag/SHA embedded in repository instead of specifying --release
```
booxter, line 194: (Just a thought) if you already successfully extracted the necessary tag from the string, why not just use it? (maybe warn but proceed?)

jaideepr97: I'm not opposed to that
I was just wondering if it creates a more confusing UX by introducing inconsistencies between how OCI and HF downloads work

booxter: (Evil thought) Maybe HF should also extract out the tag! :D

jaideepr97: sure - maybe as a post GA thing? the UX could use a lot of cleaning up and improvement in general

booxter: Oh I didn't mean this to block the PR! Just a curiosity. Feel free to ignore.

booxter (approved):

src/instructlab/model/download.py
n1hility: Yeah IMO we should revisit in the future

n1hility (approved):

nathan-weinberg (approved):

PR #1955, opened 2024-08-01
make `ilab config init` clone taxonomy by default

### LINKED ISSUE #1951: `ilab config init` can create invalid taxonomy
Ran `ilab config init` with no arguments and was prompted to clone the taxonomy repo. I pressed enter without specifying `y` or `N` and it created an empty taxonomy directory, causing `ilab data generate` to fail due to an 'invalid taxonomy'. I then ran `ilab config init` again but since the directory existed, I was not prompted to clone the taxonomy repo.

**To Reproduce**
1. run `ilab config init`
2. during the prompt for whether to clone taxonomy repo, press enter without typing anything
3. run `ilab data generate` and you will see an error due to invalid taxonomy
4. run `ilab config init` again and you will not be prompted to clone the taxonomy repo

**Expected behavior**
1. The prompt to clone taxonomy should enforce `y` or `N` as an answer
2. `ilab config init` should validate the taxonomy at the path you give so that users do not unknowingly proceed with an invalid taxonomy

khaledsulayman:
If you pressed enter without specifying `y` or `N` when asked to clone taxonomy, it would by default not clone it, which would result in an empty taxonomy directory if you used the default taxonomy path. This change makes it so that the taxonomy is cloned unless you specify `n`.

Resolves #1951

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -168,7 +168,8 @@ def get_params_default(
         clone_taxonomy_repo = False
     elif interactive:
         clone_taxonomy_repo = click.confirm(
-            f"`{taxonomy_path}` seems to not exist or is empty. Should I clone {repository} for you?"
+            f"`{taxonomy_path}` seems to not exist or is empty. Should I clone {repository} for you?",
+            default=True,
         )
 
     # clone taxonomy repo if it needs to be cloned


booxter (approved):
Validated that we now have `Y/n` in the output, as expected.

```
`/home/vagrant/.local/share/instructlab/taxonomy` seems to not exist or is empty. Should I clone https://github.com/instructlab/taxonomy.git for you? [Y/n]:
```

khaledsulayman:
@mairin set it to 18 just now since I think it's small/niche enough that it shouldn't impact anything. We can move it to 19 though if we deem that better.

nathan-weinberg (approved):

PR #1957, opened 2024-08-02
Updating default location for judge model to be a local model in the cache

danmcp:
Along the way we disabled the ability for vllm to download the model.  This default config change is reccognizing where the model should be when downloaded if using prometheus.  If we want to enable the auto download feature in the future, we should instead use the ilab download capability so the model ends up in the model cache.

Thanks to @JamesKunstle for pointing out this issue!

**Checklist:**


JamesKunstle (approved):
quickest PR-opener in the west huh?

alimaredia (approved):

PR #1959, opened 2024-08-02
Add training profiles for GPU configs

cdoern:
Add training profiles for the following GPU configs:

4x/8x A100/H100

2x A100/H100

4x L40

8x L40

8x L4

these are rendered into ~/.local/share/instructlab/internal/train_configuration/

also introduce ILAB_TRAIN_PROIFLE_DIR, and env var which when set will be used as the template dir to render the initial train profiles into ~/.local/share

The user will also now see an interactive menu on initialization to choose their train profile.

**Checklist:**

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -2,7 +2,7 @@
 
 # Standard
 from os import listdir
-from os.path import dirname, exists
+from os.path import dirname, exists, join
 import pathlib
 import typing
 
@@ -103,6 +103,29 @@ def init(
     click.echo(f"Generating `{DEFAULTS.CONFIG_FILE}`...")
     if train_profile is not None:
         cfg.train = read_train_profile(train_profile)
+    elif interactive:
+        entries = listdir(DEFAULTS.TRAIN_PROFILE_DIR)
+        click.echo("Please choose a train profile to use:")
+        for i, value in enumerate(entries):
+            click.echo(f"{i}. {value}")
+        train_profile_selection = click.prompt(
+            "Enter the number of your choice [hit enter for the default]",
+            type=int,
+            default=-1,
+        )
+        if 0 <= train_profile_selection < len(entries):
+            click.echo(f"You selected: {entries[train_profile_selection]}")
+            cfg.train = read_train_profile(
+                join(DEFAULTS.TRAIN_PROFILE_DIR, entries[train_profile_selection])
+            )
+        elif train_profile_selection == -1:
+            click.echo("Using default train profile.")
+        else:
+            click.secho(
+                "Invalid selection. Please select a valid train profile option.",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
 
     cfg.chat.model = model_path
     cfg.generate.model = model_path
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -67,6 +67,7 @@ class _InstructlabDefaults:
     # file. When set, the CLI will use the file specified in the environment variable as a sample to
     # generate the default config file.
     ILAB_GLOBAL_CONFIG = "ILAB_GLOBAL_CONFIG"
+    ILAB_TRAIN_PROFILE_DIR = "ILAB_TRAIN_PROFILE_DIR"
 
     # TODO: Consolidate --model and --model-path into one --model-path flag since we always need a path now
     MODEL_NAME_OLD = "merlinite-7b-lab-Q4_K_M"
@@ -184,6 +185,30 @@ class _InstructlabDefaults:
     @property
     def TRAIN_DEFAULT_PROFILE(self) -> str:
         return path.join(self.TRAIN_PROFILE_DIR, "default.yaml")
+
+    @property
+    def TRAIN_A100_H100_X4_PROFILE(self) -> str:
+        return path.join(self.TRAIN_PROFILE_DIR, "A100_H100_x4.yaml")
+
+    @property
+    def TRAIN_A100_H100_X8_PROFILE(self) -> str:
+        return path.join(self.TRAIN_PROFILE_DIR, "A100_H100_x8.yaml")
+
+    @property
+    def TRAIN_A100_H100_X2_PROFILE(self) -> str:
+        return path.join(self.TRAIN_PROFILE_DIR, "A100_H100_x2.yaml")
+
+    @property
+    def TRAIN_L40_X8_PROFILE(self) -> str:
+        return path.join(self.TRAIN_PROFILE_DIR, "L40_x8.yaml")
+
+    @property
+    def TRAIN_L40_X4_PROFILE(self) -> str:
+        return path.join(self.TRAIN_PROFILE_DIR, "L40_x4.yaml")
+
+    @property
+    def TRAIN_L4_X8_PROFILE(self) -> str:
+        return path.join(self.TRAIN_PROFILE_DIR, "L4_x8.yaml")
 
 
 DEFAULTS = _InstructlabDefaults()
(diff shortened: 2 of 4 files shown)


src/instructlab/configuration.py
```diff
@@ -585,13 +608,154 @@ def ensure_storage_directories_exist():
         "lora_target_modules": ["q_proj", "k_proj", "v_proj", "o_proj"],
     }
 
+    FOUR_AND_EIGHT_GPU_TRAIN_AH = _train(
```
RobotSail, line 611: Does AH here mean that this applies to both A100s and H100s?

src/instructlab/configuration.py
```diff
@@ -585,13 +608,154 @@ def ensure_storage_directories_exist():
...
+        additional_args={
+            "warmup_steps": 10,
+            "learning_rate": 2e-5,
+            "lora_dropout": 0.1,
+            "lora_alpha": 32,
+        },
+        ckpt_output_dir=os.path.join(DEFAULTS._data_dir, "checkpoints"),
+        data_output_dir=os.path.join(DEFAULTS._data_dir, "internal"),
+        data_path=os.path.join(DEFAULTS._data_dir, "datasets"),
+        nproc_per_node=4,
```
RobotSail, line 621: Why are we using only 4 procs per node when we have 8 GPUs? Should this be 8 instead in that case?

src/instructlab/configuration.py
```diff
@@ -585,13 +608,154 @@ def ensure_storage_directories_exist():
...
+            "learning_rate": 2e-5,
+            "lora_dropout": 0.1,
+            "lora_alpha": 32,
+            "deepspeed_cpu_offload_optimizer_ratio": 1,
+            "deepspeed_cpu_offload_optimizer_pin_memory": False,
+        },
+        ckpt_output_dir=os.path.join(DEFAULTS._data_dir, "checkpoints"),
+        data_output_dir=os.path.join(DEFAULTS._data_dir, "internal"),
+        data_path=os.path.join(DEFAULTS._data_dir, "datasets"),
+        nproc_per_node=4,
```
RobotSail, line 723: Shouldn't this be 8 if we have 8 L40 GPUs?

RobotSail (commented):
This looks good for the most part, I left a few comments where I had questions. 

Once those are addressed we can merge this.

src/instructlab/configuration.py
```diff
@@ -619,7 +647,228 @@ def ensure_storage_directories_exist():
...
+            tmpl_file = os.path.join(profile_dir, file)
+            train_cfg = read_train_profile(tmpl_file)
+            if not os.path.isfile(new_file):
+                with open(new_file, "w", encoding="utf-8") as outfile:
+                    d = train_cfg.model_dump_json()
+                    loaded = yaml.load(d, Loader=yaml.SafeLoader)
+                    yaml.dump(loaded, outfile)
+    else:
+        # else render the defaults
+        FOUR_GPU_TRAIN_AH = _train(
```
RobotSail, line 674: Is "FOUR_GPU_TRAIN_AH" referring to 4x A100 and H100? Why are we coupling them together in this way? We should decouple them so that 4xA100, 8xA100, 4xH100, and 8xH100 are all distinct configs. 
Otherwise, if we want to share config values between these, we should have a map setup to point to the common config being used in this case. 

Also - keep in mind that A100s can be either 40GB or 80GB, but H100s are a minimum of 80GB.

src/instructlab/configuration.py
```diff
@@ -619,7 +647,228 @@ def ensure_storage_directories_exist():
...
+            additional_args={
+                "warmup_steps": 10,
+                "learning_rate": 2e-5,
+                "lora_dropout": 0.1,
+                "lora_alpha": 32,
+            },
+            ckpt_output_dir=os.path.join(DEFAULTS._data_dir, "checkpoints"),
+            data_output_dir=os.path.join(DEFAULTS._data_dir, "internal"),
+            data_path=os.path.join(DEFAULTS._data_dir, "datasets"),
+            nproc_per_node=4,
```
RobotSail, line 709: Why is `nproc_per_node` here set to 4 if we have 8 GPUs?

RobotSail (commented):
Thank you for this PR. For some reason my earlier review comments seem to have been erased. 

For us to have an easy time maintaining configs long-term, we should address the following changes before merging:
- H100 and A100 configs seem to be coupled, we should uncouple them here. If the concern is around reusing values, then we can set up a map to point to a common object. But it may be better to duplicate here and be redundant for our own sanity. 
- A number of 8 GPU configs seem to have `nproc_per_node` set to 4 which would only utilize half of the GPUs on the system. Was this intentional? 
- the `too-many-lines` Pylint check seems pretty useless, we should just disable it in the `.pylintrc` file.

src/instructlab/configuration.py
```diff
@@ -619,7 +647,228 @@ def ensure_storage_directories_exist():
         "lora_target_modules": ["q_proj", "k_proj", "v_proj", "o_proj"],
     }
 
-    # create exper_args file for users to see/edit
+    TRAIN_DIR_EXPECTED_FILES = {
+        "A100_H100_x4.yaml",
```
booxter, line 651: (Observation) these are duplicates from above. We could probably have this code refactored so that the list of supported profiles is defined once and the rest is generated (including properties).

src/instructlab/configuration.py
```diff
@@ -619,7 +647,228 @@ def ensure_storage_directories_exist():
...
+            effective_batch_size=96,
+            lora_quantize_dtype=None,
+            lora_rank=0,
+            max_batch_len=60000,
+            max_seq_len=4096,
+            save_samples=1000,
+            deepspeed_cpu_offload_optimizer=False,
+            is_padding_free=True,
+            num_epochs=7,
+            model_path=os.path.join(
```
booxter, line 694: Shouldn't it point to `model_path=DEFAULTS.MODEL_REPO` as in some other places for the same field? (in `get_default_config`).

src/instructlab/configuration.py
```diff
@@ -619,7 +647,228 @@ def ensure_storage_directories_exist():
...
+        to_write = {
+            DEFAULTS.TRAIN_A100_H100_X4_PROFILE: FOUR_GPU_TRAIN_AH,
+            DEFAULTS.TRAIN_A100_H100_X8_PROFILE: EIGHT_GPU_TRAIN_AH,
+            DEFAULTS.TRAIN_A100_H100_X2_PROFILE: TWO_GPU_TRAIN_AH,
+            DEFAULTS.TRAIN_L40_X8_PROFILE: EIGHT_L_FORTY_GPU,
+            DEFAULTS.TRAIN_L40_X4_PROFILE: FOUR_L_FORTY_GPU,
+            DEFAULTS.TRAIN_L4_X8_PROFILE: EIGHT_L_FOUR_GPU,
+        }
+
+        for file, train_cfg in to_write.items():
```
booxter, line 864: nit: this section seems identical to how we dump above for the 'custom profile dir' scenario. It would be nice to have this deduplicated.

src/instructlab/configuration.py
```diff
@@ -619,7 +647,228 @@ def ensure_storage_directories_exist():
...
+            if not os.path.isfile(file):
+                with open(file, "w", encoding="utf-8") as outfile:
+                    d = train_cfg.model_dump_json()
+                    loaded = yaml.load(d, Loader=yaml.SafeLoader)
+                    yaml.dump(loaded, outfile)
+
+        # create exper_args file for users to see/edit
```
booxter, line 871: nit: indent?

src/instructlab/configuration.py
cdoern: fixed

src/instructlab/configuration.py
cdoern: the configurations provided to me by the training team coupled 4x and 8x a100s and h100s. There are no differences between the two from what I found. 

We are only supporting 80GB+ A100 configs, not 40

src/instructlab/configuration.py
```diff
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# pylint: disable=too-many-lines
```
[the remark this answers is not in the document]
RobotSail: Agree

RobotSail (approved):
LGTM, just make sure to add the two pylint disables into the global ignore file, since neither line length nor number of public methods should be things we constrain ourselves with.

src/instructlab/config/init.py
```diff
@@ -103,6 +103,20 @@ def init(
...
+            click.echo(f"{i}. {value}")
+        selection = click.prompt(
+            "Enter the number of your choice", type=int, default=-1
+        )
+        if 0 <= selection < len(entries):
+            click.echo(f"You selected: {entries[selection]}")
+        elif selection == -1:
+            click.echo("Using default train profile.")
+        else:
+            click.secho("Invalid selection, a valid train profile option.", fg="red")
```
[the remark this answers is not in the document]
cdoern: yeah, I will error out

src/instructlab/config/init.py
```diff
@@ -103,6 +103,26 @@ def init(
     click.echo(f"Generating `{DEFAULTS.CONFIG_FILE}`...")
     if train_profile is not None:
         cfg.train = read_train_profile(train_profile)
+    elif interactive:
+        entries = listdir(DEFAULTS.TRAIN_PROFILE_DIR)
+        click.echo("Please choose a train profile to use:")
+        for i, value in enumerate(entries):
+            click.echo(f"{i}. {value}")
+        selection = click.prompt(
```
JamesKunstle, line 111: nit: s/selection/train_profile_selection

src/instructlab/config/init.py
```diff
@@ -103,6 +103,26 @@ def init(
...
+        if 0 <= selection < len(entries):
+            click.echo(f"You selected: {entries[selection]}")
+        elif selection == -1:
+            click.echo("Using default train profile.")
+        else:
+            click.secho(
+                "Invalid selection. Please select a valid train profile option.",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
```
JamesKunstle, line 125: I think I'm missing something, but I'm not seeing that you're doing anything w/ the `selection` aside from checking that it's in range of the available options.

src/instructlab/configuration.py
JamesKunstle: ++

src/instructlab/config/init.py
booxter: Indeed. We probably need a test to confirm this actually works. The interactive part may be a bit challenging though.

src/instructlab/configuration.py
```diff
@@ -619,7 +644,228 @@ def ensure_storage_directories_exist():
...
+            save_samples=1000,
+            deepspeed_cpu_offload_optimizer=False,
+            is_padding_free=True,
+            num_epochs=7,
+            model_path=os.path.join(
+                DEFAULTS._cache_home, "models/instructlab/granite-7b-lab"
+            ),
+        )
+
+        EIGHT_GPU_TRAIN_AH = _train(
```
JamesKunstle, line 696: These configs have a lot of shared values. Would it define a single base config and additional_params and override each of these config objects with the values that change between them? It'd be easier to identify what the major changes are between the profiles and make refactoring a lot easier.

booxter:
@JamesKunstle doesn't the seemingly ignored selection justify a respin?..

src/instructlab/configuration.py
RobotSail: This was my initial thought, but for the sake of getting this across I think we should handle this in a follow-up PR. It will involve additional folding logic that we'd need to bring in, which expands the scope of this PR.

src/instructlab/config/init.py
nathan-weinberg: A test would be good - it seems from the user docs that maybe the ret value doesn't _need_ to be used, but I'm curious if it's needed at all: https://click.palletsprojects.com/en/8.1.x/api/#click.prompt

src/instructlab/configuration.py
nathan-weinberg: Agreed - I think this is fine for now but it would be a welcome future optimization

src/instructlab/configuration.py
```diff
@@ -619,7 +644,228 @@ def ensure_storage_directories_exist():
...
+        "A100_H100_x4.yaml",
+        "A100_H100_x8.yaml",
+        "A100_H100_x2.yaml",
+        "L40_x8.yaml",
+        "L40_x4.yaml",
+        "L4_x8.yaml",
+    }
+
+    profile_dir = os.environ.get(DEFAULTS.ILAB_TRAIN_PROFILE_DIR)
+    if profile_dir != "" and profile_dir is not None:
```
nathan-weinberg, line 657: What is the case we are checking for where `profile_dir` could be equal to an empty string?

cdoern: if someone messes with the env var

src/instructlab/config/init.py
cdoern: very good catch, fixed

src/instructlab/configuration.py
booxter: The `!= ""` check will work when users pass it as follows: `VAR= ilab ...` in their shell. To completely remove it from the environment (and not just to set it to an empty string), users would have to `env -u VAR ilab` which is more cumbersome. I think it's good we handle both.

src/instructlab/configuration.py
cdoern: I'm not sure how this'd work given that not everything is in `additional_args`. If we were to define a base _train class and then create other _train classes, I am not aware of a way to unwrap one into the other? I might be missing something though

booxter: You could probably have some dicts with common and specific settings, and then

```
base_config = { ... }
config1 = { ... }
config2 = { ... }
final_config = _train(**(base_config | config1))
```

JamesKunstle: Pydantic objects aren't immutable by default.

```
base_config = _train(**defaults)
EIGHT_L_FOUR_GPU = base_config.model_copy(deep=True)
EIGHT_L_FOUR_GPU.nproc_per_node = 8
EIGHT_L_FOUR_GPU.max_batch_len = 30000
EIGHT_L_FOUR_GPU.lora_rank = 0
EIGHT_L_FOUR_GPU.deepspeed_cpu_offload_optimizer=True
```

There's a tradeoff with this, which is that the input to the objects aren't as nicely formed, but you could do something like:

```
base_config = _train(**defaults)
EIGHT_GPU_TRAIN_AH = base_config.model_copy(deep=True)
EIGHT_GPU_TRAIN_AH.nproc_per_node = 8
# ... any overrides for eight GPUs of A or H

FOUR_GPU_TRAIN_AH = EIGHT_GPU_TRAIN_AH.model_copy(deep=True)
FOUR_GPU_TRAIN_AH.nproc_per_node = 4
```

JamesKunstle:
> @JamesKunstle doesn't the seemingly ignored selection justify a respin?..

wdym a respin?

booxter (approved):
This looks ok now. Thanks.

---

There's a question about why `model_path` for custom training profiles has to list full path instead of model name (as in the default config), but perhaps I misunderstand something. I will let others figure it out.

src/instructlab/config/init.py
```diff
@@ -103,6 +103,29 @@ def init(
...
+            click.echo(f"{i}. {value}")
+        train_profile_selection = click.prompt(
+            "Enter the number of your choice [hit enter for the default]",
+            type=int,
+            default=-1,
+        )
+        if 0 <= train_profile_selection < len(entries):
+            click.echo(f"You selected: {entries[train_profile_selection]}")
+            cfg.train = read_train_profile(
+                join(DEFAULTS.TRAIN_PROFILE_DIR, entries[train_profile_selection])
```
JamesKunstle, line 119: nit: generally, it's preferred to not import single-symbol references like this, and to instead use the full name like `os.path.join`. It's a python disambiguation preference that similar to naming a list of names `name_list` instead of just `names`, partially as a pre-annotations legible types thing.

booxter:
@JamesKunstle respin is an updated patch with an identified bug fixed. (See [here](https://www.quora.com/What-is-respin-in-software-testing-life-cycle) for example.) Charlie already pushed an update fixing the selection bug you found, so we are good to go now.

src/instructlab/configuration.py
```diff
@@ -619,7 +647,228 @@ def ensure_storage_directories_exist():
...
+    else:
+        # else render the defaults
+        FOUR_GPU_TRAIN_AH = _train(
+            additional_args={
+                "warmup_steps": 10,
+                "learning_rate": 2e-5,
+                "lora_dropout": 0.1,
+                "lora_alpha": 32,
+            },
+            ckpt_output_dir=os.path.join(DEFAULTS._data_dir, "checkpoints"),
```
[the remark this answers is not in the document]
JamesKunstle: ++, don't need to do the joins in the config.

JamesKunstle (approved):
I've noted the few things that I think should be updated in the future, but improvement aside, I think this is good to go in and we can attend to anything that we don't like nit-picky wise in the future.

PR #992, opened 2024-04-25
chore: update chat greeting to InstructLab

nathan-weinberg:
Right now `ilab chat` greets with `Welcome to Chat CLI` - I think it may make more sense to be `Welcome to InstructLab CLI Chat`

PR #1961, opened 2024-08-02
Refactor the CUDA container spec

### LINKED ISSUE #899: Container ships build dependencies

fabiendupont:
This change modifies the Containerfile for NVIDIA CUDA to:
- Use CentOS Stream 9 as the default base image. UBI9 also works
- Install the NVIDIA packages from NVIDIA repository for exact CUDA version
- Add NVIDIA scripts from `nvcr.io/nvidia/cuda` image for copyrights...
- Install the Python packages in a Python 3.11 virtualenv
- Install instructlab package from PyPI
- Allow specifying the instructlab version at build time
- Make the final container rootless

Resolves #899
Replaces #932
Replaces #994

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
+INSTRUCTLAB_VERSION ?=
+
 BUILD_ARGS =
 CENGINE ?= podman
 CONTAINER_PREFIX ?= localhost/instructlab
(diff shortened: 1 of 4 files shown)


containers/README.md
```diff
@@ -6,8 +6,9 @@ using dedicated GPUs. This guide shows you how to put the `ilab` CLI, all of its
 dependencies, and your GPU into a container for an isolated and easily reproducible
 experience.
 
-## Steps to build an image then run a container
+## Build the `ilab` container image
 
+<<<<<<< HEAD:containers/README.md
```
bcrochet, line 11: Need to resolve the conflict.

fabiendupont: Good catch. Thanks @bcrochet

bcrochet (commented):
What I've found so far. I'll be running a build shortly to test.

nathan-weinberg:
@fabiendupont did you mean "replaces https://github.com/instructlab/instructlab/pull/992"? seems unrelated

fabiendupont:
> @fabiendupont did you mean "replaces #992"? seems unrelated

Thanks. Reference updated.

bcrochet (changes requested):
Just a missing letter...

nathan-weinberg (approved):
one nit but otherwise LGTM

Makefile
```diff
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
+INSTRUCLAB_VERSION ?=
```
markstur, line 3: misspelled "INSTRUCLAB" here, but it has the 2nd T when used below.

markstur (changes requested):
typo on var name

Makefile
fabiendupont: Thanks

PR #1308, opened 2024-06-07
feat: Allow qna yaml files with .yml extension

gfbarros:
**Issue resolved by this Pull Request:**
Resolves #1306

**Checklist:**

- [ x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ x] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #1962, opened 2024-08-02
Use instructlab-schema package to parse qna.yaml files

bjhargrave:
We replace parsing/validation code to use the shared code from instructlab-schema package.

**Checklist:**

### DIFF
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -1,11 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
-from functools import cache, wraps
-from importlib import resources
-from importlib.abc import Traversable
+from functools import wraps
 from pathlib import Path
-from typing import Any, List, Mapping, Optional, Tuple, TypedDict
+from typing import List, Tuple, TypedDict
 from urllib.parse import urlparse
 import copy
 import glob
@@ -21,7 +19,12 @@ import tempfile
 
 # Third Party
 from git import Repo, exc
-from referencing import Resource
+from instructlab.schema.taxonomy import DEFAULT_TAXONOMY_FOLDERS as TAXONOMY_FOLDERS
+from instructlab.schema.taxonomy import (
+    TaxonomyMessageFormat,
+    TaxonomyParser,
+    TaxonomyReadingException,
+)
 import click
 import git
 import gitdb
@@ -33,16 +36,6 @@ from . import common
 
 logger = logging.getLogger(__name__)
 
-MIN_KNOWLEDGE_VERSION = 3
-
-DEFAULT_YAML_RULES = """\
-extends: relaxed
-
-rules:
-  line-length:
-    max: 120
-"""
-
 
 class SDGTokens:
     """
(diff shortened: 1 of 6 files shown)


mairin:
Right now taxonomy diff does 5 layers of checking. It will be handled by this now. 

@nathan-weinberg can you describe the 5 layers here?

This came up in our team weekly meeting today, when it was suggested taxonomy diff should check the yaml schema version and let the user know if supported or not. Right now we flag at sdg generate and that's a bit late in the process.

bjhargrave:
The parsing/validation code will check the knowledge version for v3 and report an error for v1/v2. There is a test case in this PR for this.

https://github.com/instructlab/instructlab/pull/1962/files#diff-f2fb20ddc8bf7f3b80b1a99bfe9fd3e563c5675f9faf7f914ac7f74e276d15f1R348-R368

The parsing/validation code will handle the linting and json schema validation reporting errors if found. I don't know what the 5 layers are.

nathan-weinberg:
5 layers are:
1. file extension check (is it `.yaml`)
2. content check (is the file empty)
3. YAML lint check (is the file valid yaml)
4. Schema check (does the file follow the taxonomy schema)
5. Knowledge check (does the file have everything needed for knowledge if applicable)

This PR should defintely be a `0.19.0` priority IMO - we want to align the schema checking as much as possible while minimizing code duplication (there's some code for this in the SDG library as well)

src/instructlab/utils.py
```diff
@@ -185,27 +174,29 @@ def lab_check_callback(*args, **kwargs):
...
-TAXONOMY_FOLDERS: List[str] = ["compositional_skills", "knowledge"]
-"""Taxonomy folders which are also the schema names"""
-
-
 def is_taxonomy_file(fn: str) -> bool:
     path = Path(fn)
     if path.parts[0] not in TAXONOMY_FOLDERS:
         return False
-    if path.name == "qna.yml":
+    if path.name == "qna.yaml":
```
RobotSail, line 181: Can we change this to allow for `.yml` in addition to `.yaml`? They are both valid yaml extensions, so it being restrictive here doesn't make sense.

bjhargrave: This discussion has been had multiple times. :-) The conclusion is to have a single name `qna.yaml` to avoid overlapping names: qna.yml and qna.yaml in the same folder. It is for consistency and disambiguation.

bjhargrave: See 

https://github.com/instructlab/instructlab/pull/1905
https://github.com/instructlab/instructlab/issues/1899
https://github.com/instructlab/instructlab/pull/1308
https://github.com/instructlab/instructlab/issues/1306

for example.

RobotSail: @bjhargrave I see, that makes sense. What would be the problem with checking if both files exist in the directory, and simply rejecting the newer one?

bjhargrave: That is possible but complicated. It also is somewhat ambiguous since a user can edit one and then edit another so the files can alternate in being newest. It would also be dependent on how the taxonomy files were realized for a generate run. We don't know that the realization is file metadata preserving.

If both were present, I think we would want to error out anyway since that indicates a problem.

Also this code is shared by the taxonomy repo for PR checks and in the taxonomy repo we want to enforce some level of uniformity and consistency.

RobotSail: That makes sense, thank you for providing the detailed context. In that case I think we can just assume `qna.yaml` for now, as it doesn't sound like a high priority to work that into this PR.

nathan-weinberg (approved):

PR #1966, opened 2024-08-02
train: Remove cli options that are not used

booxter:
These options seem dead end? Not read by anything. Is it a remnant of migration to train profiles or something else?

**Checklist:**


cdoern (changes requested):
these are all used! they are passed in via kargs to the train library and directly unfolded into a `TrainingArgs` and `TorcRunArgs` object via map_train_to_library

booxter:
You are right! Not via `kwargs` but via `ctx.params` indeed!

This should probably be more explicit (pass argument values into `train()` function explicitly, then construct the torch and training args directly, not via `ctx.params`, which is a bit implicit to my taste.

PR #1968, opened 2024-08-02
Increase default vllm startup attempts and make configurable

danmcp:
With larger models, the existing ~5 mins timeout proved not to be sufficient on some hardware.  The longest startup times seen so far have been in the 8 min range.  This moves the default up to 20 mins and makes the retry attempts configurable.  The new config param is being added as optional so config changes aren't required.

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -260,6 +260,7 @@ class _serve_vllm(BaseModel):
     """Class describing configuration of vllm serving backend."""
 
     llm_family: str = ""
+    max_startup_attempts: int | None = None
     # arguments to pass into vllm process
     vllm_args: list[str] | None = None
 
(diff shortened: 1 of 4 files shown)


n1hility (approved):

src/instructlab/configuration.py
```diff
@@ -260,6 +260,7 @@ class _serve_vllm(BaseModel):
     """Class describing configuration of vllm serving backend."""
 
     llm_family: str = ""
+    max_startup_attempts: typing.Optional[int] = None
```
alinaryan, line 263: @RobotSail mentioned in another review that pydantic might not always interpret typing.Optional properly so we might want something like `int | None = None` here

danmcp: Updated, thanks!

alimaredia (approved):

PR #1969, opened 2024-08-03
Fail fast when vllm errors out and exits

danmcp:
Previously, if vllm failed to start and raised an exception, the logic would have tried until its max attempts ran out to connect to a server that had already died.  This new logic detects if the server process is stopped, and if so outputs a message to use --enable-serving-output to debug further if the option is available.  So users will get a quick failure and guidance on debugging.

**Checklist:**

### DIFF
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -262,6 +262,7 @@ def generate(
                     }
                 ),
                 background=not enable_serving_output,
+                foreground_allowed=True,
             )
         except Exception as exc:
             click.secho(f"Failed to start server: {exc}", fg="red")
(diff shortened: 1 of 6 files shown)


n1hility (approved):

booxter (approved):

alinaryan (approved):

alimaredia:
@danmcp Any reason like we wouldn't want the additional log message anytime `background` is set to `True`.

Couldn't the same functionality in this PR have been accomplished with just:
```
           if vllm_server_process.poll():
                if background:
                    raise ServerException(
                        "vLLM failed to start.  Retry with --enable-serving-output to learn more about the failure."
                    )
                raise ServerException("vLLM failed to start.")
```

danmcp:
> @danmcp Any reason like we wouldn't want the additional log message anytime `background` is set to `True`.

Chat doesn't have a foreground option.  So in that case it would tell people about a --enable-serving-output option that doesn't exist.

PR #1972, opened 2024-08-03
doc: remove "testing doesn't work on Linux"

makelinux:
Already supported, see #1337

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 
 ## 📋 Requirements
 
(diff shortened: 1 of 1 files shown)


booxter (approved):
This seems reasonable but I'd love someone with more knowledge about the project support status for Linux to chime in. @cdoern @RobotSail Are there any other features beyond `model test` not available for Linux at the moment that would justify a "not fully supported" statement?

danmcp (approved):
I think this is a good change, but the story of what works on Linux vs Mac has become more complicated.

RobotSail (approved):
LGTM

PR #1975, opened 2024-08-05
feat: add pipeline into generate config

alimaredia:
Most flags have config variable associated with them. The --pipeline flag for the `ilab data generate` command does not have a config variable associated with it. Ideally the `ilab data generate` command should need as few flags as possible once the config file is set.

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -84,6 +84,7 @@ class _InstructlabDefaults:
     CONNECTION_TIMEOUT = httpx.Timeout(timeout=30.0)
     # use spawn start method, fork is not thread-safe
     MULTIPROCESSING_START_METHOD = "spawn"
+    SDG_PIPELINE = "simple"
     SDG_SCALE_FACTOR = 30
 
     # When otherwise unknown, ilab uses this as the default family
@@ -363,6 +364,7 @@ class _generate(BaseModel):
     model_config = ConfigDict(extra="ignore")
 
     # required fields
+    pipeline: StrictStr = DEFAULTS.SDG_PIPELINE
     model: StrictStr
     taxonomy_path: StrictStr
     taxonomy_base: StrictStr
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -364,7 +364,7 @@ class _generate(BaseModel):
     model_config = ConfigDict(extra="ignore")
 
     # required fields
-    pipeline: StrictStr = DEFAULTS.SDG_PIPELINE
+    pipeline: Optional[str] = DEFAULTS.SDG_PIPELINE
     model: StrictStr
     taxonomy_path: StrictStr
     taxonomy_base: StrictStr
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -143,7 +143,7 @@ logger = logging.getLogger(__name__)
 @click.option(
     "--pipeline",
     type=click.STRING,
-    default="simple",
+    cls=clickext.ConfigOption,
     help="Data generation pipeline to use. Available: simple, full, or a valid path to a directory of pipeline workflow YAML files. Note that 'full' requires a larger teacher model, Mixtral-8x7b.",
 )
 @click.option(
@@ -284,7 +284,7 @@ def generate(
 
     try:
         click.echo(
-            f"Generating synthetic data using '{model_path}' model, taxonomy:'{taxonomy_path}' against {api_base} server"
+            f"Generating synthetic data using '{pipeline}' pipeline, '{model_path}' model, '{taxonomy_path}' taxonomy, against {api_base} server"
         )
         generate_data(
             logger=logging.getLogger("instructlab.sdg"),  # TODO: remove
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -198,6 +198,18 @@ def generate(
     from instructlab.sdg.generate_data import generate_data
     from instructlab.sdg.utils import GenerateException
 
+    # if --pipeline is not used, pipeline defaults to the value of ctx.obj.config.generate.pipeline
+    # set in the config file. A user could intentionally set this to 'null' in the config file
+    # if they want to ensure --pipeline needs to be used.
+    # This would happen if the type of pipeline needs to be different across different runs of
+    # `ilab data generate`.
+    if not pipeline:
+        click.secho(
+            "Pipeline not set. Please use the --pipeline flag or set it in the config file.",
+            fg="red",
+        )
+        raise click.exceptions.Exit(1)
+
     if num_instructions != -1:
         click.secho(
             "The --num-instructions flag is deprecated. Please use --sdg-scale-factor instead.",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,8 @@
 * Legacy Linux training is now compatible with the phase07 pretraining format.
 * Add support for `ILAB_TRAIN_PROFILE_DIR` which will point to the template train profiles to be brought into the `train_configuration` directory.
 * Add interactive prompt for users to choose their train profile.
+* The `generate` section of the config now has a `pipeline` value. This value sets a default value
+  and can be overridden by the `--pipeline` flag. The default for this value is 'simple'.
 
 ### Breaking Changes
 
--- a/README.md
+++ b/README.md
@@ -450,7 +450,7 @@ Before following these instructions, ensure the existing model you are adding sk
    (venv) $ ilab data generate
    INFO 2024-07-30 19:57:44,093 numexpr.utils:161: NumExpr defaulting to 8 threads.
    INFO 2024-07-30 19:57:44,452 datasets:58: PyTorch version 2.3.1 available.
-   Generating synthetic data using '$HOME/.cache/instructlab/models/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf' model, taxonomy:'./taxonomy' against http://localhost:8000/v1 server
+   Generating synthetic data using 'simple' pipeline, '$HOME/.cache/instructlab/models/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf' model, './taxonomy' taxonomy, against http://localhost:8000/v1 server
    INFO 2024-07-30 19:57:45,084 instructlab.sdg:375: Synthesizing new instructions. If you aren't satisfied with the generated instructions, interrupt training (Ctrl-C) and try adjusting your YAML files. Adding more examples may help.
    INFO 2024-07-30 19:57:45,090 instructlab.sdg.pipeline:153: Running pipeline single-threaded
    INFO 2024-07-30 19:57:47,820 instructlab.sdg.llmblock:51: LLM server supports batched inputs: False
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -59,6 +59,7 @@ class TestConfig:
         assert cfg.generate.teacher.host_port == "127.0.0.1:8000"
         assert cfg.generate.teacher.backend is None
         assert cfg.generate.teacher.chat_template is None
+        assert cfg.generate.pipeline == "simple"
         assert cfg.generate.model == default_model
         assert cfg.generate.taxonomy_path == f"{data_dir}/taxonomy"
         assert cfg.generate.taxonomy_base == "origin/main"
@@ -128,6 +129,7 @@ class TestConfig:
                 """general:
   log_level: INFO
 generate:
+  pipeline: simple
   teacher:
     model_path: models/granite-7b-lab-Q4_K_M.gguf
     chat_template: tokenizer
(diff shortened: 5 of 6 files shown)


danmcp:
@alimaredia Can you state why this is blocker proposed?

alimaredia:
Updated the comment. One thing to note is that the instructlab-sdg library verifies the input for `pipeline` is valid, so I did not include any  validation in this PR.

danmcp (commented):
> Most flags have config variable associated with them. The --pipeline flag for the `ilab data generate` command does not have a config variable associated with it. Ideally the `ilab data generate` command should need as few flags as possible once the config file is set.

--pipeline is a lot like --benchmark for evaluate.  Meaning you could logically make it a separate subcommand.  Which makes me wonder whether it should be in config.  One question that might be helpful is: would a use cases typically want to run with different pipelines across different runs or is this a set it once an forget it setting.  I am guessing this is more likely if you are training multiple models.  You might want to use different pipelines for different models being trained on the same system.

If we are going to add this to config, we should make sure to print out which pipeline is running at the beginning of generate.

alimaredia:
I was thinking that it would be a set it and forget it. Correct me if I'm wrong, but the reason someone would use the 'simple' vs the 'full' have to do with hardware resources. The reasons for 'full' vs something that's custom that would be less black and white. Do you know who else could clarify this one for us? I was thinking you would know.

src/instructlab/data/generate.py
```diff
@@ -283,7 +283,7 @@ def generate(
 
     try:
         click.echo(
-            f"Generating synthetic data using '{model_path}' model, taxonomy:'{taxonomy_path}' against {api_base} server"
+            f"Generating synthetic data using '{pipeline}' pipeline, '{model_path}' model, '{taxonomy_path}' taxonomy, against {api_base} server"
```
alimaredia, line 286: @danmcp added the pipeline to the existing log message before `generate_data`.

aakankshaduggal (commented):
Thanks @alimaredia for initiating this. 

If I understand correctly `ilab data generate` doesn't work at all unless we define `--pipeline`. And I support the default to simple but we should also update the documentation [here](https://github.com/instructlab/instructlab/blob/main/README.md#-generate-a-synthetic-dataset) to reflect these changes. 

cc- @kelbrown20

alimaredia:
@aakankshaduggal `ilab data generate` defaults the `pipeline` to `simple`. But you can't just run a `ilab data generate` with a `full` pipeline or a custom pipeline.

danmcp:
> I was thinking that it would be a set it and forget it. Correct me if I'm wrong, but the reason someone would use the 'simple' vs the 'full' have to do with hardware resources. The reasons for 'full' vs something that's custom that would be less black and white. Do you know who else could clarify this one for us? I was thinking you would know.

Realistically I do think this this is only a concern for full vs custom vs future options.

Since it's already defaulted with the click option, I guess it's not significantly different moving the default to config.  Allowing for null in the config might be an improvement in that regard so you could force a value to be specified with the command if you wanted.

I am also questioning whether this is a blocker.

alimaredia:
When we triaged the blockers today, our thinking was that this is a blocker because it's a UX improvement in the `full` and `custom` pipelines cases of `ilab data generate` and that it's a small fix a PR already open.

danmcp:
> When we triaged the blockers today, our thinking was that this is a blocker because it's a UX improvement in the `full` and `custom` pipelines cases of `ilab data generate` and that it's a small fix a PR already open.

It is a small change, but does require a new property in the config, corresponding change in all existing config templates, re-testing, and doc updates.

If this does need to go in, my remaining request is we address whether the new config should be required (as it is in the PR currently) or whether optional should be allowed with an exit and message added to generate.

src/instructlab/data/generate.py
```diff
@@ -143,7 +143,7 @@
 @click.option(
     "--pipeline",
     type=click.STRING,
-    default="simple",
+    cls=clickext.ConfigOption,
     help="Data generation pipeline to use. Available: simple, full, or a valid path to a directory of pipeline workflow YAML files. Note that 'full' requires a larger teacher model, Mixtral-8x7b.",
```
booxter, line 147: (Observation) These should probably be enforced as enum CLI option. Same for config option allowed values.

booxter (approved):
No brainer.

src/instructlab/configuration.py
```diff
@@ -444,6 +446,7 @@ def get_default_config() -> Config:
             model=DEFAULTS.DEFAULT_MODEL,
         ),
         generate=_generate(
+            pipeline=DEFAULTS.SDG_PIPELINE,
```
booxter, line 449: nit: I think this is redundant, the model already has default set on the attribute. This will be used by default.

alimaredia:
> > When we triaged the blockers today, our thinking was that this is a blocker because it's a UX improvement in the `full` and `custom` pipelines cases of `ilab data generate` and that it's a small fix a PR already open.
> 
> It is a small change, but does require a new property in the config, corresponding change in all existing config templates, re-testing, and doc updates.
> 
> If this does need to go in, my remaining request is we address whether the new config should be required (as it is in the PR currently) or whether optional should be allowed with an exit and message added to generate.

@danmcp I've added an entry to the CHANGELOG. Is there anywhere else we need to add this new value?

danmcp:
> @danmcp I've added an entry to the CHANGELOG. Is there anywhere else we need to add this new value?

Nothing in instructlab that I am aware of.

For instructlab, my remaining question is whether the config should be optional (null allowed).

booxter:
@danmcp what would `null` mean? Right now, if the config option is omitted from the config file, the default value - `simple` would be used. If not default, then what would `null` mean?

danmcp:
> @danmcp what would `null` mean? Right now, if the config option is omitted from the config file, the default value - `simple` would be used. If not default, then what would `null` mean?

In a previous comment I was asking whether null (if allowed) should imply the user must specify.

booxter:
I think we are moving to the place where an empty config file will have defaults filled in runtime: https://github.com/instructlab/instructlab/pull/1945 so I think it's reasonable to choose a default here.

danmcp:
> I think we are moving to the place where an empty config file will have defaults filled in runtime: #1945 so I think it's reasonable to choose a default here.

I am probably missing something still, but this doesn't seem like a case that needs a default in all cases.   It seems like asking the user to specify if a correct answer isn't known is reasonable.

booxter:
Do you mean that if it's `null` (either missing in config file or set there to `null`), then `data generate` would fail? ask interactively?

If so, this could be a reasonable behavior and someone with better UX taste would have to chime in. But I think the code already defaults to `simple`, and this PR was not trying to change that?

danmcp:
> Do you mean that if it's `null` (either missing in config file or set there to `null`), then `data generate` would fail? ask interactively?


I was suggesting that if we do allow null, it would fail with a message that either --pipeline needs to be specified or added to the config.



> 
> If so, this could be a reasonable behavior and someone with better UX taste would have to chime in. But I think the code already defaults to `simple`, and this PR was not trying to change that?

It would still be the default if that's the default in the config (although I am not sure we can't change the default still).  I am asking to consider allowing null to indicate --pipeline is required as an option.  Or state why that's not a valid option.

alimaredia:
I just generated a config and changed the value of `pipeline` to `null` and tried to run `ilab data genereate`. The error below already gets thrown if that's the case. Wouldn't this invalidate the possibility of `null` for `pipeline` and having to account for it? 

```
(venv) [ali@vm instructlab (sdg-pipeline-config-var)]$ grep -R "pipeline" ~/.config/instructlab/config.yaml 
  pipeline: null
(venv) [ali@vm instructlab (sdg-pipeline-config-var)]$ ilab data generate
Usage: ilab data [OPTIONS] COMMAND [ARGS]...
Try 'ilab data --help' for help.

Error: 1 errors in /home/ali/.config/instructlab/config.yaml:
- string_type generate->pipeline: input should be a valid string
```

danmcp:
> Wouldn't this invalidate the possibility of `null` for `pipeline` and having to account for it?

My question isn't about whether the code currently accounts for it.  I am asking whether null should be allowed in which case it would indicate --pipeline is a required arg.  The reason I am asking the question is because it's not clear (to me at least) the future of --pipeline will always be to set as a config option and have a reasonable default vs the pipeline picked will vary enough that users should be asked to select a pipeline per run is common.

danmcp (approved):

danmcp (changes requested):
Removing approval, was added unintentionally

danmcp (commented):
Looks good.  Just the one nit and a rebase.

src/instructlab/configuration.py
khaledsulayman: agree w/ this. Better to just remove this line unless it's providing a different pipeline

src/instructlab/data/generate.py
khaledsulayman: +1 it's also pretty easy to implement. Something like this:
```
from enum import Enum

...

class SDGPipeline(Enum):
    simple = 0
    full = 1
...

pipeline = SDGPipeline(pipeline_arg)  # assuming pipeline_arg is the arg passed in the CLI in string format, for example, "simple"
# pipeline will be < SDGPipeline.simple: 0>
```

This way you can still do the same checking against CLI args while having a more robust approach overall with Enums.

khaledsulayman (commented):
General question: where/how do we check the architecture that `lab config init` is being run on so we can decide whether to set `simple`, `full`, or otherwise? In this PR I only see `simple` being set as the default globally, but my understanding is that the main reason for introducing this to the config is so that we can have `lab data generate` with no additional args to be able to run the correct pipeline depending on the machine it is run on.

src/instructlab/data/generate.py
danmcp: Paths to custom pipelines also need to be allowed.  Not sure if there is a way to validate enum or random str.

danmcp (approved):

src/instructlab/configuration.py
alimaredia: Removed

src/instructlab/data/generate.py
alimaredia: This might be a good follow up PR once this is merged.

src/instructlab/data/generate.py
```diff
@@ -198,6 +198,18 @@ def generate(
     from instructlab.sdg.generate_data import generate_data
     from instructlab.sdg.utils import GenerateException
 
+    # if --pipeline is not used, pipeline defaults to the value of ctx.obj.config.generate.pipeline
+    # set in the config file. A user could intentionally set this to 'null' in the config file
+    # if they want to ensure --pipeline needs to be used.
+    # This would happen if the type of pipeline neeeds to be different across different runs of
```
booxter, line 204: `neeeds`! :)

src/instructlab/data/generate.py
```diff
@@ -198,6 +198,18 @@ def generate(
     from instructlab.sdg.generate_data import generate_data
     from instructlab.sdg.utils import GenerateException
 
+    # if --pipeline is not used, pipeline defaults to the value of ctx.obj.config.generate.pipeline
```
booxter, line 201: I'm not sure this is a great direction. Are we going to add nulls to all other settings to "ensure --<option> needs to be used"? But I won't insist otherwise.

alimaredia:
> General question: where/how do we check the architecture that `lab config init` is being run on so we can decide whether to set `simple`, `full`, or otherwise? In this PR I only see `simple` being set as the default globally, but my understanding is that the main reason for introducing this to the config is so that we can have `lab data generate` with no additional args to be able to run the correct pipeline depending on the machine it is run on.

If by "check the architecture" you mean auto detect which pipeline to use based on the hardware ilab is being run on, currently we're not doing that. 

The reason for introducing this to the config is just so that you don't have to set the flag when you run `ilab data generate`. For the current version of ilab we expect users to use the same pipeline type over and over depending on their hardware. Instead of passing the flag the user can just set the value in the config and forget about it.

booxter (approved):

src/instructlab/data/generate.py
alimaredia: This comment was left as a note to remember a conversation that Dan and I had about his previous comments on the PR. After we talked I understood why he wanted to make the default for `pipeline` and `Optional[str]` instead of a string and that was to retain the flexibility for the behavior described in this comment.

Is this behavior ideal, probably not, but we can determine what's best in the future.

khaledsulayman (approved):

src/instructlab/data/generate.py
danmcp: @booxter Do you have another recommendation on how to achieve the same result?  Or are you saying we must force a default to be specified?

danmcp: Thinking about this a bit more, is this any different than we handle backend for serving?  The only difference I can think of is we try to auto detect backend.  But if it's null in config, and autodetect doesn't work, you get an error.  And you need to specify --backend for force an option.

PR #1978, opened 2024-08-05
Use separate SDG outputs for each phase of phased training

bbrowning:
Instead of using the single messages jsonl file when testing phased training, we should be using the data-mixed separate outputs designed for the separate knowledge and skills training phases.

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -306,10 +306,13 @@ test_train() {
 test_phased_train() {
     task Train the model with LAB multi-phase strategy.
 
-    # linter wants DATA_PATH and over variables declared then assigned separately.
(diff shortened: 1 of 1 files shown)


bbrowning:
This is marked as a draft because I have no way to test it locally and CI doesn't actually test phased training yet. But, this is an idea of how I think this phased training test needs to be updated to work with SDG's new data mixing output.

bbrowning:
This is still relevant,but also requires further work (or a separate PR) to get CI to actually run a phased training job. We could add this to the e2e-nvidia-a10g-x4 workflow, although that would substantially increase the time it takes to complete.

bbrowning:
Closing this as CI has evolved a lot since this was opened.

PR #1979, opened 2024-08-05
fix: wrong phased_phase2_data path into _run_phased_training

JamesKunstle:
@bbrowning caught that we were passing `phased_phase1_data` into `_run_phased_training` where it should have been `phased_phase2_data`


bbrowning (approved):
Looks good - thanks for the quick fix!

danmcp (approved):

nathan-weinberg (approved):

PR #1986, opened 2024-08-06
fix: introduce vram stability waiting

n1hility:
vLLM requires GPU vram usage to be stable when calculating its memory pools. A hard failure on start occurs when it detects this is not the case. This creates a racy failure condition for various InstructLab facilities, such as evaluation, which rely on multiple successive restarts.

Address the problem by introducing a cool down period after process termination. This will complete as early as 10s if VRAM stability is detected. In the event of in-flight reclamation this will wait as long as 30s.

Additionally address an issue where shutdown was being run twice.

No tests are included in this PR since the condition is racy. Coverage is provided by standard CI. In the future we should introduce a stress harness.

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
-from time import sleep, time
+from time import monotonic, sleep, time
 from types import FrameType
 from typing import Optional, Tuple
 import abc
@@ -10,6 +10,7 @@ import json
 import logging
 import mmap
 import multiprocessing
+import os
 import pathlib
 import signal
 import socket
@@ -39,6 +40,7 @@ SUPPORTED_BACKENDS = frozenset({LLAMA_CPP, VLLM})
 API_ROOT_WELCOME_MESSAGE = "Hello from InstructLab! Visit us at https://instructlab.ai"
 CHAT_TEMPLATE_AUTO = "auto"
 CHAT_TEMPLATE_TOKENIZER = "tokenizer"
+
 templates = [
     {
         "model": "merlinite",
@@ -279,6 +281,22 @@ def get(model_path: pathlib.Path, backend: str | None) -> str:
     return backend
 
 
+def get_max_stable_vram_wait(timeout: int) -> int:
+    # Internal env variable for CI adjustment / disablement
+    env_name = "ILAB_MAX_STABLE_VRAM_WAIT"
+    wait_str = os.getenv(env_name)
+    wait = timeout
+    try:
+        if wait_str:
+            wait = int(wait_str)
+    except ValueError:
+        logger.warning(
+            'Ignoring invalid timeout value for %s ("%s")', env_name, wait_str
+        )
+
+    return wait
+
+
 # TODO: This is only used by vLLM but should move to vllm.py
 def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
     """
@@ -303,6 +321,101 @@ def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
             f"Sending SIGKILL to vLLM server since timeout ({timeout}s) expired"
         )
         process.kill()
+
+    # Various facilities of InstructLab rely on multiple successive start/stop
+    # cycles. Since vLLM relies on stable VRAM for estimating capacity, residual
+    # reclamation activity can lead to crashes on start. To prevent this add a
+    # short delay (typically ~ 10 seconds, max 30) to verify stability.
+    #
+    # Ideally a future enhancement would be contributed to vLLM to more gracefully
+    # handle this condition.
+    wait_for_stable_vram(get_max_stable_vram_wait(30))
+
+
+def wait_for_stable_vram(timeout: int):
+    logger.info("Waiting for GPU VRAM reclamation...")
+    supported, stable = wait_for_stable_vram_cuda(timeout)
+    if not supported:
+        # TODO add support for intel
+        sleep(timeout)
+        return
+    if not stable:
+        # Only for debugging since recovery is likely after additional start delay
+        logger.debug(
+            "GPU VRAM did not stabilize during max timeout (%d seconds)", timeout
+        )
+
+
+def wait_for_stable_vram_cuda(timeout: int) -> Tuple[bool, bool]:
+    if timeout == 0:
+        logger.debug("GPU vram stability check disabled with 0 max wait value")
+        return True, True
+
+    # Third Party
+    import torch
+
+    # Fallback to a constant sleep if we don't have support for the device
+    if not torch.cuda.is_available():
+        return False, False
+    start_time = monotonic()
+    stable_samples = 0
+    last_free = 0
+    try:
+        while True:
+            free_memory = 0
+            try:
+                # TODO In the future this should be enhanced to better handle
+                # GPU partitioning. However, to do so will require that serve
+                # assign specific GPUs to vLLM, so that the same device pool is
+                # analyzed.
+                for i in range(torch.cuda.device_count()):
+                    device = torch.device(f"cuda:{i}")
+                    free_memory += torch.cuda.mem_get_info(device)[0]
+            except Exception:  # pylint: disable=broad-exception-caught
+                logger.warning(
+                    "Could not determine CUDA memory, falling back to general sleep"
+                )
+                return False, False
+
+            # Wait for free memory to stop growing indicating the release of
+            # vram after vLLM shutdown is complete. Wait for 5 successful
+            # samples where this is true, but ignore any spurious readings that
+            # occur between those samples. In the future we may be able to
+            # optimize this by checking a few strictly successive samples.
+            if free_memory <= last_free:
+                stable_samples += 1
+                logger.debug(
+                    "GPU free vram stable (stable count %d, free %d, last free %d)",
+                    stable_samples,
+                    free_memory,
+                    last_free,
+                )
+                if stable_samples > 5:
+                    logger.debug(
+                        "Successful sample recorded, (stable count %d, free %d, last free %d)",
+                        stable_samples,
+                        free_memory,
+                        last_free,
+                    )
+                    return True, True
+            else:
+                if last_free != 0:
+                    logger.debug(
+                        "GPU free vram still growing (free %d, last free %d)",
+                        free_memory,
+                        last_free,
+                    )
+
+            if monotonic() - start_time > timeout:
+                return True, False
+
+            last_free = free_memory
+            sleep(1)
+    finally:
+        try:
+            torch.cuda.empty_cache()
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.debug("Could not free cuda cache: %s", e)
 
 
 def ensure_server(
--- a/tests/test_lab_serve.py
+++ b/tests/test_lab_serve.py
@@ -1,6 +1,7 @@
 # Standard
 from unittest import mock
 from unittest.mock import MagicMock, patch
+import os
 import pathlib
 
 # Third Party
@@ -11,6 +12,7 @@ import yaml
 
 # First Party
 from instructlab import configuration, lab
+from instructlab.model.backends.backends import get_max_stable_vram_wait
 from instructlab.model.serve import warn_for_unsuported_backend_param
 
 CFG_FILE_NAME = "test-serve-config.yaml"
@@ -24,7 +26,9 @@ def vllm_setup_test(mock_popen, mock_determine, runner, args):
     mock_process.returncode = 0
     mock_determine.return_value = ("vllm", "testing")
 
-    runner.invoke(lab.ilab, args)
+    result = runner.invoke(lab.ilab, args)
+    if result.exit_code != 0:
+        print(result.output)
 
     assert len(mock_popen.call_args_list) == 1
     return mock_popen.call_args_list[0][1]["args"]
@@ -392,6 +396,15 @@ def test_vllm_args_config_with_ctx_tps(
         ],
     )
     assert_tps(args, "4")
+
+
+def test_max_stable_vram_wait():
+    with mock.patch.dict(os.environ, {"ILAB_MAX_STABLE_VRAM_WAIT": "0"}):
+        wait = get_max_stable_vram_wait(10)
+        assert wait == 0
+    with mock.patch.dict(os.environ, clear=True):
+        wait = get_max_stable_vram_wait(10)
+        assert wait == 10
 
 
 @pytest.mark.parametrize(
--- a/src/instructlab/model/backends/vllm.py
+++ b/src/instructlab/model/backends/vllm.py
@@ -125,6 +125,7 @@ class Server(BackendServer):
         # Needed when a temporary server is started
         if self.process is not None:
             shutdown_process(self.process, 20)
+            self.process = None
 
     def get_backend_type(self):
         return VLLM
--- a/tox.ini
+++ b/tox.ini
@@ -15,6 +15,7 @@ passenv =
 setenv =
     PIP_EXTRA_INDEX_URL=https://download.pytorch.org/whl/cpu
     CMAKE_ARGS={env:CMAKE_ARGS:-DLLAMA_NATIVE=off}
+    ILAB_MAX_STABLE_VRAM_WAIT=0
 package = wheel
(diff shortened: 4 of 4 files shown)


src/instructlab/model/backends/backends.py
```diff
@@ -304,6 +305,75 @@ def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
...
+                    "Could not determine CUDA memory, falling back to general sleep"
+                )
+                return False, False
+
+            # Wait for recovery of free memory to stop for 5 successive samples ~ 10s.
+            # (6 passes counting init) Additional allocations are ignored since they
+            # may be spurious.
+            if last_free - free_memory >= 0:
+                stable_samples += 1
+                if stable_samples > 5:
```
danmcp, line 363: When everything is working properly (meaning the memory is freed up in < 6 cycles), how often does this return True, True vand return True, False because it didn't lower a 5th time?

n1hility:
Looks like the failure in the A10 runner was against llama.cpp the vllm portion passed

src/instructlab/model/backends/backends.py
n1hility: Not often, I only ever saw it occur artificially when i reduced the timeout value for the method. One case where it would occur is if someone has volatile activity on other GPUs outside of instructlab, then we could end up waiting for ever so thats where the max timeout comes in.

danmcp: After some quick testing, it seems to consistently make it to the 5th iteration and it's still freeing memory.  So my follow-up question is how is 5 enough to know it's stable?  Does the % not matter at all?

danmcp: I am also surprised unstable is defined as it increased in memory usage.  Is that happening in practice?

src/instructlab/model/backends/backends.py
```diff
@@ -304,6 +305,75 @@ def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
...
+                logger.warning(
+                    "Could not determine CUDA memory, falling back to general sleep"
+                )
+                return False, False
+
+            # Wait for recovery of free memory to stop for 5 successive samples ~ 10s.
+            # (6 passes counting init) Additional allocations are ignored since they
+            # may be spurious.
+            if last_free - free_memory >= 0:
+                stable_samples += 1
```
danmcp, line 362: Adding a logger.debug statement in here might be helpful going forward

danmcp: And probably a corresponding log.debug in an else

n1hility: yeah good point i had that developing not sure what i was thinking taking it out

src/instructlab/model/backends/backends.py
danmcp: In practice I am wondering if last_free == free_memory for 1 or 2 cycles implies it's done and stable?  I don't have a good examples where it's taking a while to free the mem.

src/instructlab/model/backends/backends.py
```diff
@@ -39,6 +39,7 @@
 API_ROOT_WELCOME_MESSAGE = "Hello from InstructLab! Visit us at https://instructlab.ai"
 CHAT_TEMPLATE_AUTO = "auto"
 CHAT_TEMPLATE_TOKENIZER = "tokenizer"
+MAX_STABLE_VRAM_WAIT = 30
```
alimaredia, line 42: Are we sure 30 is enough here? Would it be worth the extra work to try to make this configurable? Both eval and phased training are using `launch_server()` in evaluate.py so we'd just need to pass the value from config into that.

n1hility: good question, i was on the fence on that one. There might be a scenario we would want to change it, but i haven't seen it yet. The timing is a fairly small window since it's a few seconds after process exit colliding with the very early start phase of vllm.  My thinking is we could add that as a follow-up if needed, and get the simple constant version in for now to get in our images and e2e testing.

danmcp: Are you seeing any cases that don't succeed in the 5 stable check loop below?

danmcp: In my other comments on this PR, I was wondering whether we could remove the magic number 5 with a check of free memory is not changing for a couple iterations.  I don't see a good reason to make the 30 secs configurable unless we see something useful happening >10-15 secs or we get a lot cases that don't exit out with the 5 stable check loop.

src/instructlab/model/backends/backends.py
n1hility: ah so the purpose of the 5 counter was only to tolerate some spurious noise. Strict successive would likely be fine, although i did observe cases like this (made up numbers):

free memory - 100
free memory - 10
free memory - 10
free memory + 5
free memory - 10
free memory - 10

n1hility: > I am also surprised unstable is defined as it increased in memory usage. Is that happening in practice?

So to be clear unstable is defined as "actively decreasing usage of GPU memory"

More specifically vLLM monitors for drops in usage during its startup. It tolerates increases because it is itself allocating.  Specifically they do

1. Initial early measurement
2. <lots of stuff in between, pool measurement>
3.  then eventually: peak_memory = self.init_gpu_memory - free_gpu_memory
      assert peak_memory > 0 (vLLM has started with more memory than it has now)

danmcp (commented):
I am a +1 with the addition of log debug statements and a comment explaining the free memory check.  Some future TODOs:

- Pass an option to not wait and use from any case that doesn't need to restart vllm right away
- Test more options to remove the magic numbers.  Ex: Exit on free memory == previous free memory
- Consider env vars to reduce change the values of the magic numbers (useful for testing)
- Move magic numbers to config if prove to be the long term solution

src/instructlab/model/backends/backends.py
```diff
@@ -304,6 +309,100 @@ def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
...
+    # cycles. Since vLLM relies on stable VRAM for estimating capacity, residual
+    # reclamation activity can lead to crashes on start. To prevent this add a
+    # short delay (typically ~ 10 seconds, max 30) to verify stability.
+    #
+    # Ideally a future enhancement would be contributed to vLLM to more gracefully
+    # handle this condition.
+    wait_for_stable_vram(MAX_STABLE_VRAM_WAIT)
+
+
+def wait_for_stable_vram(timeout: int):
```
booxter, line 322: is it `int` though?.. Where do you convert it to `int`? It's a string, it seems.

src/instructlab/model/backends/backends.py
```diff
@@ -304,6 +309,100 @@ def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
         )
         process.kill()
 
+    # Various facilities of InstructLab rely on multiple successive start/stop
+    # cycles. Since vLLM relies on stable VRAM for estimating capacity, residual
```
booxter, line 313: This seems reversed. If we need "stable VRAM" on vLLM startup, then shouldn't the wait happen when it starts? If this is the last vLLM run, the wait below will be for nothing, no?

src/instructlab/model/backends/backends.py
```diff
@@ -304,6 +309,100 @@ def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
...
+                    )
+                    return True, True
+            else:
+                logger.debug(
+                    "GPU free vram still growing (free %d, last free %d)",
+                    free_memory,
+                    last_free,
+                )
+
+            if time() - start_time > timeout:
```
booxter, line 395: I think this is unsafe if clock changes. Use `monotonic`?

src/instructlab/model/backends/backends.py
```diff
@@ -304,6 +309,100 @@ def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
...
+    last_free = 0
+    try:
+        while True:
+            free_memory = 0
+            try:
+                # TODO In the future this should be enhanced to better handle
+                # GPU partitioning. However, to do so will require that serve
+                # assign specific GPUs to vLLM, so that the same device pool is
+                # analyzed.
+                for j in range(torch.cuda.device_count()):
```
booxter, line 358: (jk) `j`! I immediately started looking for the other `for` that I have missed. Nada.

booxter (changes requested):

src/instructlab/model/backends/backends.py
```diff
@@ -304,6 +309,100 @@ def shutdown_process(process: subprocess.Popen, timeout: int) -> None:
...
+                if stable_samples > 5:
+                    logger.debug(
+                        "Successful sample recorded, (stable count %d, free %d, last free %d)",
+                        stable_samples,
+                        free_memory,
+                        last_free,
+                    )
+                    return True, True
+            else:
+                logger.debug(
```
danmcp, line 389: Can we add a if last_free != 0: here?

danmcp (changes requested):

src/instructlab/model/backends/backends.py
danmcp: The while loop is the symbolic i :)

src/instructlab/model/backends/backends.py
n1hility: Sure we could do it either way, we have a similar problem were on start the waiting may not be necessary. The best long term fix would be to do a patch to vlllm to alter its mechanism. The only reason I leaned toward doing this on stop was just that the start sequence is incredibly long as it is.

src/instructlab/model/backends/backends.py
n1hility: LOL sorry good point

src/instructlab/model/backends/backends.py
danmcp: I would vote for not changing it given the amount of testing that has gone into this approach.  In either case we would probably want to start passing flags indicating whether the check was necessary.  That should eliminate most of the concern.  And hopefully we can get a fix in vllm to avoid this in the long run.

booxter: I see how this is catch-22, one way or another it slows things down. I don't insist on this to change.

booxter: @n1hility is this reported somewhere in vllm? Would be nice to add a link to the report in comments so that readers can follow up if / when this is fixed.

src/instructlab/model/backends/backends.py
```diff
@@ -39,6 +40,10 @@
 API_ROOT_WELCOME_MESSAGE = "Hello from InstructLab! Visit us at https://instructlab.ai"
 CHAT_TEMPLATE_AUTO = "auto"
 CHAT_TEMPLATE_TOKENIZER = "tokenizer"
+
+# Internal env variable for CI adjustment / disablement
+MAX_STABLE_VRAM_WAIT = int(os.getenv("ILAB_MAX_STABLE_VRAM_WAIT", "30"))
```
booxter, line 45: this will raise on import. I think we will have to handle `ValueError`s when an invalid value is passed. I'd suggest to move the type coercion from import time to runtime, perhaps introduce a separate helper function called `get_vram_stable_wait_timeout()` and hide the error handling there.

src/instructlab/model/backends/backends.py
n1hility: sure will do

src/instructlab/model/backends/backends.py
```diff
@@ -39,6 +40,13 @@
 API_ROOT_WELCOME_MESSAGE = "Hello from InstructLab! Visit us at https://instructlab.ai"
 CHAT_TEMPLATE_AUTO = "auto"
 CHAT_TEMPLATE_TOKENIZER = "tokenizer"
+
+# Internal env variable for CI adjustment / disablement
+# pylint: disable=unnecessary-direct-lambda-call
+MAX_STABLE_VRAM_WAIT = (lambda value: int(value) if value and value.isdigit() else 30)(
```
booxter, line 46:
```
>>> "𐩃".isdigit()
True
>>> int("𐩃")
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '𐩃'
```

I think it's safer to just:

```
try:
    MAX_STABLE_VRAM_WAIT = int(value)
except ValueError: # perhaps some more types?
    MAX_STABLE_VRAM_WAIT = 30
```

And yes, I realize that this is a very weird case that probably no one would even consider.

booxter (approved):
This is good. But the envvar conversion is not 100% correct. (Though no one probably cares.)

src/instructlab/model/backends/backends.py
booxter: Or use `isdecimal` maybe.

n1hility: @booxter great catch i was thinking decimal and for some reason typed digit 🤦 .it turns out to make this easier to test I need to just make this a function

src/instructlab/model/backends/backends.py
```diff
@@ -279,6 +281,22 @@ def get(model_path: pathlib.Path, backend: str | None) -> str:
     return backend
 
 
+def get_max_stable_vram_wait(timeout: int) -> int:
+    # Internal env variable for CI adjustment / disablement
+    env_name = "ILAB_MAX_STABLE_VRAM_WAIT"
```
cdoern, line 286: this should probably be a constant as the other envs are, but this is a nit

cdoern (approved):
one nit, but overall logic looks good. We can follow up to make the env a constant

cdoern:
@danmcp WDYT about the current state of this?

danmcp:
> @danmcp WDYT about the current state of this?

I added a simple unit test and tried out the latest.  Without setting the env var in the test cases, each test would take 30s waiting for the timeout.  If everyone is good with those changes, I am a +1 on the PR as a whole.

tests/test_lab_serve.py
```diff
@@ -11,10 +12,13 @@
 
 # First Party
 from instructlab import configuration, lab
+from instructlab.model.backends.backends import get_max_stable_vram_wait
 from instructlab.model.serve import warn_for_unsuported_backend_param
 
 CFG_FILE_NAME = "test-serve-config.yaml"
 
+os.environ["ILAB_MAX_STABLE_VRAM_WAIT"] = "0"
```
booxter, line 20: Is it needed just for these test cases? I think you should isolate the effect then, by mocking it and reverting back after tests. Otherwise you potentially pollute environment for other test cases that run in the same process.

```
@mock.patch.dict(os.environ, {"ILAB_MAX_STABLE_VRAM_WAIT": "0"})
def test_max_stable_vram_wait():
    ...
```

Thoughts?
```

**Comment by booxter** [comment]
The patch is fine. The test is good. It can be improved (and I think it should be), but this can be a follow-up. You decide.

File: tests/test_lab_serve.py (hunk 12 above)
Comment (danmcp) [reply tests/test_lab_serve.py:20]: We would want it for any unit test.  I was tempted to add it to init but figured that wouldn't be loved either.  The mock better, but I set it more globally to avoid anyone introducing 30s delays and not realizing it.

Comment (booxter) [reply tests/test_lab_serve.py:20]: Ah I think I see why you need to set it to 0, sorry missed your comment. This is to speed up tests. It makes sense but why not just pass it to pytest from tox env? There's `setenv` directive for this. The location of this magic setting is a bit hidden (why this and not other test module?)

File: tests/test_lab_serve.py [hunk 13]
```diff
@@ -394,6 +400,15 @@ def test_vllm_args_config_with_ctx_tps(
     assert_tps(args, "4")
 
 
+def test_max_stable_vram_wait():
+    del os.environ["ILAB_MAX_STABLE_VRAM_WAIT"]
+    wait = get_max_stable_vram_wait(10)
+    assert wait == 10
```
booxter, line 406: So if this fails, you leave default 30 seconds for the rest of tests. This is wrong, no? That's why you should really mock patch it.

danmcp: Switched to patch.  If it did fail, we would obviously notice, but patch is a much better solution.  Thanks!

tests/test_lab_serve.py
danmcp: Moved to setenv.  Thanks for the suggestion!

booxter (approved):
This is solid. Thanks for updating the test case.

danmcp:
> This is solid. Thanks for updating the test case.

Thanks for the review!

danmcp (approved):

PR #1990, opened 2024-08-06
fix: replace default endpoint with config value reading

RobotSail:
Currently, when launching ilab model chat, we check to see if the user is serving
on the endpoint url before proceeding. The issue is that the endpoint url is set
to a default value equivalent to http://localhost:8000 - which also happens to
be the default value that we set in the serve config. Instead of checking for
this default, we should instead read the host port from the users serve config
and check that to see if anything is being served. This will provide cleaner
logic - as we can assume that the user just wants us to connect directly
when endpoint url is provided and skip the precheck which is currently used
to determine whether or not we should start a server ourselves

Resolves #1882 (partially) 

**Checklist:**

### DIFF
--- a/src/instructlab/model/chat.py
+++ b/src/instructlab/model/chat.py
@@ -132,7 +132,6 @@ def is_openai_server_and_serving_model(
     "--endpoint-url",
     type=click.STRING,
     help="Custom URL endpoint for OpenAI-compatible API. Defaults to the `ilab model serve` endpoint.",
-    default=DEFAULT_ENDPOINT,
 )
 @click.option(
     "--api-key",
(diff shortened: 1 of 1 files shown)


src/instructlab/model/chat.py
```diff
@@ -181,33 +180,36 @@ def chat(
...
+    tls_client_key,
+    tls_client_passwd,
     model_family,
 ):
     """Runs a chat using the modified model"""
     # pylint: disable=import-outside-toplevel
     # First Party
     from instructlab.model.backends.llama_cpp import is_temp_server_running
 
+    users_endpoint_url = cfg.get_api_base(ctx.obj.config.serve.host_port)
```
booxter, line 195: this is just `users_endpoint_url = ctx.obj.config.serve.api_base()`

booxter (approved):
This is saner.

cdoern (approved):
lgtm, thank you for the clear Pr description

PR #1992, opened 2024-08-06
Check if a model server is already running when doing ilab data generate

RobotSail:
Resolves #1882

Today in ilab model serve, it is required to provide an endpoint url in order to override the
default behavior of ilab launching a separate model server to generate data. The other
alternative is to specify the model we are generating from, which is redundant since
the user had to have already done this when they initially served the model.

Therefore; this PR uses the users host port to check if a server at that address
is running, OpenAI-compliant, and actively serving a model. If all of these checks
succeed, we default to the user-provided endpoint URL when no other endpoint URL
is passed.

**Checklist:**

### DIFF
--- a/src/instructlab/data/generate.py
+++ b/src/instructlab/data/generate.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+from pathlib import Path
 import copy
 import logging
 import os.path
@@ -10,7 +11,9 @@ import click
 
 # First Party
 from instructlab import clickext
+from instructlab.client import ClientException, list_models
 from instructlab.configuration import DEFAULTS
+from instructlab.utils import HttpClientParams, is_valid_path
 
 # Local
 from ..utils import http_client
@@ -18,7 +21,48 @@ from ..utils import http_client
 logger = logging.getLogger(__name__)
 
 
+def check_if_user_wants_to_use_model_at_server(
+    url: str, api_key: str, http_params: HttpClientParams
+) -> bool:
+    """
+    Returns a boolean indicating whether or not the given URL is
+    serving a teacher model and if the user wants to use it.
+    """
+    try:
+        res = list_models(
+            url,
+            api_key=api_key,
+            http_client=http_client(http_params),
+        )
+    except ClientException:
+        # do nothing
+        return False
+
+    if not res.data:
+        return False
+
+    discovered_model = res.data[0].id
+    presented_model = discovered_model
+    if is_valid_path(discovered_model):
+        # when the model is a long path, e.g. /home/user/.cache/instructlab/models/mistralai/Mixtral-8x7B_v0.1
+        # we truncate it for user-friendly presentation
+        discovered_model_as_path = Path(discovered_model)
+        presented_model = discovered_model_as_path.name
+
+    return click.confirm(
+        f"We've discovered the model {presented_model} being served at {url}\nWould you like to use this as the teacher model?",
+        default=True,
+        show_default=True,
+    )
+
+
 @click.command()
+@click.option(
+    "--interactive/--non-interactive",
+    default=True,
+    show_default=True,
+    help="Be prompted to confirm model selection when `ilab` detects one is already running.",
+)
 @click.option(
     "--model",
     "model_path",
@@ -168,6 +212,7 @@ logger = logging.getLogger(__name__)
 @clickext.display_params
 def generate(
     ctx,
+    interactive: bool,
     model_path,
     num_cpus,
     num_instructions,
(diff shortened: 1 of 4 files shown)


src/instructlab/data/generate.py
```diff
@@ -215,8 +216,22 @@ def generate(
         batch_size = 8
 
     backend_instance = None
+    user_endpoint_url = get_api_base(ctx.obj.config.serve.host_port)
```
alimaredia, line 219: This should be using the `host_port` here from `ctx.obj.config.generate.teacher.host_port`. See #1777 for why the `teacher` section was added to the config.

RobotSail: @alimaredia Can we have it checking for both? When people are manually serving the teacher model, it will be done via `ilab model serve` and so it'll be using those defaults rather than the ones in `generate.teacher.host_port`.

src/instructlab/data/generate.py
```diff
@@ -215,8 +216,33 @@ def generate(
         batch_size = 8
 
     backend_instance = None
+    # check the user's host ports to see if anything is actively running there,
+    # priority is given by the ordering of the hostports
+    user_host_ports = [
```
booxter, line 234: there's already `Model.api_base()` that will give you the relevant `api_base` string, no need to extract `host_port` and pass it to `get_api_base` explicitly.

src/instructlab/data/generate.py
```diff
@@ -215,8 +216,33 @@ def generate(
...
+        ctx.obj.config.generate.teacher.host_port,
+        ctx.obj.config.serve.host_port,
+    ]
+    active_url: str | None = None
```
booxter, line 238: nit: you could tighten types a bit if you'd use `""` as the default value (then it could be just `str`).

booxter (commented):
I am not sure this is a good idea to silently switch to using the `serve` model when the teacher endpoint is not up (perhaps temporarily.) I'd expect a failure in this scenario. The `serve` model may not be what the user actually wants, so they should deal with the error. (by reconfiguring `ilab` or fixing their server.)

RobotSail:
@booxter Please see my comment here regarding the teacher model:
https://github.com/instructlab/instructlab/pull/1992#discussion_r1706939593

The `teacher.host_port` is not a value we'd want to necessarily check in the first place. Consider the context in which it would be used:

1. If a user uses `model serve` to launch the teacher model, then the teacher model would by default be on `model serve` - the user should then expect this to be the model that we connect to by default.
2. When the `teacher.host_port` setting _IS_ used, we don't have a teacher model started and would be launching it when `ilab data generate` itself is called. Therefore, this value isn't relevant here. 

The only reason I've updated this to include the value is in *very* rare case when it's set to an external server and we want to check it before launching our own model.

src/instructlab/data/generate.py
RobotSail: The best practice in this case would be keeping it as `str | None` since it's clear when the value is evaluating to False.

src/instructlab/data/generate.py
```diff
@@ -228,8 +229,33 @@ def generate(
         batch_size = 8
 
     backend_instance = None
+    # check the user's host ports to see if anything is actively running there,
+    # priority is given by the ordering of the hostports
+    user_host_ports = [
+        ctx.obj.config.generate.teacher.host_port,
+        ctx.obj.config.serve.host_port,
```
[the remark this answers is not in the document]
RobotSail: What if the teacher model is actively being served via `ilab model serve`?

src/instructlab/data/generate.py
```diff
@@ -228,8 +229,33 @@ def generate(
...
+            },
+        ):
+            active_url = api_base
+
     if endpoint_url:
         api_base = endpoint_url
+    elif active_url:
+        # if one of the user's host ports is actively serving, we prefer it to
+        # spinning up our own model
+        api_base = active_url
```
cdoern, line 258: maybe a log here?

cdoern (commented):
looks good, but I agree with @booxter on his comments

src/instructlab/data/generate.py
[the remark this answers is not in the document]
cdoern: I think we should proceed but debug log, is that fair?

src/instructlab/data/generate.py
cdoern: +1 with oleg here

src/instructlab/data/generate.py
cdoern: @RobotSail thoughts on this?

src/instructlab/data/generate.py
RobotSail: Yep Ill add it

src/instructlab/data/generate.py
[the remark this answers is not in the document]
RobotSail: > I think we should proceed but debug log, is that fair?

Yep that will make it clear to the users what's happening.

src/instructlab/data/generate.py
RobotSail: That makes sense to me.

src/instructlab/data/generate.py
```diff
@@ -215,8 +216,33 @@ def generate(
...
+            api_base,
+            api_key=api_key,
+            http_params={
+                "tls_insecure": tls_insecure,
+                "tls_client_cert": tls_client_cert,
+                "tls_client_passwd": tls_client_passwd,
+                "tls_client_key": tls_client_key,
+            },
+        ):
+            active_url = api_base
```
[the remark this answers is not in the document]
RobotSail: Done.

src/instructlab/data/generate.py
```diff
@@ -228,8 +229,33 @@ def generate(
...
+
     if endpoint_url:
         api_base = endpoint_url
+    elif active_url:
```
JamesKunstle, line 255: Is it possible that we'd be serving the wrong model? i.e. we want to be serving Mixtral but we're serving Mistral on that port. We might want to identify _which_ model is being served and ask the user if they want to use that, or if that server should be shut down and a new one booted.

RobotSail: This is true and could actually become a common mistake. I'll introduce some logic here to verify that this is the indended model to use. If not, we'll just continue on with the rest of the function. 

The only concern I have here is that the model we prompt will be in the format of `/very/long/path/to/.local/share/instructlab/models/mixtral` - should this be presented as-is, or should we trim out the base path if we detect that the model is in this format?

JamesKunstle: I mean, could totally do:
```
click.secho(f"Name of model currently served: {active_url.split("/")[-1]}")
click.secho(f"This model's path is: {active_url}")
click.confirmation("If you would like to keep using this server, please confirm. Otherwise, please shut down this server and try again.", abort=True)
```

RobotSail: Discussed offline, but the other consideration that we need to make here is when `--non-interactive` is passed, we will skip finding the server automatically and provide a warning in red about the fact we skipped this check. We will include a recommendation here that the user needs to be explicit about the teacher model.

ktam3:
@RobotSail - do you have an update on this?

RobotSail:
@ktam3 We'll need to rework this PR as it was not chosen to go into 0.18.0

cdoern:
@RobotSail is this still relevant?

RobotSail:
@cdoern Doubt it, closing

PR #1999, opened 2024-08-06
`ilab` global config fixes

cdoern:
`ILAB_GLOBAL_CONFIG` and `ILAB_TRAIN_PROFILE_DIR` had two unexpected behaviors:

1. `ilab config init` overwrites the cfg without warning the user when the env is set
2. get_params_from_env would overwrite all model paths with the serve model which it should not
3. `ilab config init` did NOT overwrite the train profiles, prompt the user that this is going to happen and add the functionality to overwrite

**Checklist:**

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -154,7 +154,7 @@ ilab --version
 ilab system info
 
 # pipe 3 carriage returns to ilab config init to get past the prompts
-echo -e "\n\n\n" | ilab init
+echo -e "y\n\n\n\n" | ilab init
 
 # Enable Debug in func tests with debug level 1
 sed -i.bak -e 's/log_level:.*/log_level: DEBUG/g;' "${ILAB_CONFIG_FILE}"
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -92,7 +92,7 @@ def init(
         param_source = ctx.parent.parent.get_parameter_source("config_file")
 
     if param_source == click.core.ParameterSource.ENVIRONMENT:
-        model_path, taxonomy_path, taxonomy_base, cfg = get_params_from_env(ctx.obj)
+        taxonomy_path, taxonomy_base, cfg = get_params_from_env(ctx.obj)
     else:
         try:
             model_path, taxonomy_path, cfg = get_params_default(
@@ -133,12 +133,14 @@ def init(
             )
             raise click.exceptions.Exit(1)
 
-    cfg.chat.model = model_path
-    cfg.generate.model = model_path
-    cfg.serve.model_path = model_path
+    # we should not override all paths with the serve model if special ENV vars exist
+    if param_source != click.core.ParameterSource.ENVIRONMENT:
+        cfg.chat.model = model_path
+        cfg.generate.model = model_path
+        cfg.serve.model_path = model_path
+        cfg.evaluate.model = model_path
     cfg.generate.taxonomy_path = taxonomy_path
     cfg.generate.taxonomy_base = taxonomy_base
-    cfg.evaluate.model = model_path
     cfg.evaluate.mt_bench_branch.taxonomy_path = taxonomy_path
     write_config(cfg)
 
@@ -150,12 +152,11 @@ def init(
 
 def get_params_from_env(
     obj: typing.Optional[typing.Any],
-) -> typing.Tuple[str, str, str, Config]:
+) -> typing.Tuple[str, str, Config]:
     if obj is None or not hasattr(obj, "config"):
         raise ValueError("obj must not be None and must have a 'config' attribute")
 
     return (
-        obj.config.serve.model_path,
         obj.config.generate.taxonomy_path,
         obj.config.generate.taxonomy_base,
         obj.config,
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -18,6 +18,7 @@ from instructlab.configuration import (
     ensure_storage_directories_exist,
     get_default_config,
     read_train_profile,
+    recreate_train_profiles,
     write_config,
 )
 
@@ -90,7 +91,10 @@ def init(
     # the parameter source from the parent of the parent command.
     else:
         param_source = ctx.parent.parent.get_parameter_source("config_file")
-
+    try:
+        overwrite = check_if_configs_exist()
+    except click.exceptions.Exit as e:
+        ctx.exit(e.exit_code)
     if param_source == click.core.ParameterSource.ENVIRONMENT:
         taxonomy_path, taxonomy_base, cfg = get_params_from_env(ctx.obj)
     else:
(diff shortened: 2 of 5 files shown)


src/instructlab/config/init.py
```diff
@@ -132,12 +132,14 @@ def init(
...
 
-    cfg.chat.model = model_path
-    cfg.generate.model = model_path
-    cfg.serve.model_path = model_path
+    # we should not override all paths with the serve model if special ENV vars exist
+    if param_source != click.core.ParameterSource.ENVIRONMENT:
+        cfg.chat.model = model_path
+        cfg.generate.model = model_path
+        cfg.serve.model_path = model_path
+        cfg.evaluate.model = model_path
```
jaideepr97, line 140: I feel like this all needs to be revisited after GA - seems like a band aid solution

jaideepr97 (commented):
@cdoern don't we need to add the overwrite warning for the train profiles as well? or is that in a different PR

src/instructlab/config/init.py
cdoern: yes, I agree

cdoern:
@jaideepr97 I think we decided we are not overriding train profiles once they exist, right?

jaideepr97:
>@jaideepr97 I think we decided we are not overriding train profiles once they exist, right?

Did we? I'm personally of the opinion that we should be consistent between the train profiles and the config file itself. I don't know why we would want to treat them differently since the train profiles are just a subsection of the config eventually anyway 

but no need to block on this either way

src/instructlab/config/init.py
booxter: Indeed. I think the culprit is the fact that the CLI argument has an "acceptable" default that overrides regardless of what is in the config. The default value for `model-path` should be made `None`, in which case we should not touch what's pulled from the config (either env or default).

booxter: (Something like [this](https://github.com/instructlab/instructlab/pull/2014).)

scripts/functional-tests.sh
```diff
@@ -151,7 +151,7 @@ ilab --version
 ilab system info
 
 # pipe 3 carriage returns to ilab config init to get past the prompts
-echo -e "\n\n\n" | ilab init
+echo -e "\n\n\n\n" | ilab init
```
booxter: (Observation) could as well just `--non-interactive` and never deal with the problem anymore. (Though maybe we should still have a test case for the interactive mode - but more elaborate than just typing Enters.)

cdoern: I changed it to y\n\n\n\n to account for all cases

src/instructlab/configuration.py
```diff
@@ -804,6 +804,18 @@ def ensure_storage_directories_exist():
         "lora_target_modules": ["q_proj", "k_proj", "v_proj", "o_proj"],
     }
 
+    recreate_train_profiles()
```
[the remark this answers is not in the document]
booxter: I mean, **logically** it probably belongs to the code that (re)creates profile files. But this is a minor point. Feel free to ignore.

src/instructlab/config/init.py
```diff
@@ -92,7 +93,12 @@ def init(
         param_source = ctx.parent.parent.get_parameter_source("config_file")
 
     if param_source == click.core.ParameterSource.ENVIRONMENT:
-        model_path, taxonomy_path, taxonomy_base, cfg = get_params_from_env(ctx.obj)
+        try:
+            taxonomy_path, taxonomy_base, cfg = get_params_from_env(
+                interactive, ctx.obj
+            )
+        except click.exceptions.Exit as e:
+            ctx.exit(e.exit_code)
```
booxter, line 101: I believe this is done by the `click` runner when you let the `Exit` bubble up. I don't think this handler is necessary. But perhaps I am wrong?

src/instructlab/config/init.py
```diff
@@ -147,14 +158,28 @@ def init(
     )
 
 
+def check_if_configs_exist():
+    for entry in (DEFAULTS.CONFIG_FILE, DEFAULTS.TRAIN_PROFILE_DIR):
+        if exists(entry):
+            overwrite = click.confirm(
+                f"Found {DEFAULTS.CONFIG_FILE}, do you still want to continue?"
```
booxter, line 165: this should interpolate `{entry}`

booxter (dismissed):

src/instructlab/config/init.py
```diff
@@ -92,7 +93,12 @@ def init(
         param_source = ctx.parent.parent.get_parameter_source("config_file")
 
     if param_source == click.core.ParameterSource.ENVIRONMENT:
```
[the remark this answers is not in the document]
cdoern: good point! will do that

src/instructlab/config/init.py
cdoern: this was existing behavior so i'd prefer to keep it just for ease of merging this before the e2e

src/instructlab/config/init.py
```diff
@@ -90,9 +91,14 @@ def init(
     # the parameter source from the parent of the parent command.
     else:
         param_source = ctx.parent.parent.get_parameter_source("config_file")
-
+    try:
+        overwrite = False
+        if interactive:
+            overwrite = check_if_configs_exist()
+    except click.exceptions.Exit as e:
+        ctx.exit(e.exit_code)
```
RobotSail, line 99: `ctx.exit` prints the usage here when we exit, is this what we'd want in this case?

cdoern: same behavior as below

cdoern:
(fixed failing test that was just changed)

nathan-weinberg (approved):

src/instructlab/config/init.py
RobotSail: Let's keep it as-is then to be consistent.

RobotSail (approved):
LGTM

alinaryan (approved):

alinaryan (approved):

khaledsulayman (approved):

src/instructlab/config/init.py
```diff
@@ -148,14 +161,26 @@ def init(
...
+def check_if_configs_exist() -> bool:
+    if exists(DEFAULTS.CONFIG_FILE):
+        overwrite = click.confirm(
+            f"Found {DEFAULTS.CONFIG_FILE}, do you still want to continue?"
+        )
+        if not overwrite:
+            raise click.exceptions.Exit(0)
+    if exists(DEFAULTS.TRAIN_PROFILE_DIR):
+        return click.confirm(
+            f"Found {DEFAULTS.TRAIN_PROFILE_DIR}, do you still want to continue?"
```
booxter, line 173: "Continue what"? Yes, I, the user, want to continue to initialize the config. How is the presence of train profile directory relevant here? Should the message be something like "Found <dir>, do you want to re-initialize it?"

src/instructlab/config/init.py
booxter: Sure. I've posted [this](https://github.com/instructlab/instructlab/pull/2017) as a follow-up.

PR #2006, opened 2024-08-07
Expand env vars and user dirs read from config

danmcp:
Expands env vars (Ex: $HOME) and user vars (Ex: ~) for config and training profiles.

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -32,7 +32,7 @@ import platformdirs
 import yaml
 
 # Local
-from . import log
+from . import log, utils
 
 ILAB_PACKAGE_NAME = "instructlab"
 CONFIG_FILENAME = "config.yaml"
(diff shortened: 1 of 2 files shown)


n1hility (approved):
Also tested in image

nathan-weinberg (approved):

RobotSail:
It seems like this PR introduced some circular import issues that I'm now experiencing when adding new code paths into the designated files. For reference, see the failure in this CI runner:

https://github.com/instructlab/instructlab/actions/runs/10285276171/job/28463143941#step:11:57

nathan-weinberg:
How did this not fail on the PR checks here?

bjhargrave:
Perhaps this is an interaction of multiple PRs which landed near the same time?

RobotSail:
@bjhargrave @nathan-weinberg This seems to be an issue only when we move code into utils which imports from ilabclient - since that's what the breaking PR is doing. 

This PR is perfectly fine and doing the right thing, it just seems like this is a very Python-specific issue since we need to interpret all of our paths. 

The better solution to this issue would be to move all of the defaults into a dedicated file, and that way we wouldn't have any of these circular dependency problems.

PR #2007, opened 2024-08-07
hotfix: fix circular dependency issue

RobotSail:
Currently we have an issue with a circular dependency inside the configuration code,
where we cannot include util code that imports from ilabclient in the utils module.
The reason why this happens is ilabclient also imports from configuration. But if we
have code which is technically a util, we cannot move it there for this reason.

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -32,7 +32,7 @@ import platformdirs
 import yaml
 
 # Local
-from . import log, utils
+from . import log
 
 ILAB_PACKAGE_NAME = "instructlab"
 CONFIG_FILENAME = "config.yaml"
@@ -749,7 +749,8 @@ def _expand_paths(content: dict | list):
 
 def _expand_value(value):
     if isinstance(value, str):
-        return utils.expand_path(value)
+        expanded_value = os.path.expanduser(value)
+        return os.path.expandvars(expanded_value)
     if isinstance(value, (dict, list)):
         _expand_paths(value)
     return None


src/instructlab/configuration.py
```diff
@@ -749,7 +749,9 @@ def _expand_paths(content: dict | list):
 
 def _expand_value(value):
     if isinstance(value, str):
-        return utils.expand_path(value)
+        value = os.path.expanduser(value)
+        value = os.path.expandvars(value)
+        return value
```
nathan-weinberg, line 754: Can we use diff vars here? Having them all called `value` is not great

nathan-weinberg (changes requested):

danmcp (approved):

src/instructlab/configuration.py
danmcp: This is the same code that's in utils as well.  Could do it with no vars.

```
return os.path.expandvars(os.path.expanduser(value))
```

RobotSail: Consider it renamed

nathan-weinberg (changes requested):

nathan-weinberg (approved):

cdoern (approved):

RobotSail:
This PR is more-or-less just a temporary fix. The better solution which I've outlined in https://github.com/instructlab/instructlab/issues/2009  will be to move all of the default values to exist in a separate file.

PR #2008, opened 2024-08-07
lora followup to #1917

alinaryan:
This commit fixes the lora mappings to the train options and
sets the 'lora_quantize_dytpe' to a default of None.

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -500,7 +500,7 @@ class _train(BaseModel):
     deepspeed_cpu_offload_optimizer: bool
 
     lora_rank: int | None = None
-    lora_quantize_dtype: str | None
+    lora_quantize_dtype: str | None = None
 
     is_padding_free: bool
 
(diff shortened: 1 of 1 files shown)


src/instructlab/configuration.py
```diff
@@ -1124,9 +1124,10 @@ def map_train_to_library(ctx, params):
     )
 
     lora_args = LoraOptions()
+    lora = False
     if params["lora_rank"] is not None:
-        lora = True
-        lora_args.rank = params["lora_rank"]
+        if not lora:
```
RobotSail, line 1129: You can just keep `lora = True` here because it doesn't really benefit us to be performing the `if not lora` check prior to setting it. 

So all of the blocks should now be set to:

```diff
- if not lora:
-   lora = True
+ lora = True
```

RobotSail (approved):
One minor comment but otherwise LGTM

src/instructlab/configuration.py
alinaryan: Updated all the blocks!

cdoern (approved):

Issue #1976, opened 2024-08-05
Idea: Deprecate and remove `--model-path` from `config init` command


booxter:
**What is the current behavior of feature you are looking to change?**

Right now, `config init` allows to specific `--model-path` to override the default chosen in code. When it's passed, the same model is set for all sub-sections. In most scenarios, this is wrong, and the user will have to choose different models for different tasks.

**What is the proposed behavior of feature you are looking to change?**

We may replace the existing single `--model-path` option with separate options, one per "task". (`--judge-model`, `--teacher-model` etc.) Then deprecate remove the `--model-path` option.

But perhaps what should be done instead is just leave this setting out of CLI options list. Instead, instruct the users to run `config init`, then set the relevant config attributes in their text editor, as needed. (Basically, update documentation to reflect this workflow; maybe update the greetings output of `config init` to give some pointers / guidance on what should be done.)

**What is the benefit of the design change you are proposing?**

One fewer useless "knob" to support and list in the `config init` usage message.

**If applicable, how would backwards-compatability be handled?**

We'll follow the accepted deprecation strategy for the project. (To do: understand if there's such a thing. We may have to write it down, if not already.)

nathan-weinberg:
Excellent use of the design template :pray: thank you :smile:

PR #2014, opened 2024-08-07
Don't override models set in initialized config source

booxter:


jaideepr97:
@booxter IMO we should get rid of `--model-path` for config init altogether
I dont think anyone would find value in blanket overriding all default models with a given model that should be used for all the various purposes
Currently we specify different default models for chat/generate/train because those models are best suited for those purposes

If users want to override them they should do so in the config, or if it must be done via flags there should be individual flags for each model to be overridden 

thoughts? 
cc @cdoern

booxter:
@jaideepr97 Yes. This is tracked in https://github.com/instructlab/instructlab/issues/1976 I will close this PR and we can focus on removal instead whenever we give this a priority.

PR #2035, opened 2024-08-09
fix: phased training fixes

cdoern:
set the defaults in pydantic for effective batch size
set the click flag range for save samples to be > 0 rather than > 1. Also change pydantic defaults
because save samples per phase is off by default, set checkpoint_at_each_epoch to true

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -497,7 +497,7 @@ class _train(BaseModel):
     num_epochs: int
     effective_batch_size: int
     save_samples: int
-    checkpoint_at_epoch: bool = False
+    checkpoint_at_epoch: bool = True
 
     deepspeed_cpu_offload_optimizer: bool
 
@@ -515,12 +515,16 @@ class _train(BaseModel):
     # TODO: could move into its own object.
     # Not strictly necessary for a correct training object.
     phased_phase1_num_epochs: int | None = OPTIONAL_POSITIVE_INTEGER
-    phased_phase1_samples_per_save: int | None = OPTIONAL_POSITIVE_INTEGER
-    phased_phase1_effective_batch_size: int | None = OPTIONAL_POSITIVE_INTEGER
+    # phased_phase1_samples_per_save is disabled when the value is 0.
+    # anything greater than 0 enables samples_per_save for the phase.
+    phased_phase1_samples_per_save: int = Field(ge=0, default=0)
+    phased_phase1_effective_batch_size: int | None = 128
 
     phased_phase2_num_epochs: int | None = OPTIONAL_POSITIVE_INTEGER
-    phased_phase2_samples_per_save: int | None = OPTIONAL_POSITIVE_INTEGER
-    phased_phase2_effective_batch_size: int | None = OPTIONAL_POSITIVE_INTEGER
+    # phased_phase2_samples_per_save is disabled when the value is 0.
+    # anything greater than 0 enables samples_per_save for the phase.
+    phased_phase2_samples_per_save: int = Field(ge=0, default=0)
+    phased_phase2_effective_batch_size: int | None = 3840
 
     phased_mt_bench_judge: str | None = None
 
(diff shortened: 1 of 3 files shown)


src/instructlab/configuration.py
```diff
@@ -514,12 +514,12 @@ class _train(BaseModel):
     # TODO: could move into its own object.
     # Not strictly necessary for a correct training object.
     phased_phase1_num_epochs: int | None = OPTIONAL_POSITIVE_INTEGER
-    phased_phase1_samples_per_save: int | None = OPTIONAL_POSITIVE_INTEGER
-    phased_phase1_effective_batch_size: int | None = OPTIONAL_POSITIVE_INTEGER
+    phased_phase1_samples_per_save: int | None = -1
```
RobotSail, line 517: I don't disagree with the approach, but it's not obvious what the difference between `-1`, `0`, and `None` is in this context. We should add a docstring or comment here explaining what they do.

JamesKunstle: yeah good point

JamesKunstle: -1 and 0 have same effect: turns save_samples saving off.
None: use the default `save_samples` value from general training config
>0, not None: overrides general default

JamesKunstle:
`if samples_per_save:` -> `if samples_per_save is not None:` in _run_training_phase @cdoern

src/instructlab/configuration.py
cdoern: fixed after slack discussion

src/instructlab/model/train.py
```diff
@@ -355,19 +355,19 @@ def clickpath_setup(is_dir: bool) -> click.Path:
...
     cls=clickext.ConfigOption,
-    type=click.IntRange(min=1),
+    type=click.IntRange(min=0),
     help="Number of epochs to run for second phase of end-to-end training.",
 )
 @click.option(
     "--phased-phase2-samples-per-save",
     cls=clickext.ConfigOption,
-    type=click.IntRange(min=1),
+    type=click.IntRange(min=-1),
```
RobotSail, line 364: Should this have a minimum of -1 if the minimum in the corresponding field is 0?

src/instructlab/model/train.py
```diff
@@ -338,7 +338,7 @@ def clickpath_setup(is_dir: bool) -> click.Path:
 @click.option(
     "--phased-phase1-samples-per-save",
     cls=clickext.ConfigOption,
-    type=click.IntRange(min=1),
+    type=click.IntRange(min=-1),
```
RobotSail, line 341: Same here

src/instructlab/model/train.py
```diff
@@ -355,19 +355,19 @@ def clickpath_setup(is_dir: bool) -> click.Path:
 @click.option(
     "--phased-phase2-num-epochs",
     cls=clickext.ConfigOption,
-    type=click.IntRange(min=1),
+    type=click.IntRange(min=0),
```
JamesKunstle, line 358: this needs to be 1

src/instructlab/model/train.py
JamesKunstle, line 341: this is 0

src/instructlab/model/train.py
JamesKunstle, line 364: this is 0

src/instructlab/model/train.py
```diff
@@ -355,19 +355,19 @@ def clickpath_setup(is_dir: bool) -> click.Path:
...
     cls=clickext.ConfigOption,
-    type=click.IntRange(min=1),
+    type=click.IntRange(min=-1),
     help="Number of samples to train on between saves for second phase of end-to-end training.",
 )
 @click.option(
     "--phased-phase2-effective-batch-size",
     cls=clickext.ConfigOption,
-    type=click.IntRange(min=1),
+    type=click.IntRange(min=0),
```
JamesKunstle, line 370: this is 1

src/instructlab/model/train.py
cdoern: my bad, fixed

src/instructlab/model/train.py
cdoern: changed the wrong thing, my bad. fixed

src/instructlab/model/train.py
cdoern: fixed

RobotSail (approved):
LGTM

booxter:
The commit message is vague on the reasoning for these changes. I assume some other team members may have more details to justify it, but these should be shared in the commit message and/or in an issue that is linked to the patch. Without it, it's impossible to assess if these changes are the right thing to do.

The commit message is probably incorrect when it says that "set the click flag range for save samples to be > 0 rather than > 1" (because the ">=" operator should be used I think.)

khaledsulayman (approved):

PR #2040, opened 2024-08-09
Fix Traceback when `--gpus` is used and `vllm_args` is `null` in config file

booxter:
This should fix the following Traceback from data generate and similar commands:

```
INFO 2024-08-09 17:44:41,707 numexpr.utils:161: NumExpr defaulting to 4 threads.
INFO 2024-08-09 17:44:42,097 datasets:58: PyTorch version 2.3.1+cpu available.
Traceback (most recent call last):
  File "/home/vagrant/git/instructlab/.tox/py311-unit/bin/ilab", line 8, in <module>
    sys.exit(ilab())
             ^^^^^^
  File "/home/vagrant/git/instructlab/.tox/py311-unit/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/git/instructlab/.tox/py311-unit/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/vagrant/git/instructlab/.tox/py311-unit/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/git/instructlab/.tox/py311-unit/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/git/instructlab/.tox/py311-unit/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/git/instructlab/.tox/py311-unit/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/git/instructlab/.tox/py311-unit/lib/python3.11/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
[...]

PR #2041, opened 2024-08-09
vllm_args cannot be none

cdoern:
vllm_args must always be an empty list, not NoneType

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -302,7 +302,7 @@ class _serve_vllm(BaseModel):
     max_startup_attempts: int | None = None
     gpus: Optional[int] = None
     # arguments to pass into vllm process
-    vllm_args: list[str] | None = None
+    vllm_args: list[str] | None = []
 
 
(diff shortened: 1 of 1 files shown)


src/instructlab/configuration.py
```diff
@@ -302,7 +302,7 @@ class _serve_vllm(BaseModel):
     max_startup_attempts: int | None = None
     gpus: Optional[int] = None
     # arguments to pass into vllm process
-    vllm_args: list[str] | None = None
+    vllm_args: list[str] | None = []
```
booxter, line 305: this is not fixing the issue - the type can still be null so you can have it set in config file.

booxter:
See here https://github.com/instructlab/instructlab/pull/2040

src/instructlab/configuration.py
booxter: Btw this is also unsafe in python because then every instance of a config will reuse the same list object, so if you'd e.g. add something to it in runtime, other copies would immediately see it.

PR #2043, opened 2024-08-09
honor ~/.config/instructlab/config.yaml

### LINKED ISSUE #2042: `ctx.obj.config` does not honor ~/.config once it exists
due to 

https://github.com/instructlab/instructlab/blob/ac229679366d86f4940b00c487027f24993f501a/src/instructlab/configuration.py#L1085

which will _always_ pull its value from the env var first, the global config supercedes the local config in all scenarios. This makes it so that if users edit the ~/.config version, their changes are not used when running cmds and do not show up in cmds like `ilab config show` and `ilab config edit`.

We need to edit the logic of `--config` since it should only really be applicable for `ilab config init` and nothing else.

cdoern:
currently if $ILAB_GLOBAL_CONFIG is set, it supercedes the ~/.config/instructlab/config.yaml in ALL scenarios. We only want $ILAB_GLOBAL_CONFIG to win when running `ilab config init`.

Therefore the `envvar=` should only be used on the command itself, and ignored everywhere else. This makes it so that `init` (a different function from ilab config init`) sets the ctx.obj.cfg value to the LOCAL file rather than the global one when running other commands

resolves #2042

**Checklist:**

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -17,6 +17,7 @@ from instructlab.configuration import (
     Config,
     ensure_storage_directories_exist,
     get_default_config,
+    read_config,
     read_train_profile,
     recreate_train_profiles,
     write_config,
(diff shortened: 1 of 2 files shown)


RobotSail (approved):
LGTM

cdoern:
removed a leftover comment, made no other changes in most recent push

alinaryan (approved):

nathan-weinberg (approved):

PR #2046, opened 2024-08-09
tests: Validate that null vllm_args are accepted

booxter:
This is a follow-up to #2040.

Note to reviewers: `common.py` is largely (refactored) code from `test_lab_serve.py`

### DIFF
--- a/tests/test_lab_evaluate.py
+++ b/tests/test_lab_evaluate.py
@@ -11,6 +11,9 @@ from git import Repo
 # First Party
 from instructlab import lab
 
+# Local
+from . import common
+
 
 def gen_qa_pairs(odd):
     i = 1
@@ -540,3 +543,24 @@ def test_invalid_model_path_mt_bench(cli_runner: CliRunner, tmp_path):
         in result.output
     )
     assert result.exit_code != 0
+
+
+def test_vllm_args_null(cli_runner: CliRunner):
+    fname = common.setup_gpus_config(section_path="serve", vllm_args=lambda: None)
+    args = common.vllm_setup_test(
+        cli_runner,
+        [
+            f"--config={fname}",
+            "model",
+            "evaluate",
+            "--benchmark",
+            "mt_bench",
+            "--model",
+            "instructlab/granite-7b-lab",
+            "--judge-model",
+            "instructlab/merlinite-7b-lab",
+            "--gpus",
+            "4",
+        ],
+    )
+    common.assert_tps(args, "4")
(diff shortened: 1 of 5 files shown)


booxter:
Note to reviewers: `common.py` is largely (refactored) code from `test_lab_serve.py`

tests/test_lab_evaluate.py
```diff
@@ -540,3 +543,26 @@ def test_invalid_model_path_mt_bench(cli_runner: CliRunner, tmp_path):
         in result.output
     )
     assert result.exit_code != 0
+
+
+def test_vllm_args_None(cli_runner: CliRunner):
+    fname = common.setup_gpus_config(
+        section_path="generate.teacher", vllm_args=lambda: None
```
danmcp, line 550: Why is generate.teacher in evaluate testing?  Typo?

tests/test_lab_evaluate.py
```diff
@@ -540,3 +543,26 @@ def test_invalid_model_path_mt_bench(cli_runner: CliRunner, tmp_path):
         in result.output
     )
     assert result.exit_code != 0
+
+
+def test_vllm_args_None(cli_runner: CliRunner):
```
danmcp, line 548: Is this a common pattern to use case like this?  Hurts my eyes but might just be me.

danmcp (changes requested):

tests/test_lab_evaluate.py
booxter: I was trying to reflect the `None` name, but I can change to `null` which is how it's represented in file.

tests/test_lab_serve.py
```diff
@@ -58,46 +41,19 @@ def assert_template(args, expect_chat, path_chat, chat_value):
...
-
-@patch("time.sleep", side_effect=Exception("Intended Abort"))
-@patch("instructlab.model.backends.backends.determine_backend")
-@patch("subprocess.Popen")
-def test_chat_auto(mock_popen, mock_determine, _, cli_runner: CliRunner):
-    args = vllm_setup_test(
-        mock_popen,
-        mock_determine,
+def test_chat_auto(cli_runner: CliRunner):
+    args = common.vllm_setup_test(
```
danmcp, line 45: Very nice!

tests/test_lab_evaluate.py
booxter: Ooooops. :) Yes it should be `serve`. (New version is pushed; I also confirmed that the test case fails without the original fix as expected.)

danmcp (approved):

PR #2051, opened 2024-08-12
do not re-create train profiles on fresh install

cdoern:
since creation of the train profiles is handled by `ilab config init` as well as
the persistent storage dir creation process, on a new system the train profiles dir
is always detected as existing. This creates a double re-write and prompts the user
asking them if they want to overwrite train profiles they didn't know existed.

Fix this by checking if ANY of the train profiles DNE when being created. If this scenario is
detected, do not prompt the user to recreate the profiles since they have just been created for the first time.

Also, `ilab config init` should be the only place that persistent storage dirs are created. Currently each time ANY ilab command is run, the ~/.local and ~/.config dirs are created if they do not exist. This PR makes it so that only `ilab config init` creates the persistent storage dirs and any other ilab command errors out if `ilab config init` has not been run

some exceptions to this are:

if you specify --config DEFAULT or --config <some_non_default>.yaml
certain commands that dont need the dirs ex:
just running `ilab model` or any top level cmd
`ilab system info`

Some test fixes were needed since in most scenarios we are testing, we no longer prompt the user to override their profiles since they are fresh systems

**Checklist:**

### DIFF
--- a/src/instructlab/clickext.py
+++ b/src/instructlab/clickext.py
@@ -15,6 +15,9 @@ from click.core import ParameterSource
 from click_didyoumean import DYMGroup
 import click
 
+# First Party
+from instructlab.configuration import DEFAULTS, storage_dirs_exist
+
 logger = logging.getLogger(__name__)
 
 
@@ -76,6 +79,17 @@ class ExpandAliasesGroup(LazyEntryPointGroup):
                 "in a future release. Please consider using "
                 f"`ilab {group} {primary}` instead"
             )
+            # if some storage dirs do not exist
+            # AND the command is not `ilab init`
+            # AND the --config flag is not customized, then we error
+            if not storage_dirs_exist() and (
+                primary != "init" and ctx.params["config_file"] == DEFAULTS.CONFIG_FILE
+            ):
+                click.secho(
+                    "Some ilab storage directories do not exist yet. Please run `ilab config init` before continuing.",
+                    fg="red",
+                )
+                raise click.exceptions.Exit(1)
             return cmd
         return super().get_command(ctx, cmd_name)
 
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -93,12 +93,12 @@ def init(
     config_file,
 ):
     """Initializes environment for InstructLab"""
-    ensure_storage_directories_exist()
+    fresh_install = ensure_storage_directories_exist()
     param_source = ctx.get_parameter_source("config_file")
     try:
-        overwrite = False
+        overwrite_profile = False
         if interactive:
-            overwrite = check_if_configs_exist()
+            overwrite_profile = check_if_configs_exist(fresh_install)
     except click.exceptions.Exit as e:
         ctx.exit(e.exit_code)
     else:
@@ -114,7 +114,7 @@ def init(
             )
         except click.exceptions.Exit as e:
             ctx.exit(e.exit_code)
-    if overwrite:
+    if overwrite_profile:
         click.echo(
             f"Generating `{DEFAULTS.CONFIG_FILE}` and `{DEFAULTS.TRAIN_PROFILE_DIR}`..."
         )
@@ -165,18 +165,19 @@ def init(
     )
 
 
-def check_if_configs_exist() -> bool:
+def check_if_configs_exist(fresh_install) -> bool:
     if exists(DEFAULTS.CONFIG_FILE):
         overwrite = click.confirm(
             f"Found {DEFAULTS.CONFIG_FILE}, do you still want to continue?"
         )
         if not overwrite:
             raise click.exceptions.Exit(0)
-    if exists(DEFAULTS.TRAIN_PROFILE_DIR):
+    if exists(DEFAULTS.TRAIN_PROFILE_DIR) and not fresh_install:
         return click.confirm(
             f"Found {DEFAULTS.TRAIN_PROFILE_DIR}, do you also want to reset existing profiles?"
         )
-    return True
+    # default behavior should be do NOT overwrite files that could have just been created
+    return False
 
 
 def get_params_from_env(
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -90,6 +90,20 @@ class _InstructlabDefaults:
 
     # When otherwise unknown, ilab uses this as the default family
     MODEL_FAMILY = "merlinite"
+    ADDITIONAL_ARGS_DEFAULTS = {
+        "learning_rate": 2e-5,
+        "warmup_steps": 25,
+        "random_seed": 42,
+        "node_rank": 0,
+        "nnodes": 1,
+        "rdzv_id": 123,
+        "rdzv_endpoint": "127.0.0.1:12222",
+        "deepspeed_cpu_offload_optimizer_ratio": 1,
+        "deepspeed_cpu_offload_optimizer_pin_memory": False,
+        "lora_alpha": 32,
+        "lora_dropout": 0.1,
+        "lora_target_modules": ["q_proj", "k_proj", "v_proj", "o_proj"],
+    }
 
     def __init__(self):
         self._reset()
@@ -796,7 +810,7 @@ def get_model_family(family, model_path):
     return guess if guess in MODEL_FAMILIES else DEFAULTS.MODEL_FAMILY
 
 
-def ensure_storage_directories_exist():
+def ensure_storage_directories_exist() -> bool:
     """
     Ensures that the default directories used by ilab exist.
     """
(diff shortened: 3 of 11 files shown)


nathan-weinberg:
Please run checks locally to reduce CI time, TIA

src/instructlab/configuration.py
```diff
@@ -1167,13 +1182,54 @@ def map_train_to_library(ctx, params):
...
+        DEFAULTS.MODELS_DIR,
+        DEFAULTS.TAXONOMY_DIR,
+        DEFAULTS.TRAIN_CONFIG_DIR,
+        DEFAULTS.TRAIN_PROFILE_DIR,
+        DEFAULTS.TRAIN_ADDITIONAL_OPTIONS_DIR,
+        DEFAULTS.PHASED_DIR,
+    ]
+    return all(os.path.exists(dirpath) for dirpath in dirs_to_make)
+
+    # return all(os.path.exists(dirpath) for dirpath in dirs_to_make)
```
jaideepr97, line 1235: probably wanna get rid of this

src/instructlab/configuration.py
```diff
@@ -1167,13 +1182,54 @@ def map_train_to_library(ctx, params):
...
+        }
+    _expand_paths(additional_args_and_defaults)
+    for key, val in additional_args_and_defaults.items():
         if key not in current_additional:
             current_additional[key] = val
 
     return current_additional
+
+
+def check_if_storage_dirs_exist() -> bool:
```
jaideepr97, line 1215: nit: can we rename this to `storage_dirs_exist`? the bool value then makes more sense when you read it in places where this function gets called

src/instructlab/configuration.py
```diff
@@ -1167,13 +1182,54 @@ def map_train_to_library(ctx, params):
...
+    dirs_to_make = [
```
jaideepr97, line 1216: nit: maybe call this `dirs_to_check`?

src/instructlab/configuration.py
nathan-weinberg, line 1235: This needs to be removed

src/instructlab/configuration.py
```diff
@@ -1167,13 +1182,54 @@ def map_train_to_library(ctx, params):
...
+            "node_rank": 0,
+            "nnodes": 1,
+            "rdzv_id": 123,
+            "rdzv_endpoint": "127.0.0.1:12222",
+            "deepspeed_cpu_offload_optimizer_ratio": 1,
+            "deepspeed_cpu_offload_optimizer_pin_memory": False,
+            "lora_alpha": 32,
+            "lora_dropout": 0.1,
+            "lora_target_modules": ["q_proj", "k_proj", "v_proj", "o_proj"],
+        }
```
nathan-weinberg, line 1206: Why are there values here that are only initialized in a corner case? shouldn't these be getting pulled from somewhere else?

src/instructlab/configuration.py
```diff
@@ -1075,17 +1085,22 @@ def init(
...
+            # only create dirs if the user passed --config <some_file_other_than_default>
+            if config_file != DEFAULTS.CONFIG_FILE:
+                ensure_storage_directories_exist()
             config_obj = read_config(config_file)
         except ConfigException as e:
             # delayed, so ilab config init can override a broken config.
             config_obj = get_default_config()
             error_msg = str(e)
     else:
+        # if the user is running a cmd without --config then we should NOT create the persistent storage dirs for them, they must run ilab config init
```
nathan-weinberg, line 1103:
```suggestion
        # if the user is running a cmd without --config then we should NOT create the persistent storage dirs for them, they must run `ilab config init`
```

src/instructlab/configuration.py
```diff
@@ -1039,10 +1045,14 @@ def recreate_train_profiles(overwrite: bool = False):
 
         for file, train_cfg in to_write.items():
             if not os.path.isfile(file) or overwrite:
+                # If any of the train profiles are missing treat this as a new system. We do not want to
```
nathan-weinberg, line 1048:
```suggestion
                # If any of the train profiles are missing, treat this as a new system. We do not want to
```

src/instructlab/configuration.py
```diff
@@ -866,6 +869,9 @@ def recreate_train_profiles(overwrite: bool = False):
             tmpl_file = os.path.join(profile_dir, file)
             train_cfg = read_train_profile(tmpl_file)
             if not os.path.isfile(new_file):
+                # If any of the train profiles are missing treat this as a new system. We do not want to
```
nathan-weinberg, line 872:
```suggestion
                # If any of the train profiles are missing, treat this as a new system. We do not want to
```

src/instructlab/configuration.py
cdoern: this whole persistent storage setup is a race condition waiting to happen, which is why this has to be done this way. These are initialized elsewhere but for the purpose of being written to disk. I'll try to unify the declaration of the type.

src/instructlab/config/init.py
```diff
@@ -165,18 +165,19 @@ def init(
...
         )
         if not overwrite:
             raise click.exceptions.Exit(0)
-    if exists(DEFAULTS.TRAIN_PROFILE_DIR):
+    if exists(DEFAULTS.TRAIN_PROFILE_DIR) and not fresh_install:
         return click.confirm(
             f"Found {DEFAULTS.TRAIN_PROFILE_DIR}, do you also want to reset existing profiles?"
         )
-    return True
+    # default behavior should not be do NOT overwrite files that could have just been created
```
alinaryan, line 179: nit: fixup the comment

cdoern:
I am waiting to push until all tests pass, but once they do I will be force-pushing again FYI

src/instructlab/clickext.py
```diff
@@ -76,6 +79,15 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None
                 "in a future release. Please consider using "
                 f"`ilab {group} {primary}` instead"
             )
+            # only trigger this if check_storage_dirs_exist results in an error AND you are NOT calling `ilab init` AND --config is something other than the default file (meaning the user did not specify it)
+            if not check_if_storage_dirs_exist() and (
```
n1hility, line 83: Is this the right check? Isn't what you want to test that DEFAULTS.CONFIG_FILE in ~/.config/instructlab exists?

The directories might be lingering from a past install for example

cdoern: what this is meant to be checking is:

1. are there missing dirs from `check_if_storage_dirs_exist`
2. the cmd isnt `init` which will create all of this now so we can skip the error
3. the --config flag is the default, meaning it isnt  `--config=<something_custom>.yaml`, if they did we should also skip this since we will create the dirs for them.

I think this accomplishes this? Its not meant to be checking the default file exists but rather the content of the `--config` flag

src/instructlab/clickext.py
```diff
@@ -76,6 +79,15 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None
                 "in a future release. Please consider using "
                 f"`ilab {group} {primary}` instead"
             )
+            # only trigger this if check_storage_dirs_exist results in an error AND you are NOT calling `ilab init` AND --config is something other than the default file (meaning the user did not specify it)
```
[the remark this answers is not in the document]
cdoern: I am changing it to this:

```
            # if some storage dirs do not exist
            # AND the command is not `ilab init`
            # AND the --config flag is not customized, then we error
```

src/instructlab/configuration.py
cdoern: sure

src/instructlab/config/init.py
```diff
@@ -165,18 +165,19 @@ def init(
...
-def check_if_configs_exist() -> bool:
+def check_if_configs_exist(fresh_install) -> bool:
     if exists(DEFAULTS.CONFIG_FILE):
         overwrite = click.confirm(
             f"Found {DEFAULTS.CONFIG_FILE}, do you still want to continue?"
         )
         if not overwrite:
             raise click.exceptions.Exit(0)
-    if exists(DEFAULTS.TRAIN_PROFILE_DIR):
+    if exists(DEFAULTS.TRAIN_PROFILE_DIR) and not fresh_install:
```
jaideepr97, line 175: do we want to add this check to the config file as well? dont know if we want to keep them consistent

src/instructlab/configuration.py
jaideepr97: think you mightve missed this

tests/test_lab.py
```diff
@@ -92,12 +93,12 @@ def has_debug_params(self) -> bool:
     Command(("model", "test")),
     Command(("model", "train")),
     Command(("model", "list")),
-    Command(("data",), needs_config=False),
+    Command(("data",), needs_config=False, should_fail=False),
```
jaideepr97, line 96: is this change needed in this PR if we never test the `should_fail=True` use case in this PR?

jaideepr97: seems to me like `should_fail` would default to false even if not specified so all the cases are probably defaulting to that right now so im having trouble understanding what this change is adding

alinaryan (approved):
LGTM. Nice work!
I think the persistent storage dir fix could have been moved to a sep pr to keep the changes smaller. I understood it as a separate issue, but lmk if that's wrong

tests/test_lab.py
cdoern: should_fail is default True above.

cdoern:
> I think the persistent storage dir fix could be moved to a sep pr in the future to keep the changes smaller. I understood it as a separate issue, 

I am confused what you mean by this @alinaryan ? The persistent storage dir fix most likely needed to be in this PR because without it, `ensure_persistent_storage_directories_exist` runs twice and once you get to `ilab config init`, its impossible to tell if you are running on a fresh install or if the files previously existed since `init` found in `configuration.py` also used to run the function

tests/test_lab.py
jaideepr97: oh sorry, missed that 
though i feel like thats an odd default tbh

jaideepr97: i think it might be a bit more intuitive to inverse it personally
not a blocker though

src/instructlab/config/init.py
cdoern: no because only the train profiles/ train profile dir is in the persistent storage creation func. The config yaml is handled only here.

cdoern:
fixed the dirs_to_check @jaideepr97 sorry I missed that

src/instructlab/configuration.py
jaideepr97: did you miss this? not sure why this still shows up to me. maybe it was resolved somewhere else

alinaryan:
@cdoern gotcha. thanks for clarifying!! lgtm

src/instructlab/taxonomy/taxonomy.py
```diff
@@ -1,10 +1,16 @@
...
+import logging
+
 # Third Party
 import click
 
 # First Party
 from instructlab import clickext
+from instructlab.configuration import storage_dirs_exist
+
+logger = logging.getLogger(__name__)
```
jaideepr97, line 13: nit: I see logger instance initializations in a few places in this PR but im not seeing them getting used anywhere 
dont know if we need them but not a blocker

jaideepr97 (commented):
just one last thing i could think of for now - happy to approve after that

src/instructlab/taxonomy/taxonomy.py
cdoern: agh, apologies I was using the logger at first. I will re-push and rm

khaledsulayman (approved):

khaledsulayman (approved):

src/instructlab/clickext.py
n1hility: Why are the storage directories relevant? The purpose seems to be to detect whether or not the user should run ilab config init. Isn't the reason to run ilab config init based on whether they have a config, and not whether they have directories?

PR #2053, opened 2024-08-12
Add ilab shell option that provides interactive REPL (Read-Eval-Print Loop) interface for the ilab CLI

eranco74:
This change allows interactive ilab CLI and command discovery

**Checklist:**


eranco74:
This is what it looks like:
Screenshot from 2024-08-12 22-21-46

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Features
 
+* A new `ilab shell` command is introduced as a convenience feature, which stat an interactive
```
leseb, line 5:
```suggestion
* A new `ilab shell` command is introduced as a convenience feature, which starts an interactive
```

?

leseb (commented):
What's a typical use case for this? Thanks

eranco74:
> What's a typical use case for this? Thanks
See this [demo](https://drive.google.com/file/d/1ysiuVBnwp_nXHqG413nGdSE4EYRRIxEk/view?usp=sharing)

I created a user and set the login shell to start ilab in shell mode.
This allows me to let the user use ilab on the system without exposing the underlying RHEL/Linux.

src/instructlab/lab.py
```diff
@@ -48,3 +49,17 @@ def ilab(ctx, config_file, debug_level: int = 0):
...
     """
     cfg.init(ctx, config_file, debug_level)
+
+
+# Register the REPL with a custom prompt
+def ilab_prompt():
+    return "ilab # "
+
+
+@ilab.command()
```
cdoern, line 59: can this belong to a cmd group? maybe system? We are trying to group all cmds.

leseb (commented):
I'm casting a blank vote on this. Honestly, I don't quite see its usefulness, but that could just be me. I can't think of a scenario where I'd personally find it helpful.

That said, it might offer new users a more user-friendly, interactive way to explore and get acquainted with the CLI. 🤷🏻 

Thanks for your contribution.

bjhargrave:
> I don't quite see its usefulness

Same for me. REPL makes more sense when there is significant state in the runtime (like Python has namespaces, variable values, function definitions, etc.). ilab basically has zero state which is why it is just a bunch of subcommands. Any state between running commands is stored in the file system and not volatile memory.

Adding a REPL is just additional (cognitive) complexity but does not add any real value to the user.

For command discovery, we already have `--help`. If we want command completion, we can supply bash/zsh/... completions.

PR #2060, opened 2024-08-14
chore: remove unecessary @model_validator

leseb:
All the defaults are coming from the Field(default), thus using the
@model_validator to populate default values is not needed anymore.

Also, the @model_validator is typically used for custom logic, like if
you have more complex logic for setting defaults, such as computing them
based on other fields or some external data, the model_validator could
be necessary. However, in this case, the logic is simply setting the
default values, which Pydantic can already do through Field().

A few options with default_factory were incorrect too, we don't need a
lambda here. The default_factory should directly reference the callable
that returns an instance of the class.

Co-Authored-By: Ihar Hrachyshka <ihrachys@redhat.com>

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -321,13 +321,6 @@ class _chat(BaseModel):
         description="The maximum number of tokens that can be generated in the chat completion. Be aware that larger values use more memory.",
     )
 
-    @model_validator(mode="before")
-    def fill_defaults(
-        cls, values: typing.Dict[str, typing.Any]
-    ) -> typing.Dict[str, typing.Any]:
-        defaults = {"model": DEFAULTS.DEFAULT_MODEL}
-        return finish_cfg_section(defaults, values)
-
 
 class _serve_vllm(BaseModel):
     """Class describing configuration of vllm serving backend."""
@@ -378,12 +371,12 @@ class _serve(BaseModel):
     model_config = ConfigDict(extra="ignore", protected_namespaces=())
     # vllm configuration
     vllm: _serve_vllm = Field(
-        default_factory=lambda: _serve_vllm,
+        default_factory=_serve_vllm,
         description="vLLM serving settings.",
     )
     # llama-cpp configuration
     llama_cpp: _serve_llama_cpp = Field(
-        default_factory=lambda: _serve_llama_cpp,
+        default_factory=_serve_llama_cpp,
         description="llama-cpp serving settings.",
     )
     model_path: StrictStr = Field(
@@ -414,17 +407,6 @@ class _serve(BaseModel):
     def api_base(self):
         """Returns server API URL, based on the configured host and port"""
         return get_api_base(self.host_port)
-
-    @model_validator(mode="before")
-    def fill_defaults(
-        cls, values: typing.Dict[str, typing.Any]
-    ) -> typing.Dict[str, typing.Any]:
-        defaults = {
-            "model_path": DEFAULTS.DEFAULT_MODEL,
-            "llama_cpp": {"gpu_layers": -1, "max_ctx_size": 4096, "llm_family": ""},
-            "vllm": {"llm_family": "", "vllm_args": [], "max_startup_attempts": 120},
-        }
-        return finish_cfg_section(defaults, values)
 
 
 class _generate(BaseModel):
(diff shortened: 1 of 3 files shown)


booxter:
Is it trying to do the same thing as https://github.com/instructlab/instructlab/pull/1985 ?

leseb:
> Is it trying to do the same thing as #1985 ?

Oh boys, that's what happens when you go on Holidays and miss some history there... Actually https://github.com/instructlab/instructlab/pull/2055 is the starting block. Sorry about the inconvenience, how do you want to proceed?

booxter:
FYI we discussed with @leseb the overlap and we agreed that we'll go with Seb's PRs but he'll just add some prior art comment in commit message to acknowledge authorship of the overlapping work. And also, `get_default_config` should be squashed to just `return Config()` at this point, since all defaults are now in fields.

src/instructlab/configuration.py
```diff
@@ -786,125 +682,10 @@ class Config(BaseModel):
...
-            ),
-            mmlu=_mmlu(
-                few_shots=5,
-                batch_size="auto",
-            ),
-            mt_bench_branch=_mtbenchbranch(taxonomy_path=DEFAULTS.TAXONOMY_DIR),
-            mmlu_branch=_mmlubranch(tasks_dir=DEFAULTS.DATASETS_DIR),
-        ),
-    )
+    return Config()
```
booxter, line 688: at this point the necessity of this function is questionable, but it can be cleaned up later if someone wants to follow up.

booxter (approved):

src/instructlab/configuration.py
leseb: Yeah, thought about that too but it's called in a bunch of places, and was trying to minimize the changes (or lazy?) :). Thanks!

nathan-weinberg (approved):
LGTM but added a `hold` as I'd like @cdoern to take a quick look before merging - thanks @leseb!

src/instructlab/configuration.py
```diff
@@ -786,125 +682,10 @@ class Config(BaseModel):
...
-            num_epochs=10,
-            effective_batch_size=3840,
-            save_samples=250000,
-            lora_quantize_dtype="nf4",
-            lora_rank=4,
-            nproc_per_node=1,
-            deepspeed_cpu_offload_optimizer=False,
-            additional_args={},
-            is_padding_free=False,
-            phased_phase1_num_epochs=10,
```
cdoern, line 887: did we lose these defaults?

tests/testdata/default_config.yaml
```diff
@@ -301,13 +301,13 @@ train:
...
   # Default: None
-  phased_mt_bench_judge: /cache/instructlab/models/prometheus-eval/prometheus-8x7b-v2.0
+  phased_mt_bench_judge:
   # Phased phase1 effective batch size.
   # Default: 128
   phased_phase1_effective_batch_size: 128
   # Number of epochs to run training for during phase1.
   # Default: None
-  phased_phase1_num_epochs: 10
+  phased_phase1_num_epochs:
```
cdoern, line 310: same here, losing the default. I am not seeing a Field for this added

tests/testdata/default_config.yaml
```diff
@@ -317,7 +317,7 @@ train:
   phased_phase2_effective_batch_size: 3840
   # Number of epochs to run training for during phase2.
   # Default: None
-  phased_phase2_num_epochs: 10
+  phased_phase2_num_epochs:
```
cdoern, line 320: This one too, I think?

cdoern (commented):
I think I found a few missing defaults that will break downstream if removed. I just want to ensure these all have field equivalents!

src/instructlab/configuration.py
leseb: Initially, the `phased_phase1_num_epochs` was declared with a default to None. So I just kept that reference. Do you want me to replace the default with the override you mentioned? Thanks!

src/instructlab/configuration.py
```diff
@@ -786,125 +682,10 @@ class Config(BaseModel):
...
-            model_path=DEFAULTS.MODEL_REPO,
-            data_path=DEFAULTS.DATASETS_DIR,
-            ckpt_output_dir=DEFAULTS.CHECKPOINTS_DIR,
-            data_output_dir=DEFAULTS.INTERNAL_DIR,
-            max_seq_len=4096,
-            max_batch_len=10000,
-            num_epochs=10,
-            effective_batch_size=3840,
-            save_samples=250000,
-            lora_quantize_dtype="nf4",
```
leseb, line 881: @cdoern about this one, the default was `nf4` but then overridden later with None, which one do you want to keep? Thanks

src/instructlab/configuration.py
```diff
@@ -786,125 +682,10 @@ class Config(BaseModel):
...
-            lora_rank=4,
```
leseb, line 882: Same for this one, default to 4 then changed to 0. Which one should we keep? @cdoern

cdoern (approved):
LGTM, thanks for the amendments!

PR #2067, opened 2024-08-15
cloud-instance: Print help when only cloud type is passed

russellb:
When running the following command,

    cloud-instance.sh ec2

the output would be empty as the script would silently exit on the
`shift`. It is desirable to output the usage help text in this case.
Adjust the `ec2` and `ibm` handlers to handle this case and print the
usage text when no command is provided.

### DIFF
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -14,7 +14,12 @@ else
 fi
 
 ec2() {
-    local cmdname=$1; shift
+    local cmdname=$1
+    if [ -z "$cmdname" ]; then
+        show_usage
+        exit 1
+    fi
+    shift
     if [ "$(type -t "ec2__$cmdname")" = "function" ] >/dev/null 2>&1; then
         INSTANCE_NAME="$EC2_INSTANCE_NAME"
         "ec2__$cmdname" "$@"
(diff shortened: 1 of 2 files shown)


danmcp:
Want to make the same fix in local-setup-sh?

scripts/infra/cloud-instance.sh
```diff
@@ -14,7 +14,12 @@ else
 fi
 
 ec2() {
-    local cmdname=$1; shift
+    local cmdname=$1;
```
booxter, line 17: nit: semicolons are redundant here.

booxter (approved):
+1 but Dan's comment applies too.

scripts/infra/cloud-instance.sh
russellb: fixed

russellb:
> Want to make the same fix in local-setup-sh?

sure, done

danmcp (changes requested):
I think if you rebase there is now 1 more case in local-setup

russellb:
> I think if you rebase there is now 1 more case in local-setup

done

booxter (approved):

nathan-weinberg (approved):

PR #2072, opened 2024-08-15
Add better tuning for max_workers based on perf tests

### LINKED ISSUE #2050: evaluate --max-workers recommendation might not work well for high-end hardware
The current evaluate --max-workers recommendation is #cpus +- 50%.  The logic behind this was:

- Testing showed that value range usually contained the best performance with 1-4 GPU systems and 8-48 core cpus.  In systems with > cpu to gpus ratios, it's possible the recommendation might not hold.
- While the majority of the work is being done by the GPUs, tuning the workers to the CPU seemed to have the strongest correlation.   Likely because the threads were CPU bound.  So smaller numbers couldn't keep the GPUs busy and larger numbers started creating thrashing on the GPU side.

The logic likely isn't sophisticated enough and needs more testing to confirm.  Two edge cases that might help to determine a more general algorithm:

- Greater ratio of CPU to GPU.  Ex: 48 CPU cores with 1 GPU vs 8 CPU cores and 1 GPU.
- Larger systems.  ~200 cores with 4-8 GPUs

The final recommendation might need to take into account the CPU:GPU ratio in addition to using CPUs as the primary tuning indicator

**To Reproduce**
1. time ilab model evaluate --model mymodelpath --benchmark mt_bench --max_workers=X --gpus=Y

**Additional context**
Note: Be sure not to count the timing of the first run because of the first model load taking longer

danmcp:
The result of those tests showed ideal values for max_workers needs to take gpu and cpu counts into account.  The new recommendation does have a magic number but hopefully the suggestion range is large enough to account for the magic number potentially not being perfect.  Removing the magic number would require understanding the relative compute power of the gpus.

Also switching from multiprocessing.cpu_count() to len(os.sched_getaffinity(0)) to avoid cases where the instance has cpus that aren't available to the current process. (Thanks @jaideepr97 )

Resolves #2050 

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/vllm.py
+++ b/src/instructlab/model/backends/vllm.py
@@ -143,9 +143,36 @@ def format_template(model_family: str, model_path: pathlib.Path) -> str:
     return prefix + template
 
 
-def contains_argument(prefix: str, arg: typing.Iterable[str]) -> bool:
+def contains_argument(prefix: str, args: typing.Iterable[str]) -> bool:
     # Either --foo value or --foo=value
-    return any(s == prefix or s.startswith(prefix + "=") for s in arg)
+    return any(s == prefix or s.startswith(prefix + "=") for s in args)
+
+
+def get_argument(prefix: str, args: typing.List[str]):
+    # Return last value in args for either --foo value or --foo=value
+    # Returns True if flag --foo is provided with no value
+
+    args_len = len(args)
+    # Traverse the args in reverse (args_len-1 to 0)
+    for i in range(args_len - 1, -1, -1):
+        s = args[i]
+        if s == prefix:
+            # Case: --foo value or --foo (with no value)
+            if i == args_len - 1:
+                # --foo is the last entry, must be a flag
+                return True
+            next_arg = args[i + 1]
+            if next_arg.startswith("-"):
+                # No value provided, must be a flag
+                return True
+            # The entry after is the value
+            return next_arg
+        v = prefix + "="
+        if s.startswith(v):
+            # Case: --foo=value
+            # Return everything after prefix=
+            return s[len(v) :]
+    return None
 
 
 def create_tmpfile(data: str):
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -5,7 +5,6 @@
 from copy import deepcopy
 import enum
 import logging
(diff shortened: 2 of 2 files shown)


src/instructlab/model/backends/vllm.py
```diff
@@ -143,9 +143,23 @@ def format_template(model_family: str, model_path: pathlib.Path) -> str:
...
 
 
-def contains_argument(prefix: str, arg: typing.Iterable[str]) -> bool:
+def contains_argument(prefix: str, args: typing.Iterable[str]) -> bool:
     # Either --foo value or --foo=value
-    return any(s == prefix or s.startswith(prefix + "=") for s in arg)
+    return any(s == prefix or s.startswith(prefix + "=") for s in args)
+
+
+def get_argument(prefix: str, args: typing.List[str]):
```
booxter, line 151: This is style suggestion, but perhaps you could adopt it (I'm a bit confused about all the indices).

```
def get_argument2(prefix: str, args):
    # Traverse the args in reverse
    args = list(reversed(args))
    for i, arg in enumerate(args):
        if arg == prefix:
            if i == 0: # last argument
                return None
            return args[i - 1]
        v = prefix + "="
        if arg.startswith(v):
            return arg.removeprefix(v)
    return None
```

danmcp: The reason for not using reversed was to avoid creating an extra list just to walk it backwards.  I agree it's harder to read.  I added the comment because of that but maybe it's still too much.

The downside of using removeprefix is that we have already done the check whether it should be removed and then removeprefix will do it again.  It could also be written as:

```
x = arg.removeprefix(v):
if x != arg:
    return x
```

I don't have a strong preference on either of these.

booxter: I am all for runtime optimization but I prefer readability. Your `removeprefix` version is also fine by me. I guess I'm not used to so many `__getitem__`s in Python code. Others may be ok with it.

danmcp: I've made a few changes to the logic to handle flags without args and added several comments.  Let me know if that's any more agreeable.  Otherwise I am happy to adjust to whatever methodology has consensus.

jaideepr97 (approved):
thanks @danmcp

PR #2082, opened 2024-08-15
Update e2e-nvidia-t4-x1.yml

jjasghar:
From the documentation, it seems these steps are now required: https://github.com/instructlab/instructlab?tab=readme-ov-file#install-with-nvidia-cuda

### DIFF
--- a/.github/workflows/e2e-nvidia-t4-x1.yml
+++ b/.github/workflows/e2e-nvidia-t4-x1.yml
@@ -88,12 +88,14 @@ jobs:
           nvidia-smi
           python3.11 -m pip cache remove llama_cpp_python
 
-          CMAKE_ARGS="-DLLAMA_CUDA=on" python3.11 -m pip install .
+          CMAKE_ARGS="-DLLAMA_CUDA=on -DLLAMA_NATIVE=off" python3.11 -m pip install .
 
           # https://github.com/instructlab/instructlab/issues/1821
           # install with Torch and build dependencies installed
           python3.11 -m pip install packaging wheel setuptools-scm
           python3.11 -m pip install .[cuda]
+
+          python3.11 -m pip install vllm@git+https://github.com/opendatahub-io/vllm@2024.08.01
 
       - name: ilab system info
         run: |


nathan-weinberg:
This is a good catch - two things are confusing me here:
1. Why does the README use the exact tag instead of pointing to https://github.com/instructlab/instructlab/blob/main/requirements-vllm-cuda.txt?
2. Why isn't vLLM being installed as part of [the CUDS-specific requirements](https://github.com/instructlab/instructlab/blob/main/requirements/cuda.txt)? 

@cdoern @n1hility do either of you know the answers to the above?

nathan-weinberg:
cc @alimaredia

cdoern:
Requirements Cuda file cannot be referenced in the Readme bc you only have that if you clone the entire Ilab repo. If you're installing from pip you need to install vllm by itself.

If you have the repo cloned, sure you can use the txt file. I put this here to cover every use case.

cdoern:
I do agree that vllm should be added to the instructlab[cuda] target though.

alimaredia (approved):
Lgtm as a stop gap. It looks like a follow up issue needs to be made to get vllm installed when `pip install .[cuda]` is run?

.github/workflows/e2e-nvidia-t4-x1.yml
```diff
@@ -88,13 +88,15 @@ jobs:
...
 
-          CMAKE_ARGS="-DLLAMA_CUDA=on" python3.11 -m pip install .
+          CMAKE_ARGS="-DLLAMA_CUDA=on -DLLAMA_NATIVE=off" python3.11 -m pip install .
 
           # https://github.com/instructlab/instructlab/issues/1821
           # install with Torch and build dependencies installed
           python3.11 -m pip install packaging wheel setuptools-scm
           python3.11 -m pip install .[cuda]
 
+          python3.11 -m pip install vllm@git+https://github.com/opendatahub-io/vllm@2024.08.01
```
n1hility, line 98: A couple of issues here:

1. Ideally this refer to vllm-requirements.txt instead of the direct reference so we have one place to bump versions/tags
2. These two install operations are not guaranteed to be compatible since each only has their respective dependency set. They must either be installed in a combined single install command, or they have to share a common constraints file.

I recommend replacing this section of install commands with the approach in the other workflows: https://github.com/instructlab/instructlab/blob/9e1770a86b5519c84f7b4d3d8bf12f2b3bd2d690/.github/workflows/e2e-nvidia-a10g-x4.yml#L124

n1hility (changes requested):

nathan-weinberg:
This PR seems abandoned and quite a few changes have been made to the E2E ecosystem since it was opened.

Closing this for now.

PR #2083, opened 2024-08-15
Update README.md

jjasghar:
python3 defaults to python3.9 if you don't specify 3.11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
    #### Install using PyTorch without CUDA bindings and no GPU acceleration
 
       ```shell
-      python3 -m venv --upgrade-deps venv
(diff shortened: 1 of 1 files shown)


bjhargrave:
> python3 defaults to python3.9

I think it depends upon how one's system is setup:

```shell
➜  ~ python3 --version
Python 3.12.5
```

But just using python3 is not specific enough since we now require with 3.10 or 3.11.

danmcp:
Maybe we should instead add a comment to explain what python3 needs to be (or to use python3.10 or python3.11 if desired).

danmcp (changes requested):
Maybe we should instead add a comment to explain what python3 needs to be (or to use python3.10 or python3.11 if desired).

nathan-weinberg:
Superceded by https://github.com/instructlab/instructlab/pull/2226

Issue #2080, opened 2024-08-15
Log file for serving


alimaredia:
This feature request has to do with logging the output of `ilab model serve` to a file and the out put of other `ilab` commands that serve models under the hood.

1. Verify the `--log-file` flag for ilab model serve works
2. Send background vllm serving output to a file instead of to /dev/null
3. Design and add config variables and flags to all commands that do background serving

leseb:
`log_file` is broken since each module as its own logger instance.

leseb:
Taking over as agreed offline.

ktam3:
@leseb - do you have any new updates on this?

leseb:
> @leseb - do you have any new updates on this?

PR was posted a while ago and I'm still waiting to get some reviews https://github.com/instructlab/instructlab/pull/2084 :)

ktam3:
Ack, let me see if I can get some movement on this

PR #2084, opened 2024-08-15
fix: enable server log to file

leseb:
Since we no longer pass loggers between modules, we've lost the ability to easily redirect stdout/stderr to a logger. Now, any encountered loggers will be redirected to the `log_file` if specified.

We had to introduce a new ServerConfig class to reduce the number of
attributes to BackendServer. However, only a few options were added to
avoid pylint failing. Later on, we will continue to move other
attributes into ServerConfig. But for now, let's keep the change
minimal.

Fixes: https://github.com/instructlab/instructlab/issues/1738, https://github.com/instructlab/instructlab/issues/2080

**Checklist:**

### DIFF
--- a/src/instructlab/log.py
+++ b/src/instructlab/log.py
@@ -4,7 +4,9 @@
 import logging
 import os
 import pathlib
-import sys
+
+# Local
+from .defaults import LOG_FORMAT
 
 
 class CustomFormatter(logging.Formatter):
@@ -18,39 +20,61 @@ class CustomFormatter(logging.Formatter):
 
 
 class LoggerWriter:
-    def __init__(self, level):
+    def __init__(self, logger, level):
+        self.logger = logger
         self.level = level
-        self.buffer = ""
 
     def write(self, message):
-        self.buffer += message
-        if message == "\n":  # Flush the buffer when an empty line is encountered
-            self.flush()
+        if message.strip():
+            self.logger.log(self.level, message)
 
     def flush(self):
-        if self.buffer:
-            self.level(self.buffer.strip())
-            self.buffer = ""
-
-    def isatty(self):
-        return False
+        pass
 
 
-def stdout_stderr_to_logger(
-    logger: logging.Logger, log_file: pathlib.Path | None, fmt: str
-):
-    if log_file:
-        # Use the existing log file if it exists and append to it
-        mode = "a"
-        # Create file handler
-        file_handler = logging.FileHandler(log_file, mode, encoding="utf-8")
-        logger.addHandler(file_handler)
-        # Create formatter and add it to the handler
-        formatter = CustomFormatter(fmt)
-        file_handler.setFormatter(formatter)
+def add_file_handler_to_logger(
+    logger: logging.Logger, log_file: pathlib.Path | None
+) -> None:
+    """
+    Adds a file handler to the specified logger to write log messages to a file.
 
-    sys.stdout = LoggerWriter(logger.info)
-    sys.stderr = LoggerWriter(logger.error)
+    This function checks if a file handler for the specified log file already exists in the logger.
+    If it does, the function does nothing. Otherwise, it creates a new file handler that appends
+    log messages to the specified file and adds it to the logger.
+
+    Args:
+        logger (logging.Logger): The logger to which the file handler should be added.
+        log_file (pathlib.Path): The path to the log file where log messages should be written.
+    """
+
+    if log_file is None:
+        return
+
+    log_file_path = os.path.abspath(log_file)
+    logger.debug(f"Checking for existing file handler for log file: {log_file}")
+
+    # Check if a file handler for the specified log file already exists
+    for handler in logger.handlers:
+        if isinstance(handler, logging.FileHandler):
+            if handler.baseFilename == log_file_path:
+                logger.debug(
+                    f"Log file already exists in handler: {log_file}, doing nothing!"
+                )
+                return
+    logger.debug(f"No file handler found for log file: {log_file}, adding one!")
+
+    # Use the "append" mode
+    mode = "a"
+
+    # Create file handler and add it to the logger
+    file_handler = logging.FileHandler(log_file, mode, encoding="utf-8")
+    logger.debug(f"Adding file handler for log file: {log_file}")
+    logger.addHandler(file_handler)
+
+    # Create formatter and set it
+    formatter = CustomFormatter(LOG_FORMAT)
+    file_handler.setFormatter(formatter)
+    logger.debug(f"Added file handler for log file: {log_file}")
 
 
 def configure_logging(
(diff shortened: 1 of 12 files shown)


leseb:
@alimaredia FYI

src/instructlab/model/backends/backends.py
```diff
@@ -77,18 +85,26 @@ def __init__(
...
         self.host = host
         self.port = port
+        self.config = config
         self.resources: list[Closeable] = []
 
+        # Write logs to a file if log_file is provided
+        if self.config.log_file:
+            self.write_logs_to_file(logger)
+
+    def write_logs_to_file(self, a_logger: logging.Logger) -> None:
```
danmcp, line 104: This seems more like a setup than a write operation?

leseb: Funny enough, I initially had `setup_logs_to_file`. I guess I'll use setup then. Thanks!

src/instructlab/model/backends/llama_cpp.py
```diff
@@ -211,8 +230,11 @@ def read_root():
     if is_temp_server_running():
         # TODO: redirect temp server logs to a file instead of hidding the logs completely
         # Redirect stdout and stderr to null
+        log_to = pathlib.Path(os.devnull)
```
danmcp, line 233: Maybe move to an else below to avoid instantiating unless needed?

src/instructlab/model/backends/vllm.py
```diff
@@ -247,6 +261,15 @@ def run_vllm(
                 env=vllm_env,
                 start_new_session=True,
             )
+        elif log_file:
+            # write both stdout and stderr to the log file
+            with log_file.open("a") as f:
+                vllm_process = subprocess.Popen(
+                    args=vllm_cmd,
+                    stdout=f,
+                    stderr=f,
```
danmcp, line 270: Does this need: env=vllm_env,

src/instructlab/model/backends/vllm.py
```diff
@@ -247,6 +261,15 @@ def run_vllm(
                 env=vllm_env,
                 start_new_session=True,
             )
+        elif log_file:
```
danmcp, line 264: I am missing how background isn't still set to true when the logfile is set for chat.  Where is that happening?

src/instructlab/model/chat.py
```diff
@@ -219,6 +231,10 @@ def chat(
             ctx.obj.config.serve, model_path=model
         )
 
+        if serving_log_file:
+            logger.debug(f"Setting serving log file to {serving_log_file}")
+            backend_instance.set_log_file(serving_log_file)
```
danmcp, line 236: Wondering whether this would be better passed to run_detached or to select_backend?

src/instructlab/model/backends/vllm.py
leseb: Simply because both are optional and can be activated independently. At this point `background` should become obsolete but I'm not sure we want to activate `log_file` through `background`. Did I get your question right?

src/instructlab/model/chat.py
leseb: Makes more sense indeed. Thanks!

src/instructlab/model/backends/vllm.py
danmcp: I'm not sure.  I am trying to parse through how background would be false at this point in the code when coming from chat.  I see chat go through run_detached accepting the default of background=True.  Which I think eventually ends up here.  I am sure I am missing something though.

danmcp: I went ahead and tried out this PR to try and and understand how it was working.  Got some strange behavior running with instructlab/granite-7b-lab:

- `ilab chat` worked as expected.  Logic went through `if background`.
- ilab chat --serving-log-file out.txt resulted in:

```
INFO 2024-08-28 23:48:07,704 instructlab.model.backends.vllm:140: Waiting for the vLLM server to start at http://127.0.0.1:58811/v1, this might take a moment... Attempt: 6/120
INFO 2024-08-28 23:48:07,708 instructlab.model.backends.vllm:147: vLLM engine successfully started at http://127.0.0.1:58811/v1
INFO 2024-08-28 23:48:07,723 instructlab.model.backends.vllm:0: ╭─────────────────────────────────────────────────────────────────────────────────── system ───────────────────────────────────────────────────────────────────────────────────╮
│ Welcome to InstructLab Chat w/ GRANITE-7B-LAB (type /h for help)                                                                                                             │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
                                                                               ]                                                    [S][default
INFO 2024-08-28 23:48:43,392 instructlab.model.backends.vllm:0: e
INFO 2024-08-28 23:48:43,652 instructlab.model.backends.vllm:0: x
INFO 2024-08-28 23:48:43,813 instructlab.model.backends.vllm:0: i
INFO 2024-08-28 23:48:44,145 instructlab.model.backends.vllm:0: t
```

every character I typed was preceded with the logging info prefix.  And the code path went through if background rather than elif logfile.  I switched the logic around to go through if logfile and got the same result.  So the problem isn't just with the if here.

Or maybe I am going something wrong?

leseb: Thanks for the testing, I'll dive into the problem.

src/instructlab/log.py
```diff
@@ -37,17 +38,59 @@ def isatty(self):
...
-    if log_file:
-        # Use the existing log file if it exists and append to it
-        mode = "a"
-        # Create file handler
-        file_handler = logging.FileHandler(log_file, mode, encoding="utf-8")
-        logger.addHandler(file_handler)
-        # Create formatter and set it for both handlers
-        formatter = CustomFormatter(FORMAT)
-        file_handler.setFormatter(formatter)
+def write_log_to_file(logger: logging.Logger, log_file: pathlib.Path | None) -> None:
```
danmcp, line 41: Wondering whether there is a better name for this?  Suggestions: `add_file_handler` or `setup_file_logger`

src/instructlab/model/backends/backends.py
```diff
@@ -59,18 +67,26 @@ def __init__(
...
+        self.config = config
         self.resources: list[Closeable] = []
 
+        # Write logs to a file if log_file is provided
+        if self.config.log_file:
+            self.setup_logs_to_file(logger)
+
+    def setup_logs_to_file(self, a_logger: logging.Logger) -> None:
+        """Write logs to a file if log_file is provided"""
+        write_log_to_file(a_logger, self.config.log_file)
```
danmcp, line 88: I am worried about how often this has to be called and how easy it will be to miss a case going forward.  Can you help me understand why this can't be done more globally?

leseb: Hum, now that I think about it, I could try to add the handler to the root logger instead all the sub-sequent loggers. Agrees that it does not look right 😓

leseb: Fixed using the root logger :)

src/instructlab/log.py
leseb: Renamed to `add_file_handler_to_logger`. Thanks

danmcp (approved):

danmcp:
@mergifyio rebase

PR #2085, opened 2024-08-15
Update A100_H100_x4 to slightly lower max_batch_len (55000->54000)

Maxusmusti:
Turns out there are edge cases where exactly 55k breaks, 54k should be fully safe

**Checklist:**


leseb (commented):
Can you add more rationale about that choice so that others can benefit from it :). Thanks!

Maxusmusti:
@leseb Yeah sure, essentially with training, we have found that the maximum value we can push that `max_batch_len` field to without the chance of hitting an OOM lies somewhere between 54k and 55k. A batch with a size above that value will overflow an 80GB card (in a 4x scenario) and crash with an OOM. Anything below that value should be within 80GB. Originally we thought that cut-off was 55k, but we realized it was slightly lower, so adjusting accordingly.

leseb:
> @leseb Yeah sure, essentially with training, we have found that the maximum value we can push that `max_batch_len` field to without the chance of hitting an OOM lies somewhere between 54k and 55k. A batch with a size above that value will overflow an 80GB card (in a 4x scenario) and crash with an OOM. Anything below that value should be within 80GB. Originally we thought that cut-off was 55k, but we realized it was slightly lower, so adjusting accordingly.

This is very helpful thank you. Could you please include this in your commit message? Thanks!

Maxusmusti:
@leseb done

leseb (approved):
Next time if you can wrap the commit body to 72 char that would be great :), thanks!

nathan-weinberg (approved):

PR #2102, opened 2024-08-19
Ignore directories with venv prefix

khaledsulayman:
Currently, `.gitignore` ignores virtual environments named `venv` or other similar patterns. I like to keep multiple local venvs for different python versions like `venv-3.10`, `venv-3.11`, etc. This PR proposes ignoring directories with `venv` as a prefix in order to facilitate keeping a variety of different local virtual environments.

**Checklist:**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,7 +121,7 @@ celerybeat.pid
 .env
 .venv
 env/
-venv/
+venv*/
 ENV/
 env.bak/
 venv.bak/


nathan-weinberg (approved):

danmcp (changes requested):
I am wondering if this is more appropriate to be done with a global .gitignore since it's up to specific development environment choices.

nathan-weinberg:
What if any scenario would we have where we wanted to commit a `venv*`?

khaledsulayman:
just want to add I confirmed that this doesn't impact files that begin with `venv`, only directories

danmcp:
> What if any scenario would we have where we wanted to commit a `venv*`?

I doubt we would ever hit the issue.  It's more about the maintenance.  The logic of .gitignore vs using a global .gitignore is to differentiate exactly these sort of use cases where developers have files specific to their environments.  Would you accept *venv because I like to name by venv dirs as <descriptor>venv

I would argue the same should apply for the dotfiles for all the various editors.  We have added support for a few already in early commits.  But if we allowed all of them to be added, we would probably have 20 or so and then we would need to maintain the list or decide which ones are important enough to add.  It's generally easier to avoid the topic and push everyone to use global config for this sort of stuff.

IMO, the things that should be in .gitignore are the things that affect a good percentage of users.  So if you require or recommend a particular dev environment or development pattern, putting it in .gitconfig makes sense.

nathan-weinberg:
> > What if any scenario would we have where we wanted to commit a `venv*`?
> 
> I doubt we would ever hit the issue. It's more about the maintenance. The logic of .gitignore vs using a global .gitignore is to differentiate exactly these sort of use cases where developers have files specific to their environments. Would you accept *venv because I like to name by venv dirs as venv
> 
> I would argue the same should apply for the dotfiles for all the various editors. We have added support for a few already in early commits. But if we allowed all of them to be added, we would probably have 20 or so and then we would need to maintain the list or decide which ones are important enough to add. It's generally easier to avoid the topic and push everyone to use global config for this sort of stuff.
> 
> IMO, the things that should be in .gitignore are the things that affect a good percentage of users. So if you require or recommend a particular dev environment or development pattern, putting it in .gitconfig makes sense.

We don't (as far as I know) have any community policies around this - perhaps it's worth adding some, but in the meantime this strikes me as a harmless addition that I don't see causing any maintenance overhead.

I'll go with whatever folks in general want to go with at the end of the day, but leaving my approval.

danmcp:
> but in the meantime this strikes me as a harmless addition that I don't see causing any maintenance overhead.

To point out how this can become a finicky maintenance problem.  If we are going to add venv*/, given what's already in the file, why wouldn't we also add:

.env*/
.venv*/
env*/
env*.bak/
venv*.bak/

Obviously using env* is a little more likely to create a future collision.

danmcp:
The other way to think about this is from the perspective of working on lots of projects.  As a dev working on multiple projects (all the instructlab projects are a good example).  Would I need to go contribute these changes to every repo in the project?

It's always seemed nicer (to me at least) that I can setup my dev settings in global config and that will apply to whatever project I start working on without me having to convince someone to accept my dev environment choices.

The maintenance finickiness is also increased by having mutliple projects as instructlab does.  Are we going to sync changes like this to all the repos?

khaledsulayman:
@danmcp I see your point and I do agree this is probably a good line to draw regardless of whether we expect the change to impact anything. I've set up a global config locally and it works great, so I'll close this PR. Thanks!

PR #2105, opened 2024-08-20
ci: improve merge commit content

leseb:
This commit introduces a commit message template in the mergify.yml
configuration file. The template includes the pull request title,
number, body, and a list of users who approved the review.

One important feature is to add the name of approvers to the PR, so the
merge commit will include fields like:

```
Approved-by: leseb
```

**Checklist:**

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -118,6 +118,14 @@ pull_request_rules:
   actions:
     merge:
       method: merge
+      commit_message_template: |
+        {{ title }} (#{{ number }})
+
+        {{ body }}
+
+        {% for user in approved_reviews_by %}
+        Approved-by: {{ user }}
+        {% endfor %}
     delete_head_branch:
 
 - name: label-cicd


nathan-weinberg (commented):
This is a great idea @leseb!

Could you share an example of what a merge commit might look like with this format?

.github/mergify.yml
```diff
@@ -118,6 +118,14 @@ pull_request_rules:
   actions:
     merge:
       method: merge
+      commit_message_template: |
+        {{ title }} (#{{ number }})
+
+        {{ body }}
+
+        {% for user in approved_reviews_by %}
```
bjhargrave, line 126: I am ok with the title and body. To me, the approved-by is just noise in the git log. If you need that level of information for a commit, you can find it in the PR.

leseb: GitHub already allows configuring the title and body for merge commits [natively](https://github.blog/changelog/2022-08-23-new-options-for-controlling-the-default-commit-message-when-merging-a-pull-request/?ref=blog.mergify.com). However, this PR is focused on adding an `Approved-By` field. This information isn't readily available by just looking at the git history, but it’s crucial for quickly identifying who authored and approved a PR. Including this field enhances accountability and traceability, making it easier to see who was responsible for key decisions.

leseb:
> This is a great idea @leseb!
> 
> Could you share an example of what a merge commit might look like with this format?

```
This is a PR title (https://github.com/leseb/supertini-rs/pull/3)

This is the PR body.

Approved-By: leseb
```

nathan-weinberg (approved):

.github/mergify.yml
bjhargrave: > This information isn't readily available by just looking at the git history

Yes but the values in the field are GitHub ids and not something general git like committer/author emails. So seeing a GitHub id in the commit message means you still need to go to GitHub to map the id back to someone. I would need to go to GitHub to "decode" `leseb` to some person.

leseb: I fixed this issue in my latest push.

.github/mergify.yml
```diff
@@ -118,6 +118,40 @@ pull_request_rules:
...
+            "jaideepr97": "Jaideep Rao <jrao@redhat.com>",
+            "joesepi": "Joe Sepi <sepi@joesepi.com>",
+            "kelbrown20": "Kelly Brown <kelbrown@redhat.com>",
+            "khaledsulayman": "Khaled Sulayman <ksulayma@redhat.com>",
+            "leseb": "Sébastien Han <seb@redhat.com>",
+            "mairin": "Máirín Duffy <duffy@redhat.com>",
+            "markmc": "Mark McLoughlin <markmc@redhat.com>",
+            "markstur": "Mark Sturdevant <mark.sturdevant@ibm.com>",
+            "nathan-weinberg": "Nathan Weinberg <nweinber@redhat.com>",
+            "RobotSail": "Oleg S <97077423+RobotSail@users.noreply.github.com>",
```
nathan-weinberg, line 147: @RobotSail is this your typical email?

nathan-weinberg (approved):
@leseb This still LGTM - my only concern is that this list will need to kept up-to-date with the Maintainers no? Should we add a note in the community repo that this needs to be updated when a Maintainer is added?

leseb:
> @leseb This still LGTM - my only concern is that this list will need to kept up-to-date with the Maintainers no? Should we add a note in the community repo that this needs to be updated when a Maintainer is added?

If this gets approved I'll update the community repo :)

.github/mergify.yml
```diff
@@ -118,6 +118,40 @@ pull_request_rules:
...
       method: merge
+      # The dict of GitHub user ID and sign-off was built from
+      # https://github.com/instructlab/community/blob/main/MAINTAINERS.md
+      # Then, the sign-off was looked up in the git history
+      commit_message_template: |
+        {{ title }} (#{{ number }})
+
+        {{ body }}
+
+        {% set approved_by = {
```
bjhargrave, line 129: > my only concern is that this list will need to kept up-to-date with the Maintainers no? Should we add a note in the community repo that this needs to be updated when a Maintainer is added?

I think this list is not sustainable. If we must include approved by in the commit message, then let's just use the GitHub id rather than maintaining this new maintainers list.

bjhargrave (changes requested):

.github/mergify.yml
leseb: Now I have built the list and it defaults to the GitHub user ID if not present, so why not use it? We don't add maintainers frequently, don't you think saying it's "not sustainable" is a bit exaggerated? Thanks!

.github/mergify.yml
RobotSail: @nathan-weinberg Yes, I use the GitHub email obfuscation

.github/mergify.yml
bjhargrave: It is a repetition of the source of truth and will deviate over time since attention won't be paid. It is also hard to copy this file to another instruct lab repo if they want to use mergify. 

If this goes in, I would appreciate removing me from the list.

leseb: Sounds good. I think we've debated this enough, and it's time for others to weigh in. If the majority approves, I'll remove your sign-off from `.github/mergify.yml`. Thanks.

danmcp:
I don't have a strong opinion, but I would vote for only including github ids.  The reasons being (most of these are nits):
- It's one more place that isn't changed after the fact if someone changes their identification in git
- It's not difficult to look up someone by id if you don't know who they are.  And the people looking for who approved will mostly know people by their ids anyway.
- (already stated) It is one less thing to maintain if you only include ids
- (this is probably a me thing) When skimming through the commit log, I would prefer the author name to stand out over the approver names and only using ids for approvers helps a bit.

leseb:
Thanks @tiran @markmc @bjhargrave @danmcp for weighing in. We now use the GitHub IDs. PTAL.

bjhargrave (approved):

danmcp (approved):

PR #2114, opened 2024-08-20
refactor: move ensure_server() to vllm.py

makelinux:
ensure_server is used only by vllm.

Move it to reduce circular dependency between backends.py and vllm.py.
Slightly adopt the function on the new place.

References:
- https://en.wikipedia.org/wiki/Circular_dependency
- https://en.wikipedia.org/wiki/Acyclic_dependencies_principle

### DIFF
--- a/src/instructlab/model/backends/vllm.py
+++ b/src/instructlab/model/backends/vllm.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+from typing import Optional, Tuple
 import json
 import logging
 import os
@@ -15,12 +16,13 @@ import typing
 import httpx
 
 # Local
+from ...client import check_api_base
 from ...configuration import get_api_base
 from .backends import (
     BackendServer,
     Closeable,
     ServerException,
-    ensure_server,
+    free_tcp_ipv4_port,
     safe_close_all,
     shutdown_process,
 )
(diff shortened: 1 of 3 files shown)


leseb:
@Mergifyio rebase

leseb (commented):
Initially the vLLM code was using `ensure_server()`, this might get revisited at some point. So perhaps it should move to `common.py` instead so we don't shuffle this each time we touch code and we can resolve the circular dependency.

makelinux:
> Initially the vLLM code was using `ensure_server()`, this might get revisited at some point. So perhaps it should move to `common.py` instead so we don't shuffle this each time we touch code and we can resolve the circular dependency.

The function has undergone several metamorphoses and no longer meets its original intents. The reusable code already has been split into `check_api_base()`, `free_tcp_ipv4_port()` and `get_api_base()`. The rest of the code it vLLM specific. 

Once moved, the function can be combined with `run_detached()` of vLLM or renamed to `start_vllm()`.

BTW, both `run_detached()` and `run_detached()` have retry loops. This is surprising. It looks redundant. Well, that's already another issue.

danmcp:
> BTW, both `run_detached()` and `run_detached()` have retry loops. This is surprising. It looks redundant. Well, that's already another issue.

Did you mean `run_detached()` and `ensure_server()`?

src/instructlab/model/backends/vllm.py
```diff
@@ -93,6 +95,64 @@ def create_server_process(self, port: int, background: bool) -> subprocess.Popen
...
+
+        logger.info(f"Trying to connect to model server at {self.api_base}")
+        if check_api_base(self.api_base, http_client):
+            return (None, self.api_base)
+        port = free_tcp_ipv4_port(self.host)
+        logger.debug(f"Using available port {port} for temporary model serving.")
+
+        host_port = f"{self.host}:{port}"
+        temp_api_base = get_api_base(host_port)
+        vllm_server_process = None
```
danmcp, line 115: Doesn't look like this is needed anymore

src/instructlab/model/backends/vllm.py
```diff
@@ -93,6 +95,64 @@ def create_server_process(self, port: int, background: bool) -> subprocess.Popen
...
+
+        # TODO: resolve how the hostname is getting passed around the class and this function
```
danmcp, line 117: Is this comment still relevant and in the right place?

danmcp (changes requested):

src/instructlab/model/backends/vllm.py
makelinux: @alimaredia ?

leseb (approved):

PR #2131, opened 2024-08-22
refactor: use global shell var in e2e script for all granite safetensors instances

nathan-weinberg:
Also changes `test_serve` to use local MODEL_PATH var - this makes the test function consistent with evaluate and phased training

**Checklist:**

### DIFF
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -182,7 +182,7 @@ test_serve() {
 
     if [ "${MODEL_TYPE}" == "base" ]; then
         if [ "${BACKEND}" == "vllm" ]; then
-            model="${CACHE_HOME}/instructlab/models/instructlab/granite-7b-lab"
+            model="${CACHE_HOME}/instructlab/models/${GRANITE_SAFETENSOR_REPO}"
         else    # if not using vLLM, use the GGUF for llama-cpp
             model="${CACHE_HOME}/instructlab/models/${GRANITE_GGUF_MODEL}"
         fi
@@ -310,7 +310,7 @@ test_phased_train() {
     DATA_PATH=$(find "${DATA_HOME}"/instructlab/datasets -name 'messages_*' | head -n 1)
 
     local MODEL_PATH
-    MODEL_PATH="${CACHE_HOME}/instructlab/models/instructlab/granite-7b-lab"
+    MODEL_PATH="${CACHE_HOME}/instructlab/models/${GRANITE_SAFETENSOR_REPO}"
 
     # general training args
     local TRAIN_ARGS
@@ -354,7 +354,7 @@ test_convert() {
 }
 
 test_evaluate() {
-    local MODEL_PATH="${CACHE_HOME}/instructlab/models/instructlab/granite-7b-lab"
+    local MODEL_PATH="${CACHE_HOME}/instructlab/models/${GRANITE_SAFETENSOR_REPO}"
     # Temporarily using merlinite as the base model to confirm the workflow executes correctly 
     local BASE_MODEL_PATH="${CACHE_HOME}/instructlab/models/instructlab/merlinite-7b-lab"
     task Evaluate the model
--- a/scripts/basic-workflow-tests.sh
+++ b/scripts/basic-workflow-tests.sh
@@ -177,31 +177,34 @@ test_list() {
 }
 
 test_serve() {
-    MODEL_TYPE="$1"
-    SERVE_ARGS=()
+    local model_type
+    local serve_args
+    local model_path
 
-    if [ "${MODEL_TYPE}" == "base" ]; then
+    model_type="$1"
+    serve_args=()
+    if [ "${model_type}" == "base" ]; then
         if [ "${BACKEND}" == "vllm" ]; then
-            model="${CACHE_HOME}/instructlab/models/${GRANITE_SAFETENSOR_REPO}"
+            model_path="${CACHE_HOME}/instructlab/models/${GRANITE_SAFETENSOR_REPO}"
         else    # if not using vLLM, use the GGUF for llama-cpp
-            model="${CACHE_HOME}/instructlab/models/${GRANITE_GGUF_MODEL}"
+            model_path="${CACHE_HOME}/instructlab/models/${GRANITE_GGUF_MODEL}"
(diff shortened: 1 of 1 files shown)


scripts/basic-workflow-tests.sh
```diff
@@ -179,29 +179,30 @@ test_list() {
 test_serve() {
     MODEL_TYPE="$1"
     SERVE_ARGS=()
+    local MODEL_PATH
```
danmcp, line 182: Would the local be better in lowercase?

JamesKunstle: +1 since it's not a global definition

nathan-weinberg: I pushed up a commit showing what that would look like - can remove it if desired

danmcp: > Not opposed but this would be inconsistent with how serve, evaluate and phased training does it currently - I think if I change it here I should change it everywhere - wdyt?

+1



> I pushed up a commit showing what that would look like

Consistency is probably most important.  Any reason not to change the other locals if we are going to change model_path?

nathan-weinberg: Like do the rest of the locals in this PR? I can for sure, just need to take a closer look at the script

nathan-weinberg: I wonder if we can also enforce this with shellcheck somehow?

scripts/basic-workflow-tests.sh
```diff
@@ -253,48 +257,52 @@ test_taxonomy() {
 
 test_generate() {
     task Generate synthetic data
-    GENERATE_ARGS+=("--endpoint-url" "http://localhost:8000/v1")
+    local generate_args
```
danmcp, line 260: what is the GENERATE_ARGS at the top of the file used for?

scripts/basic-workflow-tests.sh
```diff
@@ -253,48 +257,52 @@ test_taxonomy() {
...
-        # the train profile specified in test_init overrides the majority of TRAIN_ARGS, including things like num_epochs. While it looks like much of those settings are being lost, they just have different values here.
-        TRAIN_ARGS=("--device=cuda" "--model-path=${GRANITE_SAFETENSOR_REPO}" "--data-path=${DATA}" "--lora-quantize-dtype=nf4" "--4-bit-quant" "--effective-batch-size=4" "--is-padding-free=False")
+        # the train profile specified in test_init overrides the majority of train_args, including things like num_epochs. While it looks like much of those settings are being lost, they just have different values here.
+        train_args=("--device=cuda" "--model-path=${GRANITE_SAFETENSOR_REPO}" "--data-path=${data}" "--lora-quantize-dtype=nf4" "--4-bit-quant" "--effective-batch-size=4" "--is-padding-free=False")
         if [ "${BACKEND}" != "vllm" ]; then
-            TRAIN_ARGS+=("--gguf-model-path" "${CACHE_HOME}/instructlab/models/${GRANITE_GGUF_MODEL}")
+            train_args+=("--gguf-model-path" "${CACHE_HOME}/instructlab/models/${GRANITE_GGUF_MODEL}")
         fi
 
         ilab model train "${TRAIN_ARGS[@]}"
```
danmcp, line 297: train_args?

scripts/basic-workflow-tests.sh
```diff
@@ -253,48 +257,52 @@ test_taxonomy() {
...
         if [ "$LEGACYTRAIN" -eq 0 ]; then
-            TRAIN_ARGS+=("--4-bit-quant")
+            train_args+=("--4-bit-quant")
         fi
         if [ "${BACKEND}" != "vllm" ]; then
-            TRAIN_ARGS+=("--gguf-model-path" "${CACHE_HOME}/instructlab/models/${GRANITE_GGUF_MODEL}")
+            train_args+=("--gguf-model-path" "${CACHE_HOME}/instructlab/models/${GRANITE_GGUF_MODEL}")
         fi
 
         ilab model train "${TRAIN_ARGS[@]}"
```
danmcp, line 308: train_args?

Same issue below.

scripts/basic-workflow-tests.sh
```diff
@@ -253,48 +257,52 @@ test_taxonomy() {
...
     else
         # TODO Only cuda for now
-        TRAIN_ARGS+=("--legacy" "--device=cuda")
+        train_args=("--legacy" "--device=cuda")
```
danmcp, line 300: This was appended to the global TRAIN_ARGS which is still there.  It's also appended to in set_defaults.  Maybe this one shouldn't switch to a local?

nathan-weinberg: Yeah you're right, I'm going to leave this one be

scripts/basic-workflow-tests.sh
nathan-weinberg: I think it's the same as `TRAIN_ARGS` where it's being used as a base global arg - going to go the same route and leave this be for now

scripts/basic-workflow-tests.sh
```diff
@@ -304,18 +310,18 @@ test_train() {
...
+    local data_path
+    data_path=$(find "${DATA_HOME}"/instructlab/datasets -name 'messages_*' | head -n 1)
 
-    local MODEL_PATH
-    MODEL_PATH="${CACHE_HOME}/instructlab/models/instructlab/granite-7b-lab"
+    local model_path
+    model_path="${CACHE_HOME}/instructlab/models/${GRANITE_SAFETENSOR_REPO}"
 
     # general training args
     local TRAIN_ARGS
```
danmcp, line 322: This seems like it's a separate TRAIN_ARGS from the global one?  Might help for it to be renamed to train_args to differentiate.

nathan-weinberg: Yeah raised that here https://github.com/instructlab/instructlab/pull/2131#discussion_r1731540461, waiting to hear from @JamesKunstle

nathan-weinberg: But I agree

danmcp (approved):

JamesKunstle (approved):
if this test runs as expected it's great w/ me. I definitely prefer using these local variables rather than global

PR #2152, opened 2024-08-26
enable nvidia-toolkit-setup.service

sallyom:
I noticed after rebooting my instance after running `nvidia-setup.sh`, the nvidia-toolkit-setup.service was not running, which made the command
`python -m vllm.entrypoints.openai.api_server --model instructlab/granite-7b-lab --tensor-parallel-size 1` fail. 
Any reason not to enable the service?


leseb (approved):
Feels like this should have been there :), thanks!

danmcp (approved):

PR #2162, opened 2024-08-28
refactor: move Closeable and ServerException to common

makelinux:

Move Closeable and ServerException to reduce the circular dependencies.

Current state of [Circular dependency](https://en.wikipedia.org/wiki/Circular_dependency):
```mermaid
graph LR;
  backends --> llama_cpp
  llama_cpp --> backends
  backends --> vllm
  vllm --> backends
```
Desired state of [Acyclic dependencies](https://en.wikipedia.org/wiki/Acyclic_dependencies_principle):

```mermaid
graph LR;
  backends --> llama_cpp
  llama_cpp -->common
  backends --> common
  backends --> vllm
  vllm --> common
```

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -28,19 +28,11 @@ import uvicorn
 # Local
 from ...configuration import _serve as serve_config
 from ...utils import split_hostport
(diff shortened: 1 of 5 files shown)


danmcp:
@mergifyio rebase

PR #2168, opened 2024-08-29
fix: remove signatures while copying OCI artifacts

jaideepr97:
This PR supplies the `--remove-signatures` flag to skopeo by default when copying OCI artifacts during model download. Without doing this model download from OCI registries will fail if a sigstore signature is detected because signature storage is not supported in the OCI folder that is created in the interim 

**Checklist:**


bbrowning:
Before your PR, I run `ilab model download` against an OCI registry that uses signatures and hit a fatal error trying to push signatures to the destination OCI path. After your PR, the model downloads successfully for me with no signature error.

It's just one data point and a manual test on my machine, but this appears to work as intended.

nathan-weinberg (approved):

nathan-weinberg:
@mergifyio backport release-v0.18

dhellmann:
Are there other intermediate formats that would support the signature? If so, would we switch to one of those for a long-term fix?

jaideepr97:
> Are there other intermediate formats that would support the signature? If so, would we switch to one of those for a long-term fix?

Good question, I am unsure myself
in the long run however, we have discussed implementing a custom client to do this rather than relying on skopeo so we may have more flexibility to do that if that comes to pass

bbrowning:
Yes, the `dir` destination supports signatures, so we could do something like
```
skopeo copy docker://my.registry.host/path:1.2.3 dir:bar
Getting image source signatures
Checking if image destination supports signatures
Copying blob 1c57ec4b54dj done   | 
Copying config 33045af216 done   | 
Writing manifest to image destination
Storing signatures
```

You then end up with files on disk, just not in the oci repository format. But there's still a manifest.json we can parse to map the blobs on disk to their contents.

dhellmann:
> Yes, the `dir` destination supports signatures, so we could do something like
> 
> ```
> skopeo copy docker://my.registry.host/path:1.2.3 dir:bar
> Getting image source signatures
> Checking if image destination supports signatures
> Copying blob 1c57ec4b54dj done   | 
> Copying config 33045af216 done   | 
> Writing manifest to image destination
> Storing signatures
> ```
> 
> You then end up with files on disk, just not in the oci repository format. But there's still a manifest.json we can parse to map the blobs on disk to their contents.

That seems better than writing a custom client that would have to be maintained.

n1hility (approved):

n1hility:
@dhellmann @bbrowning long term we definitely want to have direct oci artifact fetching capability in one of the container tools. which we can then reuse instead. since we post process the stored  fs layout we dont do anything with the signatures anyway so not too much to be gained by switching formats.

Issue #2171, opened 2024-08-29
Extend free disk space action to be compatable with MacOS


nathan-weinberg:
We currently have a GitHub Action that frees disk space on our CI runners: https://github.com/instructlab/instructlab/blob/main/.github/actions/free-disk-space/action.yml

However, currently this only works for our Linux runners and not Mac OS. We've recently see CI failures due to disk space issues on the Mac OS runners.

As such, we should extend this action to support Mac OS as well.

PR #2173, opened 2024-08-29
chore: delete check that platform is not macos

khaledsulayman:
Upon resolving #2171, the 'Free disk space' GitHub action will be compatible with MacOS. When that is ready, we should remove this check so that it runs on all platforms

**Checklist:**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,8 +64,6 @@ jobs:
           fetch-depth: 0
 
       - name: Free disk space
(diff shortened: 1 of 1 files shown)


danmcp:
Can we close this in favor of https://github.com/instructlab/instructlab/pull/2177 ?

@khaledsulayman I added you as a co-author on that commit.

khaledsulayman:
@danmcp sure, sounds good! I originally put this up and set it to hold so we remember to delete that line once we add freeing disk space for MacOS, but I like your solution for now and it makes it easy for me to pickup from there. Thanks!

PR #2177, opened 2024-08-31
ci: Add check for disk space on mac to debug when runner is out of space

danmcp:
Related to #2171 

Even though there isn't cleanup to be done on the mac runners, we can at least print out the disk info for debug purposes.
This change includes a global df as well as checking where $TMPDIR (and the corresponding runner sandbox) is located.

Output on mac runner:
```
Prepare all required actions
Run ./.github/actions/free-disk-space
Run df -h
Filesystem        Size    Used   Avail Capacity iused ifree %iused  Mounted on
/dev/disk3s1s1   [2](https://github.com/instructlab/instructlab/actions/runs/10648266505/job/29517133109?pr=2177#step:5:2)95Gi   9.6Gi    19Gi    35%    404k  194M    0%   /
devfs            19[5](https://github.com/instructlab/instructlab/actions/runs/10648266505/job/29517133109?pr=2177#step:5:6)Ki   195Ki     0Bi   100%     674     0  100%   /dev
/dev/disk3s6     295Gi    20Ki    19Gi     1%       0  194M    0%   /System/Volumes/VM
/dev/disk3s2     295Gi   4.9Gi    19Gi    22%     [15](https://github.com/instructlab/instructlab/actions/runs/10648266505/job/29517133109?pr=2177#step:5:17)3  194M    0%   /System/Volumes/Preboot
/dev/disk3s4     295Gi   288Ki    19Gi     1%      23  194M    0%   /System/Volumes/Update
/dev/disk1s2     500Mi    20Ki   495Mi     1%       0  5.1M    0%   /System/Volumes/xarts
/dev/disk1s1     500Mi    96Ki   495Mi     1%      20  5.1M    0%   /System/Volumes/iSCPreboot
/dev/disk1s3     500Mi   196Ki   495Mi     1%      [17](https://github.com/instructlab/instructlab/actions/runs/10648266505/job/29517133109?pr=2177#step:5:19)  5.1M    0%   /System/Volumes/Hardware
/dev/disk3s5     295Gi   261Gi    19Gi    94%    2.2M  194M    1%   /System/Volumes/Data
map auto_home      0Bi     0Bi     0Bi   100%       0     0     -   /System/Volumes/Data/home
df -h for /var/folders/m4/5dz5h26x329cqq4fx333f8gm0000gn/T/
Filesystem      Size    Used   Avail Capacity iused ifree %iused  Mounted on
[...]

PR #2178, opened 2024-08-31
ci: Remove oci download from mac runner

### LINKED ISSUE #2171: Extend free disk space action to be compatable with MacOS

danmcp:
This change removes the OCI download test from the mac runner test path since space is limited.  And since the mac runners are running inside a sandbox, there isn't much we can do to cleanup the existing space.

The other change we could consider making would be to upload a much smaller model that would save on download time/cost and still leave enough space to test on mac runners.

Resolves #2171 with #2177

**Checklist:**

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -8,7 +8,33 @@
 # As soon as TEST_DIR is set to any non-empty value, the test directory will NOT be removed
 # This assumes you control your test directory TEST_DIR and its cleanup
 
-set -ex
+set -e
+
+SKIP_MAC_OCI_DOWNLOAD=0
+
+usage() {
+    echo "Usage: $0"
+    echo "  -s  skip oci download for mac"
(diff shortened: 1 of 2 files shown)


khaledsulayman (approved):

leseb (changes requested):
I have an alternative approach here https://github.com/instructlab/instructlab/pull/2179, which free up the macOS runner and print the stderr on subprocess error. PTAL.

danmcp:
Closing in favor of https://github.com/instructlab/instructlab/pull/2179

PR #2180, opened 2024-09-02
feat: better troubleshooting on internal server error

leseb:
We try to surface as much information as we can using DEBUG level logging when the temporary server crashes. This is important when running the chat without a server or a generation.

**Checklist:**

### DIFF
--- a/src/instructlab/model/chat.py
+++ b/src/instructlab/model/chat.py
@@ -7,6 +7,7 @@ import logging
 import os
 import sys
 import time
+import traceback
 
 # Third Party
 from openai import OpenAI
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):

src/instructlab/model/chat.py
```diff
@@ -667,16 +668,23 @@ def start_prompt(
...
             raise KeyboardInterrupt from e
         except KeyboardInterrupt as e:
             raise e
         except httpx.RemoteProtocolError as e:
+            logger.debug("Connection to the server was closed", exc_info=True)
             self.console.print("Connection to the server was closed", style="bold red")
             self.info["messages"].pop()
             raise ChatException("Connection to the server was closed") from e
         except Exception as e:
+            if logger.getEffectiveLevel() == logging.DEBUG:
```
danmcp, line 683: Suggestion: Would it be better to use the if to form a different message but still raise in the exception in the same place?  I would worry the code paths might accidentally diverge and change behavior (not just messages) when debug is enabled.

danmcp (approved):

PR #2183, opened 2024-09-03
ci: add AWS tag to show github ref for all jobs

nathan-weinberg:
right now we can see what type of runner and which repository an AWS runner is associated with, but not which commit triggered it

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x1.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x1.yml
@@ -38,7 +38,8 @@ jobs:
           aws-resource-tags: >
             [
               {"Key": "Name", "Value": "instructlab-ci-github-medium-runner"},
-              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"}
+              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"},
+              {"Key": "GitHubRef", "Value": "${{ github.ref }}"}
             ]
 
   e2e:
--- a/.github/workflows/e2e-nvidia-a10g-x4.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x4.yml
@@ -40,7 +40,8 @@ jobs:
           aws-resource-tags: >
             [
               {"Key": "Name", "Value": "instructlab-ci-github-large-runner"},
-              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"}
+              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"},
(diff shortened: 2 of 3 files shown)


.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -38,7 +38,8 @@ jobs:
           aws-resource-tags: >
             [
               {"Key": "Name", "Value": "instructlab-ci-github-medium-runner"},
-              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"}
+              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"},
+              {"Key": "GitHubRef", "Value": "${{ github.ref }}"}
```
danmcp, line 42: I assume this is the commit ref?  Is the PR# avail?  That could be really helpful.

nathan-weinberg: PR number was my original idea, but I wasn't sure if that would be a problem for the E2Es running against non-PR events (e.g. merges to `main`) so I figured this would allow us to see which jobs were running in both contexts

I was using this as a reference: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs

Definitely open to updating this with other info!

danmcp: Found the doc: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

Looks like it might already have the pr in there.

danmcp: Sorry, crossed paths (I wish github would show comment updates live).  It looks like ref is as good as we can get.

danmcp (approved):

.github/workflows/e2e-nvidia-a10g-x1.yml
russellb: I think this may only work how you expect for the `pull_request` event, not `pull_request_target`. My guess is that `github.ref` is just going to show `refs/head/main` in most cases here. You may want to consider using:

* `github.event.pull_request.number`
* `github.event.pull_request.head.sha`

cdoern (approved):

.github/workflows/e2e-nvidia-a10g-x1.yml
nathan-weinberg: @russellb i'll checkout the instances and will update accordingly, appreciate it!

PR #2201, opened 2024-09-04
feat: use default and description from Pydantic model

leseb:
Our Pydantic model, referred to internally as Config, serves as the source of truth for default values and descriptions. When a Click CLI option is also present in the Config, it should inherit its default value and description from the corresponding Config field. When applying the cls=clickext.ConfigOption configuration to a Click option, fields such as “default” and “help” should not be specified, as these will automatically be derived from the Config object.

So this is how you should write an option that exists in the Config (ultimately in config.yaml) AND has a CLI option:

```
my_config_option: StrictStr = Field(
    default="my_default_value",
    description="Great description of what the option does",
)

@click.option(
    "--my-config-option",
    type=click.STRING,
    cls=clickext.ConfigOption,
)
```

That's all. Later we could try to do the same for the type, so we don't need `type=click.STRING`. It's a bit more tedious and is a larger change.

**Checklist:**

### DIFF
--- a/src/instructlab/clickext.py
+++ b/src/instructlab/clickext.py
@@ -16,7 +16,7 @@ from click_didyoumean import DYMGroup
 import click
 
 # First Party
-from instructlab.configuration import DEFAULTS, storage_dirs_exist
+from instructlab.configuration import DEFAULTS, BaseModel, storage_dirs_exist
 
 logger = logging.getLogger(__name__)
 
@@ -111,11 +111,11 @@ class ExpandAliasesGroup(LazyEntryPointGroup):
 class ConfigOption(click.Option):
     """A click.Option with extended default lookup help
 
-    The defaults can be looked up in a subsection of the context's
-    default_map.
+    The defaults and description can be looked up in a subsection of the context's
+    config object.
 
     The help output of ConfigOption class reads defaults from the context's
-    default_map and shows the name of the config settings.
+    config object field and shows the name of the config settings.
 
     Example:
         Options:
@@ -133,10 +133,14 @@ class ConfigOption(click.Option):
         super().__init__(*args, **kwargs)
         if self.show_default:
             raise ValueError("show_default must be False")
+        if self.help:
+            raise ValueError(
+                f"help must not be set for '{self.name}', it is derived from the Field description in the Config class"
+            )
         if config_sections:
-            self.config_sections = tuple(config_sections.split("."))
+            self.config_sections = list(config_sections.split("."))
         else:
-            self.config_sections = ()
+            self.config_sections = []
 
     def get_help_record(self, ctx: click.Context) -> tuple[str, str] | None:
         result = super().get_help_record(ctx)
@@ -145,25 +149,29 @@ class ConfigOption(click.Option):
             # missing default map (clickman, sphinx)
             return result
 
-        # get default from default_map
         cmd = ctx.command.name
         name = self.name
+
         if self.config_sections:
-            config_name = f"{cmd}.{'.'.join(self.config_sections)}.{name}"
+            config_identifier = (
+                [str(cmd)]
+                + [
+                    str(section)
+                    for section in self.config_sections
+                    if section is not None
+                ]
+                + [str(name)]
+            )
         else:
-            config_name = f"{cmd}.{name}"
+            config_identifier = [str(cmd), str(name)]
 
         if typing.TYPE_CHECKING:
             assert name is not None
 
-        section = ctx.default_map
-        for secname in self.config_sections:
-            section = section.get(secname, {})
-        if self.name not in section:
-            raise ValueError(f"{config_name} not in default_map {ctx.default_map}")
-
         # create help extra
-        default_value = self.get_default(ctx)
+        description, default_value = get_default_and_description(
+            ctx.obj.config, config_identifier
+        )
         if default_value is None:
             default_string = "<None>"
         elif isinstance(default_value, (list, tuple)):
@@ -171,14 +179,23 @@ class ConfigOption(click.Option):
         else:
             default_string = str(default_value)
 
-        default_msg = f"default: {default_string}; config: '{config_name}'"
+        default_msg = (
+            f"default: {default_string}; config: '{'.'.join(config_identifier)}'"
+        )
 
         # extend message
         prefix, msg = result
-        if msg.endswith("]"):
-            msg = f"{msg[:-1]}; {default_msg}]"
+        # If the 'help' field on click.Option is not set, then 'msg' will be None
+        # unless options like 'required' are set
+        if msg:
+            # Needed to handle [required] option for example
+            if msg.endswith("]"):
+                msg = f"{description} {msg[:-1]}; {default_msg}]"
+            else:
+                msg = f"{description} {msg} [{default_msg}]"
         else:
-            msg = f"{msg}  [{default_msg}]"
+            # do not append description if msg is None to avoid space
+            msg = f"{description} [{default_msg}]"  # type: ignore
         return prefix, msg
 
     def consume_value(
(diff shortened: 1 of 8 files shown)


src/instructlab/clickext.py
```diff
@@ -133,6 +133,10 @@ def __init__(
         super().__init__(*args, **kwargs)
         if self.show_default:
             raise ValueError("show_default must be False")
+        if self.help:
+            raise ValueError(
+                "help must not be set, it is derived from the Field description in the Config class"
+            )
```
jaideepr97, line 139: not quite sure I'm following this - are we exposing this error to the user here, when the cause of error might be an action taken by a developer of the code elsewhere in the codebase? 
i think this could be confusing to users

src/instructlab/clickext.py
```diff
@@ -175,10 +174,17 @@ def get_help_record(self, ctx: click.Context) -> tuple[str, str] | None:
 
         # extend message
         prefix, msg = result
-        if msg.endswith("]"):
-            msg = f"{msg[:-1]}; {default_msg}]"
+        # If the 'help' field on click.Option is not set, then 'msg' will be None
+        # unless options like 'required' are set
+        if msg:
+            # Needed to handle [required] option for example
+            if msg.endswith("]"):
```
jaideepr97, line 181: could we use regex here instead so we definitely only catch the `[required]` case? that would be a bit more future proof IMO

src/instructlab/clickext.py
```diff
@@ -321,3 +327,43 @@ def wrapper(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
...
+    Returns:
+        typing.Tuple[str | None, typing.Any]: A tuple containing the field's description
+        and default value. If the field is not found, a ValueError is raised.
+
+    Raises:
+        ValueError: If the specified config_name is not found in the model.
+    """
+    # Loop through the fields of the model
+    for field_name, field in cfg.model_fields.items():
+        if field_name == config_name[0]:
```
jaideepr97, line 356: hmmm, if we break on the first instance of finding the config name, any chance we catch the wrong field here if there happen to be multiple fields with the same name but serving different purposes? 
Like `chat.model` and `evalulate.model` - both contain field names as "model" but they serve different purposes and should have different descriptions

jaideepr97 (commented):
Thanks @leseb! I like the direction of this PR 
Just a couple questions from my side

src/instructlab/clickext.py
leseb: It's not about catching `[required]`, I added it as an example of when this would happen. It's anything that ends with closing bracket really.

src/instructlab/clickext.py
leseb: If `help="text"` is set in `click.Option` that uses `cls=clickext.ConfigOption,`, this will fail when the command is instantiated. So it's impossible to miss it. The CI will report that error. Does that help?

leseb: Most of the unit tests cover that naturally as well.

src/instructlab/clickext.py
leseb: If this was not working, the current `default_config.yaml` would have been modified with incorrect descriptions :).
Imagine `config_name` is `chat.model`, once you find `chat`, you go down one level and the `model` you get is the once from the `_chat` BaseModel.

Hope that clarifies. Thanks!

jaideepr97: actually I just noticed that `ilab model serve -h` shows the following for `--model-path`:

```
(venv) ➜  instructlab git:(0e0417b) ilab model serve --help 
Usage: ilab model serve [OPTIONS]

  Starts a local server

  The vLLM backend accepts additional parameters in the form of extra
  arguments after "--" separator:

    $ ilab model serve ... --backend=vllm -- --dtype=auto --enable-lora

  vLLm parameters are documented at
  https://docs.vllm.ai/en/stable/serving/openai_compatible_server.html

Options:
  --model-path PATH           Path to the model used during generation.
                              [required; default: /Users/jrao/Library/Caches/i
                              nstructlab/models/instructlab/merlinite-7b-lab-
                              GGUF/main/merlinite-7b-lab-Q4_K_M.gguf; config:
                              'serve.model_path']
```

whereas the `model-path` description in the default config's `serve.model_path` and in `configuration.py` is 
```
  # Directory where model to be served is stored.
  # Default: /cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
  model_path: /cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
```

src/instructlab/clickext.py
leseb: The [required] word comes from click when `required=True` is set in click.Option, like so:

```
@click.option(
    "-m",
    "--model",
    cls=clickext.ConfigOption,
)
```

So you won't find this in `configuration.py`. It's generated

Be reminded that the ConfigOption class in `src/instructlab/clickext.py` is a custom extension of the [click.Option](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) class from the Click library. It provides additional functionality for handling default values and help messages based on a configuration object.

Does that help?

src/instructlab/clickext.py
leseb: > > If this was not working, the current default_config.yaml would have been modified with incorrect descriptions :).
> 
> does this code influence what goes into the config yaml though? my understanding was that the correct way to verify this is working is by making sure the help text in the command line is properly reflected, and that the config yaml comments are coming straight from the config class descriptions anyway

My bad, you're right! The way to verify this is indeed by looking at the help text.

leseb: > I tested this and noticed that `ilab model serve -h` shows the following for `--model-path`:
> 
> ```
> (venv) ➜  instructlab git:(0e0417b) ilab model serve --help 
> Usage: ilab model serve [OPTIONS]
> 
>   Starts a local server
> 
>   The vLLM backend accepts additional parameters in the form of extra
>   arguments after "--" separator:
> 
>     $ ilab model serve ... --backend=vllm -- --dtype=auto --enable-lora
> 
>   vLLm parameters are documented at
>   https://docs.vllm.ai/en/stable/serving/openai_compatible_server.html
> 
> Options:
>   --model-path PATH           Path to the model used during generation.
>                               [required; default: /Users/jrao/Library/Caches/i
>                               nstructlab/models/instructlab/merlinite-7b-lab-
>                               GGUF/main/merlinite-7b-lab-Q4_K_M.gguf; config:
>                               'serve.model_path']
> ```
> 
> whereas the `model-path` description in the default config's `serve.model_path` and in `configuration.py` is
> 
> ```
>   # Directory where model to be served is stored.
>   # Default: /cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
>   model_path: /cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
> ```

It's interesting, that's not the behavior I'm observing on my machine:

```
Usage: ilab model serve [OPTIONS]

  Starts a local server

  The vLLM backend accepts additional parameters in the form of extra
  arguments after "--" separator:

    $ ilab model serve ... --backend=vllm -- --dtype=auto --enable-lora

  vLLm parameters are documented at
  https://docs.vllm.ai/en/stable/serving/openai_compatible_server.html

Options:
  --model-path PATH           Directory where model to be served is stored.
                              [required; default: /Users/leseb/Library/Caches/
                              instructlab/models/merlinite-7b-lab-Q4_K_M.gguf;
                              config: 'serve.model_path']
```

Let me double-check.

src/instructlab/clickext.py
jaideepr97: yup, thanks for the context!
endswith "]" just feels like a weak check and a bit hacky to me personally
if there is a closing bracket, there must be an opening bracket as well, would it be better to have a regex that checks for `\[[^\[\]]*\]` for example?

src/instructlab/clickext.py
jaideepr97: indeed 
I just tried it again - checked out the same commit and ran the same command but this time it is displaying the correct thing for me 

```
(venv) ➜  instructlab git:(0e0417b) ilab model serve --help
Usage: ilab model serve [OPTIONS]

  Starts a local server

  The vLLM backend accepts additional parameters in the form of extra
  arguments after "--" separator:

    $ ilab model serve ... --backend=vllm -- --dtype=auto --enable-lora

  vLLm parameters are documented at
  https://docs.vllm.ai/en/stable/serving/openai_compatible_server.html

Options:
  --model-path PATH           Directory where model to be served is stored.
                              [required; default: /Users/jrao/Library/Caches/i
                              nstructlab/models/merlinite-7b-lab-Q4_K_M.gguf;
                              config: 'serve.model_path']
```

src/instructlab/clickext.py
leseb: I understand the concern, but this code was present before, I just changed the indent with the new condition :). Do you mind if we address this separately?

jaideepr97: fair enough

jaideepr97 (approved):

src/instructlab/clickext.py
```diff
@@ -145,40 +149,51 @@ def get_help_record(self, ctx: click.Context) -> tuple[str, str] | None:
             # missing default map (clickman, sphinx)
             return result
 
-        # get default from default_map
         cmd = ctx.command.name
         name = self.name
+
         if self.config_sections:
-            config_name = f"{cmd}.{'.'.join(self.config_sections)}.{name}"
+            config_name = (
```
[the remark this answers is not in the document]
leseb: done.

danmcp (approved):
LGTM.  Would appreciate more eyeballs on this PR at least for awareness.

PR #2202, opened 2024-09-05
chore: replace platformdirs with xdg-base-dirs 

### LINKED ISSUE #1747: Use of `~/Library/Application Support` on Mac is user-unfriendly
**Describe the bug**

**To Reproduce**
With the changes in #1471, `platformdirs` is used to identify platform specific folders to use. On Mac this results in putting files in `~/Library/Application Support/instructlab`. This includes the models folder, the datasets (formerly generated) folder, and the taxonomy folder. 

On Mac, the `~/Library` folder is hidden in Finder by default. (See <https://www.macworld.com/article/671158/how-to-show-hidden-files-on-a-mac.html>). Which means users won't be able to easily navigate to the `~/Library/Application Support/instructlab` folder in finder to see the files in the their taxonomy or the generated training data.

It is also an awkward file path to type in terminal given the space in the folder name. One must escape or quote.

The support for Mac also does not respect the XDG_ env vars if set. More sophisticated developers will use XDG_ and would expect it to be honored.

`~/Library/Application Support` is for files the user will generally never have to directly touch or see. But there are many files in InstructLab (e.g. taxonomy repo) which we do expect to user to see and touch.

**Expected behavior**

I would expect that the folders are in a place the user can easily access from both Finder and terminal. 

Respect XDG_ env vars if set.

**Screenshots**

 - Hardware Specs: M3 Max, 64GB
```sh
$ ilab system info
sys.version: 3.11.9 | packaged by conda-forge | (main, Apr 19 2024, 18:34:54) [Clang 16.0.6 ]
sys.platform: darwin
os.name: posix
platform.release: 23.5.0
platform.machine: arm64
instructlab.version: 0.17.1.dev494
instructlab-dolomite.version: 0.0.1
instructlab-eval.version: 0.0.9
instructlab-quantize.version: 0.1.0
instructlab-schema.version: 0.2.0
instructlab-sdg.version: 0.1.2
instructlab-training.version: 0.0.5.1
torch.version: 2.3.1
torch.version.cuda: None
torch.version.hip: None
torch.cuda.available: False
llama_cpp_python.version: 0.2.79
```
**Additional context**

jaideepr97:
Currently we use the `platformdirs` package to automatically pick the platform specific system directories that are used by instructlab. While this works fine on linux with XDG dirs, On macOS this means usage of `/Users/<user>/Library/Application Support/...` which is awkward and hard to use
This PR replaces `platformdirs` with `xdg-base-dirs` to standardize the dirs across linux and macOS

acc to https://pypi.org/project/xdg-base-dirs/ `xdg-base-dirs` functions take XDG env vars into account so we get env var support for free with this package 

There is a consideration to be made that we are making a change to the project's dependencies by making this switch so that should be taken into account 

Resolves #1747 

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## v0.19
+
+### Breaking Changes
+
+* InstructLab now uses XDG-based directories on macOS, similar to Linux.
+  Users are advised to re-initialize their config files and remove cached models.
+
 ## v0.18.1
 
 ### Features
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,7 +20,6 @@ numpy>=1.23.5,<2.0.0 ; python_version == '3.10'
 numpy>=1.26.4,<2.0.0 ; python_version >= '3.11'
 openai>=1.13.3
 peft>=0.9.0
-platformdirs>=4.2
 prompt-toolkit>=3.0.38
 pydantic>=2.7.4
 pydantic_yaml>=1.2.0
@@ -41,3 +40,4 @@ transformers>=4.40.0 ; python_version == '3.10'
 transformers>=4.41.2 ; python_version >= '3.11'
 trl>=0.9.4
 wandb>=0.16.4
+xdg-base-dirs>=6.0.1
--- a/README.md
+++ b/README.md
@@ -332,8 +332,8 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
    ```shell
    Welcome to InstructLab CLI. This guide will help you to setup your environment.
    Please provide the following values to initiate the environment [press Enter for defaults]:
-   Path to taxonomy repo [/home/user/.local/share/instructlab/taxonomy]: 
-   Path to your model [/home/user/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf]: 
+   Path to taxonomy repo [/home/user/.local/share/instructlab/taxonomy]:
+   Path to your model [/home/user/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf]:
    Generating `/home/user/.config/instructlab/config.yaml`...
    Please choose a train profile to use:
    [0] No profile (CPU-only)
@@ -343,7 +343,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
    [4] L40_x4.yaml
    [5] L40_x8.yaml
    [6] L4_x8.yaml
-   Enter the number of your choice [hit enter for the default CPU-only profile] [0]:  
+   Enter the number of your choice [hit enter for the default CPU-only profile] [0]:
    Using default CPU-only train profile.
    Initialization completed successfully, you're ready to start using `ilab`. Enjoy!
    ```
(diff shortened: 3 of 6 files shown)


jaideepr97:
Not sure if this will cause problems on windows, but we are not really actively focussed on windows at the moment

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## v0.19
+
+### Fixes
+
+* instructLab now uses XDG based directories on macOS, similar to Linux.
```
leseb, line 5: We've intentionally avoided using a `Fixes` section to keep the CHANGELOG focused solely on major changes. This will included in the list of fix commits but not a notable change. Otherwise, could you use the Feature section? Even though it's technically not a feature.

Also, this is somehow a breaking change existing users. We need a note to remove the previous config and cached models too.

nathan-weinberg:
> Not sure if this will cause problems on windows, but we are not really actively focussed on windows at the moment

We don't support Windows at all AFAIK

requirements.txt
```diff
@@ -41,3 +40,4 @@ transformers>=4.40.0 ; python_version == '3.10'
 transformers>=4.41.2 ; python_version >= '3.11'
 trl>=0.9.4
 wandb>=0.16.4
+xdg-base-dirs>=6.0.1
```
leseb, line 43: Was the line last removed too?

leseb (commented):
Happy to see this change, a few nits

requirements.txt
jaideepr97: added a new line 
thanks

CHANGELOG.md
jaideepr97: updated to breaking changes section and added the note
thanks!

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## v0.19
+
+### Breaking Changes
+
+* instructLab now uses XDG based directories on macOS, similar to Linux.
```
leseb, line 5:
```suggestion
* InstructLab now uses XDG-based directories on macOS, similar to Linux.
```

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+  Users are advised to re-initialize their config files and cached models.
```
leseb, line 6:
```suggestion
  Users are advised to re-initialize their config files and remove cached models.
```

cdoern (approved):
lgtm

bjhargrave (approved):
I assume we will need similar PRs for the other projects like sdg?

jaideepr97:
> I assume we will need similar PRs for the other projects like sdg?

oh yeah, hadn't thought of that - good point
sdg does seem to also be using platformdirs - although IMO the library shouldn't be doing any FS manipulation, all the FS interaction should ideally be happening in the CLI

leseb (approved):

danmcp (approved):

nathan-weinberg:
> > I assume we will need similar PRs for the other projects like sdg?
> 
> oh yeah, hadn't thought of that - good point sdg does seem to also be using platformdirs - although IMO the library shouldn't be doing any FS manipulation, all the FS interaction should ideally be happening in the CLI

Agree very much with that

nathan-weinberg (approved):

PR #2212, opened 2024-09-06
refactor: move free_tcp_ipv4_port() to common

makelinux:

Move free_tcp_ipv4_port() to reduce the circular dependencies.

Current state of [Circular dependency](https://en.wikipedia.org/wiki/Circular_dependency):
```mermaid
graph LR;
  subgraph backends
    x[free_tcp_ipv4_port]
    style x opacity:0
    select_backend
    style select_backend opacity:0
  end
  backends --> llama_cpp & vllm
  linkStyle 0,1 opacity:0
  select_backend --> llama_cpp & vllm --> x
 ```
Desired state of [Acyclic dependencies](https://en.wikipedia.org/wiki/Acyclic_dependencies_principle):

```mermaid
graph LR;
  subgraph backends
    a[ ]
    style a opacity:0
    select_backend
    style select_backend opacity:0
  end
  backends --> llama_cpp & vllm
  linkStyle 0,1 opacity:0
  select_backend --> llama_cpp & vllm -->x
  subgraph common
    x[free_tcp_ipv4_port]
    style x opacity:0
  end
```

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -12,7 +12,6 @@ import multiprocessing
 import os
 import pathlib
 import signal
-import socket
 import struct
 import subprocess
 import sys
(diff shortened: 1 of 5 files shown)


danmcp (commented):
The changes seem fine but it's not clear how this is fixing a circular dependency.  I see free_tcp_ipv4_port defined in backends and referenced from vllm and llamacpp.

makelinux:
@danmcp , I've updated the description. The right dependency is **backends** -> **vllm**,  **llamacpp**.

danmcp (approved):

nathan-weinberg (approved):
Thanks!

Issue #2109, opened 2024-08-20
Remove some deprecated aliased commands


nathan-weinberg:
**What is the current behavior of feature you are looking to change?**
As previously agreed upon, the aliased commands that predated the model engine redesign started in https://github.com/instructlab/instructlab/pull/990 are meant to go away as part of `0.19.0`.

**What is the proposed behavior of feature you are looking to change?**
Aliased commands will no longer work and the new commands must be used.

**What is the benefit of the design change you are proposing?**
The commands are simple aliases introduced to ease the transition from the old commands - but we don't want to maintain these indefinitely. The new ones being the only ones are better from both a UX and maintainability POV.

**If applicable, how would backwards-compatability be handled?**
[We called this out as part of the `0.17.0` release](https://github.com/instructlab/instructlab?tab=readme-ov-file#-whats-new), so backwards-compatability is not a concern here.

**Additional context**
Related to https://github.com/instructlab/instructlab/issues/1419

PR #2214, opened 2024-09-06
Deprecate aliased commands

alinaryan:
This change removes a majority of the aliased commands, except `chat`,
`generate`, `serve`, and `train`. When a user attempts to use an alias
that was removed, it errors out. If a user types an alias that was kept,
it proceeds with the command.
Also updates the relevant test files.

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@
 * InstructLab now uses XDG-based directories on macOS, similar to Linux.
   Users are advised to re-initialize their config files and remove cached models.
 * Removed unused argument `--rouge-threshold` of `ilab data generate`
+* Removed the following aliased commands:
+  * `convert`
+  * `diff`
+  * `download`
+  * `evaluate`
+  * `init`
+  * `list`
+  * `sysinfo`
+  * `test`
 
 ## v0.18.1
 
--- a/src/instructlab/clickext.py
+++ b/src/instructlab/clickext.py
@@ -73,17 +73,10 @@ class ExpandAliasesGroup(LazyEntryPointGroup):
             cmd = self.alias_eps[cmd_name].load()
             if typing.TYPE_CHECKING:
                 assert isinstance(cmd, click.Command)
-            group, primary = self.get_alias_info(cmd_name)
-            click.echo(
-                "You are using an aliased command, this will be deprecated "
-                "in a future release. Please consider using "
-                f"`ilab {group} {primary}` instead"
-            )
             # if some storage dirs do not exist
-            # AND the command is not `ilab init`
             # AND the --config flag is not customized, then we error
             if not storage_dirs_exist() and (
-                primary != "init" and ctx.params["config_file"] == DEFAULTS.CONFIG_FILE
+                ctx.params["config_file"] == DEFAULTS.CONFIG_FILE
             ):
                 click.secho(
                     "Some ilab storage directories do not exist yet. Please run `ilab config init` before continuing.",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,17 +70,9 @@ issues = "https://github.com/instructlab/instructlab/issues"
 
 [project.entry-points."instructlab.command.alias"]
 "chat" = "instructlab.model.chat:chat"
-"convert" = "instructlab.model.convert:convert"
-"diff" = "instructlab.taxonomy.diff:diff"
-"download" = "instructlab.model.download:download"
-"evaluate" = "instructlab.model.evaluate:evaluate"
 "generate" = "instructlab.data.generate:generate"
-"init" = "instructlab.config.init:init"
 "serve" = "instructlab.model.serve:serve"
-"sysinfo" = "instructlab.system.info:info"
-"test" = "instructlab.model.test:test"
 "train" = "instructlab.model.train:train"
-"list" = "instructlab.model.list:model_list"
 
 [tool.setuptools_scm]
 version_file = "src/instructlab/_version.py"
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -559,7 +559,7 @@ ilab --version
 ilab system info
 
 # pipe 3 carriage returns to ilab config init to get past the prompts
-echo -e "\n\n\n" | ilab init
+echo -e "\n\n\n" | ilab config init
 
 # Enable Debug in func tests with debug level 1
 sed -i.bak -e 's/log_level:.*/log_level: DEBUG/g;' "${ILAB_CONFIG_FILE}"
--- a/tests/test_lab.py
+++ b/tests/test_lab.py
@@ -134,16 +134,8 @@ subcommands: list[Command] = [
 aliases = [
     "serve",
     "train",
-    "convert",
     "chat",
-    "test",
-    "evaluate",
-    "init",
-    "download",
-    "diff",
     "generate",
-    "sysinfo",
-    "list",
 ]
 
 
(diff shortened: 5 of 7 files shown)


bjhargrave (commented):
I think a smoother approach would be to continue to recognize the aliases but just exit after telling the user the replacement command (instead of executing the replacement command). This can help retrain users by telling them the new command.

Then in a later release, we can just remove all handling of the old aliases.

cdoern:
@bjhargrave for each aliased command though we have had a warning that these would be removed in 0.19. I feel like enough warning has been given that warrants deletion of the code we no longer want to maintain. Open to what others think on this though!

bjhargrave:
> we have had a warning that these would be removed in 0.19

I think my suggestion conforms with removing them since they no longer do anything useful and just emit a message with the proper command to use. My suggested is just a baby step to removing them by nudging the users to the proper command. 

I still see many people using the old command forms since they are not forced to use the new forms. (I am not eve sure they read the message :-) Just removing them without a nudge to the new command is harsher than it needs to be.

alinaryan:
> I think a smoother approach would be to continue to recognize the aliases but just exit after telling the user the replacement command (instead of executing the replacement command). This can help retrain users by telling them the new command.
Then in a later release, we can just remove all handling of the old aliases.

@bjhargrave We basically have been doing this except we don't exit - just warn. The [issue](https://github.com/instructlab/instructlab/issues/2109) called for removing all handling of the old aliases, but I thought adding the transitional class with the new deprecation warning was a good compromise.

alinaryan:
> continue to recognize the aliases but just exit after telling the user the replacement command

@bjhargrave I could do this though! I'll see what others think as well

alinaryan:
Example output running ilab and then trying to run a removed command:
```

(venv) alina@fedora:~/dev/instruct-lab/cli/instructlab$ ilab

Options:
                 /home/alina/.config/instructlab/config.yaml]

Commands:

Aliases:
  chat      model chat
  generate  data generate
  serve     model serve
  train     model train
(venv) alina@fedora:~/dev/instruct-lab/cli/instructlab$ ilab evaluate
Command 'evaluate' not found. If this was an aliased command, please note that aliases have been deprecated in 0.19.0.
Please refer to the updated documentation for the new command structure.
Try 'ilab --help' for help.

Error: No such command 'evaluate'.
```

bjhargrave:
> Aliases:
>   chat      model chat
>   generate  data generate
>   serve     model serve
>   train     model train

I think the help output should not mention the aliases. But if the user uses an old alias, the error message should state the replacement command to nudge the user.

src/instructlab/clickext.py
```diff
@@ -75,9 +75,8 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None
                 assert isinstance(cmd, click.Command)
             group, primary = self.get_alias_info(cmd_name)
             click.echo(
-                "You are using an aliased command, this will be deprecated "
-                "in a future release. Please consider using "
-                f"`ilab {group} {primary}` instead"
+                "You are using an aliased command. Please consider using "
+                f"`ilab {group} {primary}` instead."
```
bjhargrave, line 79:
```suggestion
                "You used a deprecated command alias. Please use the replacement command "
                f"`ilab {group} {primary}` instead."
```

cdoern:
> Aliases:
> chat model chat
> generate data generate
> serve model serve
> train model train

If we are supporting key aliases forever I think having aliases printed in help text makes sense

alinaryan:
> I think the help output should not mention the aliases. But if the user uses an old alias, the error message should state the replacement command to nudge the user.

I will keep the `chat`, `generate`, `serve`, and `train` aliases and remove the other aliases. It makes sense to keep aliases that can perform on one dataset (you can only generate on data or you can only train on models, but we may support evaluating on more than models). To stay in line with this, I think we should print aliases that we still support.

alinaryan:
@leseb I addressed the tests, PTAL

cdoern (approved):
looks good. Thank you for doing this! if all the tests are happy and the key aliases are gone, I am happy :pray:

src/instructlab/clickext.py
alinaryan: We are keeping a few of the aliases, so I removed this error since they aren't part of the deprecated list.

leseb (commented):
One last thing, could you please update the CHANGELOG to reflect that the aliases are gone? Thanks!

alinaryan:
> One last thing, could you please update the CHANGELOG to reflect that the aliases are gone? Thanks!

@leseb Updated the changelog, PTAL!

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@
...
 * Removed unused argument `--rouge-threshold` of `ilab data generate`
+* Removed the following aliased commands:
+  * `convert`
+  * `diff`
+  * `download`
+  * `evaluate`
+  * `init`
+  * `list`
+  * `info`
+  * `test`
```
leseb, line 16: I think `sysinfo` is missing.

alinaryan: thanks for catching, just fixed

src/instructlab/clickext.py
```diff
@@ -73,25 +73,28 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None
...
+        if command is None:
+            click.echo(
+                f"Command '{cmd_name}' not found. If this was an aliased command, please note that aliases have been deprecated in 0.19.0.",
+                err=True,
+            )
+            click.echo(
+                "Please refer to the updated documentation for the new command structure.",
+                err=True,
+            )
+        return command
```
leseb, line 97: Do we have a test to validate this addition? Isn't this outside of the scope of removing aliases? Am I missing something?

leseb (commented):
Sorry one more question!

src/instructlab/clickext.py
alinaryan: We are removing aliases
  * `convert`
  * `diff`
  * `download`
  * `evaluate`
  * `init`
  * `list`
  * `sysinfo`
  * `test`
  
 and keeping aliases
  * `train`
  * `generate`
  * `chat` 
  * `serve`
  
Since we are keeping some aliases, I suppose it doesn't make sense to say that aliases have been deprecated so I will remove `If this was an aliased command, please note that aliases have been deprecated in 0.19.0.` from the error.

leseb: Resolved offline, we are removing this part. Thanks!

src/instructlab/clickext.py
```diff
@@ -73,25 +73,24 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None
...
                 raise click.exceptions.Exit(1)
             return cmd
-        return super().get_command(ctx, cmd_name)
+        command = super().get_command(ctx, cmd_name)
+        if command is None:
+            click.echo(
+                f"Command '{cmd_name}' not found.",
+                err=True,
+            )
+        return command
```
leseb, line 93: Please revert this too, this is natively handled by click when the command does not exist. Thanks!

src/instructlab/clickext.py
```diff
@@ -73,25 +73,18 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None
...
+                ctx.params["config_file"] == DEFAULTS.CONFIG_FILE
             ):
                 click.secho(
                     "Some ilab storage directories do not exist yet. Please run `ilab config init` before continuing.",
                     fg="red",
                 )
                 raise click.exceptions.Exit(1)
             return cmd
-        return super().get_command(ctx, cmd_name)
+        return None
```
booxter, line 87: I don't think you should return None here. The caller expects a command. I think super() works. What is this change trying to resolve?

leseb, line 87:
```suggestion
return super().get_command(ctx, cmd_name)
```

Please revert entirely.

booxter: I think None should also be dropped from the typing hint for the return value for this function, seems incorrect. (I understand it's not part of this PR, but maybe you can clean it up.)

alinaryan: Resolved this offline-
I reverted it back to `return super().get_command(ctx, cmd_name)`

leseb (approved):

PR #2217, opened 2024-09-09
refactor: move BackendServer to common

makelinux:

Move BackendServer to reduce the circular dependencies.

Signed-off-by: Costa Shulyupin <costa.shul@redhat.com>


danmcp (commented):
Would BackendServer make more sense in a separate file?

nathan-weinberg (approved):

PR #2251, opened 2024-09-11
fix: report server errors for chat/generate/eval

leseb:
When the internal server crashes, the errors will be displayed in the chat, as well as the progress logs from generate and other related commands.

Fixes: https://github.com/instructlab/instructlab/issues/889

**Checklist:**

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -387,31 +387,6 @@ test_no_chat_logs(){
         '
 }
 
-
-test_temp_server_ignore_internal_messages(){
-    # shellcheck disable=SC2016
-    expect -c '
-        set timeout 60
-        spawn ilab model chat
-        expect "Starting a temporary server at"
-        send "hello!\r"
-        sleep 5
-        send "\003"
-        expect {
-            "Disconnected from client (via refresh/close)" { exit 1 }
-        }
-        send "exit\r"
-        expect {
-            "Traceback (most recent call last):" { set exp_result 1 }
-            default { set exp_result 0 }
-        }
-        if { $exp_result != 0 } {
-            puts stderr "Error: ilab model chat command failed"
-            exit 1
-        }
-    '
-}
-
 test_server_welcome_message(){
     # test that the server welcome message is displayed
     ilab model serve --log-file serve.log &
@@ -631,8 +606,6 @@ cleanup
 test_temp_server_sigint
 cleanup
 test_no_chat_logs
-cleanup
-test_temp_server_ignore_internal_messages
 cleanup
 test_server_chat_template
 cleanup
--- a/src/instructlab/model/backends/llama_cpp.py
+++ b/src/instructlab/model/backends/llama_cpp.py
@@ -1,12 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
-from contextlib import redirect_stderr, redirect_stdout
 from time import sleep
 from typing import Optional, cast
 import logging
 import multiprocessing
-import os
 import pathlib
 
 # Third Party
@@ -20,7 +18,7 @@ import llama_cpp.server.app as llama_app
 # Local
 from ...client import check_api_base
 from ...configuration import get_api_base
-from .backends import UvicornServer, get_uvicorn_config, is_temp_server_running
+from .backends import UvicornServer, get_uvicorn_config
 from .common import (
     API_ROOT_WELCOME_MESSAGE,
     CHAT_TEMPLATE_AUTO,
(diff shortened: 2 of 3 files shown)


danmcp (commented):
Couple questions:

- Is there a recommended way to test? I tried killing the serving process but it didn't behave any differently than before
- How should we think about this change relative to the behavior with vllm?

leseb:
> Couple questions:
> 
> * Is there a recommended way to test? I tried killing the serving process but it didn't behave any differently than before

The best way is to do what the func test used to do to validate stderr is not surfaced. So essentially:

```bash
ilab model chat
<start chatting>
<send ctrl+C during a response>
<llama-cpp prints 'disconnected' in the chat
```

> * How should we think about this change relative to the behavior with vllm?
We would need to revert this https://github.com/instructlab/instructlab/commit/b4a1af5c6a93cff598bde854663f798bb863176b, but it might cause other issues :/

danmcp:
> > * How should we think about this change relative to the behavior with vllm?
> 
> We would need to revert this [b4a1af5](https://github.com/instructlab/instructlab/commit/b4a1af5c6a93cff598bde854663f798bb863176b), but it might cause other issues :/

Not following.  We would want to revert that change so the logging output wouldn't show up in chat?  Isn't vllm is pretty chatting even without logging enabled?

danmcp:
> > > > * How should we think about this change relative to the behavior with vllm?
> > > 
> > > 
> > > We would need to revert this [b4a1af5](https://github.com/instructlab/instructlab/commit/b4a1af5c6a93cff598bde854663f798bb863176b), but it might cause other issues :/
> > 
> > 
> > Not following. We would want to revert that change so the logging output wouldn't show up in chat? Isn't vllm is pretty chatting even without logging enabled?
> 
> We would want to revert this to show the server's internal errors to the chat window yes. vLLM is outputting to stdout/stderr, right?

If not set to background, it will output stdout/err.  Currently it's set to background for chat.  If we didn't do that, we would get a constant stream of messages from vllm in the chat.  If logging is enabled (we are accepting the default of enabled with the change you referenced), it will be even more chatty.  But I don't see how the logging being enabled or not changes the decision.  In either case I don't see how we can let all the vllm output flow to console.

Just a note regarding the logging being enabled.  The reason we have that logic in place is because vllm logging wrecks havoc on instructlab logging when imported through lm_eval.  So we have to disable it for instructlab, but we accept the default when launching vllm.

General thoughts about this change:
- It's probably fine given how llamacpp behaves (doesn't seem to output anything but errors).
- Not possible with the current impl of vllm.  We would need to suppress everything but errors.
- Don't love that we are going to have completely different behaviors between the two
- Wondering whether there is another change that could be made consistently to only show the serving output (or a tail of the output) when a failure occurred.

leseb:
> > > > > * How should we think about this change relative to the behavior with vllm?
> > > > 
> > > > 
> > > > We would need to revert this [b4a1af5](https://github.com/instructlab/instructlab/commit/b4a1af5c6a93cff598bde854663f798bb863176b), but it might cause other issues :/
> > > 
> > > 
> > > Not following. We would want to revert that change so the logging output wouldn't show up in chat? Isn't vllm is pretty chatting even without logging enabled?
> > 
> > 
> > We would want to revert this to show the server's internal errors to the chat window yes. vLLM is outputting to stdout/stderr, right?
> 
> If not set to background, it will output stdout/err. Currently it's set to background for chat. If we didn't do that, we would get a constant stream of messages from vllm in the chat. If logging is enabled (we are accepting the default of enabled with the change you referenced), it will be even more chatty. But I don't see how the logging being enabled or not changes the decision. In either case I don't see how we can let all the vllm output flow to console.
> 
> Just a note regarding the logging being enabled. The reason we have that logic in place is because vllm logging wrecks havoc on instructlab logging when imported through lm_eval. So we have to disable it for instructlab, but we accept the default when launching vllm.
> 
> General thoughts about this change:
> 
> * It's probably fine given how llamacpp behaves (doesn't seem to output anything but errors).
> * Not possible with the current impl of vllm.  We would need to suppress everything but errors.
> * Don't love that we are going to have completely different behaviors between the two
> * Wondering whether there is another change that could be made consistently to only show the serving output (or a tail of the output) when a failure occurred.

Thanks for the analysis. In conclusion, I believe we can move forward with this, right? Thanks!

danmcp:
> Thanks for the analysis. In conclusion, I believe we can move forward with this, right? Thanks!

If my understanding is correct (I put it out there to make sure I was following the change and current state correctly), I don't object to this change going in.

leseb:
> > Thanks for the analysis. In conclusion, I believe we can move forward with this, right? Thanks!
> 
> If my understanding is correct (I put it out there to make sure I was following the change and current state correctly), I don't object to this change going in.

It is correct :)

danmcp (approved):

nathan-weinberg (approved):

PR #2264, opened 2024-09-13
use pretty color for config show

reidliu41:
Improve the `show` command:
- add `color`, may view better
- add `json` output
- add key search e.g.  chat.model

**Checklist:**

- [ x] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.


reidliu41:
pretty_color

src/instructlab/config/show.py
```diff
@@ -19,10 +21,14 @@
 @clickext.display_params
 def show(ctx: click.Context) -> None:
     """Displays the current config as YAML"""
-    # TODO: make this use pretty colors like jq/yq
+    # use pretty colors like jq/yq
```
leseb, line 24:
```suggestion
    # use pretty colors
```

src/instructlab/config/show.py
```diff
@@ -19,10 +21,14 @@
...
+    console = Console()
     try:
         commented_map = configuration.config_to_commented_map(ctx.obj.config)
+        stream = compat.StringIO()
+        yaml.dump(commented_map, stream)
+        syntax = Syntax(stream.getvalue(), "yaml", theme="monokai", line_numbers=True)
+        console.print(syntax)
     except YAMLError as e:
```
leseb, line 32: I think we need to handle `except Exception as e:` now.

leseb (commented):
The main concern with this patch is that not all terminals handle rich formatting and colors consistently. In minimal terminal environments that don't support ANSI colors or handle rich output poorly, the colors may not render correctly, resulting in a suboptimal user experience. It would be beneficial to implement a fallback mechanism for cases where rich formatting isn't supported.

Additionally, I'm not entirely convinced that this change significantly enhances the user experience.

reidliu41:
thank you for you advice.
1. Personally, I saw the "to do list"， I also hesitated, but I felt better after I tried to change it, compared the same color and the highlight situation, it is different, maybe easier to view.
2. Why choose `rich`,  because instructlab installed rich by default, and use it in `chat.py`
rich>=13.3.1
https://github.com/instructlab/instructlab/blob/0f681b61d7b42966922f7c38d4f1e07a57bc1841/src/instructlab/model/chat.py#L17
3. I tried to set it without color by default, detect first if use '--color'

reidliu41:
```
 ilab config show --help
Usage: ilab config show [OPTIONS]

  Displays the current config as YAML or JSON, with an option to extract
  specific key.

Options:
  -c, --color     Enable colored output.
  -j, --json      Output in JSON format instead of YAML.
  -k, --key TEXT  Extract a specific key from the configuration.
  --help          Show this message and exit.


$ ilab config show -j
{
    "chat": {
        "context": "default",
        "greedy_mode": false,
        "logs_dir": "/Users/xx/.local/share/instructlab/chatlogs",
        "max_tokens": null,
        "model": "/Users/xx/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf",
        "session": null,
        "vi_mode": false,
        "visible_overflow": true
    },
    "evaluate": {
        "base_branch": null,
        "base_model": "instructlab/granite-7b-lab",
        "branch": null,
        "gpus": null,
        "mmlu": {
            "batch_size": "auto",
            "few_shots": 5
        },
        "mmlu_branch": {

$ ilab config show -k version
1.0.0

$ ilab config show -k chat.model
/Users/xx/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
....


```

reidliu41:
hi, please feel free to share more ideas.

alinaryan:
Thanks for opening this! We started another convo in https://github.com/instructlab/instructlab/pull/2495#issuecomment-2445065944 regarding color scheme. I think adding pretty color to the config is a great idea, but we will want to be consistent across the ilab flow. Could you propose the idea of using Monokai theme across ilab in a dev doc in https://github.com/instructlab/dev-docs? 
After that, this PR could be part of that effort to have a consistent color scheme across the workflow.

reidliu41:
> Thanks for opening this! We started another convo in [#2495 (comment)](https://github.com/instructlab/instructlab/pull/2495#issuecomment-2445065944) regarding color scheme. I think adding pretty color to the config is a great idea, but we will want to be consistent across the ilab flow. Could you propose the idea of using Monokai theme across ilab in a dev doc in https://github.com/instructlab/dev-docs? After that, this PR could be part of that effort to have a consistent color scheme across the workflow.

Thank you for your reply. How about the other part of the pr, e.g. json output, search key？
If ok,I can remove the color part,and consider later.

reidliu41:
> Thanks for opening this! We started another convo in [#2495 (comment)](https://github.com/instructlab/instructlab/pull/2495#issuecomment-2445065944) regarding color scheme. I think adding pretty color to the config is a great idea, but we will want to be consistent across the ilab flow. Could you propose the idea of using Monokai theme across ilab in a dev doc in https://github.com/instructlab/dev-docs? After that, this PR could be part of that effort to have a consistent color scheme across the workflow.

@alinaryan I agreed, but I just want to highlight with color(no matter what color is it), so simply change to green.

cdoern:
I agree with @alinaryan here, I think this is a change we need to standardize across commands. I don't think this is something we can just have in `show` and not other places like `init`, `list`, `edit`,

JustinXHale:
thanks for raising this concern! I agree this issue needs addressing, we actually need to address it across the cli. there are inconsistencies, resulting in a less than inviting experience.

reidliu41:
I think the color option is hard to make a decision. Seems more different options. Not sure when will get the result.  I will remove the `color` option currently, if you guys think -j and -k is ok, please consider merge.

reidliu41:
sorry, for [new one](https://github.com/instructlab/instructlab/pull/2590)

Issue #2265, opened 2024-09-13
container use stable tag very old


clopnis:
**Describe the bug**
stable tag is pointing v0.17 and is used for build container
https://github.com/instructlab/instructlab/blob/main/containers/cuda/Containerfile#L17

It should use pip and not github tags

**Expected behavior**
Should be on newest version

mdevino:
Hi @nathan-weinberg, @clopnis, can I work on this? It would be my first issue.

nathan-weinberg:
Yes @mdevino thank you! Assigned the issue to you

mdevino:
Hello. PR opened. I've asked a couple of questions in it.

PR #2270, opened 2024-09-14
fix: Use pip to install InstructLab on container images

mdevino:
This PR addresses #2265.

What I tested:
1. Built the InstructLab image without making changes to the container file - it worked and installed ilab 0.18.4. I tried using `torch.cuda.is_enabled()` but it returned False.
2. Changed the container file to use pip and was able to spin up a container to verify that the ilab version instaleld was 0.18.4.

A couple of questions:
1. Are there any tests I need to run on the image?
2. I have removed the conditional to install `llama-cpp-python` with CUBLAS. Is this done by the pip installation, or would I need to find a way to get the InstructLab version being installed from pip to force the `llama-cpp-python` reinstall?
3. Does the "type" of the commit message sound right?

**Checklist:**

### DIFF
--- a/containers/cuda/Containerfile
+++ b/containers/cuda/Containerfile
@@ -14,10 +14,9 @@ RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cu
     && export XLA_TARGET=cuda120 \
     && export XLA_FLAGS=--xla_gpu_cuda_data_dir=/usr/local/cuda
 
-ARG GIT_TAG=stable
-RUN if [[ "$(uname -m)" != "aarch64" ]]; then export CFLAGS="-mno-avx"; fi && \
-    CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install -r https://raw.githubusercontent.com/instructlab/instructlab/${GIT_TAG}/requirements.txt --force-reinstall --no-cache-dir llama-cpp-python
-RUN python3.11 -m pip install git+https://github.com/instructlab/instructlab.git@${GIT_TAG}
+#RUN python3.11 -m pip install instructlab
+RUN CMAKE_ARGS="-DLLAMA_CUDA=on -DLLAMA_NATIVE=off" python3.11 -m pip install instructlab packaging wheel \
+   && python3.11 -m pip install instructlab[cuda]
 
 CMD ["/bin/bash"]
 


seroyer:
I believe the reason why the GPU wasn't usable is because the environment variables set in the RUN statement on line 11 are not persisted.  You cannot use `export` inside of a `RUN` statement and hope that it will persist after that statement ends.  I changed that `RUN` statement into a series of `ENV` statements and also fixed the cudnn8 package to be cudnn9 and now the GPU works for me.
As an aside, there is another PR that completely refactors how this Containerfile works: https://github.com/instructlab/instructlab/pull/1961

containers/cuda/Containerfile
```diff
@@ -14,10 +14,7 @@ RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cu
     && export XLA_TARGET=cuda120 \
     && export XLA_FLAGS=--xla_gpu_cuda_data_dir=/usr/local/cuda
 
-ARG GIT_TAG=stable
-RUN if [[ "$(uname -m)" != "aarch64" ]]; then export CFLAGS="-mno-avx"; fi && \
-    CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install -r https://raw.githubusercontent.com/instructlab/instructlab/${GIT_TAG}/requirements.txt --force-reinstall --no-cache-dir llama-cpp-python
-RUN python3.11 -m pip install git+https://github.com/instructlab/instructlab.git@${GIT_TAG}
+RUN python3.11 -m pip install instructlab
```
tiran, line 17: This will build llama-cpp-python without CUDA support and does not install InstructLab's CUDA dependencies.

mdevino: Hi @tiran, thanks for the reply. Given the PR @seroyer linked, does it make sense to pursue this? 

If so, would `CMAKE_ARGS="DLLAMA_CUBLAS=on" python3.11 -m pip install instructlab[cuda]` install CUDA dependencies?

tiran: It's `-D` to set a cmake define.

Installation is a bit more complicated, because one dependencies (`flash-attn`) does not specify its build and installation requirements correctly. You have to install instructlab and two supports packages, then install instructlab's CUDA dependencies:

```
CMAKE_ARGS="-DLLAMA_CUDA=on -DLLAMA_NATIVE=off" python3.11 -m pip install instructlab packaging wheel \
   && python3.11 -m pip install instructlab[cuda]
```

containers/cuda/Containerfile
```diff
@@ -14,10 +14,9 @@ RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cu
     && export XLA_TARGET=cuda120 \
     && export XLA_FLAGS=--xla_gpu_cuda_data_dir=/usr/local/cuda
 
-ARG GIT_TAG=stable
-RUN if [[ "$(uname -m)" != "aarch64" ]]; then export CFLAGS="-mno-avx"; fi && \
-    CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3.11 -m pip install -r https://raw.githubusercontent.com/instructlab/instructlab/${GIT_TAG}/requirements.txt --force-reinstall --no-cache-dir llama-cpp-python
-RUN python3.11 -m pip install git+https://github.com/instructlab/instructlab.git@${GIT_TAG}
+#RUN python3.11 -m pip install instructlab
```
leseb, line 17: why is this commented?

mdevino:
Hi @tiran,

I believe this PR is obsolete. Could you confirm?

mdevino:
Hi, @courtneypacheco .Thanks for reaching out. My understanding is that this PR became obsolete, so we could close it.

courtneypacheco:
Thank you! As always, you can reopen if you need to at any time. 😄

PR #2271, opened 2024-09-15
add sys prompt to config

reidliu41:
```
$ ilab model chat --help
Usage: ilab model chat [OPTIONS] [QUESTION]...

  Runs a chat using the modified model

Options:
  --system-prompt TEXT            Custom system prompt. If not provided, it
                                  will use the default context. [default:
                                  <None>; config: 'chat.system_prompt']
```

**Checklist:**

Easy updates: can change the system prompt without touching the code.
Customization: Users can set their own prompt to fit different needs.
Consistency: Ensures the same prompt is used everywhere.
Flexibility: Prepares for future use of multiple prompts.
Better organization: Keeps settings separate from the code.

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -184,6 +184,10 @@ class _chat(BaseModel):
         default=1.0,
         description="Controls the randomness of the model's responses. Lower values make the output more deterministic, while higher values produce more random results.",
     )
+    system_prompt: Optional[str] = Field(
+        default=None,
+        description="Custom system prompt. If not provided, it will use the default context.",
+    )
 
 
 class _serve_vllm(BaseModel):
--- a/tests/testdata/default_config.yaml
+++ b/tests/testdata/default_config.yaml
@@ -19,6 +19,9 @@ chat:
   # Filepath of a dialog session file.
   # Default: None
   session:
+  # Custom system prompt. If not provided, it will use the default context.
+  # Default: None
+  system_prompt:
   # Controls the randomness of the model's responses. Lower values make the output
   # more deterministic, while higher values produce more random results.
   # Default: 1.0
--- a/src/instructlab/cli/model/chat.py
+++ b/src/instructlab/cli/model/chat.py
@@ -154,6 +154,11 @@ logger = logging.getLogger(__name__)
     is_flag=True,
     help="Disable decorations for chat responses.",
 )
+@click.option(
+    "--system-prompt",
+    type=click.STRING,
+    cls=clickext.ConfigOption,
+)
 @click.pass_context
 @clickext.display_params
 def chat(
@@ -180,6 +185,7 @@ def chat(
     embedding_model_path,
     top_k,
     no_decoration,
+    system_prompt,
 ):
     """Runs a chat using the modified model"""
 
@@ -219,6 +225,7 @@ def chat(
         embedding_model_path,
         top_k,
         no_decoration,
+        system_prompt,
         backend_type=ctx.obj.config.serve.server.backend_type,
         host=ctx.obj.config.serve.server.host,
         port=ctx.obj.config.serve.server.port,
--- a/src/instructlab/model/chat.py
+++ b/src/instructlab/model/chat.py
@@ -589,6 +589,7 @@ def chat_model(
     embedding_model_path,
     top_k,
     no_decoration,
+    system_prompt,
     backend_type,
     host,
     port,
@@ -764,6 +765,7 @@ def chat_model(
             backend_type=backend_type,
             params=params,
             no_decoration=no_decoration,
+            system_prompt=system_prompt,
         )
     except ChatException as exc:
         print(f"{RED}Executing chat failed with: {exc}{RESET}")
@@ -794,6 +796,7 @@ def chat_cli(
     visible_overflow,
     params,
     no_decoration,
+    system_prompt,
 ):
     """Starts a CLI-based chat with the server"""
     client = OpenAI(
@@ -828,7 +831,12 @@ def chat_cli(
         logger.info(f"Context {context} not found in the config file. Using default.")
         context = "default"
     loaded["name"] = context
-    sys_prompt = CONTEXTS.get(context, "default")(get_model_arch(pathlib.Path(model)))
+    if system_prompt:
+        sys_prompt = system_prompt
+    else:
+        sys_prompt = CONTEXTS.get(context, "default")(
+            get_model_arch(pathlib.Path(model))
+        )
     loaded["messages"] = [{"role": "system", "content": sys_prompt}]
 
     # Instantiate retriever if RAG is enabled
--- a/tests/test_model_chat.py
+++ b/tests/test_model_chat.py
@@ -15,6 +15,7 @@ from instructlab import lab
 from instructlab.configuration import model_info
 from instructlab.feature_gates import FeatureGating, FeatureScopes, GatedFeatures
 from instructlab.model.chat import ChatException, ConsoleChatBot
+from instructlab.utils import get_sysprompt
 from tests.test_feature_gates import dev_preview
 
 # Local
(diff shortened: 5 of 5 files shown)


src/instructlab/configuration.py
```diff
@@ -111,6 +111,13 @@ class _InstructlabDefaults:
         "lora_target_modules": ["q_proj", "k_proj", "v_proj", "o_proj"],
     }
 
+    # System Prompt
+    SYS_PROMPT = (
```
leseb, line 115: Why not re-use the prompt in src/instructlab/common.py?

src/instructlab/configuration.py
```diff
@@ -661,6 +665,15 @@ class _train(BaseModel):
     )
 
 
+class _prompts(BaseModel):
```
cdoern, line 668: is this used anywhere? I don't see this being use directly in a cmd so I am hesitant to add a field without the usecase

cdoern (changes requested):

src/instructlab/configuration.py
reidliu41: Please correct me if I am misunderstanding, here is what I check:
1. Currently, the `SYS_PROMPT` is using in `chat` from the handcode `common.py`
2. Due to different models, or different situations for the user, so the prompt might be need to modify by user frequently.
3. Change from here: define a default one from `common.py`  -->     write into `config.yaml` with init --> read the prompt from the config for `chat` --> user can use `ilab config edit` to modify the prompt

nathan-weinberg: @cdoern can you respond to the above please

reidliu41: thanks for response.  I do see a user ask for change(system prompt) in `users` channel, may consider set it in config.

src/instructlab/configuration.py
```diff
@@ -564,6 +578,11 @@ class Config(BaseModel):
         description="Configuration file structure version.",
         frozen=True,  # don't allow this to be changed anywhere in the code
     )
+    # prompts configuration
+    prompts: _prompts = Field(
```
alimaredia, line 582: Since this field is only being used in chat, it should only be included in `_chat`.

src/instructlab/common.py
```diff
@@ -1,3 +1,4 @@
 # SPDX-License-Identifier: Apache-2.0
 
 SYS_PROMPT = "I am, Red Hat® Instruct Model based on Granite 7B, an AI language model developed by Red Hat and IBM Research, based on the Granite-7b-base language model. My primary function is to be a chat assistant."
+CLI_HELPER_PROMPT = "You are an expert for command line interface and know all common commands. Answer the command to execute as it without any explanation."
```
alimaredia, line 4: What if we just added these prompts to where you're setting them in `defaults.py` and removed this file. I tried to look at the git history for this file and didn't find much.

Removing this file would mean a change would have to be made to `utils.py`.

src/instructlab/utils.py
```diff
@@ -367,7 +367,9 @@ def get_sysprompt(model=None):
     Returns:
         str: The system prompt for the model being used
     """
-    return common.SYS_PROMPT
+    config = read_config()
```
alimaredia, line 370: It looks like you can get rid of the `model` argument here too, As of this change it doesn't look like we have plans to implement it.

alimaredia (changes requested):

src/instructlab/utils.py
reidliu41: hey thanks for your feedback. Seems I see @jaideepr97 working on another pr that related to this prompt? Should I continue on this?

alimaredia: Is this the PR you're mentioning: https://github.com/instructlab/instructlab/pull/2592? 

From what I can tell, it looks like you're trying to give the user the ability to set the system prompt if they wish, which I think is great, and I don't see that functionality in #2582.

If that's the case, see how he feels about the changes here, and possibly expect to rebase and merge this PR in after he's done with his.

reidliu41: > It looks like you can get rid of the `model` argument here too, As of this change it doesn't look like we have plans to implement it.

yeah, model put it in there for a long time

src/instructlab/common.py
reidliu41: yeah, should only use in chat so far. I think is ok.

reidliu41: if just put it in `default.py`, seems no need to use `get_sysprompt `. Actually, seems even in common, seems also no need to use `get_sysprompt `, may be different purpose?

jaideepr97:
@reidliu41 I think a left a comment about BYO sys prompt on a different PR
https://github.com/instructlab/instructlab/pull/2577#pullrequestreview-2424214066

TL;DR we may need to wait for some research input on what the consequences of allowing BYO sys prompt are and what the best way to expose this feature to users would be 

cc @Maxusmusti

jaideepr97 (commented):
@reidliu41 I think a left a comment about BYO sys prompt on a different PR
https://github.com/instructlab/instructlab/pull/2577#pullrequestreview-2424214066

TL;DR we may need to wait for some research input on what the consequences of allowing BYO sys prompt are and what the best way to expose this feature to users would be

cc @Maxusmusti

tests/testdata/default_config.yaml
```diff
@@ -19,6 +19,9 @@ chat:
   # Filepath of a dialog session file.
   # Default: None
   session:
+  # Custom system prompt. If not provided, it will use the default context.
+  # Default: ''
+  system_prompt: ''
```
RobotSail, line 24: Do we really want the default system system prompt to be set to an empty string? It seems like a better default would be simply not including it / having it set to `None`.

alimaredia: The default in pydantic is `None` (https://github.com/instructlab/instructlab/blob/main/src/instructlab/configuration.py#L527). Not sure why it's rendering to an empty string here.

reidliu41: thanks, updated

tests/testdata/default_config.yaml
```diff
@@ -19,6 +19,9 @@ chat:
   # Filepath of a dialog session file.
   # Default: None
   session:
+  # Custom system prompt. If not provided, it will use the default context.
+  # Default: None
+  system_prompt:
```
RobotSail, line 24: Is there any way for it to not appear in the config at all? Or is this a limitation of the current config system?

RobotSail (commented):
I think this is a good change, the only issues is that we have a bunch of instances where the system prompt is referenced throughout the codebase. I'm seeing that the amount of lines changed is only 23, are we sure that we're covering all of them? 

If not, I think it's fine to leave it as something that's specific to chat, but we should probably add some tests if possible to validate the behavior.

tests/testdata/default_config.yaml
reidliu41: I think so, seems  a limitation of the current config system. Tried `exclude=True`, still will appear..

reidliu41:
```
    of.gguf.write_tensor_data(ndarray)
  File "/actions-runner/_work/instructlab/instructlab/venv/lib64/python3.11/site-packages/gguf/gguf_writer.py", line 417, in write_tensor_data
    tensor.tofile(fout)
[ 89/291] Writing tensor blk.9.ffn_down.weight                  | size   4096 x  11008  | type F32  | T+  65
[ 90/291] Writing tensor blk.9.attn_norm.weight                 | size   4096           | type F32  | T+  66
```

reidliu41:
> I think this is a good change, the only issues is that we have a bunch of instances where the system prompt is referenced throughout the codebase. I'm seeing that the amount of lines changed is only 23, are we sure that we're covering all of them?
> 
> If not, I think it's fine to leave it as something that's specific to chat, but we should probably add some tests if possible to validate the behavior.

yes, so far only use for chat. 
I reviewed the others, seems some use the same , some just similar(should be different), so now separate it , might be good...

tests/testdata/default_config.yaml
reidliu41: lol, yeah, but seems evaluate do it too:'
https://github.com/instructlab/instructlab/blob/5922a76266ad65b4cfc1bb7ffeee9de56dd96df6/tests/testdata/default_config.yaml#L112

reidliu41: I saw @alimaredia submit a `redesign config`, maybe he has ideas about this.

RobotSail: Yeah I suppose it's fine, it just looks very strange.

RobotSail (approved):
LGTM

reidliu41:
@alimaredia hi, can you take a look?

booxter (approved):

reidliu41:
hi team, can someone take a look?

mairin:
@reidliu41 it looks like it's blocked because @cdoern and @alimaredia requested changes... are those still applicable?

reidliu41:
thanks, yeah, so if someone has time, might help to review..

courtneypacheco:
Hey @reidliu41!

A PR was opened recently that is similar to yours: https://github.com/instructlab/instructlab/pull/3197

The difference is that PR focuses on decoupling the notion of hardcoding the specific model for each different workflow (training, sdg, eval, etc). Can you take a look?

reidliu41:
I think https://github.com/instructlab/instructlab/pull/3197 is not related to `system_prompt` in `chat`,  please correct me if I misunderstood, @RobotSail

reidliu41:
@RobotSail thank you so much for your explanation. It's good to wait for https://github.com/instructlab/instructlab/pull/3197 finish.

RobotSail:
@reidliu41 We should rebase onto the new main and update this PR

reidliu41:
@RobotSail updated

PR #2242, opened 2024-09-10
add a model remove subcommand

reidliu41:
```
$ ilab model remove --help
Usage: ilab model remove [OPTIONS]

  Remove model

Options:
  -m, --model TEXT            Specify the model name to remove (check with
                              'ilab model list' first), e.g.,
                              merlinite-7b-lab-Q4_K_M.gguf or
                              instructlab/merlinite-7b-pt.  [required]
  -f, --force                 Used to forcefully remove a model. Use with
                              caution as it permanently deletes the model from
                              the storage directory without further
                              confirmation.
  -md, --model-dir DIRECTORY  Specify the base directory for models. [default:
                              /Users/user/.cache/instructlab/models]
  --help                      Show this message and exit.

$ ilab model remove --model
Error: Option '--model' requires an argument.

$ ilab model list
+------------------------------+---------------------+---------+
| Model Name                   | Last Modified       | Size    |
+------------------------------+---------------------+---------+
| instructlab/granite-7b-lab   | 2024-09-10 18:49:32 | 12.6 GB |
| merlinite-7b-lab-Q4_K_M.gguf | 2024-09-10 18:43:54 | 4.1 GB  |
+------------------------------+---------------------+---------+

$ ilab model remove --model merlinite-7b-lab-Q4_K_M.gguf
==> Removing model file: merlinite-7b-lab-Q4_K_M.gguf
==> Model file merlinite-7b-lab-Q4_K_M.gguf has been removed.

$ ilab model list
+----------------------------+---------------------+---------+
| Model Name                 | Last Modified       | Size    |
+----------------------------+---------------------+---------+
| instructlab/granite-7b-lab | 2024-09-10 18:49:32 | 12.6 GB |
+----------------------------+---------------------+---------+

$ ilab model remove --model xx
Error: Model xx does not exist in .........
```

PR #2279, opened 2024-09-17
Defer running tests until linting workflows successfully complete

### LINKED ISSUE #2245: Reduce CI churn
When you submit a PR the current behavior is all the tests kick off at once.  This wastes a lot of CI cycles if there are issues early in the process.  Exs:

- Missing sign-off
- Linting/Spellcheck failures
- Unit test failures

Since some tests are fast to run, requiring the early stages to pass before kicking off later stages would save a lot of churn.  Suggested update:
- Don't kick off unit, functional, and e2e tests until signoff, linting, spellchecking has succeeded

Larger change:
- Today unit and functional tests run together.  They do so, I believe, because they share the same machine setup steps which takes ~10 mins before the unit tests start.  If we created an image that contained all the prereqs, we could launch from that image, update it, then run unit tests.  And that test could then be a prereq before kicking off functional tests.  Technically the functional tests could also share the same starter image.  Debatable whether the risk of updating vs installing from scratch is worth it.

Kicking off functional and e2e tests at the same time is probably still appropriate.  Separating them would be too time consuming and slow down the test/merge time fairly significantly.

**Additional context**

bjhargrave:
We use a reusable workflow to enable a workflow to wait for jobs in other workflows to complete successfully. If the jobs fail or do not complete in the timeout period, the reusable workflow's job will fail.

Resolves #2245

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-nvidia-t4-x1.yml
+++ b/.github/workflows/e2e-nvidia-t4-x1.yml
@@ -30,9 +30,33 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+env:
+  LC_ALL: en_US.UTF-8
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+
 jobs:
+  e2e-ready:
+    permissions:
+      checks: read
+    uses: ./.github/workflows/status-checks.yml
+    with:
+      job_names: >- # Space-separated job names to wait on for status checks
+        DCO
+        actionlint
+        markdown-lint
+        shellcheck
+        spellcheck
+        lint-workflow-complete
+
   start-runner:
     name: Start external EC2 runner
+    needs: ["e2e-ready"]
     runs-on: ubuntu-latest
     outputs:
       label: ${{ steps.start-ec2-runner.outputs.label }}
--- a/.github/workflows/status-checks.yml
+++ b/.github/workflows/status-checks.yml
@@ -1,0 +1,60 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Status Checks Reusable Workflow
+
+on:
+  workflow_call:
+    inputs:
+      job_names:
+        description: 'Space-separated job names to wait on for status checks'
+        required: true
+        type: string
+      delay:
+        description: 'Period in seconds to wait before first poll of GitHub Check Runs'
+        required: false
+        type: number
+        default: 0
+      interval:
+        description: 'Interval or period in seconds between polling GitHub Check Runs'
+        required: false
+        type: number
+        default: 10
+      timeout:
+        description: 'Timeout in seconds to complete polling GitHub Check Runs'
+        required: false
+        type: number
+        default: 3600
+
+env:
+  LC_ALL: en_US.UTF-8
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  checks: read
+
+jobs:
+  status-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Harden Runner"
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7 # v2.10.1
+        with:
+          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
+
+      - name: "Set status check variables"
+        id: set_variables
+        run: |
+          echo "match_pattern=$(jq -nr '[$ARGS.positional | .[] | split("\\s"; null) | map(select(. != ""))] | flatten | join("|")' --args "${{ inputs.job_names }}")" >> "$GITHUB_OUTPUT"
+
+      - name: "Wait for status checks"
+        uses: poseidon/wait-for-status-checks@6988432d64ad3f9c2608db4ca16fded1b7d36ead # v0.5.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          match_pattern: ${{ steps.set_variables.outputs.match_pattern }}
+          ignore_pattern: '/ ${{ github.job }}' # use pattern to ignore this nested job
+          delay: ${{ inputs.delay }}
+          interval: ${{ inputs.interval }}
+          timeout: ${{ inputs.timeout }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,8 +38,22 @@ permissions:
   contents: read
 
 jobs:
+  test-ready:
+    permissions:
+      checks: read
+    uses: ./.github/workflows/status-checks.yml
+    with:
+      job_names: >- # Space-separated job names to wait on for status checks
+        DCO
+        actionlint
+        markdown-lint
+        shellcheck
+        spellcheck
+        lint-workflow-complete
+
   test:
     name: "${{ matrix.python }} on ${{ matrix.platform }}"
+    needs: ["test-ready"]
     runs-on: "${{ matrix.platform }}"
     strategy:
       matrix:
@@ -144,6 +158,7 @@ jobs:
           pip cache remove llama_cpp_python
 
   docs:
+    needs: ["test-ready"]
     runs-on: ubuntu-latest
     steps:
       - name: "Harden Runner"


.github/workflows/e2e-nvidia-t4-x1.yml
```diff
@@ -30,9 +30,34 @@ concurrency:
...
+    uses: ./.github/workflows/status-checks.yml
+    with:
+      job_names: >- # Space-separated job names to wait on for status checks
+        DCO
+        actionlint
+        markdown-lint
+        shellcheck
+        spellcheck
+        lint-workflow-complete
+      delay: 5 # wait while jobs get started
```
danmcp, line 56: What happens without the delay?  I assume the logic is that it fails immediately if they don't exist?

bjhargrave: The delay allows the jobs queue to build up before deciding if the jobs are complete (or not present because they are not triggered). With a zero delay, it is possible this job could start first, see only a subset of the actual jobs and then exit with success since the named jobs are not yet present in the jobs queue. So I picked 5 seconds but we can pick another delay value if desired.

leseb (commented):
Do we have to document this?

bjhargrave:
> Do we have to document this?

The 5 seconds delay? We can. In the alternate, I can change the default for delay in the reusable workflow from 0 to 5 and then users of the reusable workflow will get that default.

leseb:
> > Do we have to document this?
> 
> The 5 seconds delay? We can. In the alternate, I can change the default for delay in the reusable workflow from 0 and then users of the reusable workflow will get that default.

I was thinking about editing this document actually https://github.com/instructlab/instructlab/blob/main/docs/ci.md#ci-for-instructlab

.github/workflows/e2e-nvidia-t4-x1.yml
danmcp: Here is a current example where the workflow tests are kicking off for a first time user and the other jobs haven't:  https://github.com/instructlab/instructlab/pull/2242

I was mostly curious what would happen in that case.

bjhargrave: The wait-for-status-checks action does not require that a job has run. It only requires that if [check-runs](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#list-check-runs-for-a-git-reference) reports the job, the job is skipped or successful. 

In the mentioned PR, check-runs shows:

```sh
➜  ~ curl -s -L -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/instructlab/instructlab/commits/3c951f5a62a402523fb38ee0195e933a86b2003e/check-runs  | jq '.check_runs[] | {(.name): .status}'
{
  "e2e-workflow-complete": "completed"
}
{
  "Stop external EC2 runner": "completed"
}
{
  "E2E Test": "completed"
}
{
  "Summary": "completed"
}
{
  "Start external EC2 runner": "completed"
}
{
  "DCO": "completed"
}
```

so the wait-for-status-checks action would end with success.

I will make another change to split the e2e workflow into 2 parts: one pull_request, the other workflow_run. The first will be subject to approval for first time contributors and the second will have access to needed secrets. We used this technique  in the taxonomy repository for a workflow having similar concerns.

bjhargrave:
I updated to split the e2e workflow into 2 workflows. However, in the CI run for this PR, e2e-nvidia-t4-x1.yml is still pull_request_target in the main branch and so is used from the main branch and not the PR branch. Once this PR is merged and the updated e2e-nvidia-t4-x1.yml is in main as a workflow_run workflow, then we can see this correctly operating. This change will prevent the e2e from starting for PRs from new contributors until approved by a maintainer and also until the status checks are all met (linting, etc.).

.github/workflows/e2e-starter.yml
```diff
@@ -0,0 +1,71 @@
...
+      - main
+      - release-*
+    paths:
+      # note this should match the merging criteria in 'mergify.yml'
+      - '**.py'
+      - 'pyproject.toml'
+      - 'requirements**.txt'
+      - 'scripts/basic-workflow-tests.sh' # Used by this workflow
+      - 'scripts/test-data/**'
+      - '.github/workflows/e2e-nvidia-t4-x1-gate.yml' # This workflow
```
nathan-weinberg, line 23:
```suggestion
      - '.github/workflows/e2e-starter.yml' # This workflow
```
old name?

.github/workflows/e2e-starter.yml
```diff
@@ -0,0 +1,71 @@
...
+      checks: read
+    uses: ./.github/workflows/status-checks.yml
+    with:
+      job_ids: >- # Space-separated job ids to wait on for status checks
+        DCO
+        actionlint
+        markdown-lint
+        shellcheck
+        spellcheck
+        lint-workflow-complete
```
nathan-weinberg, line 49: I may be misunderstanding the implementation, but seems the flow here is:

`Lints pass` => `Unit/Functional/E2E tests run`

If I'm correct, do we want that? Or something like

`Lints pass` => `Unit/Functional tests pass` => `E2E starts`

nathan-weinberg (commented):
Thanks for doing this @bjhargrave! Very happy to see this change

.github/workflows/e2e-starter.yml
nathan-weinberg: To be clear, I realize the original issue calls for the former implementation - not opposed to that route, just figured I'd bring up the topic

.github/workflows/e2e-starter.yml
bjhargrave: I specifically left the name of the actual workflow so that it changes, the starter workflow will run. If the starter workflow changes there is no real need to run the e2e.

.github/workflows/e2e-starter.yml
bjhargrave: At this stage I think the current ordering is fine. We can refine to run e2e after test in a followup PR if desired.

.github/workflows/e2e-starter.yml
nathan-weinberg: What I'm saying is that there is no file named `e2e-nvidia-t4-x1-gate`: https://github.com/bjhargrave/instructlab_instructlab/tree/issues/2245/.github/workflows

So, maybe it's supposed to be `e2e-nvidia-t4-x1.yml`?

.github/workflows/e2e-starter.yml
nathan-weinberg: SGT, :+1:

.github/workflows/e2e-starter.yml
nathan-weinberg: No worries @bjhargrave we can wait til Wed, enjoy your time off!

nathan-weinberg (approved):

danmcp:
@bjhargrave Would you mind updating the description explaining the new logic

bjhargrave:
> Would you mind updating the description explaining the new logic

Done.

> I was thinking about editing this document actually https://github.com/instructlab/instructlab/blob/main/docs/ci.md#ci-for-instructlab

I also updated this doc.

.github/workflows/e2e-starter.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: |
+          jq '{"ref": $ref, "number": $number, "event": .}' --arg ref "${{ github.ref }}" --arg number "${{ github.event.number || '0' }}" -- "$GITHUB_EVENT_PATH" > starter.json
+
+      - name: "Upload Starter Metadata artifact"
+        uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
+        with:
+          name: e2e-starter
+          path: starter.json
+          if-no-files-found: error
+          retention-days: 5
```
danmcp, line 71: What happens if:
- PR created and fails on e2e
- 6 days later someone kicks off the e2e job again

I assume it will fail to find starter.json?

bjhargrave: Yes, if GitHub reclaimed the file. Of course, only a maintainer can re-run a failed job and it is likely the e2e will fail again if nothing changed in the PR.

90 days is max retention, so this is possible for any retention value + 1 day.

danmcp: Thanks!  Given it's just the one file, a longer retention might be worth it.

danmcp (approved):

nathan-weinberg:
This didn't automerge due to a bug in the Mergify config; opened a fix here: https://github.com/instructlab/instructlab/pull/2318

Merging this manually

PR #2283, opened 2024-09-18
add .bin support to `ilab model download`

### LINKED ISSUE #2022: `ilab model download` reports confusing error when downloading non-GGUF repo without `.safetensors` file
**Describe the bug**
Note that currently,  `ilab model download` errors out on not finding a file `merlinite-7b-lab-Q4_K_M.gguf` when trying to download a non-GGUF repo that does not have a `.safetensors` file. This might be confusing to the user, who expects the whole repo to be downloaded in such cases, not InstructLab looking for a specific file.

**To Reproduce**
1. Run `ilab model download --repository=UnicomLLM/Unichat-llama3-Chinese-8B-28K`. (This is a repo containing the model in `.bin` format).
2. `ilab model download` errors out with `Entry Not Found for url: https://huggingface.co/UnicomLLM/Unichat-llama3-Chinese-8B-28K/resolve/main/merlinite-7b-lab-Q4_K_M.gguf.`, confusing the user because this is not a GGUF repo, and InstructLab can download other non-GGUF repos, like `instructlab/granite-7b-lab`.

**Expected behavior**
Error that the repo does not contain a `.safetensors` file nor a GGUF file, and thus the download is not supported.

**Screenshots**

 - Hardware Specs: ThinkPad T14s Gen 2i
 - OS Version: Fedora 40
 - Python Version: Python 3.12.4
 - InstructLab Version: ilab, version 0.17.1

**Additional context**
It is unclear from the interface that there are two code paths: one when there is a `.safetensors` file, then, the repo is considered a non-GGUF one and the whole repo is downloaded, and the other one where there is no such file where the repo is considered a GGUF repo and InstructLab automatically tries to download a GGUF file.

cdoern:
if an HF repo doesn't have .safetensors, just .bin, `ilab model download` should still understand that it needs to download the entire repo.

add support for downloading .bin models

resolves #2022

**Checklist:**

### DIFF
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -80,8 +80,8 @@ class HFDownloader(Downloader):
             if self.ctx.obj is not None:
                 hf_logging.set_verbosity(self.ctx.obj.config.general.log_level.upper())
             files = list_repo_files(repo_id=self.repository, token=self.hf_token)
-            if any(".safetensors" in fname for fname in files):
-                self.download_safetensors()
+            if any(re.search(r"\.(safetensors|bin)$", fname) for fname in files):
+                self.download_entire_hf_repo()
             else:
                 self.download_gguf()
 
@@ -109,7 +109,7 @@ class HFDownloader(Downloader):
             )
             raise click.exceptions.Exit(1)
 
-    def download_safetensors(self) -> None:
+    def download_entire_hf_repo(self) -> None:
         try:
             local_dir = os.path.join(self.download_dest, self.repository)
             os.makedirs(name=local_dir, exist_ok=True)


src/instructlab/model/download.py
```diff
@@ -80,8 +80,10 @@ def download(self):
             if self.ctx.obj is not None:
                 hf_logging.set_verbosity(self.ctx.obj.config.general.log_level.upper())
             files = list_repo_files(repo_id=self.repository, token=self.hf_token)
-            if any(".safetensors" in fname for fname in files):
-                self.download_safetensors()
+            if any(".safetensors" in fname for fname in files) or any(
+                ".bin" in fname for fname in files
+            ):
```
khaledsulayman, line 85:
```suggestion
            if (any(".safetensors" in fname for fname in files) or
                any(".bin" in fname for fname in files)):
```

src/instructlab/model/download.py
```diff
@@ -80,8 +80,10 @@ def download(self):
             if self.ctx.obj is not None:
                 hf_logging.set_verbosity(self.ctx.obj.config.general.log_level.upper())
             files = list_repo_files(repo_id=self.repository, token=self.hf_token)
-            if any(".safetensors" in fname for fname in files):
-                self.download_safetensors()
+            if any(".safetensors" in fname for fname in files) or any(
+                ".bin" in fname for fname in files
```
khaledsulayman, line 84: just making sure: In this case, it will download the entire repo even if there's both safetensors _and_ bin. Is this expected/desired behavior?

src/instructlab/model/download.py
```diff
@@ -80,8 +80,10 @@ def download(self):
             if self.ctx.obj is not None:
                 hf_logging.set_verbosity(self.ctx.obj.config.general.log_level.upper())
             files = list_repo_files(repo_id=self.repository, token=self.hf_token)
-            if any(".safetensors" in fname for fname in files):
-                self.download_safetensors()
+            if any(".safetensors" in fname for fname in files) or any(
```
jaideepr97, line 83: can we use regexp here instead of duplicating the any check?

src/instructlab/model/download.py
khaledsulayman: nit but easier to read

jaideepr97, line 85:
```suggestion
         if any(re.search(r"\.(safetensors|bin)$", fname) for fname in files):
```

src/instructlab/model/download.py
```diff
@@ -109,7 +111,7 @@ def download_gguf(self) -> None:
             )
             raise click.exceptions.Exit(1)
 
-    def download_safetensors(self) -> None:
+    def download_safetensors_or_bin(self) -> None:
```
khaledsulayman, line 114: minor minor minor nit: I prefer calling it something along the lines of `download_entire_hf_repo` or just something that indicates that's what it's doing

the logic inside `download` decides when to download the whole repo (i.e. safe tensors or bin) and then just calls the function that downloads the whole repo

khaledsulayman (commented):
left some minor comments. All in all looks great! Thanks!

src/instructlab/model/download.py
cdoern: regexp makes me sad but I will use this :laughing:

cdoern: used @jaideepr97 's regexp

src/instructlab/model/download.py
cdoern: yes. these often overlap

src/instructlab/model/download.py
cdoern: renamed to `download_entire_hf_repo`

nathan-weinberg (approved):

nathan-weinberg:
I would like to see a test added, but won't block on that

jaideepr97 (approved):

PR #2289, opened 2024-09-19
Begin decoupling listing logic from model list implementation

RobotSail:
We currently handle all operations within the implementations of the CLI commands
themselves. This commit begins to refactor this by moving the implementation of
ilab model list to have the raw machinery exist in a seperate module named `machinery`

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -2,10 +2,8 @@
 
 # Standard
 from typing import Optional, Tuple
-import json
 import logging
 import pathlib
-import struct
 import sys
 
 # Third Party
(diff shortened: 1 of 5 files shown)


cdoern (commented):
we should probably talk about this in an open forum before we decide if its needed in a PR like this.

I think this was done as a step towards our REST API work, but if implemented properly the REST API probably won't necessitate sep machinery files as the API should almost be a wrapper of everything that currently exists. It can also be viewed as another entry point to the underlying click cmds.

jaideepr97:
> we should probably talk about this in an open forum before we decide if its needed in a PR like this.
> 
> I think this was done as a step towards our REST API work, but if implemented properly the REST API probably won't necessitate sep machinery files as the API should almost be a wrapper of everything that currently exists. It can also be viewed as another entry point to the underlying click cmds.

@cdoern This was done as a first step towards https://github.com/instructlab/dev-docs/pull/131

RobotSail:
Yeah @jaideepr97 is correct, this is so we can make a follow-up PR updating our logic of selecting models.

src/instructlab/machinery/listing.py
```diff
@@ -0,0 +1,116 @@
...
+                [
+                    actual_model_name,
+                    modification_time_readable,
+                    f"{adjusted_all_sizes:.1f} {magnitude}",
+                ]
+            )
+    return models
+
+
+def list_models(model_dirs: List[Path], list_checkpoints: bool):
```
jaideepr97, line 92: Can we specify the return type here?

jaideepr97 (commented):
thanks @RobotSail 
is `machinery` supposed to function similar to a `utils` package then?

src/instructlab/machinery/listing.py
RobotSail: Done

RobotSail (commented):
@jaideepr97 I've moved the logic to exist under `utils.py`, but the way Python works causes us to encounter circular dependencies when the `is_model_gguf` and `is_model_safetensors` files are imported.

jaideepr97 (approved):
Thanks @RobotSail!

cdoern (approved):

PR #2290, opened 2024-09-19
feat: make flash-attention toggle a first-class option

RobotSail:
Previously, the `disable_flash_attn` training flag was set as an additional argument that was hidden
away. However; we run into problems when running E2E testing in training where it cannot run on a T4
system (Volta GPU). This commit resolves that issue by promoting `disable_flash_attn` as a
first-class argument that can be passed as a CLI argument

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -448,6 +448,10 @@ class _train(BaseModel):
         default=1,
         description="Number of GPUs to use for training. This value is not supported in legacy training or MacOS.",
     )
+    disable_flash_attn: Optional[bool] = Field(
+        default=False,
+        description="Whether or not we should disable the use of flash-attention during training. This is useful when using older GPUs.",
+    )
     additional_args: dict[str, typing.Any] = Field(
         default_factory=dict,
         description="Additional arguments to pass to the training script. These arguments are passed as key-value pairs to the training script.",
--- a/src/instructlab/model/train.py
+++ b/src/instructlab/model/train.py
@@ -296,6 +296,7 @@ def clickpath_setup(is_dir: bool) -> click.Path:
     config_sections=ADDITIONAL_ARGUMENTS,
     required=True,  # default from config
 )
+@click.option("--disable-flash-attn", is_flag=True, cls=clickext.ConfigOption)
 @click.option(
     "--strategy",
     type=click.Choice(
--- a/tests/testdata/default_config.yaml
+++ b/tests/testdata/default_config.yaml
@@ -264,6 +264,10 @@ train:
   # Allow CPU offload for deepspeed optimizer.
   # Default: False
   deepspeed_cpu_offload_optimizer: false
+  # Whether or not we should disable the use of flash-attention during training.
+  # This is useful when using older GPUs.
+  # Default: False
+  disable_flash_attn: false
   # Pick a distributed training backend framework for GPU accelerated full fine-
   # tuning.
   # Default: deepspeed


src/instructlab/model/train.py
```diff
@@ -282,6 +282,12 @@ def clickpath_setup(is_dir: bool) -> click.Path:
     is_flag=True,
     help="if true, enables the legacy linux training code path from release 0.17.0 and prior.",
 )
+@click.option(
+    "--disable-flash-attn",
+    is_flag=True,
+    help="If true, this disables the use of flash attention, resulting in slower training. This is useful on older systems where flash-attention isn't supported.",
```
JamesKunstle, line 288: the default and the help can come from the configuration

src/instructlab/configuration.py
```diff
@@ -414,6 +414,10 @@ class _train(BaseModel):
         default=1,
         description="Number of GPUs to use for training. This value is not supported in legacy training or MacOS.",
     )
+    disable_flash_attn: Optional[bool] = Field(
```
JamesKunstle, line 417:
```suggestion
    disable_flash_attn: bool = Field(
```

since we have it in the configuration and it gets passed down, it's not optional imo

src/instructlab/model/train.py
```diff
@@ -282,6 +282,12 @@ def clickpath_setup(is_dir: bool) -> click.Path:
     is_flag=True,
     help="if true, enables the legacy linux training code path from release 0.17.0 and prior.",
 )
+@click.option(
```
JamesKunstle, line 285: you're just unwrapping this into the TrainingArgs right? we don't need to add a function argument?

src/instructlab/model/train.py
RobotSail: Good catch, I'll get rid of this

src/instructlab/configuration.py
RobotSail: @JamesKunstle The `Optional` here is a bit misleading, it's not that its existence is optional but that it isn't required to appear in the config to be set. This way, it can be backwards-compatible with existing configs. The way this works is if you have a training config, e.g.

```py
config = TrainingConfig()
config.disable_flash_attn
> False
```

Then this field will automatically be set to False. Otherwise, you're able to specify its value to be whatever you want:

```py
config = TrainingConfig(
  disable_flash_attn = True
)
config.disable_flash_attn
> True
```

src/instructlab/model/train.py
RobotSail: I removed its presence from the function argument, but it does need to be a flag we set so that we can disable it from within E2E tests.

JamesKunstle: yeah that makes sense

src/instructlab/configuration.py
RobotSail: @cdoern Said that the `bool | None` is used for a separate reason though, so it may be good to get his thoughts on this as well.

src/instructlab/model/train.py
```diff
@@ -282,6 +282,10 @@ def clickpath_setup(is_dir: bool) -> click.Path:
...
```
cdoern, line 285: can we add some help text? most people will not know how to interpret this

tests/testdata/default_config.yaml
```diff
@@ -260,6 +260,10 @@ train:
   # Allow CPU offload for deepspeed optimizer.
   # Default: False
   deepspeed_cpu_offload_optimizer: false
+  # Whether or not we should disable the use of flash-attention during training.
```
cdoern, line 263: this is good text to also populate into the flag. Or does the clickext class do this?

src/instructlab/model/train.py
cdoern: if you are looking for this to inherit from the cfg, you'll need the clickext.ConfigOption class on here.

src/instructlab/configuration.py
cdoern: what oleg said here is spot on, the `optional[bool]` plays better w. backwards compat of the cfg. Though, if its a field, it shouldn't be an issue as a default is always populated.

cdoern (commented):
thanks for adding this, just some thoughts

src/instructlab/model/train.py
RobotSail: Gotcha, I'll make sure this is on it

tests/testdata/default_config.yaml
RobotSail: I believe @JamesKunstle mentioned that clickext does this automatically, but I'll double-check.

RobotSail: Yeah, it's added automatically here: [github.com/RobotSail/instructlab](https://github.com/RobotSail/instructlab/blob/dedf9b43ca27b06ad5798bab704997f754e8af6a/src/instructlab/configuration.py#L639)

src/instructlab/configuration.py
JamesKunstle: ++

JamesKunstle (approved):
lgtm

src/instructlab/model/train.py
```diff
@@ -291,6 +291,12 @@ def clickpath_setup(is_dir: bool) -> click.Path:
     config_sections=ADDITIONAL_ARGUMENTS,
     required=True,  # default from config
 )
+@click.option(
+    "--legacy",
+    is_flag=True,
+    help="if true, enables the legacy linux training code path from release 0.17.0 and prior.",
+)
```
jaideepr97, line 298: @RobotSail this was removed with @cdoern's training rewrite 
maybe you need to rebase?

jaideepr97 (changes requested):

src/instructlab/model/train.py
JamesKunstle: ++ ah definitely

jaideepr97 (approved):

nathan-weinberg:
The goal here is to get the training library run on the T4s, but I thought those systems didn't have enough juice to run with the training library?

cdoern (commented):
Is this handled implicitly by `map_train_to_library`? If so I am good with this

BTW. Please account for the full pipeline. This always needs to be true.

RobotSail:
@nathan-weinberg That's a great point actually. The T4's are the only runners that we currently have as CI jobs in the training library, but if we could update those to have at least 24GB of capacity, then we could potentially run a LoRA job on them. This still may be a good option though if running on older harder, I believe you could still fit it within 16GB if quantizing with bitsandbytes.

RobotSail:
@cdoern Yes, `disable_flash_attn` is a top-level field in the `TrainingArgs` object so it would be mapped here: [train_args](https://github.com/RobotSail/instructlab/blob/f2807fe259c433625412e200b9c86efdeb599a88/src/instructlab/configuration.py#L1252)

nathan-weinberg:
@RobotSail we could upgrade the runner in the training library to a A10x1 - that might be enough to run the E2E with library training? Would be a great integration test to have

cdoern (approved):

PR #2292, opened 2024-09-20
cloud-instance.sh: Add ssh config management

russellb:
When using stop/start to turn off and on dev instances, the public IP
address will change. I keep an entry in my ssh config to make it easy
to connect to these instances (both with ssh manually, as well as with
vscode over ssh). I found myself using "ec2 details" to get the new IP
and then updating my config. After a handful of times of this, I
decided to automate it.

Once an instance is running, you can run `ec2 update-ssh-config` and
it will either add the instance to the ssh config, or update it with
the latest address (via the hostname derived from the public IP, since
there was already code in the script to grab that).

### DIFF
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -189,6 +189,49 @@ ec2_calculate_instance_public_dns() {
     PUBLIC_DNS=$(echo "$PUBLIC_DNS" | xargs)
 }
 
+calculate_instance_public_dns() {
+    local cloud_type=$1
+    if [ "$cloud_type" = 'ec2' ]; then
+        ec2_calculate_instance_public_dns
+    elif [ "$cloud_type" = 'ibm' ]; then
+        ibm_calculate_instance_public_dns
+    else
+        echo "calculate_cloud_public_dns: Unknown cloud type: $cloud_type"
+        exit 1
+    fi
+}
+
+update_ssh_config() {
+    handle_help_and_instance_name_opts "$@"
+    calculate_instance_public_dns "$1"
+    local ssh_config="$HOME/.ssh/config"
+    local user_name
+    user_name=$(instance_user_name "$1")
+
+    if ! grep -q "Host ${INSTANCE_NAME}" "$ssh_config"; then
+        cat << EOF >> "$ssh_config"
+
+Host ${INSTANCE_NAME}
+    HostName ${PUBLIC_DNS}
+    User ${user_name}
+    IdentityFile ${EC2_KEY_LOCATION}
+    StrictHostKeyChecking no
+    UserKnownHostsFile /dev/null
+EOF
+        echo "Added entry for ${INSTANCE_NAME} in ${ssh_config}"
+    else
+        local temp_file
+        temp_file=$(mktemp)
+        awk -v ip="$PUBLIC_DNS" '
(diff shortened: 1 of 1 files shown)


scripts/infra/cloud-instance.sh
```diff
@@ -189,6 +189,35 @@ ec2_calculate_instance_public_dns() {
     PUBLIC_DNS=$(echo "$PUBLIC_DNS" | xargs)
 }
 
+ec2__update_ssh_config() {
```
danmcp, line 192: Any reason not to do this for all cloud types?  Seems like it could be handled generically.  Would just need to handle the dns setup, users, and keys in a general way.

scripts/infra/cloud-instance.sh
```diff
@@ -20,9 +20,9 @@ ec2() {
         exit 1
     fi
     shift
-    if [ "$(type -t "ec2__$cmdname")" = "function" ] >/dev/null 2>&1; then
+    if [ "$(type -t "ec2__${cmdname//-/_}")" = "function" ] >/dev/null 2>&1; then
```
danmcp, line 23: This wouldn't be necessary if the method was done for all cloud types.  If we do want to make this change, should also change it for ibm()

danmcp (changes requested):

scripts/infra/cloud-instance.sh
```diff
@@ -189,6 +189,35 @@ ec2_calculate_instance_public_dns() {
...
+    User ec2-user
+    IdentityFile ${EC2_KEY_LOCATION}
+    StrictHostKeyChecking no
+    UserKnownHostsFile /dev/null
+EOF
+        echo "Added entry for ${INSTANCE_NAME} in ~/.ssh/config"
+    else
+        local temp_file
+        temp_file=$(mktemp)
+        awk -v ip="$PUBLIC_DNS" '
```
leseb, line 211: This `awk` is not the easiest thing to read. It seems that the ssh_config is always updated regardless if `PUBLIC_DNS` actually changed, right? Can we avoid that?

Another approach I have in mind for clarity:

* update `~/.ssh/config` to add `Include ~/.ssh/config.d/*`, create `~/.ssh/config.d/` if it does not exist
* only do what the first portion of the condition does, which is creating an ssh config file and put it in `~/.ssh/config.d/`

Feel free to give it a try or not :)

russellb: I guess that would work, too, but I probably won't update this significantly. I wrote it in a separate script at first, so I don't mind not merging it if it's too convoluted or not interesting to enough people.

scripts/infra/cloud-instance.sh
russellb: sure, it looks like supporting both isn't that bad, though I won't be able to test it with IBM cloud. i don't have it set up

scripts/infra/cloud-instance.sh
russellb: ok - changing to support both

danmcp: So this change isn't needed since update_ssh_config is now generic. But if we want this to go in for the future, just need to make the corresponding change in ibm().

russellb: sorry, i thought i removed this. i think I lost it in a rebase

russellb: removed it for real this time!

danmcp (approved):

PR #2311, opened 2024-09-25
vllm: Add support for a vllm virtualenv

russellb:
This change adds a new configuration option for vllm: a path to a
Python virtual environment to use when running vllm. This allows you
to run a custom version of vllm for development and testing purposes,
even if that version of vllm has incompatible python dependencies with
the current version of instructlab.

As an example of how this is useful, I wanted to try the following
vllm PR to run mixtral on a host with less GPU memory than is
typically required:

    https://github.com/vllm-project/vllm/pull/8751

To install vllm into its own venv:

    python3.11 -m venv venv-vllm
    . venv-vllm/bin/activate
    git clone https://github.com/vllm-project/vllm
    cd vllm
    git fetch origin pull/8751/head
    git checkout FETCH_HEAD
    export CUDA_HOME=/usr/local/cuda
    export PATH=${CUDA_HOME}/bin:${PATH}
    pip install -e .

Back in my instructlab venv ...

    . venv-ilab/bin/activate
    ilab model download --repository mistralai/Mixtral-8x7B-Instruct-v0.1
    ilab config edit

Edit vllm config. Note the value of `--tensor-parallel-size` is for
the number of GPUs I have available.

    serve:
      backend: vllm
      model_path: /home/ec2-user/.cache/instructlab/models/mistralai/Mixtral-8x7B-Instruct-v0.1
      vllm:
        vllm_args: ['--tensor-parallel-size', '4', '-q', 'fp6_weights']
        vllm_venv: "/home/ec2-user/venv"

This now works as mixtral is quantized at load-time, cutting the gpu
memory requirement in half.


src/instructlab/model/backends/vllm.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         background: bool = False,
         vllm_args: typing.Iterable[str] | None = (),
         max_startup_attempts: int | None = None,
+        vllm_venv: Optional[str] = "",
```
danmcp, line 50: Would `None` be a more obvious default for each of these new args?

leseb (commented):
It seems like this introduces functionality in production that is more suited for development, or at most, debugging purposes. So, I'd prefer not to name it `venv`. I believe what you're aiming for is more of a "bring your own vLLM" scenario. Therefore, it might be better to call it `vllm_binary_path` (even though that is not a binary perse). Also, I think we need additional logging to warn users. Thoughts?

russellb:
> It seems like this introduces functionality in production that is more suited for development, or at most, debugging purposes. So, I'd prefer not to name it `venv`. I believe what you're aiming for is more of a "bring your own vLLM" scenario. Therefore, it might be better to call it `vllm_binary_path` (even though that is not a binary perse). Also, I think we need additional logging to warn users. Thoughts?

Yes, "bring your own vllm" is a fair summary. I think it could be useful beyond just development and debugging. If someone wants to try a brand new model as a teacher model that's only supported in a brand new release of `vllm`, this would allow that, as one example.

It would have to be a path to `python` instead of `vllm` because of how it's executed. Do you know if there was a good reason to run it via `python -m vllm.entrypoints.openai.api_server` instead of `vllm serve` ?

leseb:
> > It seems like this introduces functionality in production that is more suited for development, or at most, debugging purposes. So, I'd prefer not to name it `venv`. I believe what you're aiming for is more of a "bring your own vLLM" scenario. Therefore, it might be better to call it `vllm_binary_path` (even though that is not a binary perse). Also, I think we need additional logging to warn users. Thoughts?
> 
> Yes, "bring your own vllm" is a fair summary. I think it could be useful beyond just development and debugging. If someone wants to try a brand new model as a teacher model that's only supported in a brand new release of `vllm`, this would allow that, as one example.
> 
> It would have to be a path to `python` instead of `vllm` because of how it's executed. Do you know if there was a good reason to run it via `python -m vllm.entrypoints.openai.api_server` instead of `vllm serve` ?

I think the implementation started with v0.5.2 and `vllm serve` came out during `v0.5.3` I believe. We could switch to using `vllm serve` now. Unless someone wants to try an old version of vLLM with this patch 😆.

russellb:
> > > It seems like this introduces functionality in production that is more suited for development, or at most, debugging purposes. So, I'd prefer not to name it `venv`. I believe what you're aiming for is more of a "bring your own vLLM" scenario. Therefore, it might be better to call it `vllm_binary_path` (even though that is not a binary perse). Also, I think we need additional logging to warn users. Thoughts?
> > 
> > 
> > Yes, "bring your own vllm" is a fair summary. I think it could be useful beyond just development and debugging. If someone wants to try a brand new model as a teacher model that's only supported in a brand new release of `vllm`, this would allow that, as one example.
> > It would have to be a path to `python` instead of `vllm` because of how it's executed. Do you know if there was a good reason to run it via `python -m vllm.entrypoints.openai.api_server` instead of `vllm serve` ?
> 
> I think the implementation started with v0.5.2 and `vllm serve` came out during `v0.5.3` I believe. We could switch to using `vllm serve` now. Unless someone wants to try an old version of vLLM with this patch 😆.

Ah, I see. The version pointed to by the requirements file is still based on 0.5.2, I believe.

leseb:
> > > > It seems like this introduces functionality in production that is more suited for development, or at most, debugging purposes. So, I'd prefer not to name it `venv`. I believe what you're aiming for is more of a "bring your own vLLM" scenario. Therefore, it might be better to call it `vllm_binary_path` (even though that is not a binary perse). Also, I think we need additional logging to warn users. Thoughts?
> > > 
> > > 
> > > Yes, "bring your own vllm" is a fair summary. I think it could be useful beyond just development and debugging. If someone wants to try a brand new model as a teacher model that's only supported in a brand new release of `vllm`, this would allow that, as one example.
> > > It would have to be a path to `python` instead of `vllm` because of how it's executed. Do you know if there was a good reason to run it via `python -m vllm.entrypoints.openai.api_server` instead of `vllm serve` ?
> > 
> > 
> > I think the implementation started with v0.5.2 and `vllm serve` came out during `v0.5.3` I believe. We could switch to using `vllm serve` now. Unless someone wants to try an old version of vLLM with this patch 😆.
> 
> Ah, I see. The version pointed to by the requirements file is still based on 0.5.2, I believe.

No, our current version has `vllm serve`, and version is v0.5.3.

russellb:
Changed to `vllm_path` instead of `vllm_venv`

russellb:
marked as a draft as i need to fix the tests

src/instructlab/configuration.py
```diff
@@ -161,6 +161,10 @@ class _serve_vllm(BaseModel):
             ["--lora-alpha", "32"],
         ],
     )
+    vllm_path: Optional[str] = Field(
+        default=None,
+        description="Path to the vLLM virtual environment.",
```
leseb, line 166: I don't think the virtual env applies anymore, right?

src/instructlab/model/backends/vllm.py
```diff
@@ -359,13 +364,19 @@ def build_vllm_cmd(
...
             vllm_cmd (list of str): The command to run the vllm server
             tmp_files: a list of temporary files necessary to launch the process
     """
-    vllm_cmd = [sys.executable, "-m", "vllm.entrypoints.openai.api_server"]
+
+    if contains_argument("--model", vllm_args):
+        # The model is a positional argument to `vllm serve`
+        raise ValueError("The --model argument is not allowed in vllm_args")
+
+    vllm_cmd = [vllm_path if vllm_path else "vllm", "serve", os.fspath(model_path)]
```
leseb, line 379: I think this change is a bit out of the scope and diverges from the initial intent. The PR should make it more explicit.

src/instructlab/configuration.py
russellb: right, I'll update the text here

russellb: done

src/instructlab/model/backends/vllm.py
russellb: just to summarize our discussion from chat -- you requested that I do the change to `vllm serve` in its own commit and then add the new `vllm_path` option as a second commit on top.

src/instructlab/model/serve.py
```diff
@@ -196,6 +196,7 @@ def serve(
             vllm_args=vllm_args,
             host=host,
             port=port,
+            vllm_path=ctx.obj.config.serve.vllm.vllm_path,
```
alimaredia, line 199: @russellb would you want vllm_path to be a flag for the `ilab model serve` command too?

tests/test_backends.py
```diff
@@ -239,15 +239,13 @@ def test_build_vllm_cmd_with_defaults(tmp_path: pathlib.Path):
     chat_template.touch()
     vllm_args: list[str] = []
     expected_cmd = [
-        sys.executable,
-        "-m",
-        "vllm.entrypoints.openai.api_server",
+        "vllm",
+        "serve",
+        str(model_path),
```
alimaredia, line 244: @russellb why was `str(model_path)` used here and `os.fspath(model_path)` used in the test below?

tests/test_backends.py
```diff
@@ -269,14 +267,13 @@ def test_build_vllm_cmd_with_args_provided(tmp_path: pathlib.Path):
...
         "--distributed-executor-backend",
         "ray",
     ]
     expected_cmd = [
-        sys.executable,
-        "-m",
-        "vllm.entrypoints.openai.api_server",
+        "vllm",
+        "serve",
+        os.fspath(model_path),
```
alimaredia, line 276: @russellb why was os.fspath(model_path) used here and str(model_path) used for the other tests you changed here?

tests/test_lab_serve.py
```diff
@@ -18,9 +18,8 @@
 
 
 def assert_vllm_args(args):
-    assert "python" in args[0]
-    assert args[1] == "-m"
-    assert args[2] == "vllm.entrypoints.openai.api_server"
+    assert "vllm" in args[0]
```
alimaredia, line 21: why use `in` here and not `==`, and vice versa for the assert below? `args[0]` should always be `vllm` and `args[1]` should always be `serve`, right?

russellb:
if anyone else is interested in this, feel free to apply some changes on top - I'm unlikely to come back to this soon

tests/test_lab_serve.py
russellb: I'm sure I did this because the old line was `"python" in args[0]`. You're right that `==` would be a little better.

tests/test_backends.py
russellb: I don't remember. I'm guessing this gets converted to a string automatically so the result is the same. agree that str() would be better for consistency.

russellb: actually, it was because that's what the code in the vllm backend does

```
    vllm_cmd = [vllm_path if vllm_path else "vllm", "serve", os.fspath(model_path)]
```

which probably come from how the model path was handled in this removed code:

```
    if not contains_argument("--model", vllm_args):
        vllm_cmd.extend(["--model", os.fspath(model_path)])
```

and this is the correct way to convert the type to a string in this case

PR #2320, opened 2024-09-27
Added the --lora-enabled option to the commands: ilab model train and ilab model test through the CLI

### LINKED ISSUE #1947: Provide an easy way to disable LoRA

When we train today, the CLI assumes that the training method will be a LoRA. As such, it's very difficult to change the existing config in a way that will allow `ilab` to easily use full fine-tuning as opposed to the LoRA training.

Without changing any of the existing interpretations of the LoRA config arguments, we could introduce a new `lora_enabled` boolean flag and training config. Here's how this could be implemented in both cases:

### In the training config

`lora_enabled` could be a new field which exists under the `train` section of the current config.

When `lora_enabled` is omitted, `ilab` should assume that we are performing a full fine-tune. Otherwise, specifying `lora_enabled` should just yield the declared behavior, i.e. `lora_enabled: true` will obviously be a lora train, and `lora_enabled: false` is a full fine-tune. 

As a CLI flag it would operate as follows:
- By default, `--lora-enabled` is set to `false`
- When provided without any additional values, `--lora-enabled` enables lora as the training mode
- It can be given an explicit boolean value of `--lora-enabled {false|true}`

This makes it easy for users to enable/disable lora training without needing to edit any of the existing configs such as `--lora-rank=0`

Not sure if this is really something that can be handled in a backwards compatible way.

Sunnidhya:
Added the --lora-enabled option to the commands: ilab model train and ilab model test through the CLI

Added an additional click option in the target files train.py and test.py which will handle the --lora-enabled option being passed from the CLI

Resolves #1947 

**Checklist:**


RobotSail:
Thanks for the PR @Sunnidhya ! It looks like this only impacts the MacOS training paths but not the legacy Linux training or training library paths - could you please update the PR to include these as well? 

Also - we should also make this into a config setting in the [training config](https://github.com/instructlab/instructlab/blob/89b7457ba457845e324f33594fe8f218ab140912/src/instructlab/configuration.py#L373) object here:
https://github.com/instructlab/instructlab/blob/89b7457ba457845e324f33594fe8f218ab140912/src/instructlab/configuration.py#L373

src/instructlab/model/test.py
```diff
@@ -60,6 +60,11 @@
     "--model-family",
     help="Force model family to use when picking a generation template",
 )
+@click.option(
+    "--lora-enabled",
+    is_flag = True,
+    help="Enable or disable LORA during training"
```
jaideepr97, line 66: once this is an option in the configuration itself (as suggested in https://github.com/instructlab/instructlab/pull/2320#issuecomment-2379341121), please remove the `help` line from here, the help text is extracted directly from the description added in the config

jaideepr97 (commented):
thanks for the contribution @Sunnidhya !!

PR #2322, opened 2024-09-27
fix: two phased training fixes

alimaredia:
Phased training is broken, and this commit has the following fixes for it:

1. Phased training does not require a valid data-dir. It has phased training flags and config for the data-dir in phase 1 and phase 2.

2. The eval section from the config was trying to be read from `ctx.obj`, not `ctx.obj.config`. This lead to errors saying the `ctx.obj` does not have a `evaluate` member.

**Checklist:**

### DIFF
--- a/src/instructlab/model/train.py
+++ b/src/instructlab/model/train.py
@@ -474,10 +474,8 @@ def train(
         shutil.copy(test_files[0], test_file)
 
     if (
-        pipeline == "full"
-        or pipeline == "accelerated"
-        or strategy == SupportedTrainingStrategies.LAB_MULTIPHASE.value
-    ):
+        pipeline in ("full", "accelerated")
+    ) and strategy != SupportedTrainingStrategies.LAB_MULTIPHASE.value:
         if not os.path.isfile(data_path):
             ctx.fail(
                 f"Data path must be to a valid .jsonl file. Value given: {data_path}"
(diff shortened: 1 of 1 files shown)


cdoern (approved):
this is fine, but we need to abstract this better in the future so that we aren't calling the map function multiple times

src/instructlab/model/train.py
```diff
@@ -502,6 +500,13 @@ def get_files(directory: str, pattern: str) -> list[str]:
                 "Data for both phase1 and phase2 must be specified for phased training."
             )
 
+        # pull the trainrandom.randinting and torch args from the flags
```
Maxusmusti, line 503: Typo

cdoern:
once https://github.com/instructlab/instructlab/actions/runs/11072184232/job/30765958698 passes, I will merge this

src/instructlab/model/train.py
Maxusmusti: guessing that originated from line 484

nathan-weinberg:
> once https://github.com/instructlab/instructlab/actions/runs/11072184232/job/30765958698 passes, I will merge this

There's an outstanding comment from @Maxusmusti and @JamesKunstle should really look at this

cdoern:
e2e passed, but we might instead go with @JamesKunstle 's #2236 since it fixes the same issues

nathan-weinberg:
Superceded by https://github.com/instructlab/instructlab/pull/2236

PR #2340, opened 2024-09-30
removes MMLU evaluation from multi-phase training

JamesKunstle:
It's been experimentally idenfitied that running MMLU to pick out phase1 output checkpoint is not ideal. This removes MMLU and communicates the more optimal 7 epochs of training for phase 1.

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -457,9 +457,9 @@ class _train(BaseModel):
     # TODO: could move into its own object.
     # Not strictly necessary for a correct training object.
     phased_phase1_num_epochs: int | None = Field(
-        default=10,
+        default=7,
         gt=0,
-        description="Number of epochs to run training for during phase1.",
+        description="Number of epochs to run training for during phase1 (experimentally optimal number is 7).",
     )
     # anything greater than 0 enables samples_per_save for the phase.
     phased_phase1_samples_per_save: int = Field(
--- a/src/instructlab/model/train.py
+++ b/src/instructlab/model/train.py
@@ -560,6 +560,14 @@ def train(
         # try to load journal. May be empty.
         journal = TrainingJournal(journalfile=training_journal)
         click.secho("\n\n~~~~~~~~~~~~~STARTING MULTI-PHASE TRAINING~~~~~~~~~~~~~")
+
+        # experimental preference.
+        if phased_phase1_num_epochs != 7:
+            click.secho(
+                f"Running phased training with '{phased_phase1_num_epochs}' epochs.\nNote: 7 epochs is the recommended amount for optimal performance.",
+                fg="yellow",
+            )
+
         if journal.was_loaded:
             click.secho(
                 f"There was an existing training journal at: '{str(training_journal)}'"
@@ -1095,42 +1103,42 @@ def _run_phased_training(
 
         phase_model.ended_at_utc = TrainingJournal.now_utc()
 
-        journal.current_phase = TrainingPhases.EVAL1
+        journal.current_phase = TrainingPhases.TRAIN2
         journal.commit()
 
         logger.debug("Finished training #1\n%s", journal.print_model_rich())
     else:
         click.secho("SKIPPING: Training Phase 1/2; already in Journal", fg="cyan")
 
-    if journal.current_phase == TrainingPhases.EVAL1:
-        click.secho("MMLU evaluation for Phase 1...", fg="cyan")
+    # if journal.current_phase == TrainingPhases.EVAL1:
+    #     click.secho("MMLU evaluation for Phase 1...", fg="cyan")
 
-        # NOTE: requires hf_format sub-directory. Training sets this up.
-        phase1_checkpoints_dir = phase1_checkpoints_dir / "hf_format"
+    # NOTE: requires hf_format sub-directory. Training sets this up.
+    # phase1_checkpoints_dir = phase1_checkpoints_dir / "hf_format"
 
-        phase_model = journal.journal.eval_1
-        if phase_model is None:
-            # if it's not None, it already exists and may have 'results', so we shouldn't overwrite it.
-            phase_model = EvalPhaseModel(
-                checkpoints=list(phase1_checkpoints_dir.iterdir())
-            )
-            journal.journal.eval_1 = phase_model
-        journal.commit()
+    # phase_model = journal.journal.eval_1
+    # if phase_model is None:
+    #     # if it's not None, it already exists and may have 'results', so we shouldn't overwrite it.
+    #     phase_model = EvalPhaseModel(
+    #         checkpoints=list(phase1_checkpoints_dir.iterdir())
+    #     )
+    #     journal.journal.eval_1 = phase_model
+    # journal.commit()
 
-        best_checkpoint = _evaluate_dir_of_checkpoints(
-            eval_func=_mmlu, phase_model=phase_model, journal=journal
-        )
+    # best_checkpoint = _evaluate_dir_of_checkpoints(
+    #     eval_func=_mmlu, phase_model=phase_model, journal=journal
+    # )
 
-        phase_model.best_checkpoint = best_checkpoint
-        phase_model.ended_at_utc = TrainingJournal.now_utc()
+    # phase_model.best_checkpoint = best_checkpoint
+    # phase_model.ended_at_utc = TrainingJournal.now_utc()
 
-        journal.current_phase = TrainingPhases.TRAIN2
-        journal.commit()
-        logger.debug("Finished eval #1\n%s", journal.print_model_rich())
-    else:
-        click.secho(
-            "SKIPPING: MMLU evaluation for Phase 1; already in Journal", fg="cyan"
-        )
+    #     journal.current_phase = TrainingPhases.TRAIN2
+    #     journal.commit()
+    #     logger.debug("Finished eval #1\n%s", journal.print_model_rich())
+    # else:
+    #     click.secho(
+    #         "SKIPPING: MMLU evaluation for Phase 1; already in Journal", fg="cyan"
+    #     )
 
     if journal.current_phase == TrainingPhases.TRAIN2:
         click.secho("Training Phase 2/2...", fg="cyan")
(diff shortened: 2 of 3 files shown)


src/instructlab/model/train.py
```diff
@@ -560,6 +560,25 @@ def get_files(directory: str, pattern: str) -> list[str]:
...
+                fg="black",
+                bg="yellow",
+            )
+            click.secho(
+                f"'{phased_phase1_num_epochs}' epochs were configured.",
+                fg="black",
+                bg="yellow",
+            )
+            click.secho(
+                f"'{phased_phase1_num_epochs}' This can result in sub-optimal final checkpoint performance.",
```
danmcp, line 577: Is '{phased_phase1_num_epochs}' supposed to be at the beginning of this sentence?

src/instructlab/model/train.py
```diff
@@ -560,6 +560,25 @@ def get_files(directory: str, pattern: str) -> list[str]:
...
+                fg="black",
+                bg="yellow",
+            )
```
jaideepr97, line 580: why are these 3 separate `click.secho` statements? can they be consolidated into one?

src/instructlab/model/train.py
```diff
@@ -1141,17 +1160,26 @@ def _run_phased_training(
...
-            )
+        # if journal.journal.eval_1 is None:
+        #     raise RuntimeError(
+        #         "Training journal field 'eval_1' cannot be None for phase 'train_2'"
+        #     )
+
+        # NOTE:
+        # this is a recent change, implemented to ignore MMLU. We just look at the checkpoints
+        # from the phase 1 training and take the most recent one.
+        phase1_checkpoints = sorted(
```
alimaredia, line 1171: Are we sure this will always sort directories into the order of most recent -> least recently used? After quickly looking, I haven't seen anything that says otherwise. I'm just curious if there are any docs that explicitly say this is the best way to go.

src/instructlab/model/train.py
```diff
@@ -1095,42 +1114,42 @@ def _run_phased_training(
...
-        click.secho(
-            "SKIPPING: MMLU evaluation for Phase 1; already in Journal", fg="cyan"
-        )
+    #     journal.current_phase = TrainingPhases.TRAIN2
+    #     journal.commit()
+    #     logger.debug("Finished eval #1\n%s", journal.print_model_rich())
+    # else:
+    #     click.secho(
+    #         "SKIPPING: MMLU evaluation for Phase 1; already in Journal", fg="cyan"
+    #     )
```
jaideepr97, line 1152: is commenting the code out necessary? seems to me that setting `journal.current_phase` to `TRAIN2` effectively skips the mmlu check anyway. I think that coupled with a comment explaining why we are skipping to train 2 + maybe a debug log statement about it should be enough

src/instructlab/model/train.py
```diff
@@ -560,6 +560,25 @@ def get_files(directory: str, pattern: str) -> list[str]:
         # try to load journal. May be empty.
         journal = TrainingJournal(journalfile=training_journal)
         click.secho("\n\n~~~~~~~~~~~~~STARTING MULTI-PHASE TRAINING~~~~~~~~~~~~~")
+
+        # experimental preference.
+        if phased_phase1_num_epochs != 7:
```
RobotSail, line 565: Does it necessarily have to be a 7, or can we just take the checkpoint at the end of this phase?

RobotSail (approved):
LGTM

src/instructlab/model/train.py
alimaredia: Alternatively can we just take the 7th checkpoint directory and not do any sorting?

src/instructlab/model/train.py
JamesKunstle: We take the last checkpoint, but if the user doesn't specify 7 we let them know that they might want that value.

src/instructlab/model/train.py
JamesKunstle: Mostly because it's a hotfix- I don't want to scrape the logic from the codebase in case other people are interested in it in the immediate future.

src/instructlab/model/train.py
JamesKunstle: Multiline comments are hard

src/instructlab/configuration.py
```diff
@@ -457,9 +457,9 @@ class _train(BaseModel):
     # TODO: could move into its own object.
     # Not strictly necessary for a correct training object.
     phased_phase1_num_epochs: int | None = Field(
-        default=10,
+        default=7,
         gt=0,
-        description="Number of epochs to run training for during phase1.",
+        description="Number of epochs to run training for during phase1. (Experimentally optimal number is 7)",
```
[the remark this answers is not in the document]
RobotSail: This makes sense, @JamesKunstle I would move the warning you have from the training codebase and into this.

src/instructlab/model/train.py
```diff
@@ -1095,42 +1114,42 @@ def _run_phased_training(
...
+        journal.current_phase = TrainingPhases.TRAIN2
         journal.commit()
 
         logger.debug("Finished training #1\n%s", journal.print_model_rich())
     else:
         click.secho("SKIPPING: Training Phase 1/2; already in Journal", fg="cyan")
 
-    if journal.current_phase == TrainingPhases.EVAL1:
-        click.secho("MMLU evaluation for Phase 1...", fg="cyan")
+    # if journal.current_phase == TrainingPhases.EVAL1:
```
cdoern, line 1124: this hurts my soul a bit but I understand we want to keep this for future purposes

src/instructlab/model/train.py
```diff
@@ -1141,17 +1160,26 @@ def _run_phased_training(
...
+        #         "Training journal field 'eval_1' cannot be None for phase 'train_2'"
+        #     )
+
+        # NOTE:
+        # this is a recent change, implemented to ignore MMLU. We just look at the checkpoints
+        # from the phase 1 training and take the most recent one.
+        phase1_checkpoints = sorted(
+            list((phase1_checkpoints_dir / "hf_format").iterdir())
+        )
+        print(phase1_checkpoints)
```
cdoern, line 1174: print? is this supposed to be there

src/instructlab/model/train.py
jaideepr97: im arguing for even fewer changes to it than you lol
im not asking for it to be removed - im suggesting to leave it uncommented and steer control away from it via the current_phase - that's it
just cause im not a fan of the idea of shipping commented out code in an actual release 

not blocking tho

src/instructlab/model/train.py
```diff
@@ -560,6 +560,25 @@ def get_files(directory: str, pattern: str) -> list[str]:
         # try to load journal. May be empty.
         journal = TrainingJournal(journalfile=training_journal)
         click.secho("\n\n~~~~~~~~~~~~~STARTING MULTI-PHASE TRAINING~~~~~~~~~~~~~")
+
+        # experimental preference.
+        if phased_phase1_num_epochs != 7:
+            click.secho(
+                "Experimentally, Phase 1 of training is most effective after '7' epochs.",
```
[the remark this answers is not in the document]
JamesKunstle: no you're right, fixed

src/instructlab/configuration.py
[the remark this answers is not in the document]
JamesKunstle: ++

src/instructlab/model/train.py
JamesKunstle: It was, but on consideration I don't think that's the right move.

src/instructlab/model/train.py
JamesKunstle: OH gotcha, yeah I just want it commented so that it's clear that it's not the codepath being used.

src/instructlab/model/train.py
JamesKunstle: Can we live with it being multiple click.sechos? I won't want to retrigger the tests.

cdoern (commented):
one comment one ?, we should scope either removing the commented code or adding it back in some sort of way

src/instructlab/model/train.py
JamesKunstle: removed

src/instructlab/model/train.py
JamesKunstle: yeah I don't like it either

cdoern (approved):

src/instructlab/configuration.py
```diff
@@ -457,9 +457,9 @@ class _train(BaseModel):
     # TODO: could move into its own object.
     # Not strictly necessary for a correct training object.
     phased_phase1_num_epochs: int | None = Field(
-        default=10,
+        default=7,
```
danmcp, line 460: nit: Might be good to move 7 to a constant given the number of references

JamesKunstle: I'll make an issue for that, you're right

src/instructlab/model/train.py
```diff
@@ -1141,17 +1160,35 @@ def _run_phased_training(
...
+        # NOTE:
+        # this is a recent change, implemented to ignore MMLU. We just look at the checkpoints
+        # from the phase 1 training and take the most recent one.
+        phase1_checkpoints_dir_hf = phase1_checkpoints_dir / "hf_format"
+        if not phase1_checkpoints_dir_hf.exists():
+            raise FileNotFoundError(
+                f"{phase1_checkpoints_dir_hf} doesn't exist. This likely means that no checkpoints were saved from phase 1."
             )
 
+        phase1_checkpoints = sorted(list(phase1_checkpoints_dir_hf.iterdir()))
```
Maxusmusti, line 1177: Flip sort order

Maxusmusti (changes requested):
Hold until sort order is flipped, otherwise good

src/instructlab/model/train.py
Maxusmusti: Ok, fix is working now, picking correct ckpt

src/instructlab/model/train.py
```diff
@@ -560,6 +560,25 @@ def get_files(directory: str, pattern: str) -> list[str]:
         # try to load journal. May be empty.
         journal = TrainingJournal(journalfile=training_journal)
         click.secho("\n\n~~~~~~~~~~~~~STARTING MULTI-PHASE TRAINING~~~~~~~~~~~~~")
+
+        # experimental preference.
+        if phased_phase1_num_epochs != 7:
+            click.secho(
```
RobotSail, line 566: Would it make sense to tone this down a bit?

I feel like the presentation of this in the present state (giant text blob on a massive yellow background) gives the idea that something might be seriously wrong and may scare users away. 

What if we had it just have yellow or blue text with the phrase: "Running phased training with 2 epochs. Note: 7 epochs is the recommended amount for optimal performance."

RobotSail (approved):
One comment but otherwise LGTM

src/instructlab/model/train.py
JamesKunstle: Yeah honestly I think you're right. I'll make that change.

RobotSail: And also - this seems to apply even if the user is running more then 7 epochs, which doesn't make sense.

JamesKunstle: I think in that case we still want the message. Users are welcome to run more epochs, but we're telling them that for the technique that we're implementing, 7 is the right number.

RobotSail: That's something we should document in a user-guide but asserting that it's the absolute correct epoch I would say is still wrong, especially if going above 7. Hence why I believe the best phrasing here would be "We see you're using under 7 epochs, consider training with 7 for the best performance"

JamesKunstle: I made the update for your previous recommendation, I think that accomplishes that you're saying.

RobotSail: Let's keep it as-is, we can refine it later

RobotSail:
Aaaaand it's merged.

PR #2391, opened 2024-10-04
fix capitalization sensitivity in profile

cdoern:
the L40 profile will be detected properly but an extra print will happen saying that the profile isn't an exact match when it is.

This is because in the profile map, the "s" in L40s isn't capitalized, but torch detects the cards as "L40S".

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -1356,7 +1356,7 @@ def storage_dirs_exist() -> bool:
 
 # {"GPU Name": ({ "gpu_count": NUM_GPUS, "vram_and_config": {"vram": TOTAL_VRAM, "config": TRAIN_CONFIG}})...}
 PROFILE_MAPPINGS = {
-    "L40s": (
+    "L40S": (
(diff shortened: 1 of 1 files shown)


src/instructlab/configuration.py
```diff
@@ -1356,7 +1356,7 @@ def storage_dirs_exist() -> bool:
 
 # {"GPU Name": ({ "gpu_count": NUM_GPUS, "vram_and_config": {"vram": TOTAL_VRAM, "config": TRAIN_CONFIG}})...}
 PROFILE_MAPPINGS = {
-    "L40s": (
+    "L40S": (
```
danmcp, line 1359: Would it be better to put all the keys in as uppercase and always upper(gpu_name) before checking in the dict?  It looks like this is the only one that wasn't already uppercase and so the only other change would be to add the upper() call to the get() (add a comment).

danmcp (changes requested):

cdoern:
training profiles have been removed. this PR is no longer necessary

PR #2409, opened 2024-10-07
fix: Add dependency on instructlab-training for ROCm

fabiendupont:
**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Functional tests have been added, if necessary.
- [ ] E2E Workflow tests have been added, if necessary.

PR #2414, opened 2024-10-08
fix: mismatch between requirements action trigger and automerge criteria

nathan-weinberg:
Seen in https://github.com/instructlab/instructlab/pull/2409

Mergify expects `lint` and `test` to run on changes to requirements files at the top level of the repository as well as in the `requirements/` directory, but those actions were not getting triggered in the latter scenario.

**Checklist:**

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,7 +10,7 @@ on:
     paths:
       - '**.py'
       - 'pyproject.toml'
-      - 'requirements*.txt'
+      - 'requirements**.txt'
       - 'tox.ini'
       - .pylintrc
       - '.github/workflows/lint.yml' # This workflow
@@ -21,7 +21,7 @@ on:
     paths:
       - '**.py'
       - 'pyproject.toml'
-      - 'requirements*.txt'
+      - 'requirements**.txt'
       - 'tox.ini'
       - .pylintrc
       - '.github/workflows/lint.yml' # This workflow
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -10,7 +10,7 @@ on:
     paths:
       - '**.py'
       - 'pyproject.toml'
-      - 'requirements*.txt'
+      - 'requirements**.txt'
```
danmcp, line 13: Is this actually finding requirements.txt and requirements/*.txt?

bjhargrave: Yes. `**` glob includes `/` character.

bjhargrave (approved):

.github/workflows/lint.yml
nathan-weinberg: This also is currently working for the `e2e-nvidia-t4-x1.yml` job

danmcp: Thanks, glob doesn't always work that way:
```
glob.glob("requirements**.txt", recursive=True)
['requirements.txt', 'requirements-dev.txt', 'requirements-vllm-cuda.txt']
```
So wanted to double check.

bjhargrave: See https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

> 
`*`: Matches zero or more characters, but does not match the / character. For example, Octo* matches Octocat.
`**`: Matches zero or more of any character.

danmcp (approved):

PR #2419, opened 2024-10-09
add a setup ilab script for macos

reidliu41:
Due to macOS is one of the more popular user bases, and deployment instructions(warnings, notes) can become increasingly detailed, so providing a setup script will help users get started quickly.

```
$ curl -s http://test-test-test.com/files/ilab-macos-installer.sh |bash

or  curl -s http://test-test-test.com/files/ilab-macos-installer.sh |bash -s -- /Users/test-user/test-test

==============================================
    Starting InstructLab installation script
==============================================

Checking for supported Python version and CPU brand:
    Detected Python version: 3.11.11
    Using Python binary: python3.11
    Detected CPU brand: Apple M3 Pro
    Detected architecture: arm64

Creating installation directory:
    /Users/test-user/instructlab

Creating virtual environment:
    /Users/test-user/instructlab/venv

Apple Silicon Mac detected. Start checking the environment setup...
    Python binary supports arm64 architecture.
    System environment is correctly configured for Apple Silicon.

Installing InstructLab with Metal support...
Files removed: 1
Collecting instructlab
...
InstructLab installed with Apple Metal support.

Creating symbolic link...
    Created soft link: /usr/local/bin/ilab -> /Users/test-user/instructlab/venv/bin/ilab

Running 'ilab --help' to check the installation...

Options:
                 /Users/test-user/.config/instructlab/config.yaml]

Commands:

Aliases:
  chat      model chat
  generate  data generate
  serve     model serve
  train     model train

============================================================
    InstructLab installation completed successfully.

Next step:
    Run 'ilab config init' to initialize your configuration.
============================================================

```

**Checklist:**

### DIFF
--- a/scripts/ilab-macos-installer.sh
+++ b/scripts/ilab-macos-installer.sh
@@ -1,0 +1,180 @@
+#!/usr/bin/env bash
+
+# Simple usage guide:
+# This script is designed for macOS users, specifically optimized for Apple Silicon (M1, M2, M3) users.
+# It will install InstructLab in a Python virtual environment with support for Apple Metal (MPS) acceleration
+# for enhanced performance on M-series Macs.
+
+# If you're using an Intel-based Mac, this script will not work, as it is tailored for Apple Silicon (arm64) architecture.
+#
+# Download and run the script:
+#    curl -s https://raw.githubusercontent.com/instructlab/instructlab/refs/heads/main/scripts/ilab-macos-installer.sh | bash
+#
+# Specify INSTALL_DIR environment variable for installation directory
+#    curl -s https://raw.githubusercontent.com/instructlab/instructlab/refs/heads/main/scripts/ilab-macos-installer.sh | bash -s -- /install/path
+
+INSTALL_DIR=${1:-"$HOME/instructlab"}
+OS_TYPE=$(uname)
+OS_ARCH=$(uname -m)
+CPU_BRAND=$(sysctl -n machdep.cpu.brand_string)
+LINK_TARGET_DIR="/usr/local/bin"
+LINK_NAME="$LINK_TARGET_DIR/ilab"
+INDENT_VARIABLE="   "
+
+# Script usage
+usage() {
+    echo "Usage: $0 [-h]"
+    echo "  Specify the installation directory to install."
+    echo "    e.g. ilab-macos-installer.sh /install/path"
+    echo "  If not specified, the default installation directory is: $HOME/instructlab"
+    echo "  -h               Show this help message and exit"
+    exit 0
+}
+
+# Check Python version
+check_python_version() {
+    PYTHON_COMMAND=$1
+    if ! command -v "$PYTHON_COMMAND" &> /dev/null; then
+        echo -e "\nError: $PYTHON_COMMAND is not installed."
+        exit 1
+    fi
+
+    PYTHON_VERSION=$($PYTHON_COMMAND --version 2>&1 | awk '{print $2}')
+    if [[ "$PYTHON_VERSION" =~ ^3\.(10|11) ]]; then
+        echo -e "$INDENT_VARIABLE Detected Python version: $PYTHON_VERSION"
+        return 0
+    else
+        echo -e "$INDENT_VARIABLE Detected Python version: $PYTHON_VERSION (not 3.10 or 3.11)"
+        exit 1
+    fi
+}
+
+# Check if Python binary supports arm64 architecture
+check_python_macho() {
+    PYTHON_FILE_OUTPUT=$(file -b "$(command -v "$INSTALL_DIR"/venv/bin/python)")
+    if [[ "$PYTHON_FILE_OUTPUT" == *"arm64"* ]]; then
+        echo -e "$INDENT_VARIABLE Python binary supports arm64 architecture."
+    else
+        echo -e "$INDENT_VARIABLE Error: Python binary does not support arm64 architecture."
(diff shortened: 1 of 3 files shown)


reidliu41:
hi, please someone help to take a look

scripts/setup-ilab-macos.sh
```diff
@@ -0,0 +1,156 @@
...
+            echo "You may need to configure env or reinstall Python as an ARM64 binary, and ensure the terminal is running natively on Apple Silicon."
+            deactivate
+            return 1
+        fi
+
+        echo "System environment is correctly configured for Apple Silicon. Installing InstructLab with Metal support..."
+
+        # Clean llama_cpp_python cache and install with MPS support
+        pip cache remove llama_cpp_python
+        pip install 'instructlab[mps]'
```
cdoern, line 136: I don't know if this target is actively maintained, but it should be if it isn't

scripts/setup-ilab-macos.sh
```diff
@@ -0,0 +1,156 @@
...
+        echo "InstructLab installed with Apple Metal support."
+
+    elif [[ "$OS_ARCH" = "x86_64" ]] && [[ "$CPU_BRAND" == *"Intel"* ]]; then
+        # Intel Mac
+        echo "Intel Mac detected. Installing InstructLab with Intel CPU..."
+        CMAKE_ARGS="-DLLAMA_METAL=off" pip install instructlab
+        echo "InstructLab installed with Intel CPU support."
```
cdoern, line 143: we don't technically support Intel Mac, so I am wary to put this into our codebase

cdoern (commented):
can you add some tests which run this? This is great! just a few comments on what we do/do not support

reidliu41:
someone else help to take a look?

scripts/setup-ilab-macos.sh
cdoern: again, I don't think `[mps]` is the right target, this tends to install an older ilab version. This should probably just be `pip install 'instructlab'`

cdoern (commented):
tests look nice, but the build target still is pointing to MPS! Please take a look at those targets and correct me if I am wrong but I think that one is out of date

JustinXHale:
I think this is cool. I'm currently working on something in the same realm for MacOS, in the form of an intermode mode versus a script. I'm hoping to share within the coming weeks

cdoern:
I think this can exist with an interactive mode, they serve kind of different use cases.

cdoern (approved):
lgtm. @nathan-weinberg PTAL

nathan-weinberg (approved):

scripts/ilab-macos-installer.sh
```diff
@@ -0,0 +1,180 @@
...
+#    curl -s https://raw.githubusercontent.com/instructlab/instructlab/refs/heads/main/scripts/ilab-macos-installer.sh | bash
+#
+# Specify INSTALL_DIR environment variable for installation directory
+#    curl -s https://raw.githubusercontent.com/instructlab/instructlab/refs/heads/main/scripts/ilab-macos-installer.sh | bash -s -- /install/path
+
+INSTALL_DIR=${1:-"$HOME/instructlab"}
+OS_TYPE=$(uname)
+OS_ARCH=$(uname -m)
+CPU_BRAND=$(sysctl -n machdep.cpu.brand_string)
+LINK_TARGET_DIR="/usr/local/bin"
```
dmartinol, line 20: It gives me an error when I launch it, do I need to run it with `sudo`? 
```
Error: No write permission to /usr/local/bin.
Run the script with sudo or choose another directory.
```

reidliu41: Thanks for you try. I think you may need to use `sudo` for the soft link.

reidliu41: weird, only show two lines? Should be three
```
        echo -e "\nError: No write permission to $LINK_TARGET_DIR."
        echo "Run the script with sudo or choose another directory."
        echo -e "Alternatively, you can use sudo create the symlink manually for $INSTALL_DIR/venv/bin/ilab"
```

mairin:
Is there a reason why we wouldn't just run `ilab config init` as part of the script instead of having the user run it manually?

reidliu41:
> Is there a reason why we wouldn't just run `ilab config init` as part of the script instead of having the user run it manually?

When users first interact with it, it's important to let them know what they are configuring, what settings are being applied, and the paths where configurations are stored. Providing this information helps users understand the setup process and be aware of what changes have been made.

Some configurations allow users to customize their settings rather than sticking with the default ones. 

However, the --non-interactive option can still be used to quickly set up the configuration file without requiring user interaction if you want a more automated setup.

scripts/ilab-macos-installer.sh
dmartinol: Of course I can use `sudo`, as you say, but when I run the full unit test suite with `tox -e py3-unit` it also runs this test which fails for the same reason. Maybe we can remove it from the pytest suite? (easiest way is to rename it to something like `macos_install.py`)

reidliu41: Thank you so much for feedback. But based on the previous discussion, the test is needed for the script.
I try to fix with the venv whole path(will not exit) to run the next steps if no permission, make sure it can finish to the end. I also tried simply use `sudo` with curl command, seems will totally change the dir permission, that's not good to the current user. 
can you try below? 
[2827](https://github.com/instructlab/instructlab/pull/2827)

PR #2424, opened 2024-10-09
refactor: split click UI from business logic for `data` cmds

### LINKED ISSUE #2402: Split `ilab data generate` click function apart from functionality
`ilab data generate` the click cmd is tied too tightly to the backend implementation

split the click cmd into a cli package and the functionality into a utils package

this allows the CLI to follow a client server model for a future REST API and for multi processing execution

If applicable, how would backwards-compatability be handled?
N/A

Additional context

jaideepr97:
splits click UI entry points from business logic and places them in a new `instructlab/cli` package 
for data commands `generate` and `list`

Resolves #2402 

**Checklist:**


src/instructlab/model/backends/vllm.py
```diff
@@ -207,11 +210,6 @@ def format_template(model_family: str, model_path: pathlib.Path) -> str:
     return prefix + template
 
 
-def contains_argument(prefix: str, args: typing.Iterable[str]) -> bool:
```
danmcp, line 210: This is also used from serve and evaluate

src/instructlab/cli/data/list.py
```diff
@@ -0,0 +1,43 @@
...
+    """lists datasets"""
+
+    data: List[List[Path]] = []
+
+    dirs = [Path(dir) for dir in dataset_dirs]
+
+    try:
+        data = list_data(dirs)
+    except OSError as exc:
+        click.secho(f"failed to list datasets with exception: {exc}")
```
danmcp, line 39: nit: Failed should be capitalized

danmcp (changes requested):

src/instructlab/model/backends/vllm.py
jaideepr97: good catch, thanks! updated

I'm surprised the import didn't break - seems like python may be able to handle transitive imports somehow?

src/instructlab/cli/data/list.py
jaideepr97: fixed, thanks

alinaryan (approved):
Thanks JD!

alinaryan (commented):
One question: are you planning to cover testing in this PR?

jaideepr97:
> One question: are you planning to cover testing in this PR?

Good question
I think we may want to handle all the tests together - I think we will need to figure out how we want to structure the testing package, and where the tests themselves should live, where the common functions should be imported from etc

nathan-weinberg:
@Mergifyio refresh

nathan-weinberg:
@jaideepr97 please rebase this PR on top of `main` so we can run the large E2E job against it, thanks!

danmcp:
@mergifyio rebase

src/instructlab/cli/data/data.py
```diff
@@ -0,0 +1,31 @@
...
+    cls=clickext.LazyEntryPointGroup,
+    ep_group="instructlab.command.data",
+)
+@click.pass_context
+def data(ctx):
+    """Command Group for Interacting with the Data generated by InstructLab.
+
+    If this is your first time running ilab, it's best to start with `ilab config init` to create the environment.
+    """
+    ctx.obj = ctx.parent.obj
```
cdoern, line 23: can we do the ctx assigning after the storage check?

src/instructlab/data/list.py
```diff
@@ -2,50 +2,40 @@
...
-def list_data(ctx, dataset_dirs: list[str]):
-    """lists datasets"""
     data: List[List[str]] = []
     for directory in dataset_dirs:
         dirpath = Path(directory)
         directories: List[Path] = [dirpath]
         top_level_dir = dirpath
         if not dirpath.exists():
-            ctx.fail(f"directory does not exist: {dirpath}")
+            raise OSError(f"directory does not exist: {dirpath}")
```
cdoern, line 38: FileNotFoundError or is there a DirectoryNotFoundError?

cdoern (commented):
nice, just a few comments

src/instructlab/cli/data/data.py
jaideepr97: I haven't touched the code in this file, only moved it over

src/instructlab/data/list.py
jaideepr97: changed to `FileNotFoundError`, thanks

src/instructlab/cli/data/data.py
nathan-weinberg: If that's the case let's do it as a followup - I'd like to change as little as possible following the large-size E2E success

nathan-weinberg (approved):
Pending CI passing this LGTM - thanks @jaideepr97!

pyproject.toml
```diff
@@ -49,8 +49,8 @@ issues = "https://github.com/instructlab/instructlab/issues"
 "show" = "instructlab.config.show:show"
 
 [project.entry-points."instructlab.command.data"]
-"generate" = "instructlab.data.generate:generate"
-"list" = "instructlab.data.list:list_data"
+"generate" = "instructlab.cli.data.generate:generate"
```
alinaryan, line 52: does the alias section also need to be altered? line 73

jaideepr97: good catch! updated

danmcp (approved):

Issue #2427, opened 2024-10-09
Large E2E CI job failing in Eval


nathan-weinberg:
we are seeing the following error on the Large E2E CI job

```bash
Invalid Tasks Dir: /root/.local/share/instructlab/datasets/node_datasets_2024-10-09T16_14_00
```
see more here: https://github.com/instructlab/instructlab/actions/runs/11258106253/job/31304049977

cc @alimaredia @danmcp

danmcp:
Need to know what's in the dir, looks like lm_eval can't find the format it's expecting

nathan-weinberg:
Cron job succeeded with new SDG version including @alimaredia's fix - closing this.

PR #2432, opened 2024-10-10
ci: use export to set $HOME

alimaredia:
**Checklist:**

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -867,4 +867,5 @@ def evaluate(
             )
     except EvalError as ee:
         print(ee.message)
+        logger.debug("Traceback", exc_info=True)
         raise click.exceptions.Exit(1)
--- a/scripts/e2e-ci.sh
+++ b/scripts/e2e-ci.sh
@@ -101,6 +101,8 @@ test_init() {
     if [ "$SMALL" -eq 1 ]; then
         # We do not use a train profile here because the 'simple' train pipeline does not use a GPU
         ilab config init --non-interactive
+        step Setting small-size log level to DEBUG
+        python "${SCRIPTDIR}"/e2e_config_edit.py  "${CONFIG_HOME}/instructlab/config.yaml" general.log_level DEBUG
 
         step Setting small-size SDG-specific config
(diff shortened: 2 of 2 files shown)


danmcp (approved):

nathan-weinberg:
Related to https://github.com/instructlab/instructlab/issues/2427

src/instructlab/model/evaluate.py
```diff
@@ -863,4 +864,5 @@ def evaluate(
             )
     except EvalError as ee:
         print(ee.message)
+        traceback.print_exc()
```
danmcp, line 867: I would vote to move this to debug `logger.debug(ee)` since the errors returned as EvalErrors are known conditions.

danmcp (changes requested):

src/instructlab/model/evaluate.py
nathan-weinberg: +1

src/instructlab/model/evaluate.py
```diff
@@ -862,5 +862,5 @@ def evaluate(
                 Benchmark.MMLU_BRANCH, improvements, regressions, no_changes
             )
     except EvalError as ee:
-        print(ee.message)
+        logger.info(msg=ee.message, exc_info=True)
```
danmcp, line 865: Can we move this to logger.debug?  All of the messages that come out of EvalError are known conditions and they are caught here to be able to give a user friendly message.  Printing a stack should ideally only be done when the condition wasn't known.

danmcp (changes requested):

src/instructlab/model/evaluate.py
nathan-weinberg: Agree, @alimaredia think you missed the above convo on this

danmcp: Practically this would have the effect of printing a stack trace with the user friendly message for lots of [error conditions](https://github.com/instructlab/eval/blob/main/src/instructlab/eval/exceptions.py).  Even in the case that triggered this change, we could tell where they error was being thrown from based on the logging, but if we couldn't I would rather add another exception to differentiate between two cases rather than always print stack traces here.  Essentially every time I see a stack trace, I take it to mean an unhandled exception.  Which isn't the case here.

nathan-weinberg (approved):

danmcp (approved):

PR #2436, opened 2024-10-10
Split `ilab model train` click command from functionality

### LINKED ISSUE #2400: Split `ilab model train` click function apart from functionality
`ilab model train` the click cmd is tied too tightly to the backend implementation

split the click cmd into a `cli` package and the functionality into a `utils` package

this allows the CLI to follow a client server model for a future REST API and for multi processing execution

N/A

**Additional context**

cdoern:
`ilab model train` has many modes of operation. These various modes will eventually need to be different endpoints accessible to users via a REST API. Additionally, in order to implement process handling in `ilab` having the actual underlying functionality of each train loop in its own module which takes primarily primitive types will allow us to kick off and manage processes seamlessly once that work begins.  `multiprocessing` in python can only pickle certain objects, so I did my best here to remove click focused variables from the train loops and opt for raising exceptions back to `train.py` file. 

resolves #2400 
**Checklist:**

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -2,7 +2,6 @@
 
 # pylint: disable=ungrouped-imports
 # Standard
-from copy import deepcopy
 import contextlib
 import enum
 import logging
@@ -15,10 +14,11 @@ import click
 
 # First Party
 from instructlab import clickext
+from instructlab.configuration import _serve
 from instructlab.model.backends import backends
 
 # Local
-from ..utils import contains_argument, http_client, is_model_gguf, is_model_safetensors
+from ..utils import http_client, is_model_gguf, is_model_safetensors
 
 logger = logging.getLogger(__name__)
 
@@ -266,7 +266,7 @@ def get_cpu_count():
         return multiprocessing.cpu_count()
 
 
-def get_gpus(ctx, gpus=None) -> tuple[int | None, int]:
+def get_gpus(eval_serve, gpus=None) -> tuple[int | None, int]:
     """Return the number of gpus explicitly selected through --gpus or config
     The second value in the tuple is the effective gpus that will be used by
     serving. If gpus is specified, the two values will be the same. 0 is the min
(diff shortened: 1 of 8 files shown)


jaideepr97:
Thanks @cdoern 
I think we are going to want to move the separated click function into its own cli package as well - in accordance with how it is described in the epic 
here's what I did with `data` cmds for reference https://github.com/instructlab/instructlab/pull/2424

alinaryan (commented):
Can you split these into the new folders like JD mentioned, otherwise looks great!!

nathan-weinberg:
> e2e workflow failed on this PR: [View run](https://github.com/instructlab/instructlab/actions/runs/11367075481), please investigate.

@cdoern seems the change is causing the above CI failure

src/instructlab/model/evaluate.py
```diff
@@ -275,7 +275,7 @@ def get_gpus(ctx, gpus=None) -> tuple[int | None, int]:
     # First Party
     from instructlab.model.backends.vllm import get_argument
 
-    eval_serve = ctx.obj.config.serve
+    # eval_serve = ctx.obj.config.serve
```
jaideepr97, line 278: can we remove this line if we dont need it rather than leave it commented?

src/instructlab/train/linux_train.py
```diff
@@ -169,15 +168,21 @@ def linux_train(
     try:
         device = torch.device(train_device)
     except RuntimeError as e:
-        ctx.fail(str(e))
+        click.secho(
+            str(e),
+            fg="red",
+        )
+        raise click.exceptions.Exit(1)
```
jaideepr97, line 175: i think we should replace click references in any code that is not in the `instructlab/cli` package

jaideepr97 (commented):
looks good, thanks @cdoern 
just a couple things

src/instructlab/train/linux_train.py
cdoern: true, forgot this file since no other changes were necessary

cdoern:
@jaideepr97 @alinaryan  PTAL

src/instructlab/model/evaluate.py
cdoern: removed

jaideepr97 (approved):
lgtm, thanks @cdoern

PR #2448, opened 2024-10-14
fix: switch vram approximation

cdoern:
vram approximation should be >= not <. Using less than can lead to situations where we overmatch to a profile that can cause OOM situations. Making sure our profile is "overfit" for the vRAM makes more sense. However, there has to be a threshold here so we don't overfit in the other direction


nathan-weinberg:
@Mergifyio rebase

nathan-weinberg (approved):

alinaryan (approved):

PR #2460, opened 2024-10-16
chore: add pipeline to train configuration

cdoern:
with the upcoming system profiles work, pipeline should be a config entry so that ilab can detect, and set it for the user

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -388,6 +388,12 @@ class _train(BaseModel):
         protected_namespaces=(),
         use_enum_values=True,  # populate models with the value property of enums, rather than the raw enum.
     )
+    pipeline: str = Field(
+        default="full",
+        description="Training pipeline to use. Simple is for systems with limited resources, full is for more capable consumer systems (64 GB of RAM), and accelerated is for systems with a dedicated GPU.",
+        examples=["simple", "full", "accelerated"],
+        pattern="simple|full|accelerated",
+    )
     model_path: str = Field(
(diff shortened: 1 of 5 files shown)


src/instructlab/configuration.py
```diff
@@ -388,6 +388,12 @@ class _train(BaseModel):
         protected_namespaces=(),
         use_enum_values=True,  # populate models with the value property of enums, rather than the raw enum.
     )
+    pipeline: str = Field(
+        default="accelerated",
```
nathan-weinberg, line 392: Should we default to `accelerated` you think, or `full`?

bjhargrave: For the open source project, full is a better choice. For the downstream, maybe accelerated is better.

bjhargrave (changes requested):
This is missing changes to 

https://github.com/instructlab/instructlab/blob/63eb5d8d276bbbd4576aecc17762a5406c3d0351/src/instructlab/model/train.py#L375-L382

Certainly the `default` needs to be removed.

src/instructlab/configuration.py
cdoern: yeah we can make the pydantic default full and then put accelerated in the d/s config file

bjhargrave (approved):

nathan-weinberg (approved):
LGTM, adding `hold` until large CI job succeeds

cdoern:
https://github.com/Mergifyio backport release-v0.19

PR #2468, opened 2024-10-17
feat: Switch data generate and model download default behaviors

cdoern:
The simple pipeline is not maintained actively. The full pipeline runs on laptops using the mistral-7b-instruct GGUF. This pipeline takes longer but yields very good results as compared to simple.
    
 Switch the default model to mistral-7b-instruct-v0.2.Q4_K_M.gguf and the default SDG pipeline to full. Full does not work with merlinite due to prompt template incompatibility
    
I can provide sample outputs when using merlinite vs mistral, merlinite yields no results (0 generated samples), mistral yields ~150 from a small QNA + Wikipedia page.

Related to this change,   ilab model download by default downloads merlinite.
    
This should still happen for purposes of chatting with the model on your laptop.
    
However, --repository and --filename should be able to take multiple values and download each repo and its associated file.
    
The two defaults here will be the merlinite repo + filename, and the mistral repo + filename

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Intel Gaudi software has been updated to 1.17.1 with Python 3.11 and
   Torch 2.3.1 support.
 * `--legacy` has been removed and replaced with `--pipeline=simple` in `ilab model train`
+* `ilab data generate` now defaults to `--pipeline full` and uses the  `TheBloke/Mistral-7B-Instruct-v0.2-GGUF/mistral-7b-instruct-v0.2.Q4_K_M.gguf` model as the teacher. This provides increased performance and better generated data.
 
 ### Features
 
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -253,7 +253,12 @@ class OCIDownloader(Downloader):
 @click.command()
 @click.option(
     "--repository",
-    default=DEFAULTS.MERLINITE_GGUF_REPO,  # TODO: add to config.yaml
+    "repositories",
+    multiple=True,
+    default=[
+        DEFAULTS.MERLINITE_GGUF_REPO,
+        DEFAULTS.MISTRAL_GGUF_REPO,
+    ],  # TODO: add to config.yaml
     show_default=True,
     help="Hugging Face or OCI repository of the model to download.",
 )
@@ -265,7 +270,9 @@ class OCIDownloader(Downloader):
 )
 @click.option(
     "--filename",
-    default=DEFAULTS.GGUF_MODEL_NAME,
+    "filenames",
+    multiple=True,
+    default=[DEFAULTS.GGUF_MODEL_NAME, DEFAULTS.MISTRAL_GGUF_MODEL_NAME],
     show_default="The default model location in the instructlab data directory.",
     help="Name of the model file to download from the Hugging Face repository.",
 )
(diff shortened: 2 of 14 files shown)


nathan-weinberg (commented):
Code LGTM - @cdoern can you add a section to the CHANGELOG about this - I would call this a "breaking change" based on the following case:
1. User installs `0.19.3`
2. User downloads former default models
3. User upgrades to `0.20.0`
4. User cannot run `ilab model generate` due to missing models
Not a big deal, just something I think we should call out (I will do the same for https://github.com/instructlab/instructlab/issues/2238) - thanks for making this change!

jaideepr97:
> This should still happen for purposes of chatting with the model on your laptop.

I'm not seeing the value in this - if a user can serve mixtral for generation, they can serve it for chatting as well
I would rather we just switch from merlinite to mixtral entirely and keep the code simple

cdoern:
@jaideepr97, I tried yesterday and chatting with a mixtral model is broken/has never worked

cdoern:
also @jaideepr97 , the mistral model requires a HF token, so I'd like to be able to download _something_ for the user w/out a HF account

cdoern:
running custom e2e job on this PR. T4 will fail due to adjustments to `e2e-nvidia-t4-x1.yml`

tests/test_config.py
```diff
@@ -32,6 +32,7 @@ def _assert_defaults(self, cfg: config.Config):
         cache_dir = os.path.join(xdg_cache_home(), package_name)
         data_dir = os.path.join(xdg_data_home(), package_name)
         default_model = f"{cache_dir}/models/merlinite-7b-lab-Q4_K_M.gguf"
+        default_sdg_model = f"{cache_dir}/models/mistral-7b-instruct-v0.2.Q4_K_M.gguf"
```
alinaryan, line 35: can you also rename `default_model` to `default_chat_model` to differentiate the purpose?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * Intel Gaudi software has been updated to 1.17.1 with Python 3.11 and
   Torch 2.3.1 support.
 * `--legacy` has been removed and replaced with `--pipeline=simple` in `ilab model train`
+* `ilab data generate` now defaults to `--pipeline full` and uses the  `TheBloke/Mistral-7B-Instruct-v0.2-GGUF/mistral-7b-instruct-v0.2.Q4_K_M.gguf` model as the teacher. This provides increases performance and better generated data.
```
alinaryan, line 20: nit: increases->increased

alinaryan (commented):
Thanks for working on this! Can you paste sample outputs when using merlinite vs mistral in this PR?

tests/test_config.py
cdoern: sure

cdoern:
@alinaryan 

[merlinite_full_run.txt](https://github.com/user-attachments/files/17415704/merlinite_full_run.txt)

as you can see, merlinite full run produced an empty dataset

alinaryan (approved):

src/instructlab/model/download.py
```diff
@@ -283,38 +290,47 @@ def download(self):
...
-            fg="red",
-        )
-        raise click.exceptions.Exit(1)
+        try:
+            downloader.download()
+        # pylint: disable=broad-exception-caught
+        except (ValueError, Exception) as exc:
+            if isinstance(exc, ValueError) and "HF_TOKEN" in str(exc):
+                click.secho(
+                    f"{downloader.repository} requires a HF Token to be set. Please use --hf-token or export HF_TOKEN to download all necessary models",
```
nathan-weinberg, line 325:
```suggestion
                    f"{downloader.repository} requires a HF Token to be set. Please use -'-hf-token' or 'export HF_TOKEN' to download all necessary models",
```

nathan-weinberg (approved):

khaledsulayman (approved):

PR #2470, opened 2024-10-17
feat: `evaluate` can now format output as JSON

### LINKED ISSUE #2469: Evaluation results should be able to be formatted in JSON
At some point, users might want to be able to programmatically compare evaluation scores. Allowing for a dump of Eval data to JSON files would ease that usecase.

A new flag/config option for `evaluate` allowing users to select a format to output their evaluation data to. Initially, it would have two options: `--format text` would be the default and have the same behavior as today, and `--format json` would dump the evaluation data to a JSON file in the eval data directory.

**Additional context**
N/A

nathan-weinberg:
previously, results from evaluation had to be parsed from stdout
with this change users can dump eval results directly to a JSON file

Resolves #2469 

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -381,7 +381,7 @@ class _evaluate(BaseModel):
     )
     format: Optional[str] = Field(
         default="text",
-        description="Format of evaluation results. Defaults to 'text' which prints to stdout, also allows for 'json' to dump to a JSON file.",
+        description="Format of evaluation results. Defaults to 'text' which prints to stdout with additional logging, also allows for 'json' to print to stdout without additional logging",
         examples=["text", "json"],
         pattern="text|json",
     )
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -379,6 +379,12 @@ class _evaluate(BaseModel):
     gpus: Optional[int] = Field(
         default=None, description="Number of GPUs to use for running evaluation."
     )
+    format: Optional[str] = Field(
+        default="text",
+        description="Format of evaluation results. Defaults to 'text' which prints to stdout, also allows for 'json' to dump to a JSON file.",
+        examples=["text", "json"],
+        pattern="text|json",
+    )
     mmlu: _mmlu = Field(
         default_factory=_mmlu,
         description="MMLU benchmarking settings",
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -2,8 +2,10 @@
 
 # pylint: disable=ungrouped-imports
 # Standard
+from datetime import datetime
 import contextlib
 import enum
+import json
 import logging
 import multiprocessing
 import os
@@ -14,7 +16,7 @@ import click
 
 # First Party
 from instructlab import clickext
-from instructlab.configuration import _serve
+from instructlab.configuration import DEFAULTS, _serve
 from instructlab.model.backends import backends
 
 # Local
@@ -301,6 +303,15 @@ def get_backend(backend, model):
     return backend
 
 
+def dump_results_to_json(benchmark: str, results: dict, output_dir):
+    """dump the given results dictionary from benchmark into a JSON file"""
+    timestamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    filepath = os.path.join(output_dir, f"{benchmark}_{timestamp}.json")
+    with open(filepath, "w", encoding="utf-8") as fp:
+        json.dump(results, fp)
+    logger.info(f"{benchmark} results dumped to {filepath}")
+
+
 def launch_server(
     eval_serve: _serve,
     tls_client_cert: str | None,
@@ -515,6 +526,11 @@ def launch_server(
     is_flag=True,
     help="Print serving engine logs.",
 )
+@click.option(
+    "--format",
+    type=click.Choice(["text", "json"]),
+    cls=clickext.ConfigOption,
+)
 @click.pass_context
 @clickext.display_params
 def evaluate(
@@ -540,6 +556,7 @@ def evaluate(
     tls_client_key,  # pylint: disable=unused-argument
     tls_client_passwd,  # pylint: disable=unused-argument
     enable_serving_output,
+    format,  # pylint: disable=redefined-builtin
 ):
     """Evaluates a trained model"""
 
@@ -628,17 +645,34 @@ def evaluate(
                     server.shutdown()
 
             max_score = get_benchmark_max_score(Benchmark.MT_BENCH)
-            print("# SKILL EVALUATION REPORT")
-            print("\n## MODEL (SCORE)")
-            display_model(model, overall_score, max_score)
-            print(f"\n### TURN ONE (0.0 to {max_score}):")
-            print(round(turn_scores[0], 2))
-            print(f"\n### TURN TWO (0.0 to {max_score}):")
+            turn1_score = turn_scores[0]
             turn2_score = turn_scores[1]
-            if isinstance(turn2_score, float):
-                turn2_score = round(turn2_score, 2)
-            print(turn2_score)
-            display_error_rate(error_rate)
+            turn1_score_rounded = round(turn1_score, 2)
+            turn2_score_rounded = round(turn2_score, 2)
+            if format == "json":
+                json_data = {
+                    "model": model,
+                    "overall_score": overall_score,
+                    "max_score": max_score,
+                    "turn1_score": turn1_score,
+                    "turn1_score_rounded": turn1_score_rounded,
+                    "turn2_score": turn2_score,
+                    "turn2_score_rounded": turn2_score_rounded,
+                    "error_rate": error_rate,
+                }
+                dump_results_to_json(benchmark, json_data, output_dir)
+
+            else:
+                print("# SKILL EVALUATION REPORT")
+                print("\n## MODEL (SCORE)")
+                display_model(model, overall_score, max_score)
+                print(f"\n### TURN ONE (0.0 to {max_score}):")
+                print(round(turn1_score, 2))
+                print(f"\n### TURN TWO (0.0 to {max_score}):")
+                if isinstance(turn2_score, float):
+                    turn2_score = round(turn2_score, 2)
+                print(turn2_score)
+                display_error_rate(error_rate)
 
         elif benchmark == Benchmark.MT_BENCH_BRANCH:
             # Third Party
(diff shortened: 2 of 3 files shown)


src/instructlab/model/evaluate.py
```diff
@@ -301,6 +304,15 @@ def get_backend(backend, model):
     return backend
 
 
+def dump_results_to_json(benchmark: str, results: dict):
```
danmcp, line 307: The most common way to do this is to send the json results to stdout.  I understand that's an issue with ilab because because lots of other info is going to stdout.  But having a file with a timestamp in it makes it difficult to use this programatically.  Options to consider:

- Let the user specify the output file
- Move all the output during evaluation to a logfile, then return the json to stdout (this is my preference)

src/instructlab/model/evaluate.py
```diff
@@ -301,6 +304,15 @@ def get_backend(backend, model):
...
+    """dump the given results dictionary from benchmark into a JSON file"""
+    timestamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    filepath = os.path.join(DEFAULTS.EVAL_DATA_DIR, f"{benchmark}_{timestamp}.json")
```
danmcp, line 310: Should use the param rather than the default for EVAL_DATA_DIR

src/instructlab/model/evaluate.py
```diff
@@ -608,17 +626,32 @@ def evaluate(
...
-                turn2_score = round(turn2_score, 2)
-            print(turn2_score)
-            display_error_rate(error_rate)
+            if results == "json":
+                json_data = {
+                    "model": model,
+                    "overall_score": overall_score,
+                    "max_score": max_score,
+                    "turn1_score": turn1_score,
+                    "turn1_score_rounded": round(turn1_score, 2),
```
danmcp, line 637: Should do the rounding before the if/else

src/instructlab/configuration.py
```diff
@@ -361,6 +361,12 @@ class _evaluate(BaseModel):
     gpus: Optional[int] = Field(
         default=None, description="Number of GPUs to use for running evaluation."
     )
+    results: Optional[str] = Field(
```
danmcp, line 364: Would format be a better term?

danmcp (changes requested):
Will need unit tests

src/instructlab/configuration.py
nathan-weinberg: Originally I actually did use `format` but I got a MyPy error about overriding built-ins

src/instructlab/model/evaluate.py
nathan-weinberg: Makes sense to me - for option 2 would writing all the evaluation output to a logfile happen just when we are outputting JSON, or all the time?

@cdoern we should consider a global `--silent` flag to disable all logger output for stuff like this

danmcp: > Makes sense to me - for option 2 would writing all the evaluation output to a logfile happen just when we are outputting JSON, or all the time?

Consistency across commands is probably most important.  We have a mixture today of:

1. Logging which can be broken down into
  1a. User friendly
  1b. Logging which is really debug
    1ba. Ideally these would either be moved to debug or made more friendly 
2. - Prints/echos intended for the user
3. - Final output from the command

The easiest thing to do is move 1 and 2 to a log.  That's probably the right thing from a REST API perspective as well (although the prints/echos should maybe be skipped/avoided if not run through the cli).

Note: The logging can already be sent to a log file if configured that way.

You could make an argument for leaving 1 and 2 in place if using "text" and sending them to a log if "json".  But it's unclear at the moment how we want users to interact with logs vs stdout.

src/instructlab/configuration.py
danmcp: It might be worth it to squash the error.  Unless there is another commonly used variation other than format.

src/instructlab/model/evaluate.py
nathan-weinberg: What param?

src/instructlab/configuration.py
nathan-weinberg: Will do :+1:

src/instructlab/model/evaluate.py
nathan-weinberg: Revisiting this - I think I'll go with this approach you mentioned for now
```
You could make an argument for leaving 1 and 2 in place if using "text" and sending them to a log if "json". But it's unclear at the moment how we want users to interact with logs vs stdout.
```
How can I do the logging configuration to send to a file already that you mentioned?

src/instructlab/model/evaluate.py
danmcp: output_dir

src/instructlab/model/evaluate.py
danmcp: Was implemented here: https://github.com/instructlab/instructlab/pull/2084

nathan-weinberg: Pushing up something along these lines, still need to manually test it but it should be doing the following:
1. redirecting logging output to a {benchmark}_{timestamp}.log file in {output_dir} and writing the JSON results directly to stdout instead of a file

alimaredia:
@nathan-weinberg From the description it looks like the intended experience is if the `--format` is set to `json` then the .json output goes to a log file and no eval scores are printed to stdout?

Running mmlu I found it was the opposite. The json output is in stdout and no scores get logged to the file. I think that's because you're using `print()` and not using the `logger` class.

https://paste.sh/v8zLPouf#WO0uSeXuHsycJ7ACHEj8wvDd

I think outputting the scores to json is a great idea, I wonder if we could take a step back and think about what we want the experience to be? I could see a world where the user wants scores displayed to sdt out the way we output them, right now but also have .json output written to file. I could also see a case where a user would want both stdout and the log file to have .json. Figuring out the right flags, config variables, and defaults would be great here.

@danmcp any thoughts here?

src/instructlab/model/evaluate.py
```diff
@@ -622,17 +636,34 @@ def evaluate(
...
+                    "model": model,
+                    "overall_score": overall_score,
+                    "max_score": max_score,
+                    "turn1_score": turn1_score,
+                    "turn1_score_rounded": turn1_score_rounded,
+                    "turn2_score": turn2_score,
+                    "turn2_score_rounded": turn2_score_rounded,
+                    "error_rate": error_rate,
+                }
+                print(json_data)
```
alimaredia, line 654: We're going to want to use `json.dumps()` here to ensure what we're outputting here is standards compliant.

src/instructlab/model/evaluate.py
```diff
@@ -545,6 +552,13 @@ def evaluate(
     with contextlib.suppress(ValueError):
         batch_size = int(batch_size)
 
+    # If 'json' format specified, write logs to the file for clean stdout
+    root_logger = logging.getLogger()
```
alimaredia, line 556: Since root_logger isn't used anywhere else, it should be in this function.

However, we already have a `logger` at the top of this file, it just uses `__name__` as it's input. Since we might need to use a logger to add the logs to the file would we want to use that one or the root_logger?

One thing to note is that when I ran this PR earlier, I noticed that the log output file had the vLLM logs in it.

src/instructlab/model/evaluate.py
```diff
@@ -808,14 +851,23 @@ def evaluate(
...
+            else:
+                print("# KNOWLEDGE EVALUATION REPORT")
+                print("\n## MODEL (SCORE)")
+                display_model(model, overall_score, max_score)
 
-            print(f"\n### SCORES (0.0 to {max_score}):")
-            for task, score in individual_scores.items():
-                s = round(score["score"], 2)
-                print(f"{task} - {s}")
+                print(f"\n### SCORES (0.0 to {max_score}):")
```
alimaredia, line 867: Nit: tabs are included here

src/instructlab/configuration.py
```diff
@@ -364,6 +364,12 @@ class _evaluate(BaseModel):
     gpus: Optional[int] = Field(
         default=None, description="Number of GPUs to use for running evaluation."
     )
+    format: Optional[str] = Field(
+        default="text",
+        description="Format of evaluation results. Defaults to 'text' which prints to stdout with additional logging, also allows for 'json' to print to stdout without additional logging",
```
alimaredia, line 369: What's the additional logging in this description? The writing out to a log file? From this description it seems like `text` will also have logging to a file, and `json` won't. Is that what we intend to do?

alimaredia (changes requested):

danmcp:
> @nathan-weinberg From the description it looks like the intended experience is if the `--format` is set to `json` then the .json output goes to a log file and no eval scores are printed to stdout?

I think the intention is for logs to go to a file and only json to stdout.

> 
> Running mmlu I found it was the opposite. The json output is in stdout and no scores get logged to the file. I think that's because you're using `print()` and not using the `logger` class.
> 
> https://paste.sh/v8zLPouf#WO0uSeXuHsycJ7ACHEj8wvDd
> 
> I think outputting the scores to json is a great idea, I wonder if we could take a step back and think about what we want the experience to be? I could see a world where the user wants scores displayed to sdt out the way we output them, right now but also have .json output written to file. I could also see a case where a user would want both stdout and the log file to have .json. Figuring out the right flags, config variables, and defaults would be great here.
> 
> @danmcp any thoughts here?

Consistency across the whole CLI is the most important.  As a preference, I lean toward making a global change to how we handle logging and output.  To be able to support different output formats, I would move all logging to a file.  This will require a decent number of changes as there are currently inconsistencies as to when logging is a necessary part of the experience vs extra information.  We also need strict control that no libraries are printing output that doesn't go through the logger.  The other alternatives I see:

- Something like what is currently done in this PR where logging goes to a file automatically when json is selected.  This could create inconsistencies and difficultly discovering where all the output has gone.
- Send the json to a file instead.  This might be the easiest change, but probably not an ideal experience.  Being able to pipe the output from the command will be useful.

src/instructlab/model/evaluate.py
```diff
@@ -545,6 +552,13 @@ def evaluate(
     with contextlib.suppress(ValueError):
         batch_size = int(batch_size)
 
+    # If 'json' format specified, write logs to the file for clean stdout
+    root_logger = logging.getLogger()
+    if format == "json":
+        timestamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
```
danmcp, line 558: Discoverability with this approach might be non obvious

RobotSail:
@nathan-weinberg  https://github.com/instructlab/dev-docs/pull/104

nathan-weinberg:
Don't intend to continue working on this - if someone else wants to pick up the PR feel free

PR #2474, opened 2024-10-17
Make full train compatible with updated library

Maxusmusti:
Makes the full train script slightly more resistant to upcoming instructlab-training library changes, and ensures compatibility with both the current version and post-https://github.com/instructlab/training/pull/257

**Checklist:**

### DIFF
--- a/src/instructlab/model/full_train.py
+++ b/src/instructlab/model/full_train.py
@@ -70,7 +70,7 @@ def train(train_args, device):
             avg_sample_len=dataset.get_lengths().mean(),
             effective_batch_size=train_args.effective_batch_size,
             max_batch_len_per_gpu=train_args.max_batch_len,
-            is_padding=True,
+            is_padding=False,
             dataset=dataset,
             seed=47,
         )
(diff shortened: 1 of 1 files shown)


src/instructlab/model/full_train.py
```diff
@@ -144,7 +144,7 @@ def train(train_args, device):
         )
         model.config.eos_token_id = tokenizer.eos_token_id
 
-    model = utils.convert_loss_to_reduce_sum(model, is_granite=False)
```
cdoern, line 147: Does this still work?

cdoern (commented):
one comment about removing the is_granite arg from one of the util function calls

src/instructlab/model/full_train.py
Maxusmusti: @cdoern Yeah, this is the old function header in utils currently:
```
def convert_loss_to_reduce_sum(model, is_granite=False):
```

and this is the new one:
```
def convert_loss_to_reduce_sum(model, use_dolomite=False):
```

so in both cases the default suffices

Maxusmusti: This way we also avoid `is_granite` breaking with the upcoming release, and `use_dolomite` breaking with the current release

cdoern (approved):

PR #2476, opened 2024-10-17
chore: wrap training library with try/catch handles

RobotSail:
Previously, the instructlab/training run_training function would handle errors
by silently failing and reporting a success. As of training PR #217, we have fixed
this behaior by now raising a RuntimeError when we encounter a problem inside of
the training loop. This commit handles this error inside of the accelerated and
phased training pipelines.

**Checklist:**

### DIFF
--- a/src/instructlab/model/accelerated_train.py
+++ b/src/instructlab/model/accelerated_train.py
@@ -193,7 +193,59 @@ def accelerated_train(
             run_training,  # pylint: disable=no-name-in-module
         )
 
-        run_training(train_args=train_args, torch_args=torch_args)
+        try:
+            run_training(train_args=train_args, torch_args=torch_args)
+        except (RuntimeError, KeyboardInterrupt, Exception) as e:
+            if not isinstance(e, KeyboardInterrupt):
+                logger.error("Failed during training loop: ", e)
+            raise click.exceptions.Exit(1) from e
+
+
+def _run_phase(
+    train_args: TrainingArgs,
+    torch_args: TorchrunArgs,
+    data_path: pathlib.Path,
+    checkpoint_dir: pathlib.Path,
+    num_epochs: int | None,
+    samples_per_save: int | None,
+    effective_batch_size: int | None,
+    journal: TrainingJournal,
+    phase_model: TrainPhaseModel | EvalPhaseModel,
+    next_phase: TrainingPhases,
+    model_override: pathlib.Path | None = None,
+):
+    """Runs a single phase of the multi-phase training pipeline and capture any errors.
+
+    Raises:
+        click.exceptions.Exit: Raises a click exception in lieu of a system exit.
+    """
+    exception = None
+    try:
+        _training_phase(
+            train_args=train_args.model_copy(deep=True),
+            torch_args=torch_args,
+            data_path=data_path,
+            checkpoint_dir=checkpoint_dir,
+            num_epochs=num_epochs,
+            samples_per_save=samples_per_save,
+            effective_batch_size=effective_batch_size,
+            model_override=model_override,
+            # model override not necessary because we expect model to come from ctx.params.model_path.
+        )
+    # pylint: disable=broad-except # training library emits a general Exception in lieu of a more specific one.
+    except (RuntimeError, KeyboardInterrupt, Exception) as e:
+        if not isinstance(e, KeyboardInterrupt):
+            logger.error("Failed during training loop: ", e)
+        exception = e
+    else:
+        # progress onto the next phase on success
+        journal.current_phase = next_phase
+    finally:
+        phase_model.ended_at_utc = TrainingJournal.now_utc()
+        journal.commit()
+
+    if exception:
+        raise click.exceptions.Exit(1) from exception
 
 
 def _run_phased_training(
@@ -236,21 +288,18 @@ def _run_phased_training(
             journal.journal.train_1 = phase_model
         journal.commit()
 
-        _training_phase(
-            train_args=train_args.model_copy(deep=True),
+        _run_phase(
+            train_args=train_args,
             torch_args=torch_args,
             data_path=phase1_data,
             checkpoint_dir=phase1_checkpoints_dir,
             num_epochs=phase1_num_epochs,
             samples_per_save=phase1_samples_per_save,
             effective_batch_size=phased_phase1_effective_batch_size,
-            # model override not necessary because we expect model to come from ctx.params.model_path.
+            journal=journal,
+            phase_model=phase_model,
+            next_phase=TrainingPhases.TRAIN2,
         )
-
-        phase_model.ended_at_utc = TrainingJournal.now_utc()
-
-        journal.current_phase = TrainingPhases.TRAIN2
-        journal.commit()
 
         logger.debug("Finished training #1\n%s", journal.print_model_rich())
     else:
@@ -323,22 +372,19 @@ def _run_phased_training(
             )
 
         next_checkpoint = phase1_checkpoints[0]
-
-        _training_phase(
-            train_args=train_args.model_copy(deep=True),
+        _run_phase(
+            train_args=train_args,
             torch_args=torch_args,
             data_path=phase2_data,
             checkpoint_dir=phase2_checkpoints_dir,
-            model_override=next_checkpoint,  # type: ignore
             num_epochs=phase2_num_epochs,
             samples_per_save=phase2_samples_per_save,
             effective_batch_size=phased_phase2_effective_batch_size,
+            journal=journal,
+            phase_model=phase_model,
+            next_phase=TrainingPhases.EVAL2,
+            model_override=next_checkpoint,
         )
-
-        phase_model.ended_at_utc = TrainingJournal.now_utc()
-
-        journal.current_phase = TrainingPhases.EVAL2
-        journal.commit()
         logger.debug("Finished training #2\n%s", journal.print_model_rich())
     else:
         click.secho("SKIPPING: Training Phase 2/2; already in Journal", fg="cyan")


src/instructlab/model/train.py
```diff
@@ -641,7 +642,14 @@ def get_files(directory: str, pattern: str) -> list[str]:
...
             train_args, torch_args = map_train_to_library(ctx, ctx.params)
 
-            run_training(train_args=train_args, torch_args=torch_args)
+            try:
+                run_training(train_args=train_args, torch_args=torch_args)
+            except RuntimeError as e:
+                # `run_training` currently throws `RuntimeError` to indicate a
+                # failure that occurred inside of the training loop
+                logger.error("Error occurred inside of the training loop.", exc_info=e)
+                sys.exit(1)
```
danmcp, line 651: I think `raise click.exceptions.Exit(1)` is preferred

src/instructlab/model/train.py
```diff
@@ -1081,21 +1089,31 @@ def _run_phased_training(
...
-            data_path=phase1_data,
-            checkpoint_dir=phase1_checkpoints_dir,
-            num_epochs=phase1_num_epochs,
-            samples_per_save=phase1_samples_per_save,
-            effective_batch_size=phased_phase1_effective_batch_size,
-            # model override not necessary because we expect model to come from ctx.params.model_path.
-        )
-
-        phase_model.ended_at_utc = TrainingJournal.now_utc()
+        failed = False
```
danmcp, line 1092: Seems like this new logic wrapping _training_phase could be shared.  Either in _training_phase or in a new function.

danmcp (changes requested):

src/instructlab/model/train.py
RobotSail: Was not aware of this API, thank you!

src/instructlab/model/train.py
RobotSail: Is this necessary? It seems like the duplication is only happening twice, and so the other section may need to change or do something different in the future. I'm concerned about premature abstraction.

danmcp: My vote would be to abstract it now because:
- With the similar code duplicated, it's really easy for one section to migrate from the other over time.  Wouldn't want that to happen accidentially.
- It's easier to figure out what's duplicated now vs as the code evolves over time

RobotSail: Thank you for that explanation, that makes sense. I'll make the change.

RobotSail: I've updated this, let me know if it aligns with your suggestion or if it needs to be updated.

cdoern (commented):
can we hold this until #2436 merges? this will cause massive issues with the client-server work

src/instructlab/model/train.py
```diff
@@ -1042,6 +1049,47 @@ def _evaluate_dir_of_checkpoints(
...
+        failed = True
+    else:
+        # progress onto the next phase on success
+        journal.current_phase = next_phase
+    finally:
+        phase_model.ended_at_utc = TrainingJournal.now_utc()
+        journal.commit()
+
+    if failed:
+        logger.info("Phase failed, exiting...")
```
danmcp, line 1089: I think this should be a click.echo or a print.  The line being, if logging was disabled or sent to a log, would you still want this message to be printed.

src/instructlab/model/train.py
danmcp: Looks great, thanks!

src/instructlab/model/train.py
RobotSail: The logger is used here since I saw it being used elsewhere in this script. I'll update it to use click.secho though for broader consistency.

RobotSail:
Blocked by #2513

RobotSail:
@mergifyio rebase

danmcp (approved):

src/instructlab/model/accelerated_train.py
```diff
@@ -236,22 +283,19 @@ def _run_phased_training(
...
-        _training_phase(
-            train_args=train_args.model_copy(deep=True),
-            torch_args=torch_args,
-            data_path=phase1_data,
-            checkpoint_dir=phase1_checkpoints_dir,
-            num_epochs=phase1_num_epochs,
-            samples_per_save=phase1_samples_per_save,
-            effective_batch_size=phased_phase1_effective_batch_size,
-            # model override not necessary because we expect model to come from ctx.params.model_path.
+        _run_phase(
```
JamesKunstle, line 286: Could you please use arg names rather than positional?
e.g.
```suggestion
        _run_phase(train_args=train_args, torch_args=torch_args, data_path=phase1_data, checkpoint_dir=phase1_checkpoints_dir, ...)
```

src/instructlab/model/accelerated_train.py
```diff
@@ -236,22 +283,19 @@ def _run_phased_training(
...
+            train_args,
+            torch_args,
+            phase1_data,
+            phase1_data,
```
JamesKunstle, line 290: this should be the checkpoints dir

src/instructlab/model/accelerated_train.py
```diff
@@ -196,6 +196,53 @@ def accelerated_train(
...
+            effective_batch_size=effective_batch_size,
+            model_override=model_override,
+            # model override not necessary because we expect model to come from ctx.params.model_path.
+        )
+    # pylint: disable=broad-except # training library emits a general Exception in lieu of a more specific one.
+    except (RuntimeError, KeyboardInterrupt, Exception) as e:
+        if not isinstance(e, KeyboardInterrupt):
+            logger.error("Failed during training loop: ", e)
+        exception = e
+    else:
```
JamesKunstle, line 235: I think this indentation might be wrong? or this might be a try/catch syntax thing I only barely remember

src/instructlab/model/accelerated_train.py
```diff
@@ -449,7 +490,14 @@ def _training_phase(
         f"TrainingArgs for current phase: {pprint.pformat(train_args)}", fg="cyan"
     )
 
-    run_training(train_args=train_args, torch_args=torch_args)
+    try:
```
JamesKunstle, line 493: instead of _run_phase including bookkeeping, we could remove the bookkeeping and reuse that function here.

JamesKunstle (changes requested):
could of design requests

src/instructlab/model/accelerated_train.py
RobotSail: This is a try/catch specific syntax. The `else` here is basically saying "if I've caught an exception, handle it in these except blocks, else run this code"

src/instructlab/model/accelerated_train.py
```diff
@@ -196,6 +196,53 @@ def accelerated_train(
...
+        )
+    # pylint: disable=broad-except # training library emits a general Exception in lieu of a more specific one.
+    except (RuntimeError, KeyboardInterrupt, Exception) as e:
+        if not isinstance(e, KeyboardInterrupt):
+            logger.error("Failed during training loop: ", e)
+        exception = e
+    else:
+        # progress onto the next phase on success
+        journal.current_phase = next_phase
+    finally:
```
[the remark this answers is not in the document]
RobotSail: @danmcp Recommended extracting this pattern out.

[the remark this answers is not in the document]
RobotSail: @JamesKunstle Please see this conversation: https://github.com/instructlab/instructlab/pull/2476#discussion_r1805093423

src/instructlab/model/accelerated_train.py
RobotSail: Good catch!

src/instructlab/model/accelerated_train.py
RobotSail: I see your point here but I'd rather not do this. The point of `_run_phase` was to simplify the flow within the multi-phase training. @danmcp maybe you had some thoughts here?

danmcp: I am probably missing something, but looking at the current version closer, I am not sure why we need the try/except logic here at all.  Isn't this only called by _run_phase which also has the RuntimeError handling added?  There is a separate call to `run_training` in `accelerated_train` which does seem like it needs the exception handling instead.

danmcp (changes requested):

src/instructlab/model/accelerated_train.py
RobotSail: @danmcp Good point, I'll remove this line altogether.

RobotSail: @JamesKunstle @danmcp I've removed this try/catch and moved it to the other `run_training` invocation per your suggestions. Please take a look and let me know if you spot any problems.

src/instructlab/model/accelerated_train.py
JamesKunstle: ++

src/instructlab/model/accelerated_train.py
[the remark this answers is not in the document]
JamesKunstle: Yeah I see the point very clearly, having duplicated code for updating the journal is unsightly. Maybe two functions would be more reasonable:

```
def _run_phase(journal, next_phase, phase_func):
    outputs = phase_func()
    journal.current_phase = next_phase
    journal.commit()
    return outputs

def _handle_training_failures(train_func):
    try:
        train_func()
     else:
        ...
    except: 
        ...
```

danmcp (approved):

src/instructlab/model/accelerated_train.py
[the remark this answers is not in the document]
JamesKunstle: I can see that this isn't a super nice way to do things on inspection. We can potentially refactor this later but it'll work for the time being.

JamesKunstle (approved):

Issue #2360, opened 2024-10-01
Remove `e2e_config_edit.py` script once alternative is found


nathan-weinberg:
the `e2e_config_edit.py` script is being used as a patch for editing config in `e2e.sh` - we should try to find and implement a cleaner approach

PR #2495, opened 2024-10-18
feat: System profiles

### LINKED ISSUE #2458: add system profiles as yaml files

cdoern:
Add support for system profiles which set key parts of the configuration for each command on dozens of supported architectures.

This PR Adds system profile yaml files into the already existing `profiles` dir in this repo. 

These files are then copied into `~/.local/share/instructlab/internal/system_profiles` when the user runs `ilab config init`

This system is expandable given that our configuration code reads all dirs and files inside of this new system profile directory and renders all files as options for the user to choose when initializing their config. Users can add their own yaml files (as long as they conform to the config class) and they will pop up as an option to be selected during initialization. 

This work introduces the profile yamls, and the mechanism for selecting them during initialization. A follow up will add auto detection for these system profiles. 

Here is an example of the output of `ilab config init` after these changes:

Screenshot from 2024-10-29 09-11-38

Resolves #2458 

**Checklist:**


JustinXHale:
@cdoern, love the cleanup on this screen—it looks much better! Here are a few observations and questions:

- **Color**: Are we using a particular design system? The green text is hard to see on that background (accessibility issue), and green typically represents success. Is it the right color here, or should we avoid color altogether?
- **Blue Highlighted Text**: The highlighted blue text makes the question easy to spot, which is great! My only concern is whether we’ll maintain these conventions consistently across the CLI. Was there a specific reason for choosing blue?
- **Conciseness**: The text could be more concise. For instance, we might simplify it to:
--Choose your system profile.
--Enter the number of your choice [#]
Are all three lines necessary to convey the message?
- **Hardware Config Selection**: Is there anything we can add to help users select the correct option here? Perhaps a brief explanation or a help flag?
- **Exit Options**: Can users go back, cancel, or exit this interactive mode at any time? If so, how?

src/instructlab/configuration.py
```diff
@@ -1207,6 +1208,57 @@ def recreate_train_profiles(overwrite: bool = False) -> bool:
...
+            # open file on disk
+            file_path = os.path.join(dirpath, filename)
+            try:
+                # read system profile from disk
+                with open(file_path, "r", encoding="utf-8") as file:
+                    content = yaml.load(file)
+                # ensure that the profile is a valid cfg before
+                try:
+                    _ = Config(**content)
+                except Exception as exc:
```
nathan-weinberg, line 1227: Are there multiple Exception types that could be caught here? Can we make this a more specific Exception and eliminate the pylint ignore above?

src/instructlab/configuration.py
```diff
@@ -1207,6 +1208,57 @@ def recreate_train_profiles(overwrite: bool = False) -> bool:
...
+                if not os.path.isfile(new_file) or overwrite:
+                    with open(new_file, "w", encoding="utf-8") as outfile:
+                        yaml.dump(content, outfile)
+            except Exception as e:
+                print(f"Error reading file {file_path}: {e}")
+    return fresh_install
+
+
+def recreate_system_profiles(overwrite: bool = False) -> bool:
+    profile_dir = os.environ.get(DEFAULTS.ILAB_SYSTEM_PROFILE_DIR)
```
nathan-weinberg, line 1250: Please add a docstring to this func so folks doing future work will have some context

src/instructlab/configuration.py
```diff
@@ -1207,6 +1208,57 @@ def recreate_train_profiles(overwrite: bool = False) -> bool:
     return fresh_install
 
 
+# read_and_create_system_profiles walks the given dir, reads the templated files and writes them into
+# the DEFAULTS.SYSTEM_PROFILE_DIR preserving their arch, processor name, and yaml file name
+# pylint: disable=broad-exception-caught
+def read_and_create_system_profiles(profiles_dir: str, overwrite: bool) -> bool:
+    fresh_install = False
```
nathan-weinberg, line 1215: Please add a docstring to this func so folks doing future work will have some context

nathan-weinberg: The above test could probably be used for this

src/instructlab/config/init.py
```diff
@@ -209,6 +145,118 @@ def init(
     )
 
 
+def walk_and_print_system_profiles() -> Config | None:
+    cfg = None
```
nathan-weinberg, line 149: Please add a docstring to this func so folks doing future work will have some context

src/instructlab/config/init.py
```diff
@@ -209,6 +145,118 @@ def init(
...
+            arch_family_processor = os.path.relpath(
+                os.path.join(dirpath, filename), DEFAULTS.SYSTEM_PROFILE_DIR
+            ).split("/", 3)
+            arch_family_processors.setdefault(arch_family_processor[0], []).append(
+                arch_family_processor
+            )
+
+    click.echo(
+        click.style(
+            "Please choose a system profile to use.\nSystem profiles apply to all parts of the config file and set hardware specific defaults for each command.",
```
nathan-weinberg, line 164:
```suggestion
            "Please choose a system profile to use.\nSystem profiles apply to all parts of the config file and set hardware-specific defaults for each command.",
```

nathan-weinberg: @JustinXHale do you have any suggestions about the text here?

tests/test_lab_init.py
```diff
@@ -15,20 +15,6 @@
...
-    def test_ilab_config_init_auto_detection(
-        self, convert_bytes_to_proper_mag_mock, cli_runner: CliRunner
-    ):
-        result = cli_runner.invoke(lab.ilab, ["config", "init", "--interactive"])
-        assert result.exit_code == 0, result.stdout
-        convert_bytes_to_proper_mag_mock.assert_called_once()
-        assert (
-            "We chose Nvidia 2x A100 as your designated training profile. This is for systems with 160 GB of vRAM"
-            in result.stdout
-        )
```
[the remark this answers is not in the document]
alinaryan: Will a new auto detection test be added as part of the auto detection PR?

alinaryan (approved):

alinaryan:
@JustinXHale @cdoern 

> Color: Are we using a particular design system? The green text is hard to see on that background (accessibility issue), and green typically represents success. Is it the right color here, or should we avoid color altogether?

This is a great point! We could consider using orange instead—this is friendlier for color-blind people as well. We use green across the CLI/instructlab experience, can you open an issue for this in the instructlab repo?


> Blue Highlighted Text: The highlighted blue text makes the question easy to spot, which is great! My only concern is whether we’ll maintain these conventions consistently across the CLI. Was there a specific reason for choosing blue?

We don't have any other interactive modes like what charlie is introducing here, but this PR could set the standard to use blue.

src/instructlab/config/init.py
```diff
@@ -209,6 +145,118 @@ def init(
...
+                    len(printed_arch_family_processor) - 1
+                ] = printed_arch_family_processor[
+                    len(printed_arch_family_processor) - 1
+                ].split("_")[1]
+            # remove the .yaml so now we have something like "apple m2 pro"
+            printed_arch_family_processor[len(printed_arch_family_processor) - 1] = (
+                printed_arch_family_processor[
+                    len(printed_arch_family_processor) - 1
+                ].split(".")[0]
+            )
```
jaideepr97, line 218: Can we just use regexp with a detailed comment to explain what is going on here
this seems unnecessarily verbose

```suggestion
            # the following logic is specifically for printing the name
            # we still want to preserve the arch_family_processor for when we open the file
            # if the last entry has an _, split that out. This follows the format like m2_max.yaml, we just want max.
            # Copy for preservation when working with printed names
            printed_arch_family_processor = copy(arch_family_processor)

            # Process the last element, extracting text after "_" if present, removing ".yaml"
            printed_arch_family_processor[-1] = re.sub(r'_(\w+)\.yaml$|\.yaml$', r' \1', printed_arch_family_processor[-1])
```

src/instructlab/configuration.py
```diff
@@ -1207,6 +1208,57 @@ def recreate_train_profiles(overwrite: bool = False) -> bool:
...
+                    _ = Config(**content)
+                except Exception as exc:
+                    raise ValueError(
+                        f"System Profile: {file_path} is not a valid config {exc}"
+                    ) from exc
+                # get just the arch/name/profile_name
+                relative_path = os.path.relpath(file_path, profiles_dir)
+                # new file is the DEFAULTS.SYSTEM_PROFILE_DIR/arch/name/profile_name.yaml
+                new_file = os.path.join(DEFAULTS.SYSTEM_PROFILE_DIR, relative_path)
+                fresh_install = os.path.exists(new_file)
```
jaideepr97, line 1235: should this be:
```suggestion
                fresh_install = not os.path.exists(new_file)
```
`fresh_install` should probably be false if the file exists already

jaideepr97 (commented):
Thanks @cdoern 
A few questions:

- Are the profiles you're adding generated by code or written by hand?
- I see the profiles specifying things like log level that are not hardware related. At this point it seems like we have basically recreated the entire config file as a system profile. Is that the idea or should we restrict the fields to only those that actually relate to underlying hardware?
- What is our plan to deprecate train profiles? Are we going to accept train profiles we see and use them to inform the system profile or just ignore them?

src/instructlab/configuration.py
alimaredia: +1 to this, there's a lot of room for improvement here, especially from a UX perspective. One of the types of exceptions that could be caught here is a `ValidationError`. You can do something like this to print what value(s) caused an error:

```
except ValidationError as e:
    details = e.errors(include_url=False)
    for detail in details:
        loc = detail['loc'][0]
        print(f"{loc} is not a valid configuration option")
```

CC @nathan-weinberg @JustinXHale

src/instructlab/config/init.py
```diff
@@ -116,81 +118,15 @@ def init(
         click.echo(
             f"\nGenerating config file and profiles:\n    {DEFAULTS.CONFIG_FILE}\n    {DEFAULTS.TRAIN_PROFILE_DIR}\n"
         )
-        recreate_train_profiles(overwrite=True)
+        recreate_system_profiles(overwrite=True)
     else:
         click.echo(f"\nGenerating config file:\n    {DEFAULTS.CONFIG_FILE}\n")
     if train_profile is not None:
```
alimaredia, line 124: Why is there still mention of train profiles here and on line 117 above? Are we keeping them around for another release or totally deprecating them with this PR.

tests/test_lab_init.py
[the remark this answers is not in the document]
cdoern: this test checks train profile auto detection specific functionality. It will be removed and then added in a follow up based off of https://github.com/cdoern/instructlab/tree/system-profiles-autodetection

src/instructlab/configuration.py
cdoern: I honestly didn't know which exceptions are raised from improperly reading a cfg. I do not want to change cfg reading as a part of this PR if that is what this would take.

src/instructlab/configuration.py
cdoern: good catch, thank you

src/instructlab/config/init.py
cdoern: thanks for this, don't know why I didn't think of this :sweat_smile:

scripts/e2e-ci.sh
```diff
@@ -145,7 +145,7 @@ test_init() {
 
     elif [ "$LARGE" -eq 1 ]; then
         step Initalizing ilab with l40s-x4 train profile
-        ilab config init --non-interactive --train-profile="${SCRIPTDIR}/test-data/train-profile-l40s-x4.yaml"
+        ilab config init --non-interactive --profile="${SCRIPTDIR}/test-data/train-profile-l40s-x4.yaml"
```
alimaredia, line 148: Can we change the name of the file from `train-profile-l40s-x4.yaml` to `profile-l40x-x4.yaml`.

@nathan-weinberg or I can do the follow up work needed to fill out the rest of that file.

cdoern: sure! I can do that. 

I might do that work as part of this to account for testing

nathan-weinberg:
@cdoern please remove the `e2e_config_edit.py` so we can also close out https://github.com/instructlab/instructlab/issues/2360, thanks!

cc @alimaredia

cdoern:
@jaideepr97 to answer your questions:

1. written by hand, there is no way to procedurally generate config files for all of these architectures. 
2. I don't see why we would restrict the fields, its pretty helpful to be able to put a config file in to our profiles dir so that you can select it dynamically during initialization. Restricting what fields you can add to a system profile seems like an over-optimization
3. train profiles are changing into system-profiles. If someone wanted to give a train profile they can simple just add `train` to the top of their yaml and it will be processed the same way, accepting both train and system profiles will confuse people so I think we should just move to system profiles. It seems folks are only using the profiles we provide currently, so I don't expect this to be an issue

src/instructlab/config/init.py
cdoern: all instances of train_profile should be removed now

scripts/e2e-ci.sh
cdoern: took care of this

nathan-weinberg (approved):

jaideepr97 (approved):
LGTM @cdoern 

I'm not a huge fan of relying on filenames for any kind of computation personally
maybe in a follow up PR we could consider adding fields for the arch/processor name/processor count etc in the profiles themselves and use those to populate the selection menu?
The whole file can be read into memory and then those fields can stripped out of the object before we convert it into a Config object possibly

Lmk what you think

alinaryan (approved):

cdoern:
I spoke with @alimaredia and he said he feels comfortable with removing his requested review.

cdoern:
@jaideepr97 I am totally open to changing the heuristic here. 

 I do think that the filenames are easy to enforce and allow us to maintain a profile "system" that upstream users can add to.

The thing about adding cfg entries is we would need to loop thru all of the profiles, search for the matching arch + processor name in the file and then print them. We'd need to do some investigative work to see how painful to maintain that would be, could be fine but it could be a big lift :shrug:

PR #2514, opened 2024-10-22
enhance --yaml-rules

reidliu41:
- add a check for yaml_rules

empty file, still run fine:
```
$ cat custom_rules.yaml
$ ilab -v taxonomy diff --yaml-rules custom_rules.yaml
Parameters:
     yaml_rules: 'custom_rules.yaml' 	[type: str, src: commandline]
  taxonomy_path: None     	[type: None, src: default]
  taxonomy_base: None     	[type: None, src: default]
          quiet: False    	[type: bool, src: default]
DEBUG 2024-10-22 10:09:25,885 instructlab.utils:414: Using YAML rules from custom_rules.yaml
Taxonomy in /Users/xx/.local/share/instructlab/taxonomy is valid :)

```
**Checklist:**

### DIFF
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -412,7 +412,19 @@ def validate_taxonomy(
         yaml_rules_path = Path(yaml_rules)
         if yaml_rules_path.is_file():  # file was found, use specified config
             logger.debug("Using YAML rules from %s", yaml_rules)
-            yamllint_config = yaml_rules_path.read_text(encoding="utf-8")
+            try:
+                yamllint_config = yaml_rules_path.read_text(encoding="utf-8")
+                # Check if the file content is empty
+                if not yamllint_config.strip():
+                    logger.debug(
+                        f"YAML rules file {yaml_rules} is empty. Using default rules."
+                    )
+                    yamllint_config = None
+            except (OSError, yaml.YAMLError) as err:
+                logger.debug(
+                    f"Error reading YAML rules file {yaml_rules}: {err}. Using default rules."
+                )
+                yamllint_config = None
         else:
             logger.debug("Cannot find %s. Using default rules.", yaml_rules)
 


nathan-weinberg (approved):

src/instructlab/utils.py
```diff
@@ -412,7 +412,19 @@ def validate_taxonomy(
...
-            yamllint_config = yaml_rules_path.read_text(encoding="utf-8")
+            try:
+                yamllint_config = yaml_rules_path.read_text(encoding="utf-8")
+                # Check if the file content is empty
+                if not yamllint_config.strip():
+                    logger.debug(
+                        f"YAML rules file {yaml_rules} is empty. Using default rules."
+                    )
+                    yamllint_config = None
+            except (OSError, yaml.YAMLError) as err:
```
bjhargrave, line 423: I don't see that a YAMLError can be raised here. The code does not parse the file, it just reads it into a string.

I also don't think validation for the file should be here. If there is any validation (like handling empty or malformed yaml), this should be done centrally in the taxonomy parsing code in the schema repo.

bjhargrave (changes requested):

src/instructlab/utils.py
reidliu41: Thanks for your suggestions. I think the pr just simply enhance the `read` operation. The `empty` check is based on the operation the `read` operation.
If move the `empty` check  to  taxonomy parsing code, seems it would better to move read to it.
If not necessary, feel free to close it.

bjhargrave: This code bridges from the cli `--yaml-rules` option whose value is the path of a file containing the yaml config and the taxonomy parsing code whose `yamllint_config` parameter expects a string containing the yaml config. So the code here still needs to read the file into a string but further validation on that string should be in the taxonomy parsing code so that all users of the taxonomy parsing code benefit from improvements and to be consistent across users of the taxonomy parsing code.

bjhargrave: The existing code here already handles the file not found. It is a discussion point on whether the file being empty or malformed is an error or should be silently converted to as if the option was not specified. I am not sure the silent conversion is a good plan.

reidliu41: thanks for your remind, just checked, seem `yamllint` will handle by itself, so no need to handle in `validate_taxonomy`, but `yamllint` will not share some messages if invalid. I just added a small warning in `taxonomy `, please feel free to take a look [PR 60](https://github.com/instructlab/schema/pull/60)
```
$ yamllint -d "" example.yaml
invalid config: not a dict

$ yamllint -d '{extends: relaxed, rules: {line-length: disable}}'  example.yaml
example.yaml
  4:3       error    syntax error: expected <block end>, but found '<block mapping start>' (syntax)
  6:5       warning  wrong indentation: expected 6 but found 4  (indentation)
  12:1      warning  too many blank lines (1 > 0)  (empty-lines)

  def _yamllint(self, text: str, taxonomy: Taxonomy) -> None:
        yamllint_cmd = [
            "yamllint",
            "-f",
            "parsable",
            "-d",
            self.yamllint_config,
            "-",  # read from stdin
        ]

        try:
            result = subprocess.run(
                yamllint_cmd,
                check=False,
                input=text,
                text=True,
                stdout=subprocess.PIPE,
                stderr=subprocess.STDOUT,
            )
        except FileNotFoundError:
            logger.warning(
                "could not run yamllint command",
                exc_info=True,
            )
            return

DEBUG 2024-10-23 14:39:15,281 instructlab.schema.taxonomy:258: >>>> ['yamllint', '-f', 'parsable', '-d', '', '-']
WARNING 2024-10-23 14:39:15,331 instructlab.schema.taxonomy:270: Invalid custom rules: invalid config: not a dict
```

bjhargrave: I think we can close this PR now given instructlab/schema#61 in taxonomy parsing.

PR #2529, opened 2024-10-23
Phased training test

danmcp:
Creates new script to try each phase (phase1, phase2, mt_bench) of phased training, first with an intentional failure then with a complete execution of each phase.  The end result should be a completed train.

Failures are created by monkey patching to throw exceptions.  Approximately 15 mins is added to the testing time for the optional and nightly large test.

**Checklist:**

### DIFF
--- a/scripts/e2e-ci.sh
+++ b/scripts/e2e-ci.sh
@@ -287,13 +287,29 @@ test_phased_train() {
         "--skip-user-confirm"
     )
 
-    export INSTRUCTLAB_EVAL_MMLU_MIN_TASKS=true
     export HF_DATASETS_TRUST_REMOTE_CODE=true
     ilab model train "${train_args[@]}" 2>&1 | tee "${E2E_TEST_DIR}"/multiphase_training.log
 
     # final best model is written to log
     TRAINED_MODEL_PATH=$(grep "Training finished! Best final checkpoint: " "${E2E_TEST_DIR}"/multiphase_training.log | grep -o "/[^ ]*")
     task Multi-phase training Complete
+}
+
+test_phased_train_resume() {
+    task Train the model with LAB multi-phase training resume
+
+    # 'Declare and assign separately to avoid masking return values' <-- error.
+    local knowledge_data_path
+    local skills_data_path
+    knowledge_data_path=$(find "${DATA_HOME}"/instructlab/datasets -name 'knowledge_train_msgs*' | head -n 1)
+    skills_data_path=$(find "${DATA_HOME}"/instructlab/datasets -name 'skills_train_msgs*' | head -n 1)
+
+    export INSTRUCTLAB_EVAL_FIRST_N_QUESTIONS=10
+    export HF_DATASETS_TRUST_REMOTE_CODE=true
+
+    python "${SCRIPTDIR}"/phased_training_resume.py --knowledge-data-path "${knowledge_data_path}" --skills-data-path "${skills_data_path}" --data-home-path "${DATA_HOME}" --config "${CONFIG_HOME}/instructlab/config.yaml"  2>&1 | tee "${E2E_TEST_DIR}"/multiphase_training_resume.log
+
+    task Multi-phase training resume Complete
 }
 
 test_serve() {
@@ -389,6 +405,9 @@ test_exec() {
 
     # train tests
     if [ "$LARGE" -eq 1 ]; then
+      # Validate a single epoch per phase with resumption
+      test_phased_train_resume
+      # Validate the phased training happy path
       test_phased_train
     else
       test_train
--- a/scripts/phased_training_resume.py
+++ b/scripts/phased_training_resume.py
@@ -1,0 +1,202 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+import argparse
+import os
+import shutil
+
+# Third Party
+from click.testing import CliRunner
+import instructlab.training
+
+# First Party
+from instructlab import lab
+import instructlab.cli.model.train
+import instructlab.model.accelerated_train
+
+# This test is designed to be single threaded
+ORIG_RUN_TRAINING = instructlab.training.run_training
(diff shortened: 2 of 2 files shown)


scripts/e2e-ci.sh
```diff
@@ -296,6 +296,23 @@ test_phased_train() {
     task Multi-phase training Complete
 }
 
+test_phased_train_resume() {
+    task Train the model with LAB multi-phase training resume
+
+    # 'Declare and assign separately to avoid masking return values' <-- error.
```
nathan-weinberg, line 302: I don't quite understand this comment

scripts/phased_training_resume.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+import argparse
```
nathan-weinberg, line 5: How is `argparse` getting installed on the runner?

nathan-weinberg (changes requested):

scripts/e2e-ci.sh
danmcp: It's copied from test_phased_train.  You get a shell linting error if the local isn't defined separate from the assignment.

scripts/phased_training_resume.py
danmcp: It's part of the standard python library.  Maybe I am not following the question correctly though?

nathan-weinberg: Did not know this was part of the python standard lib! No worries then

nathan-weinberg (approved):

scripts/phased_training_resume.py
```diff
@@ -0,0 +1,202 @@
...
+    shutil.rmtree(phased_base_dir)
+
+    print("Finished phased training resume")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Phased Training")
+    parser.add_argument("--knowledge-data-path", help="Path to knowledge data path")
+    parser.add_argument("--skills-data-path", help="Path to skills data path")
+    parser.add_argument("--data-path", help="Data path")
```
JamesKunstle, line 192: Which 'data' is this? If you're passing knowledge and skills already, what's this other one?

JamesKunstle: Oh I see, it's the cache dir for phased_base_dir. Maybe changing the name to be more clear about it's usage would make sense.

scripts/phased_training_resume.py
```diff
@@ -0,0 +1,202 @@
...
+    assert "MT-Bench evaluation for Phase 2..." in result.output
+
+
+def assert_phase1_in_journal(result):
+    assert "SKIPPING: Training Phase 1/2; already in Journal" in result.output
+
+
+def assert_phase2_resumed(result):
+    assert_phase1(result)
+    assert_phase2(result)
```
JamesKunstle, line 132: I think we might also want `assert_phase1_in_journal` to be called? I may be misremembering what gets logged on restart.

JamesKunstle (commented):
Superb groundwork. I added a couple of questions but will be good to go if those are answered.

scripts/phased_training_resume.py
danmcp: It's DATA_HOME in the e2e-ci script which comes from platformdirs.user_data_dir().

Would data_home be better?

danmcp: I went ahead and switched to data-home-path

scripts/phased_training_resume.py
danmcp: I've updated the function names to hopefully make them less confusing.

scripts/e2e-ci.sh
```diff
@@ -296,6 +296,23 @@ test_phased_train() {
...
+    local knowledge_data_path
+    local skills_data_path
+    knowledge_data_path=$(find "${DATA_HOME}"/instructlab/datasets -name 'knowledge_train_msgs*' | head -n 1)
+    skills_data_path=$(find "${DATA_HOME}"/instructlab/datasets -name 'skills_train_msgs*' | head -n 1)
+
+    export INSTRUCTLAB_EVAL_MMLU_MIN_TASKS=true
```
alimaredia, line 308: We don't need this environment variable anymore since we aren't running MMLU in phased training. Also would you want to set `export INSTRUCTLAB_EVAL_FIRST_N_QUESTIONS=X` here?

These are both changes that shoud be reflected `test_phased_train()` as well.

scripts/e2e-ci.sh
```diff
@@ -389,6 +406,9 @@ test_exec() {
 
     # train tests
     if [ "$LARGE" -eq 1 ]; then
+      # Validate a single epoch per phase with resumption
```
alimaredia, line 409: Nit, we are running two epochs in each phase of phased training: https://github.com/instructlab/instructlab/blob/main/scripts/test-data/train-profile-l40s-x4.yaml#L23

danmcp: I think I am overriding it with --phased-phase1-num-epochs in phased_training_resume.py ?

scripts/e2e-ci.sh
danmcp: Good catch, I had just copied blindly without thinking about it.  I would vote for setting INSTRUCTLAB_EVAL_FIRST_N_QUESTIONS here but not with the non resume version as I would really like to validate the results with that version at some point.

scripts/e2e-ci.sh
alimaredia: Oh you're right, I missed that.

scripts/e2e-ci.sh
danmcp: Updated, thanks!

alimaredia:
This test looks good to me, but since it's for your feature @JamesKunstle, I'll let you go ahead and give the final approval.

JamesKunstle (approved):
Awesome, super glad to have this

PR #2532, opened 2024-10-24
Bump ami to include cudnn9 and nvrtc

n1hility:
Also update creation scripts

### DIFF
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -64,7 +64,7 @@ ec2__launch() {
         --security-group-ids "$EC2_SECURITY_GROUP_ID" \
         --subnet-id "$EC2_SUBNET_ID" \
         --key-name "$EC2_KEY_NAME" \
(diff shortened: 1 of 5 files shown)


scripts/infra/cloud-instance.sh
```diff
@@ -64,7 +64,7 @@ ec2__launch() {
         --security-group-ids "$EC2_SECURITY_GROUP_ID" \
         --subnet-id "$EC2_SUBNET_ID" \
         --key-name "$EC2_KEY_NAME" \
-        --block-device-mappings '{"DeviceName": "/dev/sda1","Ebs": {"VolumeSize": 300}}' \
+        --block-device-mappings '{"DeviceName": "/dev/sda1","Ebs": {"VolumeSize": 600}}' \
```
n1hility, line 67: FYI this increase is because we bumped the ami awhile back to 600, so will fail otherwise

danmcp (approved):

nathan-weinberg (approved):

PR #2538, opened 2024-10-25
split `ilab model chat` click cmds from functionality

alinaryan:
Refactored click command handling into a dedicated `cli` module and moved the core chat functionality into a separate module. This restructing lays the groundwork for future API handlers and multi-processing work.

**Checklist:**

### DIFF
--- a/src/instructlab/cli/model/chat.py
+++ b/src/instructlab/cli/model/chat.py
@@ -1,0 +1,157 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+import logging
+import pathlib
+
+# Third Party
+import click
+
+# First Party
+from instructlab import clickext
+from instructlab import configuration as cfg
+from instructlab.model.chat import chat_model
+
+logger = logging.getLogger(__name__)
+
+
+@click.command()
+@click.argument(
+    "question",
+    nargs=-1,
+    type=click.UNPROCESSED,
+)
+@click.option(
+    "-m",
+    "--model",
+    cls=clickext.ConfigOption,
+    required=True,  # default from config
+)
+@click.option(
+    "-c",
+    "--context",
+    cls=clickext.ConfigOption,
+    required=True,  # default from config
+)
+@click.option(
+    "-s",
+    "--session",
+    type=click.File("r"),
+    cls=clickext.ConfigOption,
+)
+@click.option(
+    "-qq",
+    "--quick-question",
+    is_flag=True,
+    help="Exit after answering question.",
+)
+@click.option(
+    "--max-tokens",
+    type=click.INT,
+    cls=clickext.ConfigOption,
+)
+@click.option(
+    "--endpoint-url",
+    type=click.STRING,
+    help="Custom URL endpoint for OpenAI-compatible API. Defaults to the `ilab model serve` endpoint.",
+)
+@click.option(
+    "--api-key",
+    type=click.STRING,
+    default=cfg.DEFAULTS.API_KEY,  # Note: do not expose default API key
+    help="API key for API endpoint. [default: config.DEFAULT_API_KEY]",
+)
+@click.option(
+    "--tls-insecure",
+    is_flag=True,
+    help="Disable TLS verification.",
+)
+@click.option(
+    "--tls-client-cert",
+    type=click.Path(),
+    default="",
+    show_default=True,
+    help="Path to the TLS client certificate to use.",
+)
+@click.option(
+    "--tls-client-key",
+    type=click.Path(),
+    default="",
+    show_default=True,
+    help="Path to the TLS client key to use.",
+)
+@click.option(
+    "--tls-client-passwd",
+    type=click.STRING,
+    default="",
+    help="TLS client certificate password.",
+)
+@click.option(
+    "--model-family",
+    help="Force model family to use when picking a chat template",
+)
+@click.option(
+    "--serving-log-file",
+    type=click.Path(path_type=pathlib.Path),
+    required=False,
+    help="Log file path to write server logs to.",
+)
+@click.option(
+    "-t",
+    "--temperature",
+    cls=clickext.ConfigOption,
+)
+@click.pass_context
+@clickext.display_params
+def chat(
+    ctx,
+    question,
+    model,
+    context,
+    session,
+    quick_question,
+    max_tokens,
+    endpoint_url,
+    api_key,
+    tls_insecure,
+    tls_client_cert,
+    tls_client_key,
+    tls_client_passwd,
+    model_family,
+    serving_log_file,
+    temperature,
+):
+    """Runs a chat using the modified model"""
+    chat_model(
+        question,
+        model,
+        context,
+        session,
+        quick_question,
+        max_tokens,
+        endpoint_url,
+        api_key,
+        tls_insecure,
+        tls_client_cert,
+        tls_client_key,
+        tls_client_passwd,
+        model_family,
+        serving_log_file,
+        temperature,
+        backend_type=ctx.obj.config.serve.server.backend_type,
+        host=ctx.obj.config.serve.server.host,
+        port=ctx.obj.config.serve.server.port,
+        current_max_ctx_size=ctx.obj.config.serve.server.current_max_ctx_size,
+        params=ctx.params,
+        backend_name=ctx.obj.config.serve.backend,
+        chat_template=ctx.obj.config.serve.chat_template,
+        api_base=ctx.obj.config.serve.api_base(),
+        gpu_layers=ctx.obj.config.serve.llama_cpp.gpu_layers,
+        max_ctx_size=ctx.obj.config.serve.llama_cpp.max_ctx_size,
+        vllm_model_family=ctx.obj.config.serve.vllm.llm_family,
+        vllm_args=ctx.obj.config.serve.vllm.vllm_args,
+        max_startup_attempts=ctx.obj.config.serve.vllm.max_startup_attempts,
+        logs_dir=ctx.obj.config.chat.logs_dir,
+        vi_mode=ctx.obj.config.chat.vi_mode,
+        visible_overflow=ctx.obj.config.chat.visible_overflow,
+    )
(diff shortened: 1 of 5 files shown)


alimaredia (approved):

src/instructlab/cli/model/chat.py
```diff
@@ -0,0 +1,269 @@
...
+            ctx.obj.config.serve,
+            model_path=model,
+            log_file=serving_log_file,
+        )
+        try:
+            # Run the llama server
+            api_base = backend_instance.run_detached(http_client(ctx.params))
+        except Exception as exc:
+            click.secho(f"Failed to start server: {exc}", fg="red")
+            raise click.exceptions.Exit(1)
```
jaideepr97, line 194: server invocation is also something i think maybe better left in the backend so that other clients can also trigger this action

nathan-weinberg (approved):

src/instructlab/cli/model/chat.py
```diff
@@ -0,0 +1,269 @@
...
+                click.secho(
+                    f"Failed to list models from {api_base}. Please check the API key and endpoint.",
+                    fg="red",
+                )
+                # Right now is_temp_server() does not check if a subprocessed vllm is up
+                # shut it down just in case an exception is raised in the try
+                # TODO: revise is_temp_server to check if a vllm server is running
+                if backend_instance is not None:
+                    backend_instance.shutdown()
+                raise click.exceptions.Exit(1) from exc
```
jaideepr97, line 249: interacting with an existing server that is already running and making decisions based on that also seems like something better suited for the backend IMO

alinaryan: I ended up splitting out the whole chat function :+1:

alinaryan:
@mbestavros I think you need to comment something before I can add you as a reviewer

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
         self._update_conversation(response_content.plain, "assistant")
 
 
+def chat_model(
+    ctx,
```
cdoern, line 540: ctx shouldn't be in here, when this is an API call it won't be able to pickle the context obj, please pass in the required fields from the config needed

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
...
+            "tls_client_passwd": tls_client_passwd,
+            "tls_insecure": tls_insecure,
+        },
+    ):
+        api_base = users_endpoint_url
+        if serving_log_file:
+            logger.warning(
+                "Setting serving log file (--serving-log-file) is not supported when the server is already running"
+            )
+        max_ctx_size = ctx.obj.config.serve.server.current_max_ctx_size
```
cdoern, line 587: should be passed in

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
...
+    serving_log_file,
+    temperature,
+):
+    """Runs a chat using the modified model"""
+    # pylint: disable=import-outside-toplevel
+    # First Party
+    from instructlab.model.backends.common import is_temp_server_running
+
+    max_ctx_size = None
+    backend_type = ctx.obj.config.serve.server.backend_type
```
cdoern, line 563: should be passed in

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
...
+    serve_server = ctx.obj.config.serve.server
```
cdoern, line 564: should be passed in

cdoern: this might need to be split up more even for future uses (non-primitives can't be pickled I don't think)

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
...
+                "Setting serving log file (--serving-log-file) is not supported when the server is already running"
+            )
+        max_ctx_size = ctx.obj.config.serve.server.current_max_ctx_size
+    else:
+        # If a log file is specified, write logs to the file
+        root_logger = logging.getLogger()
+        if serving_log_file:
+            log.add_file_handler_to_logger(root_logger, serving_log_file)
+
+        ctx.obj.config.serve.llama_cpp.llm_family = model_family
```
cdoern, line 594: this should be passed in

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
...
+        backend_instance = backends.select_backend(
+            ctx.obj.config.serve,
```
cdoern, line 596: same here, this might be kind of tricky. We might want to scope future work to remove the dependency on the entire cfg to select a backend

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
...
+            model_path=model,
+            log_file=serving_log_file,
+        )
+        backend_type = backend_instance.get_backend_type()
+        # if max tokens is the default, read from the serve cfg in case that has been overriden
+        # if serve has a different max_ctx_size, that is what we use.
+        if max_ctx_size is None and backend_type == backends.LLAMA_CPP:
+            max_ctx_size = ctx.obj.config.serve.llama_cpp.max_ctx_size
```
cdoern, line 604: same here, should be passed in

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
...
+
+                # Currently, we only present a single model so we can safely assume that the first model
+                server_model = models.data[0].id if models is not None else None
+
+                # override 'model' with the first returned model if not provided so that the chat print
+                # the model used by the server
+                model = (
+                    server_model
+                    if server_model is not None
+                    and server_model != ctx.obj.config.chat.model
```
cdoern, line 651: same here should be passed in

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
...
+                # TODO: revise is_temp_server to check if a vllm server is running
+                if backend_instance is not None:
+                    backend_instance.shutdown()
+                sys.exit(1)
+
+    try:
+        chat_cli(
+            ctx,
+            api_base=api_base,
+            config=ctx.obj.config.chat,
```
cdoern, line 671: same here. just look at `chat_cli` and see what fields are used and pass those in from the cli pkg

cdoern (commented):
there are a number of ctx objects leftover. These are click specific and should probably be handled in the CLI package because in the future we won't be able to parse these in an API or in sub processing

src/instructlab/model/chat.py
```diff
@@ -789,6 +536,157 @@ def start_prompt(
...
+                print(
+                    f"Failed to list models from {api_base}. Please check the API key and endpoint.",
+                    file=sys.stderr,
+                )
+                # Right now is_temp_server() does not check if a subprocessed vllm is up
+                # shut it down just in case an exception is raised in the try
+                # TODO: revise is_temp_server to check if a vllm server is running
+                if backend_instance is not None:
+                    backend_instance.shutdown()
+                sys.exit(1)
```
mbestavros, line 665: This _feels_ like a large block of code to be behind a single if statement, but I don't know if there's a better way to organize it...

mbestavros (commented):
I don't think I have a ton to add that @cdoern didn't say already - we definitely want this to be API friendly.

I left one organizational comment, but feel free to ignore.

src/instructlab/model/chat.py
alinaryan: I'm going to open up a follow up issue for this

src/instructlab/model/chat.py
```diff
@@ -789,10 +536,177 @@ def start_prompt(
...
+                )
+                logger.debug(f"Using model from server {model}")
+            except ilabclient.ClientException:
+                print(
+                    f"Failed to list models from {api_base}. Please check the API key and endpoint.",
+                    file=sys.stderr,
+                )
+                # Right now is_temp_server() does not check if a subprocessed vllm is up
+                # shut it down just in case an exception is raised in the try
+                # TODO: revise is_temp_server to check if a vllm server is running
```
courtneypacheco, line 678: Can we file a git issue for this so we have it on our radar?

src/instructlab/model/chat.py
```diff
@@ -789,10 +536,177 @@ def start_prompt(
...
+    api_base,
+    gpu_layers,
+    max_ctx_size,
+    vllm_model_family,
+    vllm_args,
+    max_startup_attempts,
+    logs_dir,
+    vi_mode,
+    visible_overflow,
+):
```
courtneypacheco, line 572: Maybe a future enhancement, but I think when we have a lot of function inputs like this, it can be easy to mess up the ordering, etc.

I'd personally prefer to have a `dataclass` as a single input to this function. Using a `dataclass` is also more flexible if we ever have to add/remove/modify function inputs because the `dataclass` (and therefore function) won't care about parameter ordering.

Example:

```python3
@dataclass
ChatModelConfig:
    question: str
    quick_question: str

    # etc etc
```

Then we can do

```python3
# Define this somewhere
chat_model_config = ChatModelConfig(
    question=question,
    quick_question=question,
    # etc
)


def chat_model(chat_model_config):
    # do stuff

```

src/instructlab/cli/model/chat.py
```diff
@@ -0,0 +1,158 @@
...
+    tls_insecure,
+    tls_client_cert,
+    tls_client_key,
+    tls_client_passwd,
+    model_family,
+    serving_log_file,
+    temperature,
+):
+    """Runs a chat using the modified model"""
+    chat_model(
```
courtneypacheco, line 125: Same here about the `dataclass`

src/instructlab/model/chat.py
cdoern: I think this is outside the scope of this PR

src/instructlab/model/chat.py
alinaryan: yes, great idea

src/instructlab/model/chat.py
alinaryan: I can file an issue for this as well

cdoern (approved):

PR #2559, opened 2024-10-31
refactor: move http_client... to client_utils.py

makelinux:
The utils.py file has become large and disorganized. Moving http_client(), HttpClientParams(),
and get_ssl_cert_config(), to client_utils.py would improve organization and better align functionality
with the file's purpose.


danmcp (commented):
+1 to this change in general.  It does leave me with a few questions:

- Is client.py meant to contain all helpers for connecting to http apis?  Or any external apis?
- Could the name be more precise?  client_utils?  http_utils?  api_utils?

Maybe a comment in the file as to its purpose would be appropriate.

danmcp (approved):

PR #2467, opened 2024-10-16
split `ilab model serve` click cmds from functionality

alinaryan:
Refactored click command handling into a dedicated `cli` module and moved the core serve functionality into a separate module. This restructing lays the groundwork for future API handlers and multi-processing work.

**Checklist:**

- [ ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Functional tests have been added, if necessary.
- [ ] E2E Workflow tests have been added, if necessary.

PR #2566, opened 2024-10-31
llama backend fix

cdoern:
llama-cpp was switched to llama_cpp but only in certain parts of the codebase making it hard for users to specify the backend they want to use.

Switch all instances of llama_cpp backend type back to llama-cpp

**Checklist:**

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -193,6 +193,12 @@ test_oci_model_download_with_vllm_backend(){
     fi
 }
 
+test_llama_backend(){
+    ilab model serve --backend llama-cpp &
+    PID_SERVE=$!
+    wait_for_server
+}
+
 # check that ilab model serve is working
 test_bind_port(){
     local formatted_script
@@ -672,5 +678,8 @@ cleanup
 test_model_print
 cleanup
 test_log_format
+cleanup
+test_llama_backend
+cleanup
 
 exit 0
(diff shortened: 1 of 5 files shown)


nathan-weinberg:
@cdoern @alinaryan we are sure this is inclusive of everything changed in https://github.com/instructlab/instructlab/pull/2467?

Also, we should add an additional unit test to ensure regressions like this can't happen again

nathan-weinberg:
@Mergifyio backport release-v0.20

cdoern:
@nathan-weinberg I dont think we need an additional unit test. The unit test to account for this was changed causing it to pass w/ the incorrect value, same could happen for any cfg entry honestly.

cdoern:
I did the side by side comparison myself, this is the extent of the breaking changes.

nathan-weinberg:
@cdoern we should add a test ensuring that the values are consistent - afraid I'll have to insist on this since there was a regression - this is standard practice when regressions occur - obviously the existing test was not sufficient to expose to @alinaryan what the root cause of the issue was in the first place - non-Maintainers are more even likely to make a similar change by accident

cdoern:
added two tests

alinaryan:
>we are sure this is inclusive of everything changed in https://github.com/instructlab/instructlab/pull/2467?

@nathan-weinberg yes I sat down and checked each change manually. this covers it

nathan-weinberg (approved):

alinaryan (approved):

PR #2577, opened 2024-11-04
add /sc cmd for chat

reidliu41:
- add /sc cmd to customize the context

**Checklist:**


reidliu41:
```
    tensor.tofile(fout)

Epoch 0: 100%|██████████| 11/11 [09:54<00:00, 54.08s/it]
+ rm -rf /tmp/tmp.a1ph0pUhqV
```

reidliu41:
hi team, anyone can help to take a look?

jaideepr97:
Hmmm, contexts are closely tied to what system prompt gets used for the model 
I think this is veering into the domain of supporting bring-your-own-system prompts. This is an area we do we want to support in the future, but might need some research on how we want to do this 
would like to get @Maxusmusti's input on this

jaideepr97 (commented):
Hmmm, contexts are closely tied to what system prompt gets used for the model
I think this is veering into the domain of supporting bring-your-own-system prompts. This is an area we do we want to support in the future, but might need some research on how we want to do this
would like to get @Maxusmusti's input on this

nathan-weinberg:
@Maxusmusti could you PTAL?

src/instructlab/model/chat.py
```diff
@@ -585,6 +587,32 @@ def _handle_list_contexts(self, _):
...
+                    "**ERROR**: Please use `/sc context_name custom_context_text`. "
+                    "Note: Use underscores (`_`) to join multiple words in `context_name`."
+                )
+            )
+            raise KeyboardInterrupt
+
+        context_name = parts[1]
+        custom_context_text = parts[2]
+
+        CONTEXTS[context_name] = custom_context_text
```
booxter, line 604: I'd rather not allow to override default contexts. Yes, you can restart chat session, but it would be better if the user can always reset to default, whatever they do.

src/instructlab/model/chat.py
```diff
@@ -585,6 +587,32 @@ def _handle_list_contexts(self, _):
         self._sys_print(Markdown(f"**Available contexts:**\n\n{context_list}"))
         raise KeyboardInterrupt
 
+    def _handle_set_custom_context(self, content):
+        parts = content.split(" ", 2)
```
booxter, line 591: please `rstrip` the input first, otherwise you allow to pass `/sc test_context ` (note the trailing space).

booxter (changes requested):

reidliu41:
doc issue: https://github.com/instructlab/instructlab/pull/3169

src/instructlab/model/chat.py
```diff
@@ -373,6 +375,42 @@ def _handle_list_contexts(self, _):
...
+                    "Note: Use underscores (`_`) to join multiple words in `context_name`."
+                )
+            )
+            raise KeyboardInterrupt
+
+        context_name = parts[1]
+        custom_context_text = parts[2]
+
+        # Prevent overriding default contexts
+        if context_name in CONTEXTS:
```
booxter, line 394: Don't you now forbid overriding an existing context too? (The message suggests it's not your intent; and I think we should be able to override a custom context.)

booxter (changes requested):

booxter (approved):

PR #2590, opened 2024-11-06
add json option for show

reidliu41:
Improve the `show` command. [origin one 2264](https://github.com/instructlab/instructlab/pull/2264)
- Background
So far, `ilab config show` only simply print the `config.yaml`, and config become more and more according to the repo become stronger.
The issue is hard to read/search the config what we want. e.g. there are paths, model, data, train ..etc.

Improvement actions:
- add `json` output without comments
```
$ ilab config show -j
{
    "chat": {
        "context": "default",
        "greedy_mode": false,
        "logs_dir": "/Users/xx/.local/share/instructlab/chatlogs",
        "max_tokens": null,
        "model": "/Users/xx/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf",
        "session": null,
        "vi_mode": false,
        "visible_overflow": true
    },
    "evaluate": {
        "base_branch": null,
        "base_model": "instructlab/granite-7b-lab",
        "branch": null,
        "gpus": null,
        "mmlu": {
            "batch_size": "auto",
            "few_shots": 5
        },
        "mmlu_branch": {
```

- add key search e.g. -k chat.model, easy to check the config, or combine `-k` with `-j` to list section
```

ilab config show -k chat -j
{
    "context": "default",
    "greedy_mode": false,
    "logs_dir": "/root/.local/share/instructlab/chatlogs",
    "max_tokens": null,
    "model": "/root/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf",
    "session": null,
    "vi_mode": false,
    "visible_overflow": true
}
```

**Checklist:**

### DIFF
--- a/src/instructlab/config/show.py
+++ b/src/instructlab/config/show.py
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
+
 # Standard
+from typing import Any, Dict
+import json
 import sys
 
 # Third Party
@@ -14,15 +17,53 @@ yaml = ruamel.yaml.YAML()
 yaml.indent(mapping=2, sequence=4, offset=2)
 
 
+def get_nested_value(data_dict: Dict[str, Any], path: str) -> Any:
+    """Retrieve nested value using dot-separated path."""
+    keys = path.split(".")
+    for key in keys:
+        data_dict = data_dict[key]
+    return data_dict
+
+
 @click.command()
 @click.pass_context
+@click.option(
+    "--json",
+    "-j",
+    "json_format",
+    is_flag=True,
+    help="Output in JSON format instead of YAML.",
+)
+@click.option(
+    "--key",
+    "-k",
+    "key_path",
+    default=None,
+    help="Show only a specific section of the configuration. Can be used in conjunction with -j. e.g. -k chat or -k chat.context",
+)
 @clickext.display_params
-def show(ctx: click.Context) -> None:
-    """Displays the current config as YAML"""
-    # TODO: make this use pretty colors like jq/yq
+def show(ctx: click.Context, json_format: bool, key_path: str) -> None:
+    """Displays the current config as YAML or JSON, with an option to extract a specific key."""
+
     try:
         commented_map = configuration.config_to_commented_map(ctx.obj.config)
-    except YAMLError as e:
-        click.secho(f"Error loading config as YAML: {e}", fg="red")
+        result = (
+            commented_map if not key_path else get_nested_value(commented_map, key_path)
+        )
+
+        if json_format:
+            json_output = json.dumps(result, indent=4)
+            click.echo(json_output)
+        else:
+            if isinstance(result, (dict, list)):
+                yaml.dump(result, sys.stdout)
+            else:
+                click.echo(result)
+
+    except KeyError:
+        click.secho(f"Key not found: '{key_path}'", fg="red")
         ctx.exit(2)
-    yaml.dump(commented_map, sys.stdout)
+
+    except YAMLError as e:
+        click.secho(f"Error loading config: {e}.", fg="red")
+        ctx.exit(2)
--- a/tests/test_lab_config.py
+++ b/tests/test_lab_config.py
@@ -1,15 +1,69 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+import json
 import logging
 import pathlib
 
 # Third Party
 from click.testing import CliRunner
+from ruamel.yaml import YAML
+import ruamel.yaml
 import yaml
 
 # First Party
 from instructlab import configuration, lab
+
+test_source_config_data = """
+# Chat configuration section.
+chat:
+  # Predefined setting or environment that influences the behavior and responses of
+  # the chat assistant. Each context is associated with a specific prompt that
+  # guides the assistant on how to respond to user inputs. Available contexts:
+  # default, cli_helper.
+  # Default: default
+  context: default
+  # Sets temperature to 0 if enabled, leading to more deterministic responses.
+  # Default: False
+  greedy_mode: false
+  # Directory where chat logs are stored.
+  # Default: /data/instructlab/chatlogs
+  logs_dir: /data/instructlab/chatlogs
+  # The maximum number of tokens that can be generated in the chat completion. Be
+  # aware that larger values use more memory.
+  # Default: None
+  max_tokens:
+  # Model to be used for chatting with.
+  # Default: /cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
+  model: /cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf
+  # Filepath of a dialog session file.
+  # Default: None
+  session:
+  # Enable vim keybindings for chat.
+  # Default: False
+  vi_mode: false
+  # Renders vertical overflow if enabled, displays ellipses otherwise.
+  # Default: True
+  visible_overflow: true
+# General configuration section.
+general:
+  # Debug level for logging.
+  # Default: 0
+  debug_level: 0
+  # Log format. https://docs.python.org/3/library/logging.html#logrecord-attributes
+  # Default: %(levelname)s %(asctime)s %(name)s:%(lineno)d: %(message)s
+  log_format: '%(levelname)s %(asctime)s %(name)s:%(lineno)d: %(message)s'
+  # Log level for logging.
+  # Default: INFO
+  log_level: INFO
+"""
+
+
+def remove_comment_indentation(yaml_output: str) -> str:
+    return "\n".join(
+        line.lstrip() if line.lstrip().startswith("#") else line
+        for line in yaml_output.splitlines()
+    )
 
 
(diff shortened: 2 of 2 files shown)


src/instructlab/config/show.py
```diff
@@ -14,15 +16,53 @@
...
+        if isinstance(result, (dict, list)):
+            if json_format:
+                json_output = json.dumps(result, indent=4)
+                click.echo(json_output)
+            else:
+                click.echo(output)
+        else:
+            click.echo(result)
+
+    except (YAMLError, KeyError) as e:
```
danmcp, line 66: Any reason not to separate KeyError here to give a more specific error message?

It might also be better to handle the keyerror more explicitly.  Currently it would print:
```
% ilab config show -k evaluate.invalid
Error loading config or key not found: 'invalid'.
```

I think printing the key up to the error might be more helpful.  In this case:

```
Error loading key: 'evaluate.invalid'
```

src/instructlab/config/show.py
```diff
@@ -14,15 +16,53 @@
...
+    "json_format",
+    is_flag=True,
+    help="Output in JSON format instead of YAML.",
+)
+@click.option(
+    "--key",
+    "-k",
+    "key_path",
+    default=None,
+    help="Extract a specific key from the configuration, e.g. -k chat or -k chat.context, also can be used with -j.",
```
danmcp, line 41: Suggestion:

Show only a specific section of the configuration.  Can be used in conjunction with -j. e.g. -k chat or -k chat.context

src/instructlab/config/show.py
```diff
@@ -14,15 +16,53 @@
...
+    keys = path.split(".")
+    for key in keys:
+        data_dict = data_dict[key]
+    return data_dict
+
+
 @click.command()
 @click.pass_context
+@click.option(
+    "--json",
```
danmcp, line 30: This option is complicated by having logging output to stdout.  I think we will need to address moving logging to a separate file before we can support json output consistently.  See https://github.com/instructlab/instructlab/pull/2470 for additional conversation on the topic.

src/instructlab/config/show.py
```diff
@@ -14,15 +16,53 @@
...
+
     try:
         commented_map = configuration.config_to_commented_map(ctx.obj.config)
-    except YAMLError as e:
-        click.secho(f"Error loading config as YAML: {e}", fg="red")
+        result = (
+            commented_map if not key_path else get_nested_value(commented_map, key_path)
+        )
+
+        stream = compat.StringIO()
```
danmcp, line 53: Looks like this section could be moved to the else below?

danmcp: And once it's moved, I think it could be just:

`yaml.dump(result, sys.stdout)`

danmcp (changes requested):
Seems like this is a good candidate for unit tests

src/instructlab/config/show.py
```diff
@@ -14,15 +17,53 @@
...
+            json_output = json.dumps(result, indent=4)
+            click.echo(json_output)
+        else:
+            if isinstance(result, (dict, list)):
+                yaml.dump(result, sys.stdout)
+            else:
+                click.echo(result)
+
+    except KeyError:
+        click.secho(f"Error loading key: '{key_path}'. Key not found.", fg="red")
```
danmcp, line 64: Putting the period and text after the key_path makes it harder to pick out

Suggestion:

`f"Key not found: '{key_path}'"`

src/instructlab/config/show.py
```diff
@@ -14,15 +17,53 @@
...
+    "key_path",
+    default=None,
+    help="Show only a specific section of the configuration. Can be used in conjunction with -j. e.g. -k chat or -k chat.context",
+)
 @clickext.display_params
-def show(ctx: click.Context) -> None:
-    """Displays the current config as YAML"""
-    # TODO: make this use pretty colors like jq/yq
+def show(ctx: click.Context, json_format: bool, key_path: str) -> None:
+    """Displays the current config as YAML or JSON, with an option to extract specific key."""
```
danmcp, line 46: extract a specific key
or
extract a specific section

tests/test_lab_config.py
```diff
@@ -22,6 +23,187 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+        },
+        "version": "1.0.0",
+    }
+
+    result_json = cli_runner.invoke(
+        lab.ilab, ["--config", str(test_config_path), "config", "show", "--json"]
+    )
+
+    assert result_json.exit_code == 0
+    parsed_json = json.loads(result_json.stdout)
```
danmcp, line 139: Note: this is what potentially wouldn't work with debug logging enabled.

tests/test_lab_config.py
```diff
@@ -22,6 +23,187 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+
+    assert parsed_json == expected_json_output
+
+
+def test_ilab_config_show_key_general(
+    cli_runner: CliRunner, testdata_path: pathlib.Path
+) -> None:
+    test_config_path = testdata_path / "default_config.yaml"
+
+    # Expected output as a JSON string for --key general
```
danmcp, line 149: I think it would be better to extract this from the config programmatically in this test.

tests/test_lab_config.py
```diff
@@ -22,6 +23,187 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+    # Compare the actual output with the expected JSON structure for general
+    assert json.loads(result_key_general.stdout) == json.loads(expected_general_output)
+
+
+def test_ilab_config_show_key_general_yaml(
+    cli_runner: CliRunner, testdata_path: pathlib.Path
+) -> None:
+    test_config_path = testdata_path / "default_config.yaml"
+    # Expected YAML structure for --key general without --json
+    expected_yaml_output = """\
```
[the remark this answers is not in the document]
reidliu41: sorry about this that caused the format issue, so keep it manually might be better.

tests/test_lab_config.py
```diff
@@ -22,6 +23,83 @@ def test_ilab_config_show(cli_runner: CliRunner) -> None:
...
+
+    logging.disable(logging.CRITICAL)
+    result_json = cli_runner.invoke(
+        lab.ilab, ["--config", str(test_config_path), "config", "show", "--json"]
+    )
+    logging.disable(logging.NOTSET)
+
+    assert result_json.exit_code == 0
+    parsed_json = json.loads(result_json.stdout)
+    assert "version" in parsed_json
```
danmcp, line 39: nit: You don't need this assert and the version number check below

tests/test_lab_config.py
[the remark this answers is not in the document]
danmcp: Can you clarify what you mean here?  Are you saying the formatting didn't match if you extract the value to compare programmatically?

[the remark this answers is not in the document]
reidliu41: yes. maybe also just compare the content.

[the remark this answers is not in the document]
danmcp: As a compromise what if you pulled a smaller section like `train.device`?

Or whatever is the smallest amount of config that isn't likely to change.

danmcp (changes requested):
To clarify my concern about the --json option. I know we can make it work with the unit tests.  My concern is more what users are going to experience if we support the option as implemented here.  AFAICT, we would have to tell users to disable logging to be able to parse the json output from commands reliably.  I think a change across the board for the CLI is necessary before we can start to support the --json option.  Some of the options on how to handle that are discussed here: https://github.com/instructlab/instructlab/pull/2470/files#r1804795659

I don't feel comfortable approving the --json changes until that's resolved.  The --key option could be separated if you don't want to hold up that change on the --json discussion/design.

Also note that https://github.com/instructlab/instructlab/pull/2470 is using the option --format.  Whatever we choose, it should be consistent across commands.

reidliu41:
> To clarify my concern about the --json option. I know we can make it work with the unit tests. My concern is more what users are going to experience if we support the option as implemented here. AFAICT, we would have to tell users to disable logging to be able to parse the json output from commands reliably. I think a change across the board for the CLI is necessary before we can start to support the --json option. Some of the options on how to handle that are discussed here: https://github.com/instructlab/instructlab/pull/2470/files#r1804795659
> 
> I don't feel comfortable approving the --json changes until that's resolved. The --key option could be separated if you don't want to hold up that change on the --json discussion/design.
> 
> Also note that #2470 is using the option --format. Whatever we choose, it should be consistent across commands.

Great, I also not very clear the json issue(will monitor [2470](https://github.com/instructlab/instructlab/pull/2470)). Let's hold this one for json. I will separate new one for `--key option`. If we can start to support the --json, I will re-check this one.

courtneypacheco:
Hi @reidliu41. Are you still working on this PR? I see that #2470  has been closed.

reidliu41:
@courtneypacheco thanks for remind, I think will not continue since https://github.com/instructlab/instructlab/pull/2470 closed,

courtneypacheco:
Thanks! Closing this PR. Feel free to reopen it if you change your mind!

PR #2612, opened 2024-11-10
Full Train Memory Optimizations

cdoern:
Currently, full train only runs on hardware with 64GB of Memory.

With some optimizations, this can be cut down to a 16GB requirement, massively increasing the amount of users able to run this workflow.

`AutoModel` and `AutoConfig` both have a torch_dtype and low_cpu_mem_usage fields. setting the dtype to `auto` and `low_cpu_mem_usage` to True on all architectures yields improved results.

The _only_ caveat here is that Intel CPUs preform siginficantly worse when using the `torch_dtype=auto` setting. We should stick with fp32 on intel CPUs

**Checklist:**

### DIFF
--- a/src/instructlab/cli/model/train.py
+++ b/src/instructlab/cli/model/train.py
@@ -364,6 +364,11 @@ def clickpath_setup(is_dir: bool) -> click.Path:
     is_flag=True,
     help="Clear phased cache (journal, checkpoints, metadata). Helpful paired with '--skip-user-confirm'",
 )
+@click.option(
+    "--optimize-memory",
+    is_flag=True,
+    help="Optimize Memory Usage on CPU and MacOS. This uses the torch_dtype='auto' instead of float32",
+)
 @click.pass_context
 @clickext.display_params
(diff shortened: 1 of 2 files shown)


cdoern:
for some extra context here: I have never been able to run `--pipeline full` on my 16GB M3. I can now run this workflow with RAM to spare.

alinaryan (commented):
Good work!! Curious, how did you find that Intel CPUs perform worse with the torch_dtype=auto setting?

nathan-weinberg (approved):

alinaryan (approved):

cdoern:
@nathan-weinberg I changed this a bit:

CPU training on both AMD and Intel can preform better OR worse with this enabled depending on their generation.

so. I am making it `--optimize-memory` (default false). This allows users on CPU to toggle this to `true` enabling the setting. This is always enable on M-Series Macs

For example, on my AMD laptop and Server this improved performance, but on the runners is degrades performance

PR #2615, opened 2024-11-11
Fix Intel Auto-Detection and Profile Selection Menu

### LINKED ISSUE #2608: System profile prompting has inconsistent numbering
**What is the current experience of ilab you've found issue with?**
We have several menus where `[0]` is the "default" option but users can't actually see what it is and get misleading or confusing behavior if they select it

Example below - I hit enter without picking a number, and ilab assumed I picked NVIDIA.

```bash
Please choose a system profile.
 Profiles set hardware-specific defaults for all commands and sections of the configuration.
First, please select the hardware vendor your system falls into
[1] AMD
[2] APPLE
[3] INTEL
[4] NVIDIA
Enter the number of your choice [0]: 
[1] NVIDIA A100 X2
[2] NVIDIA A100 X4
[3] NVIDIA A100 X8
[4] NVIDIA H100 X2
[5] NVIDIA H100 X4
[6] NVIDIA H100 X8
[7] NVIDIA L4 X8
[8] NVIDIA L40S X4
[9] NVIDIA L40S X8
Enter the number of your choice [hit enter for hardware defaults] [0]: 
No profile selected - any hardware acceleration for training must be configured manually.
```

**How do you think the described experience should be?**
The first menu should list `[0] None of the below` and default to exiting early with no profile selected with the same message we give for the second menu, i.e. `No profile selected - any hardware acceleration for training must be configured manually.`

The second menu should list `[0] Hardware defaults (no profile)`

**What is the benefit of the change you are proposing?**
Much less confusing experience for the user

cdoern:
Intel returns Intel(r) for their CPUs, so we need to do a more general check than strict matching `intel`.

Also, add a [0] option for profiles that just sets the pydantic defaults. It is unclear who would get into this scenario with the auto-detection working
but if a user is running on such a system, the pydantic defaults might just be what they're looking f

Screenshot from 2024-11-11 21-19-20

resolves #2608 

**Checklist:**

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -149,10 +149,13 @@ def init(
 
 
 # prompt_user_to_choose_vendors asks the user which hardware vendor best matches their system
-def prompt_user_to_choose_vendors(arch_family_processors: dict[str, list[list[str]]]):
+def prompt_user_to_choose_vendors(
+    arch_family_processors: dict[str, list[list[str]]],
+) -> str | None:
     """
     prompt_user_to_choose_vendors asks the user which hardware vendor best matches their system
     """
+    key = None
     click.echo(
         click.style(
             "Please choose a system profile.\n Profiles set hardware-specific defaults for all commands and sections of the configuration.",
@@ -167,9 +170,10 @@ def prompt_user_to_choose_vendors(arch_family_processors: dict[str, list[list[st
             fg="white",
         )
     )
+    click.echo("[0] NO SYSTEM PROFILE")
     keys = list(arch_family_processors.keys())
-    for idx, key in enumerate(keys, 1):
-        print(f"[{idx}] {key.upper()}")
+    for idx, k in enumerate(keys, 1):
+        click.echo(f"[{idx}] {k.upper()}")
 
     system_profile_selection = click.prompt(
         "Enter the number of your choice",
@@ -186,7 +190,11 @@ def prompt_user_to_choose_vendors(arch_family_processors: dict[str, list[list[st
                 fg="white",
             )
         )
-        click.echo("[0] No system profile")
+        click.echo("[0] NO SYSTEM PROFILE")
+    elif system_profile_selection == 0:
+        click.echo(
+            "No profile selected - ilab will use generic code defaults - these may not be optimized for your system."
+        )
     return key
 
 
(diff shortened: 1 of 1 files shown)


nathan-weinberg:
cc @JustinXHale

JustinXHale:
@cdoern with the auto detection will it bypass the [Choose system hardware] question all together? Or will users still be prompted to select?

nathan-weinberg:
@cdoern can you post the output of the menu now similar to what I did with the original issue so we can easily compare?

cdoern:
@nathan-weinberg is the unit test runner an x86 mac?

cdoern:
if so, testing this is going to be realllllly tricky

nathan-weinberg:
I don't know, but that should be a pretty simple thing to look up

cdoern:
@nathan-weinberg @JustinXHale giving up on the no `[0]` front. Having no default:

1. broke our unit tests
2. will break anyone who has automation build around this probably

making a `[0]` which sets the defaults we store in code and applies no profile

I also fixed intel auto detect. 

Find a photo of the output attached to the PR description. thanks

src/instructlab/config/init.py
```diff
@@ -213,7 +221,7 @@ def prompt_user_to_choose_profile(key, arch_family_processors) -> Config | None:
         cfg = read_config(file)
     elif system_profile_selection == 0:
         click.echo(
-            "No profile selected - any hardware acceleration for training must be configured manually."
+            "No profile selected - any hardware acceleration must be configured manually."
```
nathan-weinberg, line 224: Is it accurate to just say that hardware acceleration must be configured manually?

I would suggest maybe something like 
`No profile selected - ilab will use generic code defaults - these may not be optimized for your system.`

cdoern: that seems like kind of a mouthful, i feel like folks don't care that its the code defaults, but they do care that no acceleration was directly configured in their cfg (gpus: 0)

nathan-weinberg: But the configuration being done extends to a lot more than hardware acceleration - I think that's worth calling out, as well as _where_ the values that are being used are coming from (if I don't have a profile, where is my config coming from?)

nathan-weinberg (approved):

PR #2634, opened 2024-11-12
bump SDG minimum version

cdoern:
SDG 0.5.0 is out, 0.21 will be based off of this, bump the version

**Checklist:**

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ httpx>=0.25.0
 instructlab-eval>=0.3.1
 instructlab-quantize>=0.1.0
 instructlab-schema>=0.4.0
-instructlab-sdg>=0.3.0
+instructlab-sdg>=0.5.0
 instructlab-training>=0.5.0
 llama_cpp_python[server]==0.2.79
 mlx>=0.5.1,<0.6.0; sys_platform == 'darwin' and platform_machine == 'arm64'
--- a/src/instructlab/data/generate_data.py
+++ b/src/instructlab/data/generate_data.py
@@ -2,10 +2,6 @@
 import logging
 
 # Third Party
-from instructlab.sdg.generate_data import generate_data
-
-# pylint: disable=ungrouped-imports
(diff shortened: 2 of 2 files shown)


cdoern:
lets see if the dependency tests pass here

khaledsulayman (approved):
That was quick! 🏎️🏎️

LGTM

alimaredia (approved):

nathan-weinberg:
Please run the large E2E run against this as well, thanks!

tests/testdata/leanimports.py
```diff
@@ -5,7 +5,7 @@
 import sys
 
 # block slow imports
-for unwanted in ["deepspeed", "llama_cpp", "torch", "vllm"]:
+for unwanted in ["deepspeed", "llama_cpp", "vllm"]:
```
danmcp, line 8: Is this the only option or can we move the new torch import to a lazy load?

danmcp, line 8: Is this the only option or can we move the new torch import to a lazy load?

cdoern: the issue is coming directly from docling, so it seems like we are stuck with this change. If it were in ilab or SDG it would be an easy fix but its either coming from docling or something they import

danmcp: Would it not work to move:

```
    from instructlab.sdg.generate_data import generate_data
    from instructlab.sdg.utils import GenerateException
```

inside `gen_data` in generate_data.py?

cdoern: I figured these types of errors hinged more on the types of imports happening at lower levels. Good to know changing top level imports like this fixes these issues too! Let me give it a shot

danmcp (approved):

tests/testdata/leanimports.py
cdoern: looks like it worked, thanks

PR #2645, opened 2024-11-13
fix: add upsampling as an option

RobotSail:
In SDG today, when user-generated knowledge data is mixed in with skills it is done
without consideration for the difference in size between knowledge and skills.
As a result, when the skills dataset greatly outnumbers knowledge by several orders of magnitude,
we see model degradation in terms of knowledge learned. This commit resolves that by
allowing the CLI to accept an `upsample_amount` which controls how much of the knowledge
we upsample into the CLI. I.e, how many samples of the knowledge do we need in the skills dataset
for it to not forget. This is implemented in SDG by simply replicating those samples up to a given
number N. This commit simply exposes this setting in the CLI.

For example, if you had a knowledge dataset which you generated and was 20 samples in length,
specifying an upsample_amount of 100 would cause the data to be duplicated until it is 100 samples in length.
That is - the resultant dataset would have (ideally) 5 copies of the original dataset.

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -297,6 +297,10 @@ class _generate(BaseModel):
         default_factory=lambda: DEFAULTS.SEED_FILE,
         deprecated=True,
     )
+    upsample_amount: Optional[int] = Field(
+        default=5000,
+        description="We will duplicate generated knowledge samples up to this number when mixing them in with our skills dataset.",
+    )
 
 
 class _mmlu(BaseModel):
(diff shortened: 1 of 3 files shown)


src/instructlab/configuration.py
```diff
@@ -297,6 +297,10 @@ class _generate(BaseModel):
         default_factory=lambda: DEFAULTS.SEED_FILE,
         deprecated=True,
     )
+    upsample_amount: Optional[int] = Field(
+        default=5000,
```
cdoern, line 301: so 5000 will be passed in always to SDG, is this the behavior we want?

cdoern (commented):
one question about default, if possible please add a test once SDG allows for this

PR #2647, opened 2024-11-14
Don't fail fast when one of the unit/func test envs fails

danmcp:
Previously if 3.10 ubuntu failed (for example), 3.11 and macos wouldn't report results.  This should allow for all of them to finish.  It shouldn't waste much resource if they are all going to fail the same way anyway but now let you know if the failure is version dependent.

**Checklist:**


jaideepr97 (approved):
thanks @danmcp!

nathan-weinberg (approved):

markstur (commented):
LGTM -- easy enough to undo if we decide we liked fail-fast better

Issue #2604, opened 2024-11-08
address & port should be first-class settings in ilab model serve


RobotSail:
When I want to serve a model, especially on a system with many GPUs, I should be able to specify a port and optionally a hostname so as to not collide with the default. 
We can update this value today, but it requires going into the config. We should instead have this be a first-class flag that can be provided on the CLI.

I.e:
```bash
ilab model serve --port 8001 --host 0.0.0.0 --model-path ~/my/super/cool/model
```

Issue #1823, opened 2024-07-22
Please add command line argument for listening to different IPs, e.g. "0.0.0.0"


VfBfoerst:
Hello :)  
  
**Is your feature request related to a problem? Please describe.**
After setting up instructlab via container (Podman), I had to edit the code directly so the server would listen to 0.0.0.0. This was needed because of the port-forwarding of podman. So it must listen to 0.0.0.0 to get access from remote to the api. 

**Describe the solution you'd like**
A command-line option like "--host 0.0.0.0" and "--port 8000" would be a nice solution i think:  
```bash
ilab model serve --host 0.0.0.0 --port 8000
```
Additionally, there could be ENVs set which lead to the same result, e.g.  
```bash
ILAB_HOST=0.0.0.0
ILAB_PORT=8000
```

Thank you very much, this project seems to be pretty cool :)  
Cheers.

leseb:
https://github.com/instructlab/instructlab/pull/1446

nathan-weinberg:
Closing this as a dup of https://github.com/instructlab/instructlab/issues/2604

Once that PR is in the final issue will be closed as well

PR #2648, opened 2024-11-14
add --host and --port for serve

reidliu41:
add host_port arg to the serve_backend

```
$ ilab model serve --help
  -h, --host TEXT             Host to serve on. [default: 127.0.0.1; config:
                              'serve.server.host']
  -p, --port INTEGER          Port to serve on. [default: 8000; config:
                              'serve.server.port']
```

Seems some requests [2604](https://github.com/instructlab/instructlab/issues/2604) [1823](https://github.com/instructlab/instructlab/issues/1823)  for this arg, so update it. And sorry for the original one [1446](https://github.com/instructlab/instructlab/pull/1446) that missed the previous git env. please check this one, thanks.

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/llama_cpp.py
+++ b/src/instructlab/model/backends/llama_cpp.py
@@ -128,7 +128,7 @@ class Server(BackendServer):
         try:
             self.port = free_tcp_ipv4_port(self.host)
             # start new server
-            self.api_base = str(get_api_base(f"{self.host}:{self.port}"))
+            self.api_base = str(get_api_base(self.host, self.port))
             logger.debug(f"Starting a temporary server at {self.api_base}...")
             self.process = self.create_server_process(self.port)
             self.process.start()
--- a/tests/test_lab_serve.py
+++ b/tests/test_lab_serve.py
@@ -335,3 +335,53 @@ def test_warn_for_unsupported_backend_param(param, expected_call_count):
             mock_warning.assert_called_with(
                 f"Option '--{param.replace('_','-')}' not supported by the backend."
             )
+
+
+def test_serve_host(cli_runner: CliRunner):
+    args = common.vllm_setup_test(
+        cli_runner,
+        [
+            "--config=DEFAULT",
+            "model",
+            "serve",
+            "--model-path=foo",
+            "--host",
+            "2.4.6.8",
+        ],
+    )
+    assert "2.4.6.8" in args
+    assert "8000" in args
+
+
+def test_serve_port(cli_runner: CliRunner):
+    args = common.vllm_setup_test(
+        cli_runner,
+        [
+            "--config=DEFAULT",
+            "model",
+            "serve",
+            "--model-path=foo",
+            "--port",
+            "1234",
+        ],
+    )
+    assert "127.0.0.1" in args
+    assert "1234" in args
+
+
+def test_serve_host_port(cli_runner: CliRunner):
+    args = common.vllm_setup_test(
+        cli_runner,
+        [
+            "--config=DEFAULT",
+            "model",
+            "serve",
+            "--model-path=foo",
+            "--host",
+            "192.168.1.1",
+            "--port",
+            "1234",
+        ],
+    )
+    assert "192.168.1.1" in args
+    assert "1234" in args
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -219,7 +219,7 @@ test_bind_port(){
     }
 
     # configure a different port
-    exec sed -i.bak 's/:8000/:9999/g' %s
+    exec sed -i.bak 's/8000/9999/g' %s
 
     # check that ilab model serve is working on the new port
     # catch ERROR strings in the output
--- a/src/instructlab/cli/model/serve.py
+++ b/src/instructlab/cli/model/serve.py
@@ -83,6 +83,20 @@ def warn_for_unsupported_backend_param(ctx):
     cls=clickext.ConfigOption,
     config_sections="vllm",
 )
+@click.option(
+    "-h",
+    "--host",
+    type=str,
+    cls=clickext.ConfigOption,
+    config_sections="server",
+)
+@click.option(
+    "-p",
+    "--port",
+    type=int,
+    cls=clickext.ConfigOption,
+    config_sections="server",
+)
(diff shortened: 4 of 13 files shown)


tests/test_lab_serve.py
```diff
@@ -335,3 +335,19 @@ def test_warn_for_unsupported_backend_param(param, expected_call_count):
...
+        [
+            "--config=DEFAULT",
+            "model",
+            "serve",
+            "--model-path=foo",
+            "--host-port",
+            "192.168.1.1:8000",
+        ],
+    )
+    assert "192.168.1.1" in args
```
osilkin98, line 352: I think the config tests technically may cover this, but we should add an explicit test in this same case which validates that when host-port isn't provided then it uses the default value.

tests/test_lab_serve.py
```diff
@@ -335,3 +335,19 @@ def test_warn_for_unsupported_backend_param(param, expected_call_count):
...
+def test_host_port(cli_runner: CliRunner):
+    args = common.vllm_setup_test(
+        cli_runner,
+        [
+            "--config=DEFAULT",
+            "model",
+            "serve",
+            "--model-path=foo",
+            "--host-port",
+            "192.168.1.1:8000",
```
osilkin98, line 349: This logic looks correct, though I would advise you to use a more obscure address & port here just so we know it's truly reading this. Because `8000` is otherwise fairly common and could be getting populated from the defaults. 

```suggestion
            "2.4.6.8:1738",
```

osilkin98 (commented):
Thanks for the quick PR @reidliu41 ! I just have a few comments, once those are addressed I can approve.

tests/test_lab_serve.py
reidliu41: @osilkin98 thanks, updated

RobotSail (approved):
This looks great, thanks for the PR!

anastasds:
How about the simpler `--bind`?

dcantrell:
> How about the simpler `--bind`?

Agreed, or `--listen`.  When I read `--host-port` I assumed the argument was a port number, not a host:port specification.

Is the port number required or is there a default?  If there's a default, then `--bind` or `--listen` makes more sense.

dcantrell:
> @dcantrell there is a default: https://github.com/instructlab/instructlab/blob/main/src/instructlab/configuration.py#L224, it's `"127.0.0.1:8000"`

OK, that's good.  So thinking about it even more, I would say `--listen` or `--listen-on` is the more friendly way to name the option.  `--bind` assumes the user has knowledge of the BSD sockets API.  And while many of us do, it's kind of jargon specific to the sockets API.

reidliu41:
I am ok with `--listen`, if more people agreed, I would like to change. hi team, please feel free to share.

nathan-weinberg:
`--listen` is fine with me :+1:

danmcp:
It would be good to follow precedent on the arg name.  Are there good examples of using --listen for this use case out in the wild?

My first reaction to --listen is it sounds like a verb rather than an argument accepting host:port.  --listen-on doesn't have the verb issue.  But again, would be good to follow precendent.

alimaredia:
If we go with `listen` do we change the name of the underlying configuration variable `host_port` too? With the way the PR is written right now, the the default value for `host_port` in `serve()` is the value of `ctx.obj.config.serve.host_port`.

After a quick glance at the code changes in `src/clickext.py` would need to be made for `--listen` to take the value of `_serve.host_port` by default.

danmcp:
If we change the config we would want to make it backward compatible and still support the old param.

My vote would be:
- Accept the config name of host_port as precedent (the name isn't ideal but it's workable)
- Be consistent with the name in the arg and use --host-port

Because:
- The confusion created over a rename and the maintenance of the backwards compatible support doesn't seem worth it
unless we can point to a well formed precedent on a different naming choice


From some quick searching, the most common methodology seems to be to separate --host and --port into two different params.  Sometimes the port param is called host_port which is what makes our current option a little confusing (at least for me).  The help text, docs, and examples mitigate that enough for me to not feel like it's a big issue.

alimaredia:
Having `host` and `port` split apart is a convention I've seen used before too. What would the backwards compatibility look like if we remove `host_port` from the config and split it apart in the config and had the flags discussed here?

reidliu41:
Sorry, based on the suggestion（good examples of using） again, do some research
```
gunicorn --bind 0.0.0.0:8000 myapp:app
uvicorn app:app --host 0.0.0.0 --port 8000
flask run --host=0.0.0.0 --port=8000
```
Seems more common for `--host` and `--port`, actually, my first time is used `--host` and `--port`, changed to `host-port` due to easy to get the config from the suggestion.

Also, from the code, looks like separate , together for the operation, e.g. get the config `host_port`, and use function to separate it. If used separate options, maybe easier.

reidliu41:
lol, if really separate host_port, seems more to change..

reidliu41:
pick one?
ilab model serve --host-port 127.0.0.1:8000

reidliu41:
ilab model serve --host 127.0.0.1 --port 8000

reidliu41:
ilab model serve --bind 127.0.0.1:8000

reidliu41:
ilab model serve --listen 127.0.0.1:8000

reidliu41:
ilab model serve --listen-on 127.0.0.1:8000

reidliu41:
hi, I updated with `--host` and `--port`

src/instructlab/model/backends/llama_cpp.py
```diff
@@ -128,7 +128,7 @@ def run_detached(
         try:
             self.port = free_tcp_ipv4_port(self.host)
             # start new server
-            self.api_base = str(get_api_base(f"{self.host}:{self.port}"))
+            self.api_base = str(get_api_base(self.host, self.port))
```
RobotSail, line 131: Nice, this is a really awesome change.

RobotSail (approved):
LGTM

cdoern (approved):
this looks pretty good to me! I read over the conversation and I agree with `--host` and `--port` over something like `--listen` or `--listen-on`.

reidliu41:
@alimaredia could you please take a look?

PR #2660, opened 2024-11-15
feat: detect gguf model arch

jaideepr97:
This PR adds the ability to detect the architecture of a GGUF model file - which would allow GGUF models to get the appropriate system prompt based on their architecture

**Checklist:**

### DIFF
--- a/src/instructlab/utils.py
+++ b/src/instructlab/utils.py
@@ -921,21 +921,33 @@ def get_model_arch(model_path: pathlib.Path) -> str:
     Extract a given model's architecture from its config if available and return it
 
     args
-        model_path (str): Path to the model, used to read the model config if available
+        model_path (Path): Path to the model, used to read the model config if available
     returns
         model_arch (str): The architecture of the model as expressed in the config file
     """
 
-    # TODO: implement for GGUF models
-
     model_arch = "default"
-    try:
-        model_cfg = get_config_file_from_model(model_path, "config.json")
-        model_arch = model_cfg["model_type"]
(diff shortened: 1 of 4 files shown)


src/instructlab/utils.py
```diff
@@ -926,16 +926,28 @@ def get_model_arch(model_path: pathlib.Path) -> str:
...
     model_arch = "default"
-    try:
-        model_cfg = get_config_file_from_model(model_path, "config.json")
-        model_arch = model_cfg["model_type"]
-    except Exception as e:
-        logger.debug(
-            f"Unable to get model architecture from config for model: {model_path}: {e}. Falling back to default value"
-        )
+
+    if is_model_gguf(Path(model_path)):
```
danmcp, line 931: The typing here is unclear.  model_path is stated as Path in the method signature.  It's stated as str in the arg comments.  And here it's being recreated as a path?

This logic seems like a good use case for a unit test?

jaideepr97: ah - will fix that, I was running some manual experiments in a terminal 

as for the unit test - I tried looking into how to create a dummy GGUF file that is correct enough to serve the purpose of this test, but it seemed complicated to get the internal structure right 
just setting the GGUF magic number and the architecture in the metadata didn't seem enough for the `GGUFReader` to successfully read the dummy model in

cdoern (approved):

cdoern:
this looks like it will pass now, not seeing anything wrong in the code.

alinaryan (approved):

PR #2693, opened 2024-11-22
ci: add lora training with FSDP to large E2E job

### LINKED ISSUE #2614: Add coverage for LoRA training to E2E CI
We recently had an incident where there was an accidental temporary regression with LoRA training due to differences in DeepSpeed and FSDP support - we want to add this type of training to the E2E coverage to prevent this

nathan-weinberg:
Resolves #2614

**Checklist:**

### DIFF
--- a/scripts/test-data/profile-l40s-x4.yaml
+++ b/scripts/test-data/profile-l40s-x4.yaml
@@ -137,7 +137,7 @@ train:
   deepspeed_cpu_offload_optimizer: true
   effective_batch_size: 32
   lora_quantize_dtype: null
-  lora_rank: 0
+  lora_rank: 4
   max_batch_len: 10000
   max_seq_len: 4096
   model_path: ~/.cache/instructlab/models/instructlab/granite-7b-lab
@@ -154,5 +154,5 @@ train:
   distributed_backend: fsdp
   pipeline: accelerated
   device: cuda
+  use_dolomite: true
 version: 1.0.0
-
--- a/scripts/e2e-ci.sh
+++ b/scripts/e2e-ci.sh
@@ -263,10 +263,11 @@ test_phased_train() {
     )
 
     # << TODO: REMOVE THIS FLAG FROM THE CI SCRIPT AND HAVE IT ADDED AS AN OPTION THAT CAN BE REFERENCED IN A PROFILE >>
-    # For x-large jobs only, we run into physical memory issues due to `accelerate_full_state_at_epoch` == true,
+    # For large jobs, we need this to be disabled to be able to test with LoRA training
+    # For x-large jobs, we run into physical memory issues due to `accelerate_full_state_at_epoch` == true,
     # so we disable full-state checkpoints to ensure we have enough disk space.
-    if [ "$XLARGE" -eq 1 ]; then
-        echo "Detected x-large e2e job... Will disable full-state checkpoints to conserve physical memory."
+    if [ "$LARGE" -eq 1 ] || [ "$XLARGE" -eq 1 ]; then
+        echo "Detected large or x-large e2e job... Will disable full-state checkpoints to conserve physical memory/allow for LoRA training."
         train_args+=("--disable-accelerate-full-state-at-epoch")
     fi
 
@@ -300,10 +301,11 @@ test_skills_only_train() {
     )
 
     # << TODO: REMOVE THIS FLAG FROM THE CI SCRIPT AND HAVE IT ADDED AS AN OPTION THAT CAN BE REFERENCED IN A PROFILE >>
-    # For x-large jobs only, we run into physical memory issues due to `accelerate_full_state_at_epoch` == true,
+    # For large jobs, we need this to be disabled to be able to test with LoRA training
+    # For x-large jobs, we run into physical memory issues due to `accelerate_full_state_at_epoch` == true,
     # so we disable full-state checkpoints to ensure we have enough disk space.
-    if [ "$XLARGE" -eq 1 ]; then
-        echo "Detected x-large e2e job... Will disable full-state checkpoints to conserve physical memory."
+    if [ "$LARGE" -eq 1 ] || [ "$XLARGE" -eq 1 ]; then
+        echo "Detected large or x-large e2e job... Will disable full-state checkpoints to conserve physical memory/allow for LoRA training."
         train_args+=("--disable-accelerate-full-state-at-epoch")
     fi
 
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -1553,10 +1553,6 @@ def map_train_to_library(ctx, params):
     if params["lora_quantize_dtype"] is not None:
         lora_options = True
         lora_args.quantize_data_type = params["lora_quantize_dtype"]
-    if lora_enabled and params["is_padding_free"]:
-        ctx.fail(
-            "Cannot combine LoRA with a padding free model. Set lora_rank to 0 or disable is_padding_free"
-        )
     if lora_options and not lora_enabled:
         click.secho("LoRA is disabled (rank=0), ignoring all additional LoRA args")
 


RobotSail:
@nathan-weinberg In theory it should work, I don't think you need to change it.

scripts/test-data/profile-l40s-x4.yaml
```diff
@@ -115,7 +115,7 @@ train:
   deepspeed_cpu_offload_optimizer: true
   effective_batch_size: 32
   lora_quantize_dtype: null
-  lora_rank: 0
+  lora_rank: 4
```
RobotSail, line 118: Why are we getting rid of the non-LoRA test case? It seems like we should probably have both.

nathan-weinberg:
> @nathan-weinberg In theory it should work, I don't think you need to change it.

As you can see in the run linked above in the PR comment, it did not work: 
```
Usage: ilab model train [OPTIONS]
Try 'ilab model train --help' for help.

Error: Cannot combine LoRA with a padding free model. Set lora_rank to 0 or disable is_padding_free
```

RobotSail:
Ah, that's a relic of the past. We should be able to remove that as that check refers to the old Dolomite architecture. @Maxusmusti has tested the recent padding-free architecture which has worked. @nathan-weinberg You can just remove that check / change it to check if Dolomite is being used.

nathan-weinberg:
> Ah, that's a relic of the past. We should be able to remove that as that check refers to the old Dolomite architecture. @Maxusmusti has tested the recent padding-free architecture which has worked. @nathan-weinberg You can just remove that check / change it to check if Dolomite is being used.

Sounds good wiill do - the change is in the CLI not the Training library yes?

RobotSail:
> Sounds good wiill do - the change is in the CLI not the Training library yes?

Yes, you can find it here: https://github.com/instructlab/instructlab/blob/92099a58869a416548240a4f3abaced5a0958937/src/instructlab/configuration.py#L1368

scripts/test-data/profile-l40s-x4.yaml
RobotSail: Not ideal but we can probably just create a second profile in that case. Or you can just pass `--lora-rank` as a CLI argument to the training script

nathan-weinberg: Ah so a few things on that
1. we're intentionally not using flags, everything is going to be in the profile - we don't want to go the route of the original `basic-workflow-tests.sh` script
2. i _could_ create a second profile, but to use it consistently would mean having **two** L40S jobs, which is probably not worth it just to have a LoRA/no LoRA test running in terms of compute/time spend

to take a step back - the point of the E2Es is to test a "typical user workflow" - so for someone using an L40Sx4 system, what would make more sense, for them to use LoRA or not? 

the "typical" case should live in the E2E, and the "atypical" case can go in one of these places:
1. the unit/functional tests with mocking
2. the `e2e-custom.sh` script so we can run on-demand jobs testing this functionality
3. some testing going on in the training library

@danmcp i'd be curious about your thoughts on this as well!

nathan-weinberg: @RobotSail hmmm you mean run through both profiles in the same job? that would be an over 4 hour job, and again not really fulfilling the purpose I laid out above

i don't really get your point about the flag args, of course there is a use for them... but again, it's not really what the idea of the E2E/system profiles is, and even if we did use them it doesn't fix the actual problem you're saying we have here

nathan-weinberg: still need an answer to my question about what the typical user workflow is, btw

danmcp: > @danmcp i'd be curious about your thoughts on this as well!

I haven't formed an opinion on the CLI args question.  Regarding where/if this test belongs, everything is a trade off of time vs value.  So it's helpful to decide if the value is in executing code paths (which might be handled with a unit test) vs we need to verify the output of a full run.
- If this is a case where we need to verify the model output, then it's a question of how long does it take?  If it's short and provides significant value, I would like to understand the CLI args decision better because it would be nice to run the common e2e scenarios in one large job.
- If most of the test value comes from proving the code paths work, add a unit test and supporting this test in e2e-custom.sh seems reasonable.

If we end up with multiple code paths that need to do full e2e runs, we're probably going to end up with multiple large jobs.  That's also fine.  I think we just need to be selective about which tests provide the most coverage/value.  It's not going to be that hard to add the tests, but they consume a lot of resources and will need continual maintenance.

RobotSail: I'm against having multiple jobs because currently that would mean we run SDG multiple times which is a very expensive process and don't actually cache any of the jobs. If instead we split it up such that SGD runs once, outputs the resultant datasets somewhere and then subsequent jobs are launched then they could run in parallel without issue.

nathan-weinberg: @danmcp @RobotSail Yeah a little more context about the CLI command line args - basically, you can think of it like this:
1. Every CLI arg should have a config value
2. Every config value should have a profile value
3. Ergo, the specific CI profiles we have here should have all the configuration done there

Motivation: Previously, it was near-impossible to tell what the `basic-workflow-tests.sh` script was _actually running_ because there was so much modification being done in-place. The new approach that `e2e-ci.sh` uses has also the configuration in one place, which makes both maintenance and debugging much easier. 

We kept `e2e-custom.sh` for testing GPU-necessary corner cases that don't represent the "typical" workflow but are still thinks we want to be able to test automatically via the E2E custom job.

RobotSail: But I guess that just comes down to cost savings, since both runs would take equally long. If the cost of our e2e jobs isn't a factor we're concerned with then I'm okay with it.

danmcp: A few things to think about re multiple large jobs:
- If we end up with one 4 hour large job, repeating sdg and ending up with two 2.5 hour jobs might still be preferred to avoid debugging failures in really long jobs.
- If we did have multiple jobs, it might be a good way to test variations of sdg params across the different jobs.  An example custom pipeline would be a good one?
- Cost is certainly a factor, but 20-30 mins of extra runtime for another 1-10 large jobs would be reasonable from a cost perspective if it provides enough for the team.

RobotSail:
@nathan-weinberg From this error:

```
Arguments: (ValueError('`accelerate_full_state_at_epoch` is not currently supported when training LoRA models.'),)
```

Basically multi-phase training tries to save the entire state at every epoch so we can resume training state, but this isn't supported by LoRA. To fix this, you can set `accelerate_full_state_at_epoch: false` in the `additional_args` config list.

scripts/test-data/profile-l40s-x4.yaml
RobotSail: The whole point of LoRA is to allow devices with smaller configurations to test so it makes the most sense to train them on the lower-end jobs. That said, we can and probably should test on all if that's not an issue.

nathan-weinberg: Okay based on that feedback I'm gonna make an executive decision here we can push this forward - LoRA will be added to this job and not the XL job - we can iterate on this strategy later

nathan-weinberg:
> @nathan-weinberg From this error:
> 
> ```
> Arguments: (ValueError('`accelerate_full_state_at_epoch` is not currently supported when training LoRA models.'),)
> ```
> 
> Basically multi-phase training tries to save the entire state at every epoch so we can resume training state, but this isn't supported by LoRA. To fix this, you can set `accelerate_full_state_at_epoch: false` in the `additional_args` config list.

I seem to have hit the same error - maybe I wrote the config wrong?

RobotSail:
@nathan-weinberg That's weird, it should be getting disabled when you pass that setting in. I would debug it and see whether or not that value is being passed in both from the CLI and within the training library.

nathan-weinberg:
I have been trying to get this to work for several months with no avail - need to focus on other things. Will leave the issue open in case someone else wants to take it on.

PR #2694, opened 2024-11-22
Disable Auto-Detection for HPU and HIP

cdoern:
AMD and Intel GPUs are still triggering auto-detection because `get_chip_name` seems to be properly returning the AMD GPU names.

disable these checks for hip and hpu

**Checklist:**

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -314,6 +314,19 @@ def walk_and_print_system_profiles(
     return cfg
 
 
+def is_hpu_available() -> bool:
+    # Third Party
+    import torch
+
+    # Check for HPU availability without errors
+    try:
+        hpu_available = hasattr(torch, "hpu") and torch.device("hpu") is not None
+        return hpu_available
+    # pylint: disable=broad-exception-caught
+    except Exception:
+        return False
+
+
 def get_gpu_or_cpu() -> tuple[str, bool, bool, int, int]:
     """
     get_gpu_or_cpu figures out what kind of hardware the user has and returns the name in the form of 'nvidia l4 x4' as well as if this is a CPU and if the user is on Linux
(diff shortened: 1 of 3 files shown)


cdoern:
@nathan-weinberg it wasn't scoped so the system wasn't built for it. I am unaware of how the AMD cuda layer works/returns what type of GPUs we have.

cdoern:
Good point, no we do not. I think I put it in the release notes but I can add a note here to the readme or something

nathan-weinberg:
> Good point, no we do not. I think I put it in the release notes but I can add a note here to the readme or something

Awesome, thank you!

src/instructlab/config/init.py
```diff
@@ -314,6 +314,19 @@ def walk_and_print_system_profiles(
...
+def is_hpu_available() -> bool:
+    # Third Party
+    import torch
+
+    # Check for HPU availability without errors
+    try:
+        hpu_available = hasattr(torch, "hpu") and torch.device("hpu") is not None
+        return hpu_available
+    # pylint: disable=broad-exception-caught
+    except Exception:
```
nathan-weinberg, line 326: Do we not get an expected Exception here?

src/instructlab/config/init.py
```diff
@@ -314,6 +314,19 @@ def walk_and_print_system_profiles(
     return cfg
 
 
+def is_hpu_available() -> bool:
+    # Third Party
```
nathan-weinberg, line 318: Could you add a docstring here?

alinaryan (approved):

src/instructlab/config/init.py
nathan-weinberg: Also, should this be a `util`?

nathan-weinberg: Should have a unit tests for this new func as well

nathan-weinberg:
@Mergifyio backport release-v0.21

src/instructlab/config/init.py
cdoern: I do not believe so, looking in other parts other codebase, (vllm specifically), it seems we do this.

src/instructlab/config/init.py
cdoern: I do not think this should be a utility since it is directly used for auto-detection code, which also all lives in here, moving all of that is a different discussion

nathan-weinberg (approved):

PR #2704, opened 2024-11-26
Add skills only training strategy

danmcp:
This new strategy only runs the skills phase and corresponding eval of skills with mt_bench (skipping knowledge).  It is useful to bypass phase 1 of multiphase training if you don't have any knowledge to train the model with.

**Checklist:**
- [*] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [*] Documentation has been updated, if necessary.
- [*] Unit tests have been added, if necessary.
- [*] E2E Workflow tests have been added, if necessary.

### DIFF
--- a/src/instructlab/model/accelerated_train.py
+++ b/src/instructlab/model/accelerated_train.py
@@ -33,6 +33,11 @@ class SupportedTrainingStrategies(enum.Enum):
     """Available advanced training strategies"""
 
     LAB_MULTIPHASE: str = "lab-multiphase"
+    LAB_SKILLS_ONLY: str = "lab-skills-only"
+
+    @classmethod
+    def has_strategy(cls, value):
+        return value in cls._value2member_map_
 
 
 def accelerated_train(
@@ -60,7 +65,7 @@ def accelerated_train(
     # run_training is a dynamic attribute, pylint is not clever enough
     # to detect it.
     # Third Party
-    if strategy == SupportedTrainingStrategies.LAB_MULTIPHASE.value:
+    if SupportedTrainingStrategies.has_strategy(strategy):
         if (
             distributed_backend
             and distributed_backend not in DistributedBackend._value2member_map_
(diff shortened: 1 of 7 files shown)


danmcp:
The above e2e test was run with `--strategy=lab-skills-only`

danmcp:
Docs test failing due to: https://github.com/instructlab/instructlab/pull/2705

Will rebase once merged.

src/instructlab/model/accelerated_train.py
```diff
@@ -77,17 +82,29 @@ def accelerated_train(
...
+        else:  # strategy == SupportedTrainingStrategies.LAB_SKILLS_ONLY.value
+            if not phased_phase2_data:
+                # pylint: disable=broad-exception-raised
+                raise Exception(
+                    f"{SupportedTrainingStrategies.LAB_SKILLS_ONLY} training minimally requires: `--phased-phase2-data`"
+                )
+
+            if not phased_phase2_data.exists():
+                raise FileNotFoundError(
+                    "Data for both phase1 and phase2 must be specified for phased training."
```
alimaredia, line 106: Why mention phase 1 here if you're not checking for it and it's not needed for skills only training?

danmcp: Thanks, I missed updating this message.  Fixed now.

danmcp:
> @danmcp this seems fine to me. How were you envisioning this looking in CI? Some kind of functional test for it that's run nightly or is that going to be in a follow up PR?

I've added the unit test for now and I also tested by hacking into e2e-ci.sh for a quick verification.  I debated adding this to e2e-ci.sh permanently.  It would be super simple to add and tack another 10 mins onto the run.  It's a question of whether it's worth the overhead for each large run to functional test skills only.  I think yes but looking for feedback.

danmcp:
From the large e2e run:

```
2024-11-27T04:39:44.7699343Z Training finished! Best final checkpoint: /tmp/tmp.E9D1pcsmw0/.local/share/instructlab/skills-only/phase2/checkpoints/hf_format/samples_1221 with score: 8.05
2024-11-27T04:39:44.7700114Z Journal: /tmp/tmp.E9D1pcsmw0/.local/share/instructlab/skills-only/journalfile.yaml
2024-11-27T04:39:46.2420921Z + grep -o '/[^ ]*'
2024-11-27T04:39:46.2421436Z + grep 'Training finished! Best final checkpoint: ' /tmp/tmp.E9D1pcsmw0/skills_only_training.log
```

alimaredia (approved):

src/instructlab/model/accelerated_train.py
```diff
@@ -136,6 +154,9 @@ def accelerated_train(
                 bg="yellow",
                 fg="black",
             )
+            # Start at phase 2 for skills only training
+            if strategy == SupportedTrainingStrategies.LAB_SKILLS_ONLY.value:
```
JamesKunstle, line 158: I guess this is the crux of 'skipping phase 1' which worked out nicely

danmcp: Correct.  Or at least that's what I worked out as a reasonable approach.  And agreed it proves out the journal approach was well thought out if this is indeed the appropriate change.

src/instructlab/model/accelerated_train.py
```diff
@@ -77,17 +82,29 @@ def accelerated_train(
...
-            raise Exception(
-                "End-to-end training minimally requires: `--phased-phase1-data`, and `--phased-phase2-data`. One or more wasn't correctly specified."
-            )
-
-        # if they both exist, must be Path objects
-        if not (phased_phase1_data.exists() and phased_phase2_data.exists()):
-            raise FileNotFoundError(
-                "Data for both phase1 and phase2 must be specified for phased training."
-            )
+        if strategy == SupportedTrainingStrategies.LAB_MULTIPHASE.value:
```
JamesKunstle, line 85: One minor question is whether we'd like to take this opportunity to refactor our use of enums since we're starting to use them more. Stringifying enums in python this way is one way to accomplish this, another would be:

`if SupportedTrainingStrategies(strategy) == SupportedTrainingStrategies.LAB_MULTIPHASE`

Which is obviously less concise but seems like that the enum docs prefer.... I don't have a strong opinion either way, what do you think?

JamesKunstle (approved):
PR is very good as it is, especially inclusion of tests. simply changing the training journal order makes sense as the solution to skip phase1 training. Quick question commented about using python enums.

src/instructlab/model/accelerated_train.py
danmcp: I was just trying to be consistent with existing code.  I prefer not to use .value as well.  Although I think would prefer to change in a follow-on PR if that ok?

danmcp:
Removing the hold and will open the follow-on PR to address the desired enum handling.

PR #2708, opened 2024-11-26
feat: Implement Process Management, `ilab data generate -dt`, `ilab process list`, `ilab process attach`

### LINKED ISSUE #2689: implement the concept of processes in InstructLab
InstructLab should have the ability to understand the commands it runs as "processes".

This should be done as simply as possible, using python multiprocessing or subprocess to kick off key functionality and manage it via a multiprocessing Manager object.

This will also include an understanding of log files, PID tracking, and UUIDs in InstructLab.

### LINKED ISSUE #2690: implement detaching from processes in InstructLab
A process, as understood from InstructLab, should be detachable, meaning a user can run `ilab data generate -dt` and a UUID will simply print for the user to track over the course of the runtime

cdoern:
this PR introduces process management, with the first detachable command being `ilab data generate`

a user can run `ilab data generate -dt` and a few things happen:
1. a log file is created in ~/.local/share/instructlab/logs/generate/generate-UUID.log where UUID is generated on each SDG run
2. SDG is kicked off in its own process. This includes server creation, and the calling of the SDG library. Both of these tasks are crucial to running SDG. One lives in the InstructLab repo while the other lives in the SDG repo, causing this to be the point where an independent process must be split off.
3. The UUID and log file are printed so the user can re-attach or directly check on the log file.
4. ilab then tracks the children processes created by the new python process. These are added to the process registry along with the PID of the parent, the log file, and the start time. This registry is written to `~/.local/share/instructlab/internal/process_registry.json`. The registry is read from disk into memory on each invocation of the `process` package which manages this data
5. The user can now list, attach to, and kill processes. 

this necessitates the creation of the `ilab process` command group. This command group will begin with two commands: 
`ilab process list`
`ilab process attach --uuid UUID_HERE --latest` where the user can pass either a UUID or just attach to the latest process kicked off. 
here is what the workflow looks like:

Screenshot 2024-12-13 at 8 58 13 PM

Some concrete design goals here:

this should _not_ live in the CLI sub-directory or execute `ilab` commands directly. This is because eventual consumers of this rearchitecture will want to call `generate_data` without needing to pass through unecessary click parsing and initialization.

Keeping lower-level process functionality and workflow code separate from the upper level CLI code is a purposeful decision here. A user should be able to run `ilab data generate -dt` but also eventually be able to curl the ilab.socket without needing to pass CLI level arguments or go through that level of parsing. Our process splitting should happen on this "lower level" right before we execute actual functionality of the LAB methodology

resolves #2689
resolves #2690

**Checklist:**

### DIFF
--- a/scripts/e2e-ci.sh
+++ b/scripts/e2e-ci.sh
@@ -210,8 +210,14 @@ test_taxonomy() {
 }
 
 test_generate() {
-    task Generate synthetic data
-    ilab data generate
+    task Generate synthetic data in detached mode
+    ilab data generate -dt
+    task Data generation started
+    task List all processes
+    ilab process list
+    task Listing processes Complete
+    task Attach to the most recent process
+    ilab process attach --latest
     task Synthetic data generation Complete
 }
 
--- a/src/instructlab/cli/data/generate.py
+++ b/src/instructlab/cli/data/generate.py
@@ -293,7 +293,8 @@ def generate(
             legacy_pretraining_format,
             process_mode=process_mode,
         )
-        click.echo("ᕦ(òᴗóˇ)ᕤ Data generate completed successfully! ᕦ(òᴗóˇ)ᕤ")
+        if not detached:
+            click.echo("ᕦ(òᴗóˇ)ᕤ Data generate completed successfully! ᕦ(òᴗóˇ)ᕤ")
     except Exception as exc:
         click.secho(f"failed to generate data with exception: {exc}", fg="red")
         raise click.exceptions.Exit(1)
--- a/src/instructlab/cli/data/generate.py
+++ b/src/instructlab/cli/data/generate.py
@@ -15,6 +15,7 @@ from instructlab import clickext
 from instructlab.client_utils import HttpClientParams
 from instructlab.configuration import DEFAULTS
 from instructlab.data.generate_data import gen_data  # type: ignore
+from instructlab.defaults import ILAB_PROCESS_MODES
 from instructlab.utils import (
     contains_argument,
     get_model_arch,
@@ -160,6 +161,9 @@ logger = logging.getLogger(__name__)
     type=click.IntRange(min=512),
     cls=clickext.ConfigOption,
 )
+@click.option(
+    "-dt", "--detached", is_flag=True, help="Run ilab data generate in the background"
+)
 @click.pass_context
 @clickext.display_params
 def generate(
@@ -187,6 +191,7 @@ def generate(
     batch_size,
     gpus,
     max_num_tokens,
+    detached,
 ):
     """Generates synthetic data to enhance your example data"""
 
@@ -257,6 +262,10 @@ def generate(
             student_model_path, student_model_arch
         )
 
+    process_mode = ILAB_PROCESS_MODES.ATTACHED
+    if detached:
+        process_mode = ILAB_PROCESS_MODES.DETACHED
+
     try:
         gen_data(
             serve_cfg,
@@ -282,6 +291,7 @@ def generate(
             max_num_tokens,
             system_prompt,
             legacy_pretraining_format,
+            process_mode=process_mode,
         )
         click.echo("ᕦ(òᴗóˇ)ᕤ Data generate completed successfully! ᕦ(òᴗóˇ)ᕤ")
     except Exception as exc:
--- a/src/instructlab/cli/data/generate.py
+++ b/src/instructlab/cli/data/generate.py
@@ -292,6 +292,7 @@ def generate(
             system_prompt,
             legacy_pretraining_format,
             process_mode=process_mode,
+            log_level=ctx.obj.config.general.log_level,
         )
         if not detached:
             click.echo("ᕦ(òᴗóˇ)ᕤ Data generate completed successfully! ᕦ(òᴗóˇ)ᕤ")
--- a/src/instructlab/cli/process/attach.py
+++ b/src/instructlab/cli/process/attach.py
@@ -1,0 +1,24 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Third Party
+import click
+
+# First Party
+from instructlab import clickext
+from instructlab.process.process import attach_process, get_latest_process
+
+
+@click.command(name="attach")
+@clickext.display_params
+@click.option("--uuid", type=click.STRING, help="UUID of the process to attach to")
+@click.option("--latest", is_flag=True, help="Attach to the latest process")
+def attach(uuid, latest):
+    if latest:
+        uuid = get_latest_process()
+        if uuid is None:
+            click.secho(
+                "No processes found in registry",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
+    attach_process(local_uuid=uuid)
--- a/src/instructlab/cli/process/list.py
+++ b/src/instructlab/cli/process/list.py
@@ -1,0 +1,23 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Third Party
+import click
+
+# First Party
+from instructlab import clickext
+from instructlab.process.process import list_processes
+from instructlab.utils import print_table
+
+
+@click.command(name="list")
+@clickext.display_params
+def list():
+    process_list = list_processes()
+    if process_list is not None and len(process_list) > 0:
+        print_table(["Type", "PID", "UUID", "Log File", "Runtime"], process_list)
+    else:
+        click.secho(
+            "No processes found in registry",
+            fg="red",
+        )
+        raise click.exceptions.Exit(0)
--- a/src/instructlab/data/generate_data.py
+++ b/src/instructlab/data/generate_data.py
@@ -1,15 +1,13 @@
 # Standard
 import logging
 
-# Third Party
-import openai
-
 # First Party
-from instructlab.client_utils import HttpClientParams, http_client
+from instructlab.client_utils import HttpClientParams
 
 logger = logging.getLogger(__name__)
 
 
+# handler function to be used as an intermediary so we can manipulate args
 def gen_data(
     serve_cfg,
     model_path,
@@ -34,16 +32,111 @@ def gen_data(
     max_num_tokens,
     system_prompt,
     use_legacy_pretraining_format,
+    process_mode,
 ):
     """Generates synthetic data to enhance your example data"""
-    # Third Party
-    from instructlab.sdg.generate_data import generate_data
+    # First Party
+    from instructlab.defaults import ILAB_PROCESS_TYPES
+    from instructlab.process.process import add_process
 
-    # pylint: disable=ungrouped-imports
-    from instructlab.sdg.utils import GenerateException
+    print(http_client_params)
 
+    add_process(
+        process_mode=process_mode,
+        process_type=ILAB_PROCESS_TYPES.DATA_GENERATION,
+        target=create_server_and_generate,
+        extra_imports=[
+            ("instructlab.configuration", "_serve", "_serve_vllm", "_serve_llama_cpp")
+        ],
+        serve_cfg=serve_cfg,
+        model_name=model_path,
+        num_cpus=num_cpus,
+        num_instructions_to_generate=sdg_scale_factor,
+        taxonomy=taxonomy_path,
+        taxonomy_base=taxonomy_base,
+        output_dir=output_dir,
+        console_output=not quiet,
+        endpoint_url=endpoint_url,
+        api_key=api_key,
+        yaml_rules=yaml_rules,
+        chunk_word_count=chunk_word_count,
+        server_ctx_size=server_ctx_size,
+        tls_client_cert=http_client_params.get("tls_client_cert")
+        if http_client_params.get("tls_client_cert") != ""
+        else None,
+        tls_client_key=http_client_params.get("tls_client_key")
+        if http_client_params.get("tls_client_key") != ""
+        else None,
+        tls_client_passwd=http_client_params.get("tls_client_passwd")
+        if http_client_params.get("tls_client_passwd") != ""
+        else None,
+        tls_insecure=http_client_params.get("tls_insecure")
+        if http_client_params.get("tls_insecure") != ""
+        else None,
+        model_family=model_family,
+        pipeline=pipeline,
+        enable_serving_output=enable_serving_output,
+        batch_size=batch_size,
+        gpus=gpus,
+        checkpoint_dir=checkpoint_dir,
+        max_num_tokens=max_num_tokens,
+        system_prompt=system_prompt,
+        use_legacy_pretraining_format=use_legacy_pretraining_format,
+        # http_client_params=http_client_params,
+    )
+
+
+# actual generate_data function to be kicked off in its own process/eventually be REST API endpoint
+def create_server_and_generate(
+    serve_cfg,
+    model_name,
+    num_cpus,
+    num_instructions_to_generate,
+    taxonomy,
+    taxonomy_base,
+    output_dir,
+    console_output,
+    endpoint_url,
+    api_key,
+    yaml_rules,
+    chunk_word_count,
+    server_ctx_size,
+    tls_client_cert,
+    tls_client_key,
+    tls_client_passwd,
+    tls_insecure,
+    model_family,
+    pipeline,
+    enable_serving_output,
+    batch_size,
+    gpus,
+    checkpoint_dir,
+    max_num_tokens,
+    system_prompt,
+    use_legacy_pretraining_format,
+):
     backend_instance = None
 
+    # Third Party
+    import openai
+
+    # First Party
+    from instructlab.client_utils import HttpClientParams, http_client
+
+    print(tls_client_cert, tls_client_key, tls_client_passwd, tls_insecure)
+    http_client_params = HttpClientParams(
+        {
+            "tls_client_cert": tls_client_cert,
+            "tls_client_key": tls_client_key,
+            "tls_client_passwd": tls_client_passwd,
+            "tls_insecure": tls_insecure,
+        }
+    )
+
+    print(http_client_params)
+
+    # logger.info("testing")
+    # logger.debug("testing2")
     if endpoint_url:
         api_base = endpoint_url
     else:
@@ -51,7 +144,7 @@ def gen_data(
         from instructlab.model.backends import backends
         from instructlab.model.backends.llama_cpp import Server as llama_cpp_server
 
-        backend_instance = backends.select_backend(cfg=serve_cfg, model_path=model_path)
+        backend_instance = backends.select_backend(cfg=serve_cfg, model_path=model_name)
         if (
             backend_instance.get_backend_type() is not backends.VLLM
             and gpus is not None
@@ -82,21 +175,26 @@ def gen_data(
     client = openai.OpenAI(
         base_url=api_base, api_key=api_key, http_client=http_client(http_client_params)
     )
+    # Third Party
+    from instructlab.sdg.generate_data import generate_data
+
+    # pylint: disable=ungrouped-imports
+    from instructlab.sdg.utils import GenerateException
 
     try:
         logger.info(
-            f"Generating synthetic data using '{pipeline}' pipeline, '{model_path}' model, '{taxonomy_path}' taxonomy, against {api_base} server"
+            f"Generating synthetic data using '{pipeline}' pipeline, '{model_name}' model, '{taxonomy}' taxonomy, against {api_base} server"
         )
         generate_data(
             client=client,
             model_family=model_family,
-            model_name=model_path,
+            model_name=model_name,
             num_cpus=num_cpus,
-            num_instructions_to_generate=sdg_scale_factor,
-            taxonomy=taxonomy_path,
+            num_instructions_to_generate=num_instructions_to_generate,
+            taxonomy=taxonomy,
             taxonomy_base=taxonomy_base,
             output_dir=output_dir,
-            console_output=not quiet,
+            console_output=console_output,
             yaml_rules=yaml_rules,
             chunk_word_count=chunk_word_count,
             server_ctx_size=server_ctx_size,
--- a/src/instructlab/data/generate_data.py
+++ b/src/instructlab/data/generate_data.py
@@ -33,6 +33,7 @@ def gen_data(
     system_prompt,
     use_legacy_pretraining_format,
     process_mode,
+    log_level,
 ):
     """Generates synthetic data to enhance your example data"""
     # First Party
@@ -52,6 +53,7 @@ def gen_data(
                 "_serve_server",
             )
         ],
+        log_level=log_level,
         serve_cfg=serve_cfg,
         model_name=model_path,
         num_cpus=num_cpus,
@@ -91,7 +93,9 @@ def gen_data(
 
 
 # actual generate_data function to be kicked off in its own process/eventually be REST API endpoint
+# pylint: disable=redefined-outer-name
 def create_server_and_generate(
+    log_level,
     serve_cfg,
     model_name,
     num_cpus,
@@ -118,14 +122,26 @@ def create_server_and_generate(
     max_num_tokens,
     system_prompt,
     use_legacy_pretraining_format,
+    log_file,
 ):
     backend_instance = None
+    # we need to use the instructlab logger so that the libraries inherit the config we set up.
+    logger = logging.getLogger("instructlab")
+
+    # First Party
+    from instructlab.log import add_file_handler_to_logger
+
+    # when in a new process, logger configuration does not seem to be promised. Add the file handler pointing to our new log file
+    # and also, reenforce the level set on the instructlab logger.
+    add_file_handler_to_logger(log_file=log_file, logger=logger)
+    logger.setLevel(log_level)
 
     # Third Party
     import openai
 
     # First Party
     from instructlab.client_utils import http_client
+
     http_client_params = HttpClientParams(
         {
             "tls_client_cert": tls_client_cert,
@@ -141,6 +157,7 @@ def create_server_and_generate(
         # First Party
         from instructlab.model.backends import backends
         from instructlab.model.backends.llama_cpp import Server as llama_cpp_server
+
         backend_instance = backends.select_backend(cfg=serve_cfg, model_path=model_name)
         if (
             backend_instance.get_backend_type() is not backends.VLLM
--- a/src/instructlab/data/generate_data.py
+++ b/src/instructlab/data/generate_data.py
@@ -126,7 +126,6 @@ def create_server_and_generate(
 
     # First Party
     from instructlab.client_utils import http_client
-
     http_client_params = HttpClientParams(
         {
             "tls_client_cert": tls_client_cert,
@@ -142,7 +141,6 @@ def create_server_and_generate(
         # First Party
         from instructlab.model.backends import backends
         from instructlab.model.backends.llama_cpp import Server as llama_cpp_server
-
         backend_instance = backends.select_backend(cfg=serve_cfg, model_path=model_name)
         if (
             backend_instance.get_backend_type() is not backends.VLLM
--- a/src/instructlab/data/generate_data.py
+++ b/src/instructlab/data/generate_data.py
@@ -127,7 +127,6 @@ def create_server_and_generate(
     # First Party
     from instructlab.client_utils import http_client
 
-    print(tls_client_cert, tls_client_key, tls_client_passwd, tls_insecure)
     http_client_params = HttpClientParams(
         {
             "tls_client_cert": tls_client_cert,
--- a/src/instructlab/data/generate_data.py
+++ b/src/instructlab/data/generate_data.py
@@ -39,14 +39,18 @@ def gen_data(
     from instructlab.defaults import ILAB_PROCESS_TYPES
     from instructlab.process.process import add_process
 
-    print(http_client_params)
-
     add_process(
         process_mode=process_mode,
         process_type=ILAB_PROCESS_TYPES.DATA_GENERATION,
         target=create_server_and_generate,
         extra_imports=[
-            ("instructlab.configuration", "_serve", "_serve_vllm", "_serve_llama_cpp")
+            (
+                "instructlab.configuration",
+                "_serve",
+                "_serve_vllm",
+                "_serve_llama_cpp",
+                "_serve_server",
+            )
         ],
         serve_cfg=serve_cfg,
         model_name=model_path,
@@ -121,7 +125,7 @@ def create_server_and_generate(
     import openai
 
     # First Party
-    from instructlab.client_utils import HttpClientParams, http_client
+    from instructlab.client_utils import http_client
 
     print(tls_client_cert, tls_client_key, tls_client_passwd, tls_insecure)
     http_client_params = HttpClientParams(
@@ -133,10 +137,6 @@ def create_server_and_generate(
         }
     )
 
-    print(http_client_params)
-
-    # logger.info("testing")
-    # logger.debug("testing2")
     if endpoint_url:
         api_base = endpoint_url
     else:
--- a/src/instructlab/defaults.py
+++ b/src/instructlab/defaults.py
@@ -30,6 +30,17 @@ LOG_FORMAT = "%(levelname)s %(asctime)s %(name)s:%(lineno)d: %(message)s"
 RECOMMENDED_SCOPEO_VERSION = "1.9.0"
 
 
+class ILAB_PROCESS_MODES:
+    DETACHED: str = "detached"
+    ATTACHED: str = "attached"
+
+
+class ILAB_PROCESS_TYPES:
+    DATA_GENERATION: str = "Generation"
+
+    TRAINING: str = "Training"
+
+
 class STORAGE_DIR_NAMES:
     ILAB = "instructlab"
     DATASETS = "datasets"
@@ -42,6 +53,7 @@ class STORAGE_DIR_NAMES:
     )
     CHATLOGS = "chatlogs"
     PHASED = "phased"
+    LOGS = "logs"
 
 
 class _InstructlabDefaults:
@@ -112,6 +124,10 @@ class _InstructlabDefaults:
         self._cache_home = path.join(xdg_cache_home(), ILAB_PACKAGE_NAME)
         self._config_dir = path.join(xdg_config_home(), ILAB_PACKAGE_NAME)
         self._data_dir = path.join(xdg_data_home(), ILAB_PACKAGE_NAME)
+
+    @property
+    def LOGS_DIR(self) -> str:
+        return path.join(self._data_dir, STORAGE_DIR_NAMES.LOGS)
 
     @property
     def CHECKPOINTS_DIR(self) -> str:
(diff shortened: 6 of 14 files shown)


scripts/e2e-ci.sh
```diff
@@ -210,8 +210,14 @@ test_taxonomy() {
 }
 
 test_generate() {
-    task Generate synthetic data
-    ilab data generate
+    task Generate synthetic data in detached mode
+    ilab data generate -dt
+    task Data Generation started
```
nathan-weinberg, line 215: Capitalization should be consistent with the rest of the script
```suggestion
    task Data generation started
```

scripts/e2e-ci.sh
```diff
@@ -210,8 +210,14 @@ test_taxonomy() {
...
+    task list all processes
```
nathan-weinberg, line 216:
```suggestion
    task List all processes
```

scripts/e2e-ci.sh
```diff
@@ -210,8 +210,14 @@ test_taxonomy() {
...
+    ilab process list
+    task listing processes complete
```
nathan-weinberg, line 218:
```suggestion
    task Listing processes Complete
```

src/instructlab/cli/data/generate.py
```diff
@@ -160,6 +161,9 @@
     type=click.IntRange(min=512),
     cls=clickext.ConfigOption,
 )
+@click.option(
+    "-dt", "--detached", is_flag=True, help="Run ilab data generate in the background"
+)
```
nathan-weinberg, line 166: Is the idea we will independently add this to each command? Should we have a goal of making this a global flag or are we only scoping process management for certain commands?

src/instructlab/cli/process/attach.py
```diff
@@ -0,0 +1,24 @@
+# Third Party
+import click
+
+# First Party
+from instructlab import clickext
+
+
+@click.command(name="attach")
+@clickext.display_params
+@click.option("--uuid", type=click.STRING, help="PID of the process to attach to")
```
nathan-weinberg, line 10: Is this a true Linux PID or an InstructLab-generated UUID?

src/instructlab/cli/process/attach.py
```diff
@@ -0,0 +1,24 @@
...
+@click.option("--latest", is_flag=True, help="attach to the latest process")
```
nathan-weinberg, line 11:
```suggestion
@click.option("--latest", is_flag=True, help="Attach to the latest process")
```

src/instructlab/cli/process/attach.py
```diff
@@ -0,0 +1,24 @@
...
+def attach(uuid, latest):
+    # First Party
+    from instructlab.process.process import attach_process, get_latest_process
```
nathan-weinberg, line 14: Why not make this a global import?

src/instructlab/cli/process/list.py
```diff
@@ -0,0 +1,17 @@
...
+# First Party
+from instructlab import clickext
+from instructlab.utils import print_table
+
+
+@click.command(name="list")
+@clickext.display_params
+def list():
+    # First Party
+    from instructlab.process.process import list_processes
```
nathan-weinberg, line 13: Same question around the importing here

src/instructlab/data/generate_data.py
```diff
@@ -34,15 +32,113 @@ def gen_data(
...
+):
     backend_instance = None
+    # we need to use the instructlab logger so that the libraries inherit the config we set up.
+    logger = logging.getLogger("instructlab")
+
+    # First Party
+    from instructlab.log import add_file_handler_to_logger
+
+    add_file_handler_to_logger(log_file=log_file, logger=logger)
+    logger.setLevel(logging.INFO)
```
nathan-weinberg, line 126: What if a user has set a different log level previously? Why are we overwriting the level to `INFO`?

src/instructlab/data/generate_data.py
```diff
@@ -51,7 +147,7 @@ def gen_data(
         from instructlab.model.backends import backends
         from instructlab.model.backends.llama_cpp import Server as llama_cpp_server
 
-        backend_instance = backends.select_backend(cfg=serve_cfg, model_path=model_path)
+        backend_instance = backends.select_backend(cfg=serve_cfg, model_path=model_name)
```
nathan-weinberg, line 150: Arg rename?

src/instructlab/data/generate_data.py
```diff
@@ -82,21 +178,26 @@ def gen_data(
...
+    # Third Party
+    from instructlab.sdg.generate_data import generate_data
+
+    # pylint: disable=ungrouped-imports
+    from instructlab.sdg.utils import GenerateException
 
     try:
         logger.info(
-            f"Generating synthetic data using '{pipeline}' pipeline, '{model_path}' model, '{taxonomy_path}' taxonomy, against {api_base} server"
+            f"Generating synthetic data using '{pipeline}' pipeline, '{model_name}' model, '{taxonomy}' taxonomy, against {api_base} server"
```
nathan-weinberg, line 189: Arg rename?

src/instructlab/defaults.py
```diff
@@ -26,6 +26,17 @@
...
+class ILAB_PROCESS_MODES:
+    DETACHED: str = "detached"
+    ATTACHED: str = "attached"
+
+
+class ILAB_PROCESS_TYPES:
+    DATA_GENERATION: str = "Generation"
+
+    TRAINING: str = "Training"
+
```
nathan-weinberg, line 38: Not sure why there's an extra newline here
```suggestion
class ILAB_PROCESS_TYPES:
    DATA_GENERATION: str = "Generation"
    TRAINING: str = "Training"

```

src/instructlab/defaults.py
```diff
@@ -109,6 +121,10 @@ def _reset(self):
         self._config_dir = path.join(xdg_config_home(), ILAB_PACKAGE_NAME)
         self._data_dir = path.join(xdg_data_home(), ILAB_PACKAGE_NAME)
 
+    @property
+    def LOGS_DIR(self) -> str:
+        return path.join(self._data_dir, STORAGE_DIR_NAMES.LOGS)
```
nathan-weinberg, line 126: I like this setup!

src/instructlab/defaults.py
```diff
@@ -225,5 +241,9 @@ def TRAIN_L40_X4_PROFILE(self) -> str:
     def TRAIN_L4_X8_PROFILE(self) -> str:
         return path.join(self.TRAIN_PROFILE_DIR, "L4_x8.yaml")
 
+    @property
+    def REGISTRY_FILE(self) -> str:
+        return path.join(self.INTERNAL_DIR, "process_registry.json")
```
nathan-weinberg, line 246: I like this setup as well!

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+    """Load the process registry from a file, if it exists."""
+    # we do not want a persistent registry in memory. This causes issues when in scenarios where you switch registry files (ex, in a unit test, or with multuple users)
+    # but the registry with incorrect processes still exists in memory.
+    if os.path.exists(DEFAULTS.REGISTRY_FILE):
+        with open(DEFAULTS.REGISTRY_FILE, "r") as f:
+            data = json.load(f)
+            for key, value in data.items():
+                process_registry[key] = value
+    else:
+        logger.info("No existing process registry found. Starting fresh.")
```
nathan-weinberg, line 33: We may want this to be a `debug` statement - not sure if normal users care about a process registry

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def close(self):
+        """
+        Close the log file.
+        """
+        if self.log:
+            self.log.close()
+
+
+def add_process(process_mode: str, process_type: str, target, extra_imports, **kwargs):
```
nathan-weinberg, line 80: Can we add type hinting for `target` and `extra_imports`?

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+    """
+    Start a detached process using subprocess.Popen, logging its output.
+
+    Args:
+        task_name (str): Name of the task.
+        process_registry (dict): Shared process registry.
```
nathan-weinberg, line 86: These don't match the function args

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+
+    Returns:
+        int: PID of the started process.
```
nathan-weinberg, line 89: If we know we are returning an int you can add `-> int` to the function definition

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+                    time.sleep(0.1)  # Wait briefly before trying again
+    except KeyboardInterrupt:
+        logger.info("\nDetaching from and killing process.")
+    finally:
+        stop_process(local_uuid=local_uuid)
+
+
+def all_processes_running(pids: list[int]):
+    """
+    Returns if a process and all of its children are still running
```
nathan-weinberg, line 259: Returns a bool? If I'm right can we add `-> bool` to the function definition?

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+    return all(psutil.pid_exists(pid) for pid in pids)
+
+
+def stop_process(local_uuid):
+    """
+    Stop a running process.
+
+    Args:
+        uuid (int): uuid of the process to stop.
+    """
```
nathan-weinberg, line 272: Doesn't match function args, also I think `uuid` should be a string, not an int

tests/test_lab_process_attach.py
```diff
@@ -0,0 +1,56 @@
...
+
+# Third Party
+from click.testing import CliRunner
+
+# First Party
+from instructlab import lab
+from instructlab.defaults import DEFAULTS
+
+
+def test_process_attach(
```
nathan-weinberg, line 14: There should be a test for attaching when no processes exist

tests/test_lab_process_list.py
```diff
@@ -0,0 +1,73 @@
...
+            desired_columns = columns[:-1]
+
+            # Clean and join the remaining columns
+            clean_line = " | ".join(desired_columns)
+            entries.append(clean_line)
+
+    return "\n".join(entries)
+
+
+def test_process_list(cli_runner: CliRunner):
```
nathan-weinberg, line 34: There should be a test for listing when no processes exist

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+                local_uuid,
+                entry.get("log_file"),
+                runtime_str,
+            )
+        )
+
+    return list_of_processes
+
+
+def attach_process(local_uuid):
```
nathan-weinberg, line 214:
```suggestion
def attach_process(local_uuid: str):
```

nathan-weinberg (changes requested):
I'd like to see more unit testing as well as possibly some functional tests added, given the size of this change

src/instructlab/cli/process/attach.py
cdoern: instructlab generated UUID. using the pid as a tracker doesn't make much sense given that most of these processes spawn children and those ones are what we are really interested in

src/instructlab/cli/data/generate.py
cdoern: each command, not everything will have a detached mode. Primarily any long running command but not things like `ilab model list`

src/instructlab/data/generate_data.py
cdoern: good point! I can plumb some way here to only set it to the level in the config.

src/instructlab/data/generate_data.py
cdoern: I renamed some arg so they could be used as kwargs fed directly into SDG. `generate_data` required model_name, so I just switched it for us as well. 

this isn't user-facing.

src/instructlab/data/generate_data.py
cdoern: same as above

src/instructlab/defaults.py
cdoern: thanks!

src/instructlab/process/process.py
cdoern: true, will switch

src/instructlab/process/process.py
cdoern: hmmmm, would target be something like `: function`? Not sure if there is a type for a generic function

src/instructlab/process/process.py
cdoern: sorry, there might be a few of these. Lots of change here. Will go through and update.

src/instructlab/process/process.py
cdoern: right, thanks

src/instructlab/cli/process/attach.py
nathan-weinberg: So the help text is wrong then?

src/instructlab/cli/process/attach.py
nathan-weinberg: I'm fine with either, I just want the motivation to make sense. When is a local import preferred over a global?

src/instructlab/cli/process/list.py
nathan-weinberg: Cool let's just talk on that one and do whatever resolution we come to for both

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+    list_of_processes = []
+    for local_uuid, entry in process_registry.items():
+        now = datetime.now()
+
+        # Calculate runtime
+        runtime = now - datetime.fromisoformat(entry.get("start_time"))
+        # Convert timedelta to a human-readable string (HH:MM:SS)
+        total_seconds = int(runtime.total_seconds())
+        hours, remainder = divmod(total_seconds, 3600)
+        minutes, seconds = divmod(remainder, 60)
```
jaideepr97, line 198: This could probably be its own helper function 
```suggestion
        runtime_str = str(runtime).split(".")[0]
```

src/instructlab/data/generate_data.py
nathan-weinberg: Okay - it's a little hard as a reviewer to know what arg renaming is purposeful and why it's being done versus knowing what's a mistake.

Refactors ideally should be done prior to new features - for example, I did this here prior to OCI downloads so people could focus their attention on what was being done and why: https://github.com/instructlab/instructlab/pull/2808

Something to consider for the future.

src/instructlab/data/generate_data.py
nathan-weinberg: Ack, will resolve this but see my point in that thread.

src/instructlab/cli/process/attach.py
cdoern: generally speaking, local imports are typically used to guard specific things from being imported unless explicitly needed aka: torch. With this, nothing in process.py seems to be something we'd want to guard in this way, I can move the imports to be global

nathan-weinberg: SGTM!

src/instructlab/data/generate_data.py
cdoern: right, it kind of had to be done here due to implications of passing things as **kwargs. I can add a comment in the code as to why this exists the way it does now if that would help.

this is internal code though, nothing user facing, an implementation detail

src/instructlab/process/process.py
cdoern: nice! thanks

src/instructlab/data/generate_data.py
nathan-weinberg: No worries! Will resolve this thread - thanks for the explanation!

src/instructlab/process/process.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def close(self):
+        """
+        Close the log file.
+        """
+        if self.log:
+            self.log.close()
+
+
+def add_process(
```
alinaryan: This is a huge function, I think this should be its own class too, or broken down into smaller bits.

```
class ProcessManager:
    def __init__(self, process_registry, log_dir):
        self.process_registry = process_registry
        self.log_dir = log_dir

    def add_process(self, process_mode, process_type, target, extra_imports, **kwargs):
        # ... (implementation details)

    def _prepare_command(self, target, extra_imports, kwargs):
        # ... (command setup logic)

    def _setup_logging(self, process_type):
        # ... (logging setup logic)

    def _launch_process(self, cmd, log_file):
        # ... (process launching logic)

    def _capture_child_pids(self, parent_pid):
        # ... (child PID capturing logic)

    def _update_registry(self, pid, process_type, start_time, log_file, children_pids):
        # ... (registry update logic)
```

src/instructlab/process/process.py
cdoern: actually this was leftover, returns nothing. removed.

src/instructlab/defaults.py
cdoern: tox adds this in, keeping it

tests/common.py
```diff
@@ -61,6 +61,7 @@ def vllm_setup_test(runner, args, mock_popen, *_mock_args):
 
     return mock_popen.call_args_list[0][1]["args"]
 
+
```
alinaryan: Do we need these newlines?

alinaryan (commented):
Thanks for the hard work on this! Please see my comments.

src/instructlab/data/generate_data.py
```diff
@@ -121,7 +119,7 @@ def create_server_and_generate(
     import openai
 
     # First Party
-    from instructlab.client_utils import HttpClientParams, http_client
```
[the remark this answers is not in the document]
cdoern: updated in a future commit in sequence.

src/instructlab/data/generate_data.py
```diff
@@ -34,24 +32,119 @@ def gen_data(
...
+    system_prompt,
+    use_legacy_pretraining_format,
+):
     backend_instance = None
 
+    # Third Party
+    import openai
+
+    # First Party
+    from instructlab.client_utils import HttpClientParams, http_client
```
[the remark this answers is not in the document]
cdoern: Updated in future commit in sequence.

tests/common.py
cdoern: updated in a future commit. I can go back if you'd like and remove them

alinaryan: I see that now! No worries, I'll resolve this convo

src/instructlab/defaults.py
nathan-weinberg: How is `tox` doing this?

cdoern: `tox -e fix` adds this in if I remove it

nathan-weinberg: who am I to argue with the great and powerful `tox`

btw you can run `make fix` as shorthand (if you aren't already)

resolving this

src/instructlab/cli/process/attach.py
```diff
@@ -0,0 +1,24 @@
...
+
+# First Party
+from instructlab import clickext
+from instructlab.process.process import attach_process, get_latest_process
+
+
+@click.command(name="attach")
+@clickext.display_params
+@click.option("--uuid", type=click.STRING, help="UUID of the process to attach to")
+@click.option("--latest", is_flag=True, help="Attach to the latest process")
```
reidliu41, line 14: cool for `--latest`, but seems a new flag if use for different types.

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+
+    local_uuid = uuid.uuid4()
+    log_file = None
+
+    log_dir = os.path.join(DEFAULTS.LOGS_DIR, process_type.lower())
+
+    if not os.path.exists(log_dir):
+        os.makedirs(log_dir)
+
+    log_file = os.path.join(log_dir, f"{process_type.lower()}-{local_uuid}.log")
```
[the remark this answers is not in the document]
reidliu41: maybe add a check after a new uuid generate?

src/instructlab/cli/process/list.py
```diff
@@ -0,0 +1,23 @@
...
+from instructlab.process.process import list_processes
+from instructlab.utils import print_table
+
+
+@click.command(name="list")
+@clickext.display_params
+def list():
+    process_list = list_processes()
+    if process_list is not None:
+        print_table(["Type", "PID", "UUID", "Log File", "Runtime"], process_list)
```
reidliu41, line 17: Seems it would better to add the `start time` , `state`, `command` in the `list`, specially what commands it run and what parameters it used,  can simply to check, that might help to troubleshoot.
like the common process list `ps`:
```
]# ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0 242836 15992 ?        Ss   Dec03   1:35 /usr/lib/systemd/systemd --system
```

src/instructlab/cli/process/list.py
```diff
@@ -0,0 +1,23 @@
...
+    else:
+        click.secho(
+            "No processes found in registry",
+            fg="red",
+        )
```
reidliu41, line 22: maybe can simply print a empty table e.g. `model list`, `data list`

src/instructlab/cli/process/list.py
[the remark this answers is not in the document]
cdoern: moved these back to global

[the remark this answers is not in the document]
cdoern: I was lazy importing bc I was doing some logic inside of the proc mgmt pkg that ran some code each time the pkg was invoked

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+    if process_mode == ILAB_PROCESS_MODES.DETACHED:
+        # Prepare the subprocess command string
+        cmd = f"import {target.__module__}; {target.__module__}.{target.__name__}(**{kwargs})"
+
+        # Handle extra imports (if any)
+        if extra_imports:
+            import_statements = "\n".join(
+                [f"from {imp[0]} import {', '.join(imp[1:])}" for imp in extra_imports]
+            )
+            cmd = f"{import_statements}\n{cmd}"
```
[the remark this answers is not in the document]
cdoern: so @jaideepr97 the issue is the following:

```
import instructlab.data.generate_data; instructlab.data.generate_data.create_server_and_generate(**{'log_level': 'DEBUG', 'serve_cfg': _serve(vllm=_serve_vllm(llm_family='', max_startup_attempts=120, gpus=None, vllm_args=[]), llama_cpp=_serve_llama_cpp(gpu_layers=-1, max_ctx_size=4096, llm_family=''), model_path='/Users/charliedoern/.cache/instructlab/models/mistral-7b-instruct-v0.2.Q4_K_M.gguf', server=_serve_server(host='127.0.0.1', port=8000), chat_template='tokenizer', backend='llama-cpp'), 'model_name': '/Users/charliedoern/.cache/instructlab/models/mistral-7b-instruct-v0.2.Q4_K_M.gguf', 'num_cpus': 10, 'num_instructions_to_generate': 30, 'taxonomy': '/Users/charliedoern/.local/share/instructlab/taxonomy', 'taxonomy_base': 'origin/main', 'output_dir': '/Users/charliedoern/.local/share/instructlab/datasets', 'console_output': True, 'endpoint_url': None, 'api_key': 'no_api_key', 'yaml_rules': None, 'chunk_word_count': 1000, 'server_ctx_size': 4096, 'tls_client_cert': None, 'tls_client_key': None, 'tls_client_passwd': None, 'tls_insecure': False, 'model_family': None, 'pipeline': 'full', 'enable_serving_output': False, 'batch_size': 8, 'gpus': None, 'checkpoint_dir': '/Users/charliedoern/.local/share/instructlab/datasets/checkpoints', 'max_num_tokens': 4096, 'system_prompt': 'I am, Red Hat® Instruct Model based on Granite 7B, an AI language model developed by Red Hat and IBM Research, based on the Granite-7b-base language model. My primary function is to be a chat assistant.', 'use_legacy_pretraining_format': False, 'log_file': '/Users/charliedoern/.local/share/instructlab/logs/generation/generation-743cd73a-06c8-4ffc-b0c5-b9a8c7c011b7.log', 'extra_imports': [('instructlab.configuration._serve', 'instructlab.configuration._serve_vllm', 'instructlab.configuration._serve_llama_cpp', 'instructlab.configuration._serve_server')]})
```
this is an example command. Without directly having the import statements in the command, this will fail. 

If we use importlib in code before or after this, it doesn't really help us

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+def load_registry():
+    process_registry = {}
+    """Load the process registry from a file, if it exists."""
+    # we do not want a persistent registry in memory. This causes issues when in scenarios where you switch registry files (ex, in a unit test, or with multuple users)
+    # but the registry with incorrect processes still exists in memory.
+    if os.path.exists(DEFAULTS.REGISTRY_FILE):
+        with open(DEFAULTS.REGISTRY_FILE, "r") as f:
+            data = json.load(f)
+            for key, value in data.items():
+                process_registry[key] = value
```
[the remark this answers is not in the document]
cdoern: added a lock

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+            message (str): The message to write.
+        """
+        self.terminal.write(message)
+        self.log.write(message)
+
+    def flush(self):
+        """
+        Ensure all data is written to the terminal and the log file.
+        """
+        self.terminal.flush()
```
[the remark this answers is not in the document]
cdoern: this is the case with all concurrently running commands since they all use stdout, I think this may be beyond my ability to add thread safety, no?

src/instructlab/process/process.py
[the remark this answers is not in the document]
cdoern: these uuids are very rarely overlapping. 

I can switch to uuid1, which has zero chance since it uses date and time

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+    if target is None:
+        return None, None
+
+    local_uuid = uuid.uuid4()
+    log_file = None
+
+    log_dir = os.path.join(DEFAULTS.LOGS_DIR, process_type.lower())
+
+    if not os.path.exists(log_dir):
+        os.makedirs(log_dir)
```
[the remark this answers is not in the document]
cdoern: this is creating `log_dir/generation` we shouldn't create empty dirs for generation or training until we run generation or training. The parent log dir is handled by the ensure dirs exist check

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+                process_registry[key] = value
+    else:
+        logger.info("No existing process registry found. Starting fresh.")
+    return process_registry
+
+
+def save_registry(process_registry):
+    """Save the current process registry to a file."""
+    with open(DEFAULTS.REGISTRY_FILE, "w") as f:
+        json.dump(dict(process_registry), f)
```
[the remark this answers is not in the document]
cdoern: added a lock

src/instructlab/process/process.py
```diff
@@ -0,0 +1,296 @@
...
+            else:
+                logger.warning("No child processes detected. Tracking parent process.")
+            pid = p.pid
+            # Check if subprocess was successfully started
+            if p.poll() is not None:
+                logger.warning(f"Process {p.pid} failed to start.")
+                return None  # Process didn't start
+
+    # Add the process info to the shared registry
+    process_registry[str(local_uuid)] = {
```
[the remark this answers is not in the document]
cdoern: added proc registry class

src/instructlab/process/process.py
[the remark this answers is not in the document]
cdoern: switched to uuid1

src/instructlab/cli/process/list.py
cdoern: Is there a benefit to this? Might just clog up folks terminals

src/instructlab/process/process.py
```diff
@@ -0,0 +1,311 @@
...
+        now = datetime.now()
+
+        # Calculate runtime
+        runtime = now - datetime.fromisoformat(entry.get("start_time"))
+        # Convert timedelta to a human-readable string (HH:MM:SS)
+        total_seconds = int(runtime.total_seconds())
+        hours, remainder = divmod(total_seconds, 3600)
+        minutes, seconds = divmod(remainder, 60)
+
+        runtime_str = f"{hours:02}:{minutes:02}:{seconds:02}"
```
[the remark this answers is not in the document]
cdoern: an attached process removes itself from the list properly.

for a detached process unless we want to do IPC (out of scope for this PR), the best I can do is pause the clock the next time a user checks on the process if It is done.

src/instructlab/process/process.py
cdoern: I am confused what you're suggesting here? You want me to remove the hr, min, seconds calc but leave the below setting of runtime_str?

cdoern: Also this chunk of code is only used in one spot, so I don't see the benefit of making it a function since its 3 lines

src/instructlab/process/process.py
cdoern: I totally get the motivation here for organization, but I don't see the benefit. Are we going to be instantiating `ProcessManager` objects? If not, then grouping functions under a class is purely organizational. the way these process functions are used does not make sense to instantiate an `ProcessManager` instance IMO.

I can split up the functions a bit if it would make it clearer

cdoern: but the split up`add_process` functions will basically call themselves in sequence

cdoern: I created:

format_command
start process

update registry is a part of the `ProcessRegistry` class I just introduced

src/instructlab/cli/process/list.py
```diff
@@ -0,0 +1,23 @@
...
+def list():
+    process_list = list_processes()
+    if process_list is not None and len(process_list) > 0:
+        print_table(["Type", "PID", "UUID", "Log File", "Runtime"], process_list)
+    else:
+        click.secho(
+            "No processes found in registry",
+            fg="red",
+        )
+        raise click.exceptions.Exit(1)
```
nathan-weinberg, line 23: nit: the existence of 0 detached processes isn't necessarily a failure

src/instructlab/cli/system/info.py
```diff
@@ -0,0 +1,30 @@
+# SPDX-License-Identifier: Apache-2.0
```
nathan-weinberg, line 1: Is this file meant to be changed?

src/instructlab/cli/system/system.py
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: Apache-2.0
```
nathan-weinberg, line 1: Is this file meant to be changed?

src/instructlab/process/process.py
```diff
@@ -0,0 +1,384 @@
...
+    def load_entry(self, key, value):
+        self.processes[key] = value
+
+
+def load_registry() -> ProcessRegistry:
+    process_registry = ProcessRegistry()
+    lock_path = DEFAULTS.PROCESS_REGISTRY_LOCK_FILE
+    lock = FileLock(lock_path, timeout=1)
+    """Load the process registry from a file, if it exists."""
+    # we do not want a persistent registry in memory. This causes issues when in scenarios where you switch registry files (ex, in a unit test, or with multuple users)
```
nathan-weinberg, line 51: nit
```suggestion
    # we do not want a persistent registry in memory. This causes issues when in scenarios where you switch registry files (ex, in a unit test, or with multiple users)
```

nathan-weinberg (commented):
There are two files getting changed here I'm not sure are supposed to be, otherwise apart from my nits this is looking good to me. I ran the large job against this for some functional confidence.

src/instructlab/cli/system/info.py
cdoern: I think I ran some sort of fix that added these in automatically

src/instructlab/cli/system/system.py
cdoern: I think I ran some sort of fix that added these in automatically

src/instructlab/cli/process/list.py
cdoern: would `Exit(0)` be good?

src/instructlab/process/process.py
alinaryan: I understand not wanting to create a `ProcessManager` class if we don't plan to use it as an object, but even without a class structure, organizing it into distinct functions (even sequentially) can help clarify each part's responsibility/make it easier to test and debug.

alinaryan: Thanks for splitting it up a bit!

alinaryan (commented):
Thanks for working on this! A couple comments

src/instructlab/cli/system/system.py
cdoern: fixed and removed changes

src/instructlab/cli/system/info.py
cdoern: fixed and removed

src/instructlab/cli/process/list.py
nathan-weinberg: It's a tricky situation - it _could_ indicate an issue, just not necessarily

Since there's no code failure in this logic path, I'd lean toward exiting with 0, but if this is being used for testing somehow feel free to leave it as 1 and we can revisit this later

cdoern: No its fine I can switch to 0 and just look for the proper log

alinaryan (approved):

src/instructlab/process/process.py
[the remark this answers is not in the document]
reidliu41: yeah...IPC, thanks, got it.

src/instructlab/cli/process/attach.py
reidliu41: ah.. it always get the latest no matter `generation` or other types. Maybe another feat, what if get the latest with `type`.

cdoern:
given that this has been through 3 rounds of review from core maintainers and has been open since Nov 26, 2024. I am going to start removing requested reviewers that were requested previously in the process but have yet to review.

src/instructlab/process/process.py
jaideepr97: fair enough

src/instructlab/process/process.py
```diff
@@ -0,0 +1,384 @@
...
+        return
+
+    logger.info(f"Attaching to process {local_uuid}. Press Ctrl+C to detach and kill.")
+    all_pids = [process_info["pid"]] + process_info["children_pids"]
+    if not all_processes_running(all_pids):
+        return
+    try:
+        with open(log_file, "a+") as log:
+            log.seek(0, os.SEEK_END)  # Move to the end of the log file
+            while all_processes_running(all_pids):
```
jaideepr97, line 357: should we continue to stay attached until all child processes are finished instead of exiting immediately after any of them completes?

jaideepr97 (approved):
approving with some non blocking comments/questions
Thanks for working on this @cdoern

PR #2713, opened 2024-11-27
Introduce Metadata to the config.yaml

cdoern:
Metadata is meant to store specific information about what type of system this configuration should apply to, namely GPU information. This is a better and more maintainable heuristic for system profiles as we scale the amount of systems we want to automatically apply configurations to

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -540,6 +540,28 @@ class _train(BaseModel):
     )
 
 
+class _metadata(BaseModel):
+    # model configuration
+    model_config = ConfigDict(extra="ignore")
+    cpu_info: str | None = Field(
+        default=None,
+        description="Manufacturer, Family, and SKU of the system CPU, ex: Apple M3 Max",
+    )
+    gpu_manufacturer: str | None = Field(
+        default=None, description="Manufacturer of the system GPU, ex: Nvidia"
+    )
+    gpu_family: str | None = Field(
+        default=None, description="Family of the system GPU, ex: H100"
+    )
+    gpu_count: int | None = Field(
+        default=None, description="Amount of GPUs on the system, ex: 8"
+    )
+    gpu_sku: list[str] | None = Field(
+        default=None,
+        description="Specific SKU related information about the given GPU, ex: PCIe, NVL",
+    )
+
+
 class Config(BaseModel):
     """Configuration for the InstructLab CLI.
     Config options are defined by the respective subclasses and are loaded into a single 'Config' object here
@@ -578,6 +600,11 @@ class Config(BaseModel):
         default=CONFIG_VERSION,
         description="Configuration file structure version.",
         frozen=True,  # don't allow this to be changed anywhere in the code
+    )
+    # metadata about the configuration, specifically GPU information
+    metadata: _metadata = Field(
+        default_factory=_metadata,
+        description="Metadata pertaining to the specifics of the system which the Configuration is meant to be applied to.",
     )
 
 
--- a/tests/test_lab_init.py
+++ b/tests/test_lab_init.py
@@ -63,7 +63,7 @@ class TestLabInit:
 
     @patch(
         "instructlab.config.init.get_gpu_or_cpu",
-        return_value=("nvidia a100 x2", False, True, 0, 0),
+        return_value=("nvidia a100", 0, 2),
     )
     def test_ilab_config_init_auto_detection_nvidia(
         self, convert_bytes_to_proper_mag_mock, cli_runner: CliRunner
@@ -78,7 +78,7 @@ class TestLabInit:
 
     @patch(
         "instructlab.config.init.get_gpu_or_cpu",
-        return_value=("apple m3 max", True, False, 0, 0),
+        return_value=("apple m3 max", 0, 0),
     )
     def test_ilab_config_init_auto_detection_mac(
         self, convert_bytes_to_proper_mag_mock, cli_runner: CliRunner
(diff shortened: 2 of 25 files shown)


tests/testdata/default_config.yaml
```diff
@@ -19,7 +19,8 @@ chat:
   # Filepath of a dialog session file.
   # Default: None
   session:
-  # Controls the randomness of the model's responses.
+  # Controls the randomness of the model's responses. Lower values make the output
+  # more deterministic, while higher values produce more random results.
```
mairin, line 23: this extra line isn't related to the PR though?

tests/test_lab_init.py
```diff
@@ -117,7 +132,22 @@ def test_ilab_config_init_auto_detection_gaudi(
...
         convert_bytes_to_proper_mag_mock.assert_called_once()
         assert (
-            "We have detected the INTEL GAUDI 3 profile as an exact match for your system."
+            "We have detected the INTEL GAUDI 3 X8 profile as an exact match for your system."
+            in result.stdout
+        )
+
+    @patch(
+        "instructlab.config.init.get_gpu_or_cpu",
+        return_value=("nvidia h100 pcie", 0, 8),
```
mairin, line 141: This might be a silly point - I'm not familiar with the format of these tests. But is the return value meant to be what a mock device name string coming out of pytorch would be? If so... would it make more sense to use real life device strings from pytorch from different cards and use those for the tests? e.g. instead of "nvidia h100 pcie" using "NVIDIA H100 PCIe" and "NVIDIA H100 80GB HBM3" etc. etc. so it matches exactly output we've gotten from hardware we've used?

tests/testdata/default_config.yaml
cdoern: `make regenerate-testdata` regenerates the entire default cfg, so someone must have modified the temp help text but not run this cmd so my PR is picking that up

tests/test_lab_init.py
cdoern: the `get_gpu_or_cpu` func returns a `.lower()` version of the detected string for ease of matching later on, so this return value is what we'd get after we parse it

mairin: OK fair enough I'd missed that. Do you want to test against the core set of strings? I don't see the full set from the issue here (but might be missing them?)

src/instructlab/configuration.py
```diff
@@ -540,6 +540,32 @@ class _train(BaseModel):
...
+    )
+    cpu_info: str | None = Field(
+        default=None,
+        description="Manufacturer, Family, and SKU of the system CPU, ex: Apple M3 Max",
+    )
+    gpu_manufacturer: str | None = Field(
+        default=None, description="Manufacturer of the system GPU, ex: Nvidia"
+    )
+    gpu_family: str | None = Field(
+        default=None, description="Family the of the system GPU, ex: H100"
```
nathan-weinberg, line 558:
```suggestion
        default=None, description="Family of the system GPU, ex: H100"
```

src/instructlab/configuration.py
```diff
@@ -579,6 +605,11 @@ class Config(BaseModel):
         description="Configuration file structure version.",
         frozen=True,  # don't allow this to be changed anywhere in the code
     )
+    # metadata about the configuration, specifically GPU information
+    metadata: _metadata = Field(
+        default_factory=_metadata,
+        description="Metadata pertaining to the specifics of the system which the Configuration is meant to be applied to.",
```
nathan-weinberg, line 611: Is it worth calling this something more specific, like `system_metadata` or `sys_metadata`? Do we intend to ever use this for broader, non-system-related metadata?

nathan-weinberg (commented):
@cdoern the E2E CI profiles don't take advantage of autodetection - maybe we should have a followup where we test that?

src/instructlab/configuration.py
cdoern: I think leaving it vague is a purposeful choice, in the future I could see this being pretty useful for other config descriptors outside of system hw.

tests/test_lab_init.py
cdoern: I can add a test for each type!

cdoern: done!

src/instructlab/configuration.py
```diff
@@ -540,6 +540,32 @@ class _train(BaseModel):
     )
 
 
+class _metadata(BaseModel):
+    # model configuration
+    model_config = ConfigDict(
+        extra="ignore",
+        protected_namespaces=(),
```
alinaryan, line 547: Do we need to use protected_namespaces()?

src/instructlab/configuration.py
```diff
@@ -540,6 +540,32 @@ class _train(BaseModel):
...
+        use_enum_values=True,  # populate models with the value property of enums, rather than the raw enum.
+    )
+    cpu_info: str | None = Field(
+        default=None,
+        description="Manufacturer, Family, and SKU of the system CPU, ex: Apple M3 Max",
+    )
+    gpu_manufacturer: str | None = Field(
+        default=None, description="Manufacturer of the system GPU, ex: Nvidia"
```
alinaryan, line 555: Consider using enum if we want to keep gpu_manufacturer to a fixed set of possible values 

```
from enum import Enum

class GPUManufacturer(str, Enum):
    NVIDIA = "Nvidia"
    AMD = "AMD"
    INTEL = "Intel"

class SystemMetadata(BaseModel):
    gpu_manufacturer: GPUManufacturer | None = Field(
        default=None, description="Manufacturer of the system GPU, ex: Nvidia"
    )

```

alinaryan (commented):
Thanks for working on this!

src/instructlab/configuration.py
cdoern: This is a standard for new classes we introduce

src/instructlab/configuration.py
cdoern: I'd like to avoid using enums because they introduce issues with click.

alinaryan (approved):

nathan-weinberg (approved):

cdoern:
force merging due to known issue in CI: https://github.com/instructlab/instructlab/issues/2753. Overriding because we need this in 0.21.2

cdoern:
@Mergifyio backport release-v0.21

PR #2715, opened 2024-11-27
Update README.md to specify supported Python versions

AlexonOliveiraRH:
Updating the documentation to explicitly include the information we only support Python 3.10 and 3.11 so far.

**Checklist:**

- [ ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [x] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Functional tests have been added, if necessary.
- [ ] E2E Workflow tests have been added, if necessary.

PR #2717, opened 2024-11-28
Make it more clear in the README doc that python 3.12+ is not yet supported

kdubois:
Also add <version> to the python commands for Linux install instructions.

**Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 - C++ compiler
 - Python 3.10 or Python 3.11
 
-   📋 Python 3.12 and Python 3.13 are currently not supported. Some InstructLab dependencies don't work on Python 3.12 and above, yet.
+   ⚠️ Python 3.12+ is currently not supported. Some InstructLab dependencies don't work on Python 3.12, yet. It is recommended to use a specific version of Python in the below commands, e.g. `python3.11` instead of simply `python3`.
 
 - Minimum 250GB disk space. Approximately 500GB disk space is recommended for the entire InstructLab end-to-end process.
 
@@ -159,7 +159,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 - Install on AMD ROCm with:
 
    ```shell
-   python3 -m venv --upgrade-deps venv
+   python<version> -m venv --upgrade-deps venv
    source venv/bin/activate
    pip cache remove llama_cpp_python
    CMAKE_ARGS="-DGGML_HIPBLAS=on \
@@ -179,7 +179,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 - For the best CUDA experience, installing vLLM is necessary to serve Safetensors format models.
 
    ```shell
-   python3 -m venv --upgrade-deps venv
+   python<version> -m venv --upgrade-deps venv
    source venv/bin/activate
    pip cache remove llama_cpp_python
    CMAKE_ARGS="-DGGML_CUDA=on -DGGML_NATIVE=off" pip install 'instructlab[cuda]'


reidliu41:
already summit recently [2715](https://github.com/instructlab/instructlab/pull/2715)

AlexonOliveiraRH:
@kdubois we had the same complaint in parallel :)

kdubois:
@AlexonOliveiraRH indeed :) sorry I should've checked the existing open PRs

AlexonOliveiraRH:
> @AlexonOliveiraRH indeed :) sorry I should've checked the existing open PRs

No problem, pal. ;)

README.md
```diff
@@ -93,7 +93,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 - C++ compiler
 - Python 3.10 or Python 3.11
 
-   📋 Python 3.12 is currently not supported. Some InstructLab dependencies don't work on Python 3.12, yet.
+   ⚠️ Python 3.12+ is currently not supported. Some InstructLab dependencies don't work on Python 3.12, yet. It is recommended to use a specific version of Java in the below commands, eg. `python3.11` instead of simply `python3`.
```
cdoern, line 96:
```suggestion
   ⚠️ Python 3.12+ is currently not supported. Some InstructLab dependencies don't work on Python 3.12, yet. It is recommended to use a specific version of Python in the below commands, eg. `python3.11` instead of simply `python3`.
```

cdoern (changes requested):

nathan-weinberg:
This seems to be doing the same thing as the already-existing https://github.com/instructlab/instructlab/pull/2715

cdoern:
In my opinion, this one seems to fit the message better. I don't think we should advertise people using 0.17 as a solution if they have 3.12

kdubois:
This solution is a little more comprehensive than #2715 IMO since it also clarifies the python versions for the AMD ROCm and cuda commands.

AlexonOliveiraRH:
BTW, [docs.instructlab.ai](https://github.com/instructlab/docs.instructlab.ai/pull/22) was updated with the same content of #2715 the already (merged 2 days ago).

reidliu41:
@kelbrown20 hi could you please take a look?

cdoern:
@kelbrown20 

>BTW, https://github.com/instructlab/docs.instructlab.ai/pull/22 was updated with the same content of https://github.com/instructlab/instructlab/pull/2715 already (merged 2 days ago).


did you do this? If so, I am going to opt to merge this one instead so we should update the docs on the website accordingly. We shouldn't advertise 0.17 as a viable solution

cdoern (approved):

kelbrown20 (approved):

AlexonOliveiraRH:
> @kelbrown20
> 
> > BTW, [instructlab/docs.instructlab.ai#22](https://github.com/instructlab/docs.instructlab.ai/pull/22) was updated with the same content of #2715 already (merged 2 days ago).
> 
> did you do this? If so, I am going to opt to merge this one instead so we should update the docs on the website accordingly. We shouldn't advertise 0.17 as a viable solution

@cdoern no, I don't have permission to do it, but @kelbrown20 closed mine anyway in favor of this one, so we should proceed with this. Thank you anyway. :)

kdubois:
@nathan-weinberg FYI I fixed the merge conflicts but had to do a force push on my local branch which seems to have cancelled the automerge if I understand correctly.

README.md
```diff
@@ -96,7 +96,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
 - C++ compiler
 - Python 3.10 or Python 3.11
 
-   📋 Python 3.12 and Python 3.13 are currently not supported. Some InstructLab dependencies don't work on Python 3.12 and above, yet.
+   ⚠️ Python 3.12+ is currently not supported. Some InstructLab dependencies don't work on Python 3.12, yet. It is recommended to use a specific version of Python in the below commands, eg. `python3.11` instead of simply `python3`.
```
courtneypacheco, line 99: this is good, but just an idea:

When I look at our `pyproject.toml` file, I notice that we specify `>=3.10,<3.12`. For example:

https://github.com/instructlab/instructlab/blob/main/pyproject.toml#L13

Maybe we could be even more clear in these docs that users need to use either Python 3.10 or Python 3.11, and that the source of truth for these versions can also be found in the `pyproject.toml` file.

I only mention this in case someone tries to run `ilab` against an older Python version (like 3.9).

If we do allow versions besides 3.10 and 3.11 — namely older versions like 3.9 — I think we should explicitly list them in a table or other easily-digestible format, and identify them as "recommended", "unsupported", or "incompatible." 

btw:
 +1 on the "it is recommended to a specific version of Python in the below commands." I think that is super helpful 👍

nathan-weinberg:
@kdubois can you just fix the spellchecker before merge?

kdubois:
done!

README.md
kdubois: That's a good idea, however there is already a "hardcoded" mention of the versions above the updated sentence, so maybe that would be too confusing?

Issue #2254, opened 2024-09-11
Running chat before download gives a non obvious error


danmcp:
**Describe the bug**
If you run ilab model chat before downloading a model, the error says backend can't be determined rather than a nice message that the model doesn't exist.  It message does end with `[Errno 2] No such file or directory` but it's a bit cryptic

**To Reproduce**
1. Install ilab
2. init with defaults
3. ilab model chat

% ilab model chat   
Failed to determine backend: Cannot determine which backend to use: Failed to determine whether the model is a GGUF format: [Errno 2] No such file or directory: '/Users/dmcphers/.cache/instructlab/models/merlinite-7b-lab-Q4_K_M.gguf'

**Device Info (please complete the following information):**
% ilab system info
sys.version: 3.11.9 (main, Apr  2 2024, 08:25:04) [Clang 15.0.0 (clang-1500.3.9.4)]
sys.platform: darwin
os.name: posix
platform.release: 23.6.0
platform.machine: arm64
platform.node: Dans-MacBook-Pro
platform.python_version: 3.11.9
instructlab.version: 0.18.0rc8.dev95
instructlab-dolomite.version: 0.1.1
instructlab-eval.version: 0.2.0
instructlab-quantize.version: 0.1.0
instructlab-schema.version: 0.4.1
instructlab-sdg.version: 0.3.0
instructlab-training.version: 0.4.2
torch.version: 2.3.1
torch.backends.cpu.capability: NO AVX
torch.version.cuda: None
torch.version.hip: None
torch.cuda.available: False
torch.backends.cuda.is_built: False
torch.backends.mps.is_built: True
torch.backends.mps.is_available: True
llama_cpp_python.version: 0.2.79
llama_cpp_python.supports_gpu_offload: True

PR #2725, opened 2024-11-29
add a model exist check for backend

reidliu41:
Based on [2254](https://github.com/instructlab/instructlab/issues/2254), try to state the model doesn't exist simply if not download before `model chat`

```
From: 
$ ilab model chat
Failed to determine backend: Cannot determine which backend to use: The model file /Users/xx/.cache/instructlab/models/granite-7b-lab-Q4_K_M.gguf is not a GGUF format nor a directory containing huggingface safetensors files. Cannot determine which backend to use.
Please use a GGUF file for llama-cpp or a directory containing huggingface safetensors files for vllm.
Note that vLLM is only supported on Linux.

To:
$ ilab model chat
Model /Users/xx/.cache/instructlab/models/granite-7b-lab-Q4_K_M.gguf does not exist. Please download the model first.
```

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/backends.py
+++ b/src/instructlab/model/backends/backends.py
@@ -105,6 +105,15 @@ def get(model_path: pathlib.Path, backend: str | None) -> str:
     return backend
 
 
+def check_model_path_exists(model_path: pathlib.Path) -> None:
+    if not model_path.exists():
+        click.secho(
+            f"{model_path} does not exist. Please download model first.",
+            fg="red",
+        )
+        raise click.exceptions.Exit(1)
+
+
 def select_backend(
     cfg: serve_config,
     backend: Optional[str] = None,
@@ -118,6 +127,7 @@ def select_backend(
     logger.debug("Selecting backend for model %s", model_path)
 
     model_path = pathlib.Path(model_path or cfg.model_path)
+    check_model_path_exists(model_path)
     backend_name = backend if backend is not None else cfg.backend
     try:
         backend = get(model_path, backend_name)
--- a/tests/common.py
+++ b/tests/common.py
@@ -32,6 +32,9 @@ def setup_gpus_config(section_path="serve", gpus=None, tps=None, vllm_args=lambd
     return _CFG_FILE_NAME
 
 
+@mock.patch(
+    "instructlab.model.backends.backends.check_model_path_exists", return_value=None
+)
 @mock.patch("instructlab.model.backends.vllm.check_api_base", return_value=False)
 # ^ mimic server *not* running already
 @mock.patch(
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -419,3 +419,18 @@ def test_get_model_template(tmp_path: pathlib.Path):
     )
     assert eos == "<|endoftext|>"
     assert bos == "<|begginingoftext|>"
+
+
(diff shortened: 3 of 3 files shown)


reidliu41:
@danmcp hi, could you please take a look this one?

tests/common.py
```diff
@@ -32,6 +32,9 @@ def setup_gpus_config(section_path="serve", gpus=None, tps=None, vllm_args=lambd
     return _CFG_FILE_NAME
 
 
+@mock.patch(
```
danmcp, line 35: I am guessing this check is causing a lot of tests to make it to the previous failure point of failed to determine backend?

If that's the case, I think it would be better to change those tests to adjust to the new error.

tests/test_lab_evaluate.py
```diff
@@ -642,7 +643,9 @@ def test_invalid_model_path_mt_bench(cli_runner: CliRunner, tmp_path):
 
 
 @patch("instructlab.model.evaluate.validate_model")
-def test_vllm_args_null(_, cli_runner: CliRunner):
+def test_vllm_args_null(_, cli_runner: CliRunner, tmp_path: pathlib.Path):
+    model_dir = tmp_path / "instructlab/granite-7b-lab"
+    model_dir.mkdir(parents=True, exist_ok=True)
```
danmcp, line 648: Dummy question:  If this needs new logic to not fail the model exists check, why didn't it fail previously with "Failed to determine backend"?

reidliu41: As I understand, it will fail with `Failed to determine backend` by default if `instructlab/granite-7b-lab` not exist, actually it will not fail due to `vllm_setup_test` use `determine_backend`.
···
@mock.patch("instructlab.model.backends.vllm.check_api_base", return_value=False)
# ^ mimic server *not* running already
@mock.patch(
    "instructlab.model.backends.backends.determine_backend",
    return_value=("vllm", "testing"),   <<<
)
@mock.patch("time.sleep", side_effect=Exception("Intended Abort"))
@mock.patch("subprocess.Popen")
def vllm_setup_test(runner, args, mock_popen, *_mock_args):
···
Currently, adding a first exist check, it will check it first, no need to go the next logic.
Please correct me if I am misunderstanding.

tests/common.py
danmcp: Turns out nothing was forcing a failed to determine backend failure.  It was always being mocked.  So this change is appropriate.

src/instructlab/model/backends/backends.py
```diff
@@ -105,6 +105,15 @@ def get(model_path: pathlib.Path, backend: str | None) -> str:
     return backend
 
 
+def check_model_path_exists(model_path: pathlib.Path) -> None:
+    if not model_path.exists():
+        click.secho(
+            f"Model {model_path} does not exist. Please download model first.",
```
danmcp, line 111: nit: Does this sentence sound better if it doesn't start with Model?

danmcp (approved):
Approving with the one nit

tests/test_lab_evaluate.py
reidliu41: Never mind, thanks for always sharing suggestions.

nathan-weinberg (approved):

Issue #2720, opened 2024-11-28
Show os-release variant information in `ilab system info`


tiran:
**Is your feature request related to a problem? Please describe.**
`/etc/os-release` can contain information about an OS variant. It would be useful to include the information in the output of `ilab system info`

**Describe the solution you'd like**
For `/etc/os-release` content

```
VARIANT="Some Product Name"
VARIANT_ID=product_name
PRODUCT_NAME_VERSION_ID=1.2.3
```

show

```
os-release.VARIANT: Some Product Name
os-release.PRODUCT_NAME_VERSION_ID: 1.2.3
```

`PRODUCT_NAME_VERSION_ID` is coming from upper-case `VARIANT_ID`.

jaideepr97:
Thanks for raising this @tiran 
I agree this could be useful

tiran:
It was Rom's idea. I'm just the messenger.

reidliu41:
Checking it in fedora, looks like below, VARIANT_ID is not like 1.2.3, and `os-release.PRETTY_NAME` include `VARIANT`, not enough? Please correct if I misunderstanding..
```
VARIANT="Server Edition"
VARIANT_ID=server

# ilab system info
Platform:
  sys.version: 3.11.9 (main, Apr 17 2024, 00:00:00) [GCC 14.0.1 20240411 (Red Hat 14.0.1-0)]
  sys.platform: linux
  os.name: posix
  platform.release: 6.8.5-301.fc40.x86_64
  platform.machine: x86_64
  platform.node: fedora153
  platform.python_version: 3.11.9
  os-release.ID: fedora
  os-release.VERSION_ID: 40
  os-release.PRETTY_NAME: Fedora Linux 40 (Server Edition)
  memory.total: 72.23 GB
  memory.available: 70.54 GB
  memory.used: 0.94 GB
```
and there is no VARIANT and VARIANT_ID in rhel
```
# cat /etc/os-release
NAME="Red Hat Enterprise Linux"
VERSION="9.2 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.2"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.2 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
[...]

PR #2728, opened 2024-11-30
add os-release variant information

reidliu41:
**Checklist:**


reidliu41:
For [2720](https://github.com/instructlab/instructlab/issues/2720)

nathan-weinberg (approved):

src/instructlab/system/info.py
```diff
@@ -29,10 +29,13 @@ def _platform_info() -> typing.Dict[str, typing.Any]:
...
     if sys.platform == "linux":
         os_release = platform.freedesktop_os_release()
-        for key in ["ID", "VERSION_ID", "PRETTY_NAME"]:
+        for key in ["ID", "VERSION_ID", "PRETTY_NAME", "VARIANT"]:
             value = os_release.get(key)
             if value:
                 d[f"os-release.{key}"] = value
+        variant_id = os_release.get("VARIANT_ID")
+        if variant_id:
+            d["os-release.PRODUCT_NAME_VERSION_ID"] = variant_id
```
danmcp, line 38: I could be misreading, but this doesn't seem like what is being suggested in https://github.com/instructlab/instructlab/issues/2720

I thought it was suggesting you would be looking for:

`os_release.get(f"os-release.{variant_id.upper()}_VERSION_ID")`

and if it's set then set:

`d[f"os-release.{variant_id.upper()}_VERSION_ID"] = <to its value>`

test case could be useful here as well.

danmcp (changes requested):

src/instructlab/system/info.py
reidliu41: sorry, just found it:
```
VARIANT="TEST TEST"
VARIANT_ID=test_test
TEST_TEST_VERSION_ID='1.1.1'
```

danmcp (approved):

nathan-weinberg:
Did we want a test case for this before merging?

danmcp:
> Did we want a test case for this before merging?

My take: I wouldn't turn it down.  But there are probably better things to invest test case writing time in.  The most likely outcome of this code breaking is it just wouldn't display anything since it's pretty well guarded now.

reidliu41:
yeah, I think so too --> writing the test case would likely mean mocking out os_release. As I check, seems more focus on `product` RHELAI, not sure the test env can easy to get the info?

PR #2736, opened 2024-12-03
[FOR SHARING PURPOSES ONLY] RAG ingestion and chat pipelines

dmartinol:
**TODO list** before marking the draft as ready:
- [x] fix mypy static analysis
- [x] fix pytest
- [x] introduce factory for DocumentStore
- [x] introduce factory for Retriever
- [x] introduce factory for Document Embedders
- [x] `ilab data process` should use the `DoclingConverter` instead of sdg `DocumentChunker`
- [x] add taxonomy navigation for default `ilab data ingest` pipeline
- [ ] add unit tests
- [ ] integration tests (maybe a different PR)
- [ ] Integrate PaRAGon
- [ ] <more to come>

Hi team! I’ve put together some initial code that can serve as a starting point for our discussion on the topic in subject.
Given the current limitations on updating the CLI interface in the current version, the approach leverages configuration options to enable the ingestion and chat pipelines without requiring new commands
Details below.

**Note**:
Since we're using a development version of `instructlab-sdg` package, the following steps are needed for now:
* Clone the instructlab/sdg repo
* Import the local package like `pip install -e /path/to/sdg/repo`

## RAG document transformation and ingestion
* Transform user docs using available SDG modules (no need to define any `qna.yaml` knowledge document)
* Generate embeddings and ingest them in a configured DB

### Flags and env vars
Reflect the design document [RAG ingestion and chat pipelines](https://github.com/instructlab/dev-docs/pull/161)

### Command
From pre-processed documents:
```console
ilab data process --rag /path/to/pre-processsed/docs
```

Inclusing pre-processing pipeline:
```console
ilab data process --rag --transform --transform-output /path/to/pre-processsed/docs /path/to/user/docs
```

Output in the configured DB. Supported DB: MilvusLite.

### Configuration
```yaml
chat:
  rag:
    enable: false
    retriever:
      top_k: 20
      embedder:
        model_name: sentence-transformers/all-minilm-l6-v2
    document_store:
      type: milvuslite
      uri: embeddings.db
      collection_name: Ilab
```
### Command
```console
ilab model chat
```
Includes additional marker to display the activation status of the RAG function:
```console
% ilab model chat                        
╭───────────────────────────────────────────────────────────────────────────── system ─────────────────────────────────────────────────────────────────────────────╮
│ Welcome to InstructLab Chat w/ GRANITE-7B-LAB-Q4_K_M.GGUF (type /h for help)                                                                                     │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
>>>                                                                                                                                                [RAG][S][default]
```


requirements.txt
```diff
@@ -37,3 +37,4 @@ wandb>=0.16.4
 xdg-base-dirs>=6.0.1
 psutil>=6.0.0
 huggingface_hub[hf_transfer]>=0.1.8
+haystack-ai>=2.8
```
cdoern, line 40: I am pretty sure new requirements like these need to be run by a specific set of people.

requirements/milvus.txt
```diff
@@ -0,0 +1,4 @@
+# Cannot upgrade because of https://github.com/milvus-io/milvus-haystack/issues/39
+milvus_haystack==0.0.11
```
cdoern, line 2: same here for new requirements

src/instructlab/cli/data/ingest.py
```diff
@@ -0,0 +1,128 @@
...
+    envvar="ILAB_EMBEDDING_MODEL_NAME",
+    type=click.STRING,
+    help="The embedding model name",
+)
+@click.option(
+    "--output-dir",
+    envvar="ILAB_OUTPUT_DIR",
+    help="Directory where generated datasets are stored.",
+)
+@click.argument(
```
cdoern, line 65: we tend, for now, to stay away from using click arguments. Can we use `.option` for everything here?

src/instructlab/cli/data/process.py
```diff
@@ -0,0 +1,84 @@
...
+    type=click.Path(),
+    envvar="ILAB_TAXONOMY_PATH",
+    help="Directory where taxonomy is stored and accessed from.",
+)
+@click.option(
+    "--taxonomy-base",
+    envvar="ILAB_TAXONOMY_BASE",
+    help="Branch of taxonomy used to calculate diff against.",
+)
+@click.argument(
```
cdoern, line 42: same here for arg vs option

src/instructlab/configuration.py
```diff
@@ -162,6 +162,9 @@ class _chat(BaseModel):
         default=1.0,
         description="Controls the randomness of the model's responses. Lower values make the output more deterministic, while higher values produce more random results.",
     )
+    rag: Optional[Dict[str, Any]] = Field(
```
cdoern, line 165: This should be its own class with specific arguments we can validate and run through tests rather than a dictionary  that can contain anything. If we are putting something directly into our config, we need to have control over validation.

if we were going to separate config yaml route, we could just blindly read it in, but I think this approach is better.

src/instructlab/data/ingest_docs.py
```diff
@@ -0,0 +1,116 @@
...
+):
+    pipeline = _create_pipeline(
+        document_store_config=document_store_config,
+        embedder_config=embedder_config,
+    )
+    _connect_components(pipeline)
+    _ingest_docs(pipeline=pipeline, input_dir=input_dir)
+
+
+def _create_pipeline(
```
cdoern, line 40: if we could add docstrings to these, that would be great!

src/instructlab/haystack/docling_splitter.py
cdoern, line 1: could we add haystack as a sub-dir of src/instructlab/rag, so that the directory structure makes more sense?

src/instructlab/model/chat.py
```diff
@@ -172,6 +176,7 @@ def is_openai_server_and_serving_model(
     "--temperature",
     cls=clickext.ConfigOption,
 )
+@rag_options
```
cdoern, line 179: so we are implicitly passing these in?

the way we do this with config entries is map the nested classes to click.options where the defaults are funneled in via `cls=clickext.ConfigOption`. I would prefer to 

1. make the config options a class not a dictionary 
2. add associated flags to override in `ilab model chat` where the defaults use the above class `ConfigOption` to read defaults from the config

src/instructlab/rag/rag_configuration.py
```diff
@@ -0,0 +1,141 @@
...
+from pydantic import BaseModel, ConfigDict
+import click
+
+# First Party
+from instructlab.configuration import DEFAULTS
+
+logger = logging.getLogger(__name__)
+
+
+def rag_options(command):
```
cdoern, line 16: will this wrapper apply these to any `ilab` command using the @rag_options? I would really prefer for ilab commands directly using rag and allowing users to configure RAG, that they have their own validated set of options that read from the config. 

A second option I would be ok with, is if these `rag_options` use the `cls=clickext.ConfigOption` class to read their defaults from the config file.

src/instructlab/rag/rag_configuration.py
```diff
@@ -0,0 +1,141 @@
...
+    )(command)
+    command = click.option(
+        "--retriever-top-k",
+        default=10,
+        envvar="ILAB_RETRIEVER_TOP_K",
+        type=click.INT,
+    )(command)
+    command = click.option(
+        "--retriever-embedder-model-dir",
+        default=lambda: DEFAULTS.MODELS_DIR,
```
cdoern, line 54: things like this for example should be reading from the config, we try not to specify defaults at this level

tests/testdata/default_config.yaml
```diff
@@ -16,6 +16,9 @@ chat:
   # Model to be used for chatting with.
   # Default: /cache/instructlab/models/granite-7b-lab-Q4_K_M.gguf
   model: /cache/instructlab/models/granite-7b-lab-Q4_K_M.gguf
+  # The RAG chat configuration
+  # Default: {}
+  rag: {}
```
cdoern, line 21: yeah, this should have some sane defaults or else users will not know how to use this

cdoern (changes requested):
left some lengthy comments. I know this is in a draft state but I figured I would get a review in. Many of my comments are concerns over config defaulting, reading, and how we are using these new RAG options. 

Also, we will need e2e tests for this since it is a major change.

src/instructlab/data/ingest_docs.py
dmartinol: only to "public" functions/methods?

requirements.txt
dmartinol: could you please elaborate more on the "specific set of people" part? (for sure I miss something here)

src/instructlab/cli/data/ingest.py
dmartinol: sure, I will also update the dev-docs PR accordingly

src/instructlab/rag/rag_configuration.py
dmartinol: > will this wrapper apply these to any ilab command using the @rag_options?
This was meant to be used only in the chat command, where it makes sense to add these options.

> I would really prefer for ilab commands directly using rag and allowing users to configure RAG, that they have their own validated set of options that read from the config
👍

src/instructlab/model/chat.py
dmartinol: 1. ok, as you also requested in another comment
2. sure, I'll use the regular design then

src/instructlab/configuration.py
dmartinol: The way it's been designed follows the initial conversation we had 2 weeks ago, when we decided not to update the configuration to avoid compatibility issues in case we would have changed any of these commands, as their definition seemed "unstable".
I'd be happy if we instead decide to adopt the regular config definitions, as it is a more straightforward implementation.

src/instructlab/model/chat.py
dmartinol, line 1: @cdoern I already noticed the mypy issues detected in this file, but I was wondering what I had to do with them. Should I try to fix them even if I did not update this code (w/o introducing any side effect, of course) or can we skip these warnings?

bbrowning:
I'm concerned about some of the new exposed functionality here for processing and ingesting docs. The instructlab/sdg repo already does a lot of this work, but here we're doing a slightly different version of doc processing and ingestion in this repo in a different way? But putting these commands under `ilab data` which today only houses commands that correspond to SDG functionality?

cdoern:
>I'm concerned about some of the new exposed functionality here for processing and ingesting docs. The instructlab/sdg repo already does a lot of this work, but here we're doing a slightly different version of doc processing and ingestion in this repo in a different way? But putting these commands under ilab data which today only houses commands that correspond to SDG functionality?

we could, to be safe, introduce these commands as `ilab rag...` commands with a new group rather than under data? @bbrowning @dmartinol what do you think?

bbrowning:
An `ilab rag ...` would be a reasonable compromise on the command side, as that makes it clear these commands are only related to rag functionality and not part of data generation. That also makes it easier to cleanly separate out these more experimental features from ones that are better supported. Ideally we'd have very clear warnings or even hide this command behind some kind of feature gate, but I realize we don't yet have feature gates as a thing.

dmartinol:
> An `ilab rag ...` would be a reasonable compromise on the command side, as that makes it clear these commands are only related to rag functionality and not part of data generation. That also makes it easier to cleanly separate out these more experimental features from ones that are better supported. Ideally we'd have very clear warnings or even hide this command behind some kind of feature gate, but I realize we don't yet have feature gates as a thing.

CLI commands have been discussed in this [ADR](https://github.com/instructlab/dev-docs/pull/161), feel free to comment and propose new options. Anyway, my understanding is that this PR will not be integrated as is, since we're reviewing the whole work in the [RAG project](https://github.com/orgs/instructlab/projects/20).

dmartinol:
closing PR as most of the changes were moved into smaller PRs:
* #2832 
* #2902 
* #2886 
* #2903

PR #2738, opened 2024-12-03
ci: add disk check to script

nathan-weinberg:
the github action was cleaning up the ilab artifacts
even if we preserved the E2E test dir

add disk checks into the script itself for more
granular checking + benefiting the libraries

**Checklist:**

### DIFF
--- a/scripts/e2e-ci.sh
+++ b/scripts/e2e-ci.sh
@@ -87,6 +87,11 @@ task() {
     step "$@"
 }
 
+check_disk() {
+   task Check disk
+   df -h
+}
+
 test_smoke() {
     task InstructLab smoke test
     ilab | grep --color 'Usage: ilab'
@@ -345,6 +350,7 @@ test_exec() {
     test_taxonomy
     test_generate
     test_data_list
+    check_disk
 
     # train tests
     if [ "$LARGE" -eq 1 ]; then
@@ -355,6 +361,7 @@ test_exec() {
     else
       test_train
     fi
+    check_disk
 
     # serve + chat tests
     test_serve
@@ -372,6 +379,7 @@ test_exec() {
     fi
 
     task E2E success!
+    check_disk
     exit 0
 }
 


nathan-weinberg:
This did not seem to work based on the results from https://github.com/instructlab/instructlab/actions/runs/12148323866/job/33876674159?pr=2738 (no change in disk space)

nathan-weinberg:
Changed the approach here, details in the commit desc

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -141,10 +141,6 @@ jobs:
           . venv/bin/activate
           ./scripts/e2e-ci.sh -m
 
-      - name: Check disk after tests
```
danmcp, line 144: Would it be ok to leave these?  They might come in handy later.

bbrowning (approved):
Ahh, good catch and thanks for ensuring this works as expected!

.github/workflows/e2e-nvidia-l4-x1.yml
bbrowning: Just for clarity, are you asking for both to stay? The new one added by Nathan that runs during the CI script, and this one that runs after at least some of the CI-generated artifacts have been cleaned up?

nathan-weinberg: Yes we can keep it for sure! Will update the patchset when I get in 👍

bbrowning: Sounds reasonable - thanks for the explanation @danmcp :smile:

.github/workflows/e2e-nvidia-t4-x1.yml
```diff
@@ -140,7 +140,7 @@ jobs:
       - name: Check disk after tests
         run: |
           df -h
-
```
danmcp, line 143: nit: looks like a remnant was left around

danmcp (approved):

PR #2764, opened 2024-12-07
fix: config init -- Add retry when an invalid HW vendor is selected

LaVLaS:
During the HW vendor selection of `ilab config init`, add a retry when the user enters an invalid selection for the vendor and system profile.

CHANGE:
```
First, please select the hardware vendor your system falls into
[0] NO SYSTEM PROFILE
[1] AMD
[2] APPLE
[3] INTEL
[4] NVIDIA
Enter the number of your choice: 5                                                                                                                                                                                 
Error: 5 is not in the range 0<=x<=4.                                                                                                                                                                              
Enter the number of your choice: 4                                                                                                                                                                                 
You selected: NVIDIA

Next, please select the specific hardware configuration that most closely matches your system.                                                                                                                     
[0] NO SYSTEM PROFILE                                                                                                                                                                                              
[1] NVIDIA A100 X2                                                                                                                                                                                                 
[2] NVIDIA A100 X4                                                                                                                                                                                                 
[3] NVIDIA A100 X8                                                                                                                                                                                                 
[4] NVIDIA H100 X2                                                                                                                                                                                                 
[5] NVIDIA H100 X4                                                                                                                                                                                                 
[6] NVIDIA H100 X8                                                                                                                                                                                                 
[7] NVIDIA L4 X8                                                                                                                                                                                                   
[8] NVIDIA L40S X4                                                                                                                                                                                                 
[9] NVIDIA L40S X8                                                                                                                                                                                                 
Enter the number of your choice [hit enter for hardware defaults]: 10                                                                                                                                              
Error: 10 is not in the range 0<=x<=9.                                                                                                                                                                             
Enter the number of your choice [hit enter for hardware defaults]: -1                                                                                                                                              
Error: -1 is not in the range 0<=x<=9.                                                                                                                                                                             
Enter the number of your choice [hit enter for hardware defaults]: 11                                                                                                                                              
Error: 11 is not in the range 0<=x<=9.                                                                                                                                                                             
Enter the number of your choice [hit enter for hardware defaults]: 99                                                                                                                                              
Error: 99 is not in the range 0<=x<=9.                                                                                                                                                                             
Enter the number of your choice [hit enter for hardware defaults]: 9
```

Resolves # [RHELAI-2399](https://issues.redhat.com/browse/RHELAI-2399)

**Checklist:**

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -176,7 +176,7 @@ def prompt_user_to_choose_vendors(
 
     system_profile_selection = click.prompt(
         "Enter the number of your choice",
-        type=int,
+        type=click.IntRange(0, len(keys)),
         default=0,
     )
     if 1 <= system_profile_selection <= len(keys):
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -205,7 +205,7 @@ def prompt_user_to_choose_profile(arch_family_processors) -> Config | None:
     cfg = None
     system_profile_selection = click.prompt(
         "Enter the number of your choice [hit enter for hardware defaults]",
-        type=int,
+        type=click.IntRange(0, len(arch_family_processors)),
         default=0,
     )
     # the file is SYSTEM_PROFILE_DIR/arch_family_procesors[key][selection-1]
@@ -217,12 +217,6 @@ def prompt_user_to_choose_profile(arch_family_processors) -> Config | None:
         click.echo(
             "No profile selected - ilab will use generic code defaults - these may not be optimized for your system."
         )
-    else:
-        click.secho(
-            "Invalid selection. Please select a valid system profile option.",
-            fg="red",
-        )
-        raise click.exceptions.Exit(1)
     return cfg
 
 


src/instructlab/config/init.py
```diff
@@ -194,6 +194,12 @@ def prompt_user_to_choose_vendors(
         click.echo(
             "No profile selected - ilab will use generic code defaults - these may not be optimized for your system."
         )
+    else:
+        click.secho(
```
danmcp, line 198: Would it be better to re-ask rather than exit?

LaVLaS: That is a good suggestion.  Should that be part of another PR since that UX change to retry would need to be implemented in the vendor and hardware system selection.

Is there any requirement to fix this quickly for a `v0.21.z-stream` release and, if so, I can rebase this on top of `release-v0.21` branch then we can do another issues/PR for a retry option?  As it stands now the v0.21.z releases will reset the system profile config without warning if the user accidentally hits Enter or inputs an invalid number

danmcp: No objection if we need to do a quick fix now and update later.  Re: 

> retry would need to be implemented in the vendor and hardware system selection

Not quite following this statement.  Wouldn't the loop be around the `click.prompt` just above?

LaVLaS: `click.IntRange` did the trick

src/instructlab/config/init.py
```diff
@@ -176,8 +176,7 @@ def prompt_user_to_choose_vendors(
 
     system_profile_selection = click.prompt(
         "Enter the number of your choice",
-        type=int,
-        default=0,
```
danmcp, line 180: What happens if you set a default with IntRange also set?  Does is not still loop if you pick an invalid value?  It seems like was the original desired behavior.

src/instructlab/config/init.py
```diff
@@ -190,10 +189,6 @@ def prompt_user_to_choose_vendors(
             )
         )
         click.echo("[0] NO SYSTEM PROFILE")
-    elif system_profile_selection == 0:
```
danmcp, line 193: What was the reasoning for removing this message?  It seems unrelated to the change.

LaVLaS: Apologies.  That is an error on my part.  I intended to remove the dead 'else:' code in `prompt_user_to_choose_profile()`

src/instructlab/config/init.py
LaVLaS: This may be an overreach on my part but removing the default requires the user to be explicit in the HW family choice.

danmcp: I don't have a strong opinion.  Wanted to note it is a behavior change.

LaVLaS: I restored the `default=0` after considering upstream/downstream documentation would need to be updated for that minor change.

src/instructlab/config/init.py
```diff
@@ -205,7 +205,7 @@ def prompt_user_to_choose_profile(arch_family_processors) -> Config | None:
     cfg = None
     system_profile_selection = click.prompt(
         "Enter the number of your choice [hit enter for hardware defaults]",
-        type=int,
+        type=click.IntRange(0, len(arch_family_processors)),
```
danmcp, line 208: nit: any reason why the above one uses `min=` and `max=` and this one doesn't?

danmcp (approved):

src/instructlab/config/init.py
LaVLaS: No real reason other than the 2nd was typed from memory.

Updated to remove min/max.

nathan-weinberg (approved):

nathan-weinberg:
@Mergifyio rebase

Issue #2671, opened 2024-11-18
Implement command to push Models to S3


cdoern:
Using the over-arching `ModelUploader` interface, we should be able to add an endpoint which allows users to push models to S3. 

This should be as easy as adding a secondary implementation under the `ModelUploader` class, this one being specifically for S3, research will need to be done to figure out if there is a library to easily do this. 

A prerequisite for this, of course, is the `ModelUploader` interface/class existing in the CLI.

nathan-weinberg:
Note on the above - the `Push` class @cdoern was referencing I've updated to use the new `ModelUploader` class

PR #2768, opened 2024-12-09
feat: add `ilab model upload` command

### LINKED ISSUE #2642: Implement command to push models to HuggingFace
the HuggingFace API allows for uploading both files and folders: https://huggingface.co/docs/huggingface_hub/main/en/guides/upload

This API should be used to push both GGUF and Safetensor/.bin directories to hf. 

using the new `--model` standards, the flag should be able to figure out if this is a GGUF, a folder of safetensors, or a .bin folder.

The interface used in this command should be extendable, similarly to the download one, so that other backends like OCI can easily be added in.

nathan-weinberg:
`ilab model upload` allows for uploading safetensor and GGUF models created by InstructLab users to
supported endpoints

Hugging Face is the only supported endpoint in this commit, but provisions have been added to eventually support OCI and S3 endpoints as well

Resolves #2642 

**Checklist:**


nathan-weinberg:
Test Repos I am using:
- https://huggingface.co/nathan-weinberg/test-upload-gguf/tree/main
- https://huggingface.co/nathan-weinberg/test-upload-safetensors

nathan-weinberg:
@mbestavros would like you to take a look as well please!

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,245 @@
...
+            release=release,
+            hf_token=hf_token,
+        )
+    elif dest_type == "oci":
+        # TODO: update this when OCI uploading is supported
+        click.secho(
+            f"Uploading of type {dest_type} is not yet supported",
+            fg="yellow",
+        )
+        raise click.exceptions.Exit(0)
```
danmcp, line 215: Should this be exit 0?  Someone might think it worked.

Same question for the s3 case.

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,245 @@
...
+    try:
+        uploader.upload()
+    # pylint: disable=broad-exception-caught
+    except (ValueError, Exception) as exc:
+        if isinstance(exc, ValueError) and "HF_TOKEN" in str(exc):
+            click.secho(
+                "Uploading to Hugging Face requires a HF Token to be set.\nPlease use '--hf-token' or 'export HF_TOKEN' to upload all necessary models.",
+                fg="yellow",
+            )
+            raise click.exceptions.Exit(0)
```
danmcp, line 239: Why exit 0 here?

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,245 @@
...
+)
+@click.option(
+    "--model-dir",
+    default=lambda: DEFAULTS.CHECKPOINTS_DIR,
+    show_default="The default system model location store, located in the data directory.",
+    help="The local directory to upload the model files from.",
+)
+@click.option(
+    "--dest-type",
+    type=click.Choice(["hf", "oci", "s3"]),
```
danmcp, line 171: s3 and oci are going to require additional info of where to publish.  I am guessing this param will still make sense.  But why bake it into the API before the rest of the implementation is designed?

danmcp (changes requested):

src/instructlab/cli/model/upload.py
nathan-weinberg: My thought was that the command itself wasn't failing (expected output providing expected behavior) so 0 made more sense than 1 for an exit code

src/instructlab/cli/model/upload.py
nathan-weinberg: Same answer as above

src/instructlab/cli/model/upload.py
nathan-weinberg: I'm assigned to do OCI and S3 in the short-term following HF - I preferred skeletoning out some of that work for myself now

src/instructlab/cli/model/upload.py
danmcp: I think of it as whether a shell script calling the command should consider it a success and keep going (with -e).  I would consider calling an upload that does nothing a failure.

nathan-weinberg: Fair point - I'm fine to change this

src/instructlab/cli/model/upload.py
nathan-weinberg: Will change this as well based on the above convo

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features
 
 * `ilab train --pipeline=accelerated --strategy=lab-skills-only` supports training with only the skills phase (leaving out knowledge).
+* A new command `ilab model upload` has been introduced so users can now upload their trained models to HuggingFace via the `ilab` CLI
```
courtneypacheco, line 8: Might be useful to provide a link :)

Also, a nitpick: maybe we should put a space between `Hugging` and `Face`

```suggestion
* A new command `ilab model upload` has been introduced so users can now upload their trained models to [Hugging Face](https://huggingface.co/) via the `ilab` CLI
```

README.md
```diff
@@ -1031,6 +1032,14 @@ argument to specify your new model:
 
    If you are interested in optimizing the quality of the model's responses, please see [`TROUBLESHOOTING.md`](./TROUBLESHOOTING.md#model-fine-tuning-and-response-optimization)
 
+## ☁️ Upload the new model
+
+If you are happy with your newly-trained model and wish to upload it to a supported endpoint, you can do so with the `ilab model upload` command.
+
+InstructLab currently supports uploading to Hugging Face, with plans to support OCI and S3 uploads as well.
```
courtneypacheco, line 1039: Maybe add a quick reference link?

```suggestion
InstructLab currently supports uploading to [Hugging Face](https://huggingface.co/), with plans to support OCI and S3 uploads as well.
```

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,245 @@
...
+        Upload specified model to Hugging Face
+        """
+
+        # HF token check
+        if self.hf_token == "":
+            raise ValueError(
+                """HF_TOKEN var needs to be set in your environment to upload HF Model.
+                Alternatively, the token can be passed with --hf-token flag.
+                The HF Token is used to authenticate your identity to the Hugging Face Hub."""
+            )
```
courtneypacheco, line 78: Is it possible someone could pass in a whitespace token like `"   "`? I suggest we use `.strip()` to strip all the whitespaces

We also might want to check for the `None` case

```suggestion
        # HF token check
        if not isinstance(self.hf_token, str):
            raise ValueError(
                """The HF Token is either unset or not a valid string.
                Please ensure that either the HF_TOKEN var is set to a valid string in your environment
                or pass a valid HF token string using the --hf-token flag.
                The HF Token is used to authenticate your identity to the Hugging Face Hub."""
            )
        if len(self.hf_token.strip()) == 0:
            raise ValueError(
                """HF_TOKEN var needs to be set in your environment to upload HF Model.
                Alternatively, the token can be passed with --hf-token flag.
                The HF Token is used to authenticate your identity to the Hugging Face Hub."""
            )
```

CHANGELOG.md
nathan-weinberg: I like it!

README.md
nathan-weinberg: Done :+1:

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,245 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+from pathlib import Path
+import abc
+import logging
+import os
+
+# Third Party
+from huggingface_hub import HfApi
```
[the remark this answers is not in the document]
nathan-weinberg: Hmmm, I don't know how it's not there already... this package is previously used in `ilab`... maybe it's already being provided somewhere else?

Either way I definitely agree, but would prefer to have that done in a separate PR. This one is functional as-is.

src/instructlab/cli/model/upload.py
nathan-weinberg: I believe (@cdoern can confirm) the string check is already done by Click 

However, I will go ahead and do the strip/len check for the case you mentioned.

We may want to upload the `ilab model download` check in a separate PR as well?

src/instructlab/cli/model/upload.py
danmcp: It would be good to be certain -dest-type is still useful with oci and s3.  I assume we will also need a param for s3 url and some way to auth.  Similar for oci.  At that point dest-type would be repeat information.  Which may be fine, but worth thinking through before codifying.

Follow-on question of whether s3 implies AWS or any s3 compliant api.  And if it is any s3 compliant api, would the auth setup vary?

nathan-weinberg: Right now I'm planning on using `--dest-type` for all three - that may change in the future, but would prefer to keep it in for now.

Let's keep the convo around S3 here: https://github.com/instructlab/instructlab/issues/2671

src/instructlab/cli/model/upload.py
courtneypacheco: I think you're right about the string check. The `click` docs say:

> If no type is provided, the type of the default value is used. If no default value is provided, the type is assumed to be [STRING](https://click.palletsprojects.com/en/stable/api/#click.STRING).

Source: https://click.palletsprojects.com/en/stable/options/#basic-value-options

Since `default=""`, which is obviously a `str`, it seems `click` will assume the variable's type is `str`.

```suggestion
@click.option(
    "--hf-token",
    default="",
    envvar="HF_TOKEN",
    type=str,
    help="User access token for connecting to the Hugging Face Hub.",
)
```

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,245 @@
...
+        except Exception as exc:
+            click.secho(
+                f"\nUploading GGUF model failed with the following Hugging Face Hub error:\n{DEFAULT_INDENT}{exc}",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
+        click.secho(
+            "\nUploading GGUF model succeeded!",
+            fg="green",
+        )
```
courtneypacheco, line 133: Can we print out the `upload_file` response? It appears that the `upload_file` function returns the URL to the upload: https://github.com/huggingface/huggingface_hub/blob/v0.26.5/src/huggingface_hub/hf_api.py#L4677L4681

We can confirm that the URL gets printed out by looking at Hugging Face's unit tests: https://github.com/huggingface/huggingface_hub/blob/main/tests/test_hf_api.py

Adding this kind of logging can help users triage issues related to uploading (like if somehow there is an uncaught exception on the Hugging Face end), and let them confirm where it was uploaded to.

```suggestion
    def upload_gguf(self) -> None:
        try:
            hf_api = HfApi()
            resp = hf_api.upload_file(
                path_or_fileobj=self.local_model_path,
                path_in_repo=self.model,
                repo_id=self.upload_dest,
                revision=self.release,
                token=self.hf_token,
            )
        except Exception as exc:
            click.secho(
                f"\nUploading GGUF model failed with the following Hugging Face Hub error:\n{DEFAULT_INDENT}{exc}",
                fg="red",
            )
            raise click.exceptions.Exit(1)
        click.secho(
            f"\nUploading GGUF model succeeded! Hugging Face response:\n{resp}",
            fg="green",
        )
```

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,245 @@
...
+        except Exception as exc:
+            click.secho(
+                f"\nUploading safetensors model failed with the following Hugging Face Hub error:\n{DEFAULT_INDENT}{exc}",
+                fg="red",
+            )
+            raise click.exceptions.Exit(1)
+        click.secho(
+            "\nUploading safetensors model succeeded!",
+            fg="green",
+        )
```
courtneypacheco, line 153: Same as above, but for `upload_folder`: https://github.com/huggingface/huggingface_hub/blob/v0.26.5/src/huggingface_hub/hf_api.py#L4971L4878

```suggestion
    def upload_safetensors(self) -> None:
        try:
            hf_api = HfApi()
            resp = hf_api.upload_folder(
                folder_path=self.local_model_path,
                repo_id=self.upload_dest,
                revision=self.release,
                token=self.hf_token,
            )
        except Exception as exc:
            click.secho(
                f"\nUploading safetensors model failed with the following Hugging Face Hub error:\n{DEFAULT_INDENT}{exc}",
                fg="red",
            )
            raise click.exceptions.Exit(1)
        click.secho(
            f"\nUploading safetensors model succeeded!  Hugging Face response:\n{resp}",
            fg="green",
        )
```

courtneypacheco (commented):
Two logging/debugging suggestions that might be helpful for end users.

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,249 @@
...
+                path_or_fileobj=self.local_model_path,
+                path_in_repo=self.model,
+                repo_id=self.upload_dest,
+                revision=self.release,
+                token=self.hf_token,
+            )
+            logger.debug(f"Hugging Face response:\n{resp}")
+        except Exception as exc:
+            click.secho(
+                f"\nUploading GGUF model failed with the following Hugging Face Hub error:\n{DEFAULT_INDENT}{exc}",
```
mbestavros, line 129: Might be handy to spit out the model name or path as part of this debug message. E.g:

```suggestion
                f"\nUploading GGUF model {self.local_model_path} failed with the following Hugging Face Hub error:\n{DEFAULT_INDENT}{exc}",
```

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,249 @@
...
+            resp = hf_api.upload_folder(
+                folder_path=self.local_model_path,
+                repo_id=self.upload_dest,
+                revision=self.release,
+                token=self.hf_token,
+            )
+            logger.debug(f"Hugging Face response:\n{resp}")
+        except Exception as exc:
+            click.secho(
+                f"\nUploading safetensors model failed with the following Hugging Face Hub error:\n{DEFAULT_INDENT}{exc}",
```
mbestavros, line 150: Same here as line 129.

nathan-weinberg:
@courtneypacheco I see to be getting a couple new mypy errors now - any idea if this might be related to the version change?
```
  Error: tests/test_lab.py:128: error: Argument 3 to "Command" has incompatible type "tuple[str, str]"; expected "bool"  [arg-type]
  Error: tests/test_lab.py:129: error: Argument 4 to "Command" has incompatible type "tuple[str, str]"; expected "bool"  [arg-type]
```
@cdoern not sure how I can correct for this? The command requires 3 args from a user and I don't want to add defaults that don't make sense

tests/test_lab_upload.py
```diff
@@ -0,0 +1,204 @@
...
+from instructlab import lab
+from instructlab.defaults import DEFAULT_INDENT
+from tests.test_backends import create_safetensors_or_bin_model_files
+
+
+class TestLabUpload:
+    # When using `from X import Y` you need to understand that Y becomes part
+    # of your module, so you should use `my_module.Y`` to patch.
+    # When using `import X`, you should use `X.Y` to patch.
+    # https://docs.python.org/3/library/unittest.mock.html#where-to-patch?
```
mbestavros, line 22: Was this intended to be pushed?

mbestavros (approved):
This is great to see! Thanks, Nathan - can't wait to see this be expanded with other upload options as well.

Most of my feedback are minor nits; nothing that should block merge.

tests/test_lab_upload.py
nathan-weinberg: Yup, it's in some of the other test files

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,249 @@
...
+@click.option(
+    "--model-dir",
+    default=lambda: DEFAULTS.CHECKPOINTS_DIR,
+    show_default="The default system model location store, located in the data directory.",
+    help="The local directory to upload the model files from.",
+)
+@click.option(
+    "--dest-type",
+    type=click.Choice(["hf", "oci", "s3"]),
+    required=True,
```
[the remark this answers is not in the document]
nathan-weinberg: Fair point - I'm open to that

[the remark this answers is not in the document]
nathan-weinberg: Esp considering `download` assumes Hugging Face

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,249 @@
...
+    help="The type of destination to upload to - can be 'hf', 'oci', or 's3'",
+)
+@click.option(
+    "--upload-dest",
```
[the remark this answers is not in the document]
nathan-weinberg: That's correct yes

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,249 @@
...
+    type=str,
+    required=True,
+    help="Destination for the model to be uploaded to. Ex: for a Hugging Face repo, should be '<username>/<repo-name>'",
+)
+@click.option(
+    "--release",
```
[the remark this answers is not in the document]
nathan-weinberg: This was mostly to align with terminology Hugging Face uses as well as `ilab model download`

It will also be generalized to other upload targets (e.g. a tag for an OCI endpoint)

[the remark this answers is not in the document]
mbestavros: Makes sense, carry on!

src/instructlab/cli/model/upload.py
nathan-weinberg: Hmmm, originally I didn't include this because it was included in the original output, but looking now that doesn't actually include `local_model_path` as the var hasn't been built out yet - so makes sense to me!

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,249 @@
...
+    "-m",
+    required=True,
+    help="Local model to upload.",
+)
+@click.option(
+    "--model-dir",
+    default=lambda: DEFAULTS.CHECKPOINTS_DIR,
+    show_default="The default system model location store, located in the data directory.",
+    help="The local directory to upload the model files from.",
+)
```
jaideepr97, line 172: I'm not seeing the benefit of having both here
can we just keep `--model`? best not to have a default IMO cause it might be prone to accidental uploads cause we cant say what checkpoint/other model a user would automatically want pushed out

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,249 @@
...
+@click.option(
+    "--dest-type",
+    type=click.Choice(["hf", "oci", "s3"]),
+    default="hf",
+    help="The type of destination to upload to - can be 'hf', 'oci', or 's3'",
+)
+@click.option(
+    "--upload-dest",
```
jaideepr97, line 180: upload seems redundant seeing as its an upload command parameter
just `--dest` or `--destination` is good enough IMO

src/instructlab/cli/model/upload.py
nathan-weinberg: So my thought was it would make it easier/clearer to allow us to specify just the name of a checkpoint rather than users having to pass the absolute path, while also allowing them to pass an absolute path if they prefer

src/instructlab/cli/model/upload.py
nathan-weinberg: Makes sense to me!

alinaryan (commented):
Thanks for working on this

src/instructlab/cli/model/upload.py
nathan-weinberg: To close the loop, I've aligned the scheme used here to match what's in the dev doc - @alinaryan will be replacing this logic across the project with a utility function at a later date

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,239 @@
...
+        try:
+            hf_api = HfApi()
+            resp = hf_api.upload_file(
+                path_or_fileobj=self.local_model_path,
+                path_in_repo=self.model,
+                repo_id=self.destination,
+                revision=self.release,
+                token=self.hf_token,
+            )
+            logger.debug(f"Hugging Face response:\n{resp}")
```
reidliu41, line 122: sorry, not familiar with HfApi, is it able to check the exist model? Or it can repeat to upload?

nathan-weinberg: Here's the docs for it, we are using both `upload_file` and `upload_folder` https://huggingface.co/docs/huggingface_hub/v0.26.5/en/package_reference/hf_api#huggingface_hub.HfApi

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,239 @@
...
+                Alternatively, the token can be passed with --hf-token flag.
+                The HF Token is used to authenticate your identity to the Hugging Face Hub."""
+            )
+
+        click.echo(
+            f"Uploading model to Hugging Face:\n{DEFAULT_INDENT}Model: {self.model}\n{DEFAULT_INDENT}Destination: {self.destination}@{self.release}\n"
+        )
+
+        # set logger
+        if self.ctx.obj is not None:
```
cdoern, line 83: interesting, what logging is handled specifically by HF that isn't already printed to stdout?

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,239 @@
...
+            # throw an error if the built path is invalid
+            if not os.path.exists(self.local_model_path):
+                click.secho(
+                    f"Couldn't find model at {self.local_model_path} - are you sure it exists?",
+                    fg="red",
+                )
+                raise click.exceptions.Exit(1)
+
+        # upload either safetensors or gguf, if valid
+        if is_model_safetensors(self.local_model_path):
```
cdoern, line 101: we should support .bin, I think there is a separate function for this (list supports this, as does download I think)

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,239 @@
...
+            hf_api = HfApi()
+            resp = hf_api.upload_file(
+                path_or_fileobj=self.local_model_path,
+                path_in_repo=self.model,
+                repo_id=self.destination,
+                revision=self.release,
+                token=self.hf_token,
+            )
+            logger.debug(f"Hugging Face response:\n{resp}")
+        except Exception as exc:
```
cdoern, line 123: I understand if not, but wonder if there are documented exceptions here.

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,239 @@
...
+        try:
+            hf_api = HfApi()
+            resp = hf_api.upload_folder(
+                folder_path=self.local_model_path,
+                repo_id=self.destination,
+                revision=self.release,
+                token=self.hf_token,
+            )
+            logger.debug(f"Hugging Face response:\n{resp}")
+        except Exception as exc:
```
cdoern, line 144: same here

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,239 @@
...
+            hf_token=hf_token,
+        )
+    elif dest_type == "oci":
+        # TODO: update this when OCI uploading is supported
+        click.secho(
+            f"Uploading of type {dest_type} is not yet supported",
+            fg="yellow",
+        )
+        raise click.exceptions.Exit(1)
+    elif dest_type == "s3":
```
cdoern, line 210: lets make these constants

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,239 @@
...
+    envvar="HF_TOKEN",
+    help="User access token for connecting to the Hugging Face Hub.",
+)
+@click.pass_context
+@clickext.display_params
+def upload(ctx, model, dest_type, destination, release, hf_token):
+    """Uploads model to a specified location"""
+    uploader = None
+
+    if dest_type == "hf":
```
cdoern, line 195: a constant would be good here. `DEFAULTS` perhaps? `DEFAULTS.UPLOAD_PROVIDERS`?

src/instructlab/cli/model/upload.py
```diff
@@ -0,0 +1,239 @@
...
+        raise click.exceptions.Exit(1)
+    else:
+        click.secho(
+            f"{dest_type} matches neither Hugging Face, OCI registry, or an S3-compatable format.\nPlease supply a supported dest_type",
+            fg="red",
+        )
+        raise click.exceptions.Exit(1)
+
+    try:
+        uploader.upload()
```
cdoern, line 225: generally speaking, with the client server re-architecutre, we are trying to keep "click" functionality in `src/instructlab/cli` and actual operations in the `src/instructlab/model` folder.

I think we should enforce this since me, @alinaryan, and @jaideepr97 are in the process of re-architecting all of ilab to follow this format. That being said, I'll leave it up to you what you think is just click processing and flag reading, and what is actual operational code. I encourage  moving the operational code to `src/instructlab/model/upload.py` in order for it to be considered for Api implemetations, process mgmt, etc.

cdoern (commented):
My largest comment is on what should live in `src/instructlab/cli` and what should live in `src/instructlab/model` Overall looks great. A few other comments I think should be addressed as well.

Additionally, are we going to be e2e testing this? Unsure If I feel comfortable merging this without it being tested. I know this will be difficult, so I would be fine scooping some work for e2e testing as long as it is concrete.

src/instructlab/cli/model/upload.py
cdoern: this is another one I think should be addressed before merge besides the cli split mentioned in the review comment

src/instructlab/cli/model/upload.py
nathan-weinberg: I was explicitly told to do this last week - given that as well as the PR has been through multiple review cycles already I would prefer not to make such a large change this late in.

src/instructlab/cli/model/upload.py
nathan-weinberg: I can do a constant for the Click options similar to what we do for Evaluation

src/instructlab/cli/model/upload.py
nathan-weinberg: This is the same thing we are doing in `download`

src/instructlab/cli/model/upload.py
nathan-weinberg: Do we support `.bin` downloads?

nathan-weinberg:
> My largest comment is on what should live in `src/instructlab/cli` and what should live in `src/instructlab/model` Overall looks great. A few other comments I think should be addressed as well.
> 
> Additionally, are we going to be e2e testing this? Unsure If I feel comfortable merging this without it being tested. I know this will be difficult, so I would be fine scooping some work for e2e testing as long as it is concrete.

Just because something isn't "E2E tested" doesn't mean it's not tested - I've introduced a large number of unit tests here - we can look into more testing at a later date, but I won't be scoping any more for this PR.

I've tested it manually, feel free to do the same.

src/instructlab/cli/model/upload.py
cdoern: I will admit I didn't review immediately, sorry about that! I am curious to know where this feedback was received to explicitly do it this way?

cdoern:
> > My largest comment is on what should live in `src/instructlab/cli` and what should live in `src/instructlab/model` Overall looks great. A few other comments I think should be addressed as well.
> > Additionally, are we going to be e2e testing this? Unsure If I feel comfortable merging this without it being tested. I know this will be difficult, so I would be fine scooping some work for e2e testing as long as it is concrete.
> 
> Just because something isn't "E2E tested" doesn't mean it's not tested - I've introduced a large number of unit tests here - we can look into more testing at a later date, but I won't be scoping any more for this PR.
> 
> I've tested it manually, feel free to do the same.

Makes sense, would you be willing to make an issue to add e2e testing for model uploading? I feel since it is part of the e2e flow, it warrants a tracker

src/instructlab/cli/model/upload.py
cdoern: yes https://github.com/instructlab/instructlab/pull/2283

all trained models are also in .bin format.

src/instructlab/cli/model/upload.py
cdoern: ok thanks

src/instructlab/cli/model/upload.py
nathan-weinberg: It was actually a conversation we had - originally I was putting things in the `model` directory and having issues, and you told me to put things in the `cli` directory instead. Maybe there was a crossed wire there, but since I hadn't heard anything from @jaideepr97 and @alinaryan (who have already reviewed this PR) I assumed it was fine.

Let's talk more about this offline.

src/instructlab/cli/model/upload.py
nathan-weinberg: Like documented HF exceptions?

src/instructlab/cli/model/upload.py
nathan-weinberg: Ack, I will make this change!

src/instructlab/cli/model/upload.py
jaideepr97: I agree with @cdoern - I had only done a brief initial pass so far and was planning to look through again but yeah, I guess I could've mentioned that earlier

would be ideal to follow the split structure if we are introducing a new command

nathan-weinberg: @jaideepr97 Charlie and I are going to look at it today and see how much of a lift it will be - if we can get it done without too much headache, we'll do it as part of this PR

alinaryan: It would be ideal if you could follow the split structure, but if it's too much of a lift for this PR, then JD and I will take it on

nathan-weinberg: Sounds good - much like the `model` flag conversation, we really need to surface these conventions somewhere more obvious - otherwise we're going to discourage upstream contributions.

nathan-weinberg: To close the loop here, I will indeed be doing the splitting as part of this PR :smile:

src/instructlab/cli/model/upload.py
nathan-weinberg: I've reworked how exceptions work in the latest patchset, so going to resolve this outdated thread

src/instructlab/model/upload.py
```diff
@@ -0,0 +1,160 @@
...
+            self.model = self.model.split("/")[-1]
+        else:
+            self.local_model_path = os.path.join(DEFAULTS.CHECKPOINTS_DIR, self.model)
+            # throw an error if the built path is invalid
+            if not os.path.exists(self.local_model_path):
+                raise FileNotFoundError(
+                    f"Couldn't find model at {self.local_model_path} - are you sure it exists?"
+                )
+
+        # upload either safetensors or gguf, if valid
```
jaideepr97, line 121: may want to add a call to `is_huggingface_repo` from utils here to validate the destination

nathan-weinberg: Ooh didn't know this func existed - some endpoint checking would be good - I will do this as part of the followup model management work I have planned!

cdoern (approved):
thanks for working with me on this one, lgtm 🙏

Issue #2792, opened 2024-12-13
ilab doesnt print options list when no config exists


RobotSail:
**What is the current experience of ilab you've found issue with?**

Today in InstructLab when the user interacts with commands that require a config, they will receive a message indicating that they lack a config. However; when the user doesn't have a config set and attempts to simply view the structure of ilab commands via `ilab model evaluate --help`, they will instead see the error that they do not have a config set.

**How do you think the described experience should be?**

This error should only be presented when the user actually _runs_ the command. It's also possible to pass the config as a flag via `--config`, but the user will not see this option if they can't even see the options they're allowed to pass n.

**What is the benefit of the change you are proposing?**

This will make it easier for the user to see what the options are and adjust their usage to be proper.

**If applicable, how would backwards-compatability be handled?**
n/a

reidliu41:
[2795](https://github.com/instructlab/instructlab/pull/2795)

PR #2795, opened 2024-12-15
handle the help situation if config lacks

reidliu41:
Resolves https://github.com/instructlab/instructlab/issues/2792

https://github.com/instructlab/instructlab/blob/main/src/instructlab/configuration.py#L1315

```
    # special case: --help should always work
    if config_obj is None and "--help" in sys.argv[1:]:  # config_obj will not None, so cannot handle the help situation
        config_obj = get_default_config()
        error_msg = None
```
Before:
```
$ ilab --config=/tmp/test-test.yaml model evaluate --help
Usage: ilab model [OPTIONS] COMMAND [ARGS]...
Try 'ilab model --help' for help.

Error: `/tmp/test-test.yaml` does not exist or is not a readable file.
Please run `ilab config init` or point to a valid configuration file using `--config=<path>`.
```
after:
```
$ ilab --config=test-test.yaml model evaluate --help
Usage: ilab model evaluate [OPTIONS]

  Evaluates a trained model

Options:
  --model TEXT                    Model to be evaluated [default: <None>;
                                  config: 'evaluate.model']
```

**Checklist:**

### DIFF
--- a/scripts/functional-tests.sh
+++ b/scripts/functional-tests.sh
@@ -582,6 +582,37 @@ test_ilab_chat_server_logs(){
     fi
 }
 
+test_ilab_help_without_config() {
+    # Test the help without config file after the fix
+    result=$(ilab --config=test-test.yaml model evaluate --help)
+    if echo "$result" | grep -F 'Usage: ilab model evaluate [OPTIONS]'; then
+        echo "Test passed: ilab model evaluate help message found."
+    else
+        echo "Error: ilab model evaluate help message not found."
+        exit 1
+    fi
+
+    result=$(ilab --config=test-test.yaml config show --help)
+    if echo "$result" | grep -F 'Usage: ilab config show [OPTIONS]'; then
+        echo "Test passed: ilab config show help message found."
+    else
+        echo "Error: ilab config show help message not found."
+        exit 1
+    fi
+
+    set +e
+    # Test the command run without config file, and fail as expected
+    result=$(ilab --config=test-test.yaml config show 2>&1)
+
+    if echo "$result" | grep -F "Error: \`test-test.yaml\` does not exist or is not a readable file"; then
+        echo "Test passed: ilab config show ran as expected with error message."
+    else
+        echo "Error: ilab config show did not output the expected error message."
+        exit 1
+    fi
+    set -e
+}
+
 #########
 # SETUP #
 #########
@@ -680,6 +711,8 @@ cleanup
 test_log_format
 cleanup
 test_llama_backend
+cleanup
+test_ilab_help_without_config
 cleanup
 
 exit 0
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -1350,8 +1350,7 @@ def init(
         )
 
     # special case: --help should always work
-    if config_obj is None and "--help" in sys.argv[1:]:
-        config_obj = get_default_config()
+    if not os.path.isfile(config_file) and "--help" in sys.argv[1:]:
         error_msg = None
 
     ctx.obj = Lab(config_obj, config_file, error_msg)


reidliu41:
@RobotSail could you please help to take a look?

RobotSail:
It seems to make sense, but I would add either a unit test or functional test to validate that we're getting the intended behavior.

scripts/functional-tests.sh
```diff
@@ -582,6 +582,24 @@ test_ilab_chat_server_logs(){
     fi
 }
 
+test_ilab_help_without_config() {
+    result=$(ilab --config=test-test.yaml model evaluate --help)
```
RobotSail, line 586: @reidliu41 Thanks for adding a second case. Could we please add a case where the command fails because it doesn't have the config? That way we can make sure your test is working as expected

scripts/functional-tests.sh
```diff
@@ -582,6 +582,34 @@ test_ilab_chat_server_logs(){
...
+    # Test the help without config file, and it will show help message
+    result=$(ilab --config=test-test.yaml model evaluate --help)
+    if echo "$result" | grep -F 'Usage: ilab model evaluate [OPTIONS]'; then
+        echo "Test passed: ilab model evaluate help message found."
+    else
+        echo "Error: ilab model evaluate help message not found."
+        exit 1
+    fi
+
+    result=$(ilab --config=test-test.yaml config show --help)
```
RobotSail, line 595: Should we be passing the config file for these tests if our goal is testing that `--help` will print when it doesn't have a config?

reidliu41: I think is need, just make sure it go to `else` for testing..
```
 if config_file == "DEFAULT":

elif os.path.isfile(config_file):
 
else:
```

RobotSail: Ah I see, you're passing in a config file that doesn't exist. That makes sense.

reidliu41:
weird, seems cannot match in linux:
```
 ++2024-12-19 15:55:07 5.2.21(1)-release:./scripts/functional-tests.sh:604: test_ilab_help_without_config(): ilab --config=test-test.yaml config show
  +2024-12-19 15:55:08 5.2.21(1)-release:./scripts/functional-tests.sh:604: test_ilab_help_without_config(): result='Usage: ilab config [OPTIONS] COMMAND [ARGS]...
  Try '\''ilab config --help'\'' for help.
  
  Error: `test-test.yaml` does not exist or is not a readable file.
  Please run `ilab config init` or point to a valid configuration file using `--config=<path>`.'
  +2024-12-19 15:55:08 5.2.21(1)-release:./scripts/functional-tests.sh:1: test_ilab_help_without_config(): cleanup 2
```

reidliu41:
weird, I tried it in my local MacOS, it can create a new process, and it will return `0` and get the output. But in the test machine, seems not work.

reidliu41:
```
trap 'true' ERR
trap -- ERR 
```
Also failed

reidliu41:
@RobotSail hi , can i just write a comment for the third condition,  seems hard to capture `exit 2` error, even with new process or background to run..

reidliu41:
```
 +2024-12-21 12:38:49 5.2.21(1)-release:./scripts/functional-tests.sh:603: test_ilab_help_without_config(): set +e
  ++2024-12-21 12:38:49 5.2.21(1)-release:./scripts/functional-tests.sh:605: test_ilab_help_without_config(): ilab --config=test-test.yaml config show
  +2024-12-21 12:38:50 5.2.21(1)-release:./scripts/functional-tests.sh:605: test_ilab_help_without_config(): result='Usage: ilab config [OPTIONS] COMMAND [ARGS]...
  Try '\''ilab config --help'\'' for help.
  
  Error: `test-test.yaml` does not exist or is not a readable file.
  Please run `ilab config init` or point to a valid configuration file using `--config=<path>`.'
  +2024-12-21 12:38:50 5.2.21(1)-release:./scripts/functional-tests.sh:607: test_ilab_help_without_config(): echo 'Usage: ilab config [OPTIONS] COMMAND [ARGS]...
  Try '\''ilab config --help'\'' for help.
  
  Error: `test-test.yaml` does not exist or is not a readable file.
  Please run `ilab config init` or point to a valid configuration file using `--config=<path>`.'
  +2024-12-21 12:38:50 5.2.21(1)-release:./scripts/functional-tests.sh:607: test_ilab_help_without_config(): grep -F 'Error: `test-test.yaml` does not exist or is not a readable file'
  Error: `test-test.yaml` does not exist or is not a readable file.
  +2024-12-21 12:38:50 5.2.21(1)-release:./scripts/functional-tests.sh:608: test_ilab_help_without_config(): echo 'Test passed: ilab config show ran as expected with error message.'
  Test passed: ilab config show ran as expected with error message. <<<<<<<<<<<<<----------   passed
  +2024-12-21 12:38:50 5.2.21(1)-release:./scripts/functional-tests.sh:613: test_ilab_help_without_config(): set -e
```

scripts/functional-tests.sh
nathan-weinberg: Was this added?

RobotSail: Yeah it looks like it was added a few lines down.

nathan-weinberg (approved):

PR #2796, opened 2024-12-15
deps: add yq as installation dependency

radeksm:
There is a missing dependency on yq which cause 'ilab taxonomy diff' command to fail with below error:

  WARNING (...)  instructlab.schema.taxonomy:336: could not run yq command
  File "(...)/lib64/python3.11/site-packages/instructlab/schema/taxonomy.py", line 331, in _schema_validate
    line = subprocess.check_output(["yq", yq_expression], input=text, text=True)
  File "/usr/lib64/python3.11/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib64/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
  subprocess.CalledProcessError: Command '['yq', '.seed_examples[19].questions_and_answers | line']' returned non-zero exit status 3.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,3 +37,4 @@ wandb>=0.16.4
 xdg-base-dirs>=6.0.1
 psutil>=6.0.0
 huggingface_hub[hf_transfer]>=0.1.8
+yq>=3.0.0


reidliu41:
@radeksm may need to re-push due to the job fail with below.
```
    tensor.tofile(fout)

Epoch 0: 100%|██████████| 11/11 [10:09<00:00, 55.39s/it]
+ rm -rf /tmp/tmp.QZCoCkAa6Y
```

reidliu41:
@bjhargrave hi, could you please help to take a look at the PR ?

requirements.txt
```diff
@@ -37,3 +37,4 @@ wandb>=0.16.4
 xdg-base-dirs>=6.0.1
 psutil>=6.0.0
 huggingface_hub[hf_transfer]>=0.1.8
+yq>=3.0.0
```
bjhargrave, line 40: The [python package `yq`](https://github.com/kislyuk/yq) is not a real yaml tool. It is just a wrapper around `jq` ("jq wrapper for YAML/XML/TOML documents") and thus it does not handle the [`line` operator](https://mikefarah.gitbook.io/yq/operators/line) that we need. We need to use the proper `yq` tool, https://github.com/mikefarah/yq, which is not a python package.

bjhargrave (changes requested):
This is not a useful change.

requirements.txt
radeksm: > The [python package `yq`](https://github.com/kislyuk/yq) is not a real yaml tool. It is just a wrapper around `jq` ("jq wrapper for YAML/XML/TOML documents") and thus it does not handle the [`line` operator](https://mikefarah.gitbook.io/yq/operators/line) that we need. We need to use the proper `yq` tool, https://github.com/mikefarah/yq, which is not a python package.

Ok if that's the case so indeed my change doesn't make sense. 
I am trying to think of any way of preventing users running into similar problem, maybe some point in README.md to ask user install /usr/bin/yq, and not use Python yq wrapper?

bjhargrave: A readme mention in a useful place could be helpful. We only "need" yq to report the line number of any schema violations. So yq is not a hard requirement but a soft requirement to improve error messages when the qna.yaml is invalid.

PR #2815, opened 2024-12-17
amend A100 profile SKUs

cdoern:
the A100 profile SKUs need to match the ones here: https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/src/nvidia/generated/g_nv_name_released.h

**Checklist:**

### DIFF
--- a/src/instructlab/config/init.py
+++ b/src/instructlab/config/init.py
@@ -349,9 +349,12 @@ def get_gpu_or_cpu() -> tuple[str, int, int]:
         click.echo("Detecting hardware...")
         gpus = torch.cuda.device_count()
         for i in range(gpus):
-            properties = torch.cuda.get_device_properties(i)
-            chip_name = properties.name.lower()
-            total_vram += properties.total_memory  # memory in B
+            chip_name, vram = utils.get_cuda_device_properties(i)
+            # if the SKU is something like A100-SXM4-40GB, split it for easier reading later
+            if "-" in chip_name:
+                chip_split = chip_name.split("-")
+                chip_name = " ".join(chip_split)
+            total_vram += vram
 
     vram = int(floor(convert_bytes_to_proper_mag(total_vram)[0]))
     # only do this on a CUDA system ROCm and HPU can have strange results when getting CPU info.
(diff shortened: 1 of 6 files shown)


src/instructlab/config/init.py
```diff
@@ -349,9 +361,12 @@ def get_gpu_or_cpu() -> tuple[str, int, int]:
...
         gpus = torch.cuda.device_count()
         for i in range(gpus):
-            properties = torch.cuda.get_device_properties(i)
-            chip_name = properties.name.lower()
-            total_vram += properties.total_memory  # memory in B
+            chip_name, vram = get_cuda_device_properties(i)
+            # if the SKU is something like A100-SXM4-40GB, split it for easier reading later
+            if "-" in chip_name:
+                chip_split = chip_name.split("-")
+                chip_name = " ".join(chip_split)
```
jaideepr97, line 368: Would be better to contain all this processing within the helper function IMO 
and use regex to split on a set of predefined separators 
not blocking

src/instructlab/config/init.py
```diff
@@ -333,6 +333,18 @@ def is_hpu_available() -> bool:
         return False
 
 
+def get_cuda_device_properties(gpu_num: int) -> tuple[str, int]:
```
jaideepr97, line 336: can we move this to utils? 
I plan to create a utils package and break up the file itself into smaller related groups
would be helpful to have everything in utils.py until then

src/instructlab/config/init.py
cdoern: I did it this way for testing purposes so I could mock the return value properly, I'd prefer to keep it like this if possible 🙏

jaideepr97: I mean sure
but IMO it would make more sense to have the unit test be for the helper function and mock it there instead

jaideepr97:
@mergifyio rebase

jaideepr97 (approved):

cdoern:
ci blew up bc of the storage failure and didn't restart when I re-triggered so I just re-pushed

PR #2822, opened 2024-12-18
refactor: separate download click logic

jaideepr97:
This PR separates click logic for the model download functionality 

**Checklist:**


src/instructlab/cli/model/download.py
```diff
@@ -0,0 +1,78 @@
...
+        download_models(
+            log_level=ctx.obj.config.general.log_level.upper(),
+            repositories=repositories,
+            releases=releases,
+            filenames=filenames,
+            model_dir=Path(model_dir),
+            hf_token=hf_token,
+        )
+    except Exception as e:
+        click.secho(e)
```
nathan-weinberg, line 77: Maybe something more explicit here with some color, for example
```suggestion
        click.secho(
            f"Downloading failed with the following exception: {e}",
            fg="red"
        )
```

src/instructlab/cli/model/download.py
```diff
@@ -0,0 +1,78 @@
...
+@click.pass_context
+@clickext.display_params
+def download(ctx, repositories, releases, filenames, model_dir, hf_token):
+    try:
+        download_models(
+            log_level=ctx.obj.config.general.log_level.upper(),
+            repositories=repositories,
+            releases=releases,
+            filenames=filenames,
+            model_dir=Path(model_dir),
```
nathan-weinberg, line 73: What will happen here if an invalid Path is passed? Should we do the cast before the `download_models` call but still within the `try` block?

src/instructlab/model/download.py
```diff
@@ -28,17 +27,17 @@
 logger = logging.getLogger(__name__)
 
 
-class Downloader(abc.ABC):
+class ModelDownloader(abc.ABC):
```
nathan-weinberg, line 30: Appreciate the alignment with `upload` here!

src/instructlab/model/download.py
```diff
@@ -82,20 +84,18 @@ def download(self):
...
             else:
                 self.download_gguf()
 
         except Exception as exc:
-            click.secho(
-                f"\nDownloading model failed with the following Hugging Face Hub error:\n{DEFAULT_INDENT}{exc}",
-                fg="red",
-            )
-            raise click.exceptions.Exit(1)
+            raise ValueError(
```
nathan-weinberg, line 96: Why a `ValueError`?

src/instructlab/model/download.py
```diff
@@ -220,26 +214,20 @@ def download(self):
             f"oci:{oci_dir}",
             "--remove-signatures",
         ]
-        if self.ctx.obj is not None and logger.isEnabledFor(logging.DEBUG):
+        if self.log_level == "DEBUG" and logger.isEnabledFor(logging.DEBUG):
```
nathan-weinberg, line 217: What's the difference between these two checks?

src/instructlab/model/download.py
```diff
@@ -258,99 +246,51 @@ def download(self):
...
-@click.pass_context
-@clickext.display_params
-def download(ctx, repositories, releases, filenames, model_dir, hf_token):
+def download_models(
+    log_level: str,
+    repositories: List[str],
+    releases: List[str],
+    filenames: List[str],
+    model_dir: Path,
+    hf_token,
```
nathan-weinberg, line 255:
```suggestion
    hf_token: str,
```

src/instructlab/model/download.py
```diff
@@ -258,99 +246,51 @@ def download(self):
...
+):
     """Downloads model from a specified repository"""
-    downloader = None
+    downloader: ModelDownloader = None
```
nathan-weinberg, line 258: Is this line necessary?

src/instructlab/model/download.py
```diff
@@ -258,99 +246,51 @@ def download(self):
...
 
         try:
             downloader.download()
         # pylint: disable=broad-exception-caught
         except (ValueError, Exception) as exc:
             if isinstance(exc, ValueError) and "HF_TOKEN" in str(exc):
-                click.secho(
-                    f"\n{downloader.repository} requires a HF Token to be set.\nPlease use '--hf-token' or 'export HF_TOKEN' to download all necessary models.",
-                    fg="yellow",
+                logger.info(
```
nathan-weinberg, line 290:
```suggestion
                logger.warning(
```

src/instructlab/model/download.py
jaideepr97: we were only using the ctx obj to access the log level. Once this functionality is behind a server layer there should not an expectation of a ctx obj being passed here because other clients may not have that notion

src/instructlab/model/download.py
jaideepr97: yes, otherwise there is a type error during assignment

src/instructlab/cli/model/download.py
jaideepr97: sure

src/instructlab/model/download.py
jaideepr97: that's just what ive been defaulting to
if there is a specific error I should use Idm switching to that

src/instructlab/model/download.py
jaideepr97: first check verifies if the client asked for debug level and the second check verifies if the logger instance is actually configured to handle debug or not

nathan-weinberg: ack, makes sense

src/instructlab/model/download.py
nathan-weinberg: But why does this need to be defined at all? The downloader object is only initialized within the below logic and otherwise isn't used. I don't see how this empty initialization provides any benefit.

src/instructlab/model/download.py
nathan-weinberg: `ValueError` suggests something specific, not sure if that applies to all the usage I'm seeing here: https://docs.python.org/3/library/exceptions.html#ValueError

I've defined custom exceptions for `upload` https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/upload.py#L20-L51

src/instructlab/model/download.py
jaideepr97: without it I was getting a mypy error when trying to assign an instance of `HfDownloader` to downloader, which it was interpreting to be of type `None } OCIDownlaoder` 

my initializing it as type `ModelDownloader` mypy stopped complaining about it because now it was of type ModelDownloader which satisfies both

src/instructlab/model/download.py
```diff
@@ -70,7 +72,7 @@ def download(self):
         """
         Download specified model from Hugging Face
         """
-        click.echo(
+        logger.info(
             f"Downloading model from Hugging Face:\n{DEFAULT_INDENT}Model: {self.repository}@{self.release}\n{DEFAULT_INDENT}Destination: {self.download_dest}"
         )
```
reidliu41, line 77: So we will use `logger.info` instead of `click.echo` in the future? Seems will different if in log.

jaideepr97: @reidliu41 we are restricting click to the `src/instructlab/cli` package. That package serves as the "client" side of the architecture so click should only be used there
This file represents the logical "backend" so it should not use click anywhere

reidliu41: thanks for share.

src/instructlab/model/download.py
jaideepr97: fair enough, ive updated most of the calls with something more appropriate
custom errors seems like overkill to me

nathan-weinberg: I think custom errors should be introduced given this will be the basis for a true API, but since this is a refactor PR I won't block on this

nathan-weinberg (approved):

cdoern:
found the issue. the HF_TOKEN err we handle specially, but in order to catch it, it needs to be a `ValueError` not a `PermissionError`. Just pushed that single line change up to see if we can get this in today

cdoern (approved):

PR #2826, opened 2024-12-18
adds `--disable-accelerate-full-state-checkpoints`

### LINKED ISSUE #2453: Expose `--accelerate_full_state_at_epoch` to CLI users from training library
By default, the instructlab/training loop saves the full parameter / optimizer / lr_scheduler state every epoch. This default is configured in the TrainingArguments object, and isn't exposed to the CLI users. This contributes to higher overall disk usage, unnecessarily so if the user doesn't care about being able to restart training, or can't afford the disk utilization.

JamesKunstle:
Saving full checkpoint state takes a lot of storage, and is configurable in `instructlab/training`. This addition makes this option visible through `instructlab/instructlab` CLI. I'm choosing to not add this to the configuration for the time being but I'm open to doing that.

Resolves #2453 

**Checklist:**

### DIFF
--- a/src/instructlab/cli/model/train.py
+++ b/src/instructlab/cli/model/train.py
@@ -384,6 +384,11 @@ def clickpath_setup(is_dir: bool) -> click.Path:
     is_flag=True,
     help="Optimize Memory Usage on CPU and MacOS. This uses the torch_dtype='auto' instead of float32",
 )
+@click.option(
+    "--disable-accelerate-full-state-at-epoch",
+    is_flag=True,
+    help="Globally skips full-state (parameters, optimizer, etc.) saving per epoch. DISABLES TRAINING RESUMEABILITY.",
+)
 @click.pass_context
 @clickext.display_params
 def train(
@@ -420,6 +425,7 @@ def train(
     force_clear_phased_cache: bool,
     distributed_backend,
     optimize_memory,
+    disable_accelerate_full_state_at_epoch: bool,
     **kwargs,
 ):
     """
@@ -452,6 +458,10 @@ def train(
     if is_high_fidelity(device=device) and pipeline == "accelerated":
         train_args, torch_args = map_train_to_library(ctx, ctx.params)
 
+        train_args.accelerate_full_state_at_epoch = (
+            not disable_accelerate_full_state_at_epoch
+        )
+
         # First Party
         from instructlab.model import accelerated_train
 
@@ -483,6 +493,7 @@ def train(
         except Exception as exc:
             click.secho(f"Accelerated Training failed with {str(exc)}")
             raise click.exceptions.Exit(1)
+
     elif not is_high_fidelity(device=device) and pipeline == "full":
         # Third Party
         import torch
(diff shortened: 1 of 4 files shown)


JamesKunstle:
I'll fix the failures asap, they're minor.

nathan-weinberg (approved):

src/instructlab/cli/model/train.py
```diff
@@ -384,6 +384,11 @@ def clickpath_setup(is_dir: bool) -> click.Path:
     is_flag=True,
     help="Optimize Memory Usage on CPU and MacOS. This uses the torch_dtype='auto' instead of float32",
 )
+@click.option(
+    "--disable-accelerate-full-state-at-epoch",
```
RobotSail, line 388: If the purpose of this flag is to prevent full state at epoch (which assumes that full state at epoch is default, may not be the case in all scenarios I'm guessing), then would it make sense to simply expose the `--accelerate-full-state-at-epoch` setting as a flag, and having it be default to False? 

I see the utility in having this default to on in the profiles for large scenarios, however; I think it's currently being weighed down by the heavy requirement of storing all the checkpoints. If we have a policy to only keep the last full state checkpoint somehow then I think it would make sense to have `--full-state-at-epoch` be set to true.

RobotSail (commented):
Just one comment but otherwise everything looks good.

src/instructlab/cli/model/train.py
JamesKunstle: My first version of this did exactly as you say but I decided that the expected behavior of the flag was kinda confusing if it's 1:1. 

e.g. "User wants to _enable_ full state epochs--> _don't_ use `--full-state-at-epoch`" (defaults to True) vs.
"User wants to _disable_ full state epochs --> _use_ `--full-state-at-epoch`" (sets False)

It's kinda logically switched, so I changed it to the current version.

I do think that overwriting full state checkpoints is the correct default approach, though, so I'll open an issue for that as well. That behavior, comboed with this, seems reasonable to me.

JamesKunstle:
@RobotSail https://github.com/instructlab/training/issues/387

src/instructlab/cli/model/train.py
RobotSail: That makes sense. @cdoern Would you be okay with this? 

> I do think that overwriting full state checkpoints is the correct default approach, though, so I'll open an issue for that as well. That behavior, comboed with this, seems reasonable to me.

Yeah we can do this in a follow-up PR.

JamesKunstle: ^ created an issue for this

RobotSail (approved):
LGTM, changes look good to me. I would like @cdoern to sign off on the CLI flag if possible

src/instructlab/cli/model/train.py
cdoern: I think this is fine because it's a flag not an option. if it was an option I'd be more particular but this is fine to me

cdoern (approved):

PR #2832, opened 2024-12-20
feat: Expose document store Python API in instructlab/instructlab rag submodule

dmartinol:
## References
This PR relates to the task "[Dev][RAG] Expose document store Python API in instructlab/instructlab rag submodule" in the RAG project.
See #2872 

## Proposed interfaces
We propose two interfaces to model the ingestion and retrieval interactions with a document store, in the context of RAG pipelines:
```python
class DocumentStoreRetriever(ABC):
    @abstractmethod
    def augmented_context(self, user_query: str) -> str:...

class DocumentStoreIngestor(ABC):
    @abstractmethod
    def ingest_documents(self, input_dir) -> tuple[bool, int]:...
```

## Design notes
The actual instances are generated by factory functions which delegate the actual implementation to a factory function identified with some hardcoded conventions.

Factory functions:
```python 
def create_document_retriever(
    document_store_config: DocumentStoreConfig,
    retriever_config: RetrieverConfig,
) -> DocumentStoreRetriever:
...

def create_document_store_ingestor(
    document_store_config: DocumentStoreConfig,
    embedding_config: EmbeddingModelConfig,
) -> DocumentStoreIngestor:
...
```

Initial implementation uses a Milvus db, and there is no option to configure it differently.
The implementations are based on Haystack pipelines whose details are abstracted by the interface defined in the `document_store` module.

## Test functions
Test functions in  `tests/test_document_store_factory.py` can help to clarify how to consume these APIs.

## Updated dependencies
Requirements file has been updated to reflect the required dependencies.

## Minimize the PR
The changes include the implementation of the Haystack document stores for Milvus mainly to prove the effectiveness of the proposal. If you prefer, we can start with a minimal commit to only include the interfaces and the minimal dependencies (configuration and  unit test).

@anastasds @jwm4

**Checklist:**

- [X] Functional tests have been added, if necessary.


src/instructlab/rag/document_store_factory.py
```diff
@@ -0,0 +1,68 @@
...
+    EmbeddingModelConfig,
+    RagFramework,
+    RetrieverConfig,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def create_document_retriever(
+    framework: RagFramework,
```
anastasds, line 20: Why are we parametrizing like this? We don't need to support swapping LLM frameworks, so this machinery is nontrivial and unnecessary.

src/instructlab/rag/haystack/components/document_splitter.py
```diff
@@ -0,0 +1,102 @@
...
+        self.__content_format = content_format
+
+    @component.output_types(documents=List[Document])
+    def run(self, documents: List[Document]):
+        if not isinstance(documents, list) or (
+            documents and not isinstance(documents[0], Document)
+        ):
+            raise TypeError(
+                "DoclingDocumentSplitter expects a List of Documents as input."
+            )
```
anastasds, line 45: Are the checks/exceptions in this class reproducing existing docling checks/exceptions? If so, we can just let it throw.

src/instructlab/rag/haystack/document_store_ingestor.py
```diff
@@ -0,0 +1,95 @@
...
+
+        return True, document_store.count_documents()
+
+
+def _create_pipeline(
+    document_store_config: DocumentStoreConfig, embedding_config: EmbeddingModelConfig
+) -> Pipeline:
+    pipeline = Pipeline()
+    pipeline.add_component(instance=create_converter(), name="converter")
+    pipeline.add_component(instance=DocumentCleaner(), name="document_cleaner")
```
anastasds, line 67: Are we sure we want to apply Haystack's document cleaning?

anastasds (changes requested):
Suggested some simplification and had a few questions.

src/instructlab/rag/rag_configuration.py
```diff
@@ -0,0 +1,79 @@
...
+        extra="ignore",
+        protected_namespaces=(),
+        use_enum_values=True,  # populate models with the value property of enums, rather than the raw enum.
+    )
+    model_dir: str = Field(
+        default=DEFAULTS.MODELS_DIR,
+        description="The default system model location store, located in the data directory.",
+    )
+    model_name: str = Field(
+        # default_factory=lambda: "ibm-granite/granite-embedding-278m-multilingual",
```
anastasds, line 54: There's an ADR about model choice at https://github.com/instructlab/dev-docs/pull/169

requirements.txt
```diff
@@ -37,3 +37,7 @@ wandb>=0.16.4
 xdg-base-dirs>=6.0.1
 psutil>=6.0.0
 huggingface_hub[hf_transfer]>=0.1.8
+haystack-ai>=2.8
+milvus_haystack>=0.0.11
+docling-core[chunking]>=2.10.0
+sentence-transformers>=3.0.0
```
anastasds, line 43: Do we need this dependency?

dmartinol:
@anastasds Please note that the `document_splitter` might be replaced by the one available in the `docling-haystack`, whenever it will be available. 
ATM we cannot integrate it because of a dependency conflict with the `instructlab-sdg` package:
```
    instructlab-sdg 0.6.2 depends on docling<=2.8.3 and >=2.4.2
    docling-haystack 0.1.1 depends on docling<3.0.0 and >=2.9.0
```
I've added one comment [here](https://github.com/instructlab/sdg/issues/436#issuecomment-2574886485) to track the request of upgrading the `instructlab-sdg` dependencies.

requirements.txt
dmartinol: `sentence-transformers` package is needed if we use the `SentenceTransformersDocumentEmbedder` (at ingestion time) and `SentenceTransformersTextEmbedder` (at inference time). They are compatible with the granite embedding model and also with most of the models available on HF.
Anyway, I see the point and I'm wondering if we need to make this step configurable in some way. Assuming that we're not serving the embedding endpoint (`ilab model serve --model-path ....ibm-granite/granite-embedding-125m-english`), the alternative is what Haystack calls the [on-premise-embedders](https://docs.haystack.deepset.ai/docs/choosing-the-right-embedder#on-premise-embedders). 
What other embedding component could we use?

src/instructlab/rag/rag_configuration.py
dmartinol: Updated accordingly to `ibm-granite/granite-embedding-125m-english`

src/instructlab/rag/haystack/components/document_splitter.py
dmartinol: Actually these are the Haystack errors raised during input parameter validation, I don't think we can rely on Docling for that. Any possible exception raised by Docling will be just thrown back to the caller.

Please note that the [docling-haystack](https://github.com/DS4SD/docling-haystack/tree/main) package may probably define its own "official" version of this Haystack component, so we may dismiss ours at that time.

dmartinol: Raised [issue #4](https://github.com/DS4SD/docling-haystack/issues/4) to request the component to be defined.

src/instructlab/rag/haystack/document_store_ingestor.py
dmartinol: I'm not sure, but if we assume to feed the pipelines with un-processed user documentation (pdf), are we also assuming that they're all ready-to-use in the RAG ingestion pipeline?
BTW: I might be wrong, but I don't think Docling applies any cleanup to the original documents.

src/instructlab/rag/document_store_factory.py
dmartinol: removed the `RagFramework` type but kept the same logic to lookup the creation function based on the `DocumentStoreType` enum. Please TAL
```py
        module_name = "instructlab.rag.haystack.document_store_factory"
        module = importlib.import_module(module_name)

        function_name = f"create_{type.value}_document_retriever"
        function = getattr(module, function_name)
```

tests/test_document_store_factory.py
dmartinol, line 1: These test functions assume that the default embedding model has already been downloaded, which is not reasonable of course. What strategy do we use for similar test cases? Is there any example already in place to anticipate the requested download before the test functions are executed?

dmartinol: @anastasds @cdoern maybe you can help here

requirements.txt
jwm4: Thank you for explaining, @dmartinol .  This clarifies a lot.   I think you are right that SentenceTransformersTextEmbedder is the best option.  I will see what I can do about getting this approved.

src/instructlab/rag/document_store_factory.py
```diff
@@ -0,0 +1,65 @@
...
+    type: DocumentStoreType,
+    document_store_config: DocumentStoreConfig,
+    retriever_config: RetrieverConfig,
+) -> DocumentStoreRetriever:
+    try:
+        module_name = "instructlab.rag.haystack.document_store_factory"
+        module = importlib.import_module(module_name)
+
+        function_name = f"create_{type.value}_document_retriever"
+        function = getattr(module, function_name)
```
anastasds, line 28: I would still argue that, as a general policy, levels of abstraction such as this one are best not implemented until actually needed. Maybe the next document store required will not be supported in Haystack, for example. 

Code is a liability; less is better. I can see the argument that we expect to support multiple document stores, but right now that is simply not required, and the focus is on getting the simplest possible vertical slice working.

This is especially the case when getting into manually importing modules using string identifiers and `getattr` loading functions - these things can be very fragile.

dmartinol: 👍 I will remove both the function lookup and the store type enum, providing an hardcoded, non-configurable implementation based on milvus.

src/instructlab/rag/haystack/document_store_ingestor.py
```diff
@@ -0,0 +1,95 @@
...
+        self.document_store_config = document_store_config
+        self.embedding_config = embedding_config
+        self._pipeline = _create_pipeline(
+            document_store_config=document_store_config,
+            embedding_config=embedding_config,
+        )
+        _connect_components(self._pipeline)
+
+    def ingest_documents(self, input_dir) -> tuple[bool, int]:
+        pattern = "*json"
```
anastasds, line 43:
```suggestion
        pattern = "*.json"
```

src/instructlab/rag/haystack/document_store_retriever.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    text_embedder = create_embedder(retriever_config)
+    pipeline = Pipeline()
+    pipeline.add_component("embedder", text_embedder)
+    pipeline.add_component("retriever", document_retriever)
+    return pipeline
+
+
+def _connect_components(pipeline: Pipeline):
+    pipeline.connect("embedder.embedding", "retriever.query_embedding")
```
anastasds, line 68: I see [in Haystack docs](https://docs.haystack.deepset.ai/docs/components) that

>To connect components in a pipeline, you need to know the names of the inputs and outputs they accept. The output of one component must be compatible with the input the subsequent component accepts. 

Having magic strings as an approach gives some pause and suggests an enum, though this is probably fine for now as we're exploring.

anastasds (approved):
Pending the discussion regarding the new dependencies, looks good to me!

src/instructlab/rag/haystack/document_store_retriever.py
dmartinol: I agree, it is definitely error prone, but for now we can't do otherwise!

dmartinol:
@anastasds 
- can we make the PR ready for review?
- what about the issue with the test functions that require the download of the default embedding model?
  - make it a real unit test with mocked embedder? (not easy but probably doable)
  - make it an integration test instead?

dmartinol:
@anastasds I moved the RAG config classes to `configuration` module and made them "private". Used same code as in your PR. If you agree with the approach, you can rebase your changes to this branch

src/instructlab/rag/document_store_factory.py
```diff
@@ -0,0 +1,43 @@
...
+import logging
+
+# First Party
+from instructlab.rag.document_store import DocumentStoreIngestor, DocumentStoreRetriever
+
+logger = logging.getLogger(__name__)
+
+
+def create_document_retriever(
+    document_store_uri: str,
```
cdoern, line 11: can we add doc strings for all of these for folks unfamiliar with the functionality?

src/instructlab/rag/haystack/components/document_splitter.py
```diff
@@ -0,0 +1,101 @@
...
+    default_to_dict,
+)
+from pydantic_core._pydantic_core import ValidationError
+
+logger = logging.getLogger(__name__)
+
+
+@component
+class DoclingDocumentSplitter:
+    SUPPORTED_CONTENT_FORMATS = ["json"]
```
cdoern, line 24: should this be in `defaults.py`?

src/instructlab/rag/haystack/components/document_splitter.py
```diff
@@ -0,0 +1,101 @@
...
+        split_docs = []
+        for doc in documents:
+            if doc.content is None:
+                raise ValueError(f"Missing content for document ID {doc.id}.")
+
+            chunks = self._split_with_docling(doc.meta["file_path"], doc.content)
+            current_split_docs = [Document(content=chunk) for chunk in chunks]
+            split_docs.extend(current_split_docs)
+
+        return {"documents": split_docs}
```
cdoern, line 56: this feels like it should be a Class not just a dict

src/instructlab/rag/haystack/document_store_ingestor.py
```diff
@@ -0,0 +1,94 @@
...
+        instance=create_document_writer(
+            document_store_uri=document_store_uri,
+            document_store_collection_name=document_store_collection_name,
+        ),
+        name="document_writer",
+    )
+    return pipeline
+
+
+def _connect_components(pipeline: Pipeline):
```
cdoern, line 90: all of these haystack methods should have doc strings please

cdoern (commented):
some comments on docstrings, but nothing really major stands out here to me, looks good

src/instructlab/rag/haystack/components/document_splitter.py
dmartinol: This is how haystack works, can’t do differently

tests/test_document_store_factory.py
dmartinol: Using mock testing to avoid interacting with real models

src/instructlab/rag/document_store.py
```diff
@@ -0,0 +1,53 @@
...
+    def augmented_context(self, user_query: str) -> str:
+        """
+        Retrieve documents from the actual store matching the given `user_query` and compute the augmented context to be used
+        in a RAG chat pipeline.
+
+        Params:
+          user_query: The original user query.
+        Returns:
+          str: The augmented context to use in a RAG chat.
+        """
```
courtneypacheco, line 31: Should we raise `NotImplementedError` or add `pass` since this is an interface?

Common approach: https://github.com/google/etils/blob/6ff7408dbec072e4561e393d5abac9625476df36/etils/epath/backend.py#L41

```suggestion
    @abstractmethod
    def augmented_context(self, user_query: str) -> str:
        """
        Retrieve documents from the actual store matching the given `user_query` and compute the augmented context to be used
        in a RAG chat pipeline.

        Params:
          user_query: The original user query.
        Returns:
          str: The augmented context to use in a RAG chat.
        """
        raise NotImplementedError
```

src/instructlab/rag/document_store.py
```diff
@@ -0,0 +1,53 @@
...
+        Ingest documents from the `input_dir` location to the document store.
+
+        Params:
+          input_dir (str): The folder containing user documents (exploration is not recursive). The assumption is that these documents have been
+          processed using instructlab-sdg processing functions, based on docling, and are in JSON format.
+        Returns:
+          bool: True if the ingestion succeeded, False otherwise.
+          int: The number of documents written to the document store.
+        """
```
courtneypacheco, line 53: Same here about:

> Should we raise `NotImplementedError` or add `pass` since this is an interface?

src/instructlab/rag/haystack/component_factory.py
```diff
@@ -0,0 +1,67 @@
...
+def create_document_writer(
+    document_store_uri: str, document_store_collection_name: str
+) -> DocumentWriter:
+    return DocumentWriter(
+        create_document_store(
+            document_store_uri=document_store_uri,
+            document_store_collection_name=document_store_collection_name,
+            drop_old=True,
+        )
+    )
```
courtneypacheco, line 28: Can we add unit tests for this? Unless I missed them?

src/instructlab/rag/haystack/component_factory.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def create_retriever(
+    top_k: int,
+    document_store: InMemoryDocumentStore,
+):
+    return InMemoryEmbeddingRetriever(
+        document_store=document_store,
+        top_k=top_k,
+    )
```
courtneypacheco, line 47: Can we add unit tests for this as well?

src/instructlab/rag/haystack/components/document_splitter.py
```diff
@@ -0,0 +1,102 @@
...
+from pydantic_core._pydantic_core import ValidationError
+
+# First Party
+from instructlab.defaults import DEFAULTS
+
+logger = logging.getLogger(__name__)
+
+
+@component
+class DoclingDocumentSplitter:
```
courtneypacheco, line 26: Can we add unit tests for this class and its functions?

courtneypacheco (commented):
Looks good! Had a few comments (mostly about test coverage)

src/instructlab/rag/haystack/component_factory.py
dmartinol: it's covered by `test_document_store_factory.py`:
```
> pytest --cov=instructlab --cov-report term tests -k 'test_document_st'
...
src/instructlab/rag/__init__.py                                    0      0   100%
src/instructlab/rag/document_store.py                              7      0   100%
src/instructlab/rag/document_store_factory.py                      9      0   100%
src/instructlab/rag/haystack/component_factory.py                 22      3    86%
src/instructlab/rag/haystack/components/document_splitter.py      52     32    38%
src/instructlab/rag/haystack/document_store_factory.py             9      0   100%
src/instructlab/rag/haystack/document_store_ingestor.py           42      4    90%
src/instructlab/rag/haystack/document_store_retriever.py          27      0   100%
```

I can add an isolated test and verify it returns a not `None` instance if you prefer.

src/instructlab/rag/document_store.py
dmartinol: I agree with you, but I based this implementation from the few interfaces I found, which do the same (see https://github.com/instructlab/instructlab/blob/main/src/instructlab/llamacpp/llamacpp_convert_to_gguf.py)
I'm available to change it, of course: should we involve someone to decide b/w `pass` or `NotImplementedError`?

src/instructlab/rag/haystack/components/document_splitter.py
dmartinol: Sure, I can add more tests and increment the coverage which is quite low:
```
src/instructlab/rag/haystack/components/document_splitter.py      52     32    38%
```

I just want to anticipate you that sooner or later this implementation will be replace by the equivalent one provided by the `docling-haystack` package. See this [issue #4](https://github.com/DS4SD/docling-haystack/issues/4)

src/instructlab/rag/document_store.py
cdoern: I think if this is not a norm in ilab, we should avoid this for now and keep the current implementation.

dmartinol:
@courtneypacheco added more unit tests to increase coverage
```
src/instructlab/rag/__init__.py                                    0      0   100%
src/instructlab/rag/document_store.py                              7      0   100%
src/instructlab/rag/document_store_factory.py                      9      0   100%
src/instructlab/rag/haystack/component_factory.py                 22      0   100%
src/instructlab/rag/haystack/components/document_splitter.py      52      7    87%
src/instructlab/rag/haystack/document_store_factory.py             9      0   100%
src/instructlab/rag/haystack/document_store_ingestor.py           42      3    93%
src/instructlab/rag/haystack/document_store_retriever.py          27      0   100%
```

cdoern:
@dmartinol this is looking great! can you squash the commits before merging?

dmartinol:
> @dmartinol this is looking great! can you squash the commits before merging?
✅

cdoern (approved):
thanks for the hard work on this!

nathan-weinberg (approved):

Issue #2837, opened 2024-12-25
Does InstructLab support offline training?


orgC:
**Is your feature request related to a problem? Please describe.**
InstructLab could train data in a disconnected environment.

**Describe the solution you'd like**
It is recommended to have a parameter that supports training in an offline environment.

**Additional context**

When training, I tried adding the --local parameter, but it didn’t work.

I am trying to train a model in an offline environment and encountered an issue. Below are the steps I followed:

I executed the following command:

```
ilab model train --local --data-path ~/.local/share/instructlab/datasets/knowledge_train_msgs_2024-12-24T17_57_37.jsonl
```
However, I received the following error message:

```
...
  File "/root/instructlab/venv/lib/python3.11/site-packages/transformers/configuration_utils.py", line 574, in get_config_dict
    config_dict, kwargs = cls._get_config_dict(pretrained_model_name_or_path, **kwargs)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/instructlab/venv/lib/python3.11/site-packages/transformers/configuration_utils.py", line 633, in _get_config_dict
    resolved_config_file = cached_file(
                           ^^^^^^^^^^^^
  File "/root/instructlab/venv/lib/python3.11/site-packages/transformers/utils/hub.py", line 446, in cached_file
    raise EnvironmentError(
OSError: We couldn't connect to 'https://huggingface.co' to load this file, couldn't find it in the cached files and it looks like instructlab/granite-7b-lab is not the path to a directory containing a file named config.json.
Checkout your internet connection or see how to run the library in offline mode at 'https://huggingface.co/docs/transformers/installation#offline-mode'.
```

reidliu41:
Did you download the model to local? And simply check `--local` should just use for `--pipeline simple`.

orgC:
@reidliu41 , I added `--pipeline simple`, but the result is still the same 

```
(venv) [root@rhel-ai-offline datasets]# ilab model train --pipeline simple --local
INFO 2024-12-25 19:20:32,919 numexpr.utils:162: NumExpr defaulting to 16 threads.
INFO 2024-12-25 19:20:33,306 datasets:59: PyTorch version 2.4.1 available.
LINUX_TRAIN.PY: NUM EPOCHS IS:  10
[...]

PR #2838, opened 2024-12-26
add tips for --local option usage

reidliu41:
Checking `train` , only `pipeline simple` use `--local`, but users thought they all works, so add tips for it.

https://github.com/instructlab/instructlab/issues/2837

**Checklist:**

### DIFF
--- a/src/instructlab/cli/model/train.py
+++ b/src/instructlab/cli/model/train.py
@@ -95,7 +95,7 @@ def clickpath_setup(is_dir: bool) -> click.Path:
 @click.option(
     "--local",
     is_flag=True,
-    help="Whether or not `model_dir` is remote from HuggingFace.",
+    help="Whether or not `model_dir` is local vs remote from HuggingFace (use with '--pipeline simple' on MacOS).",
 )
 @click.option(
     "-sq",


src/instructlab/cli/model/train.py
```diff
@@ -95,7 +95,7 @@ def clickpath_setup(is_dir: bool) -> click.Path:
 @click.option(
     "--local",
     is_flag=True,
-    help="Whether or not `model_dir` is remote from HuggingFace.",
+    help="Whether or not `model_dir` is remote from HuggingFace (use with '--pipeline simple').",
```
danmcp, line 98: The description (including before your change) seems to indicate the opposite of what the arg does.

Would this be better as something like:

```Whether or not `model_dir` is local vs remote from Hugging Face (use with '--pipeline simple')```

?

reidliu41: Thank you for your suggestion.
And I check again, seems it(`local`) also only use in `MacOS`, should I mention it too?
https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/simple_train.py#L17

danmcp (approved):

courtneypacheco (approved):

Issue #2287, opened 2024-09-19
OpenAPI is not working for data generation using Ollama http://localhost:11434


aseelert:
**Describe the bug**
```bash
ilab model serve --endpoint-url http://localhost:11434
````
is not working with Ollama. According the Ollama/WebUI documentation the curl commands are working as expected. I tried also with API key, bu

```bash
sudo docker run -d \
--network=host \
--gpus=all \
-v ollama:/root/.ollama \
-v open-webui:/app/backend/data \
--name open-webui \
--restart=always \
ghcr.io/open-webui/open-webui:ollama
```

**To Reproduce**
```bash
ilab data generate --pipeline full --gpus 1 --num-cpus 16 --sdg-scale-factor 500 \
--endpoint-url http://127.0.0.1:11434 --api-key 500sk-0a726e394d28462a8f891c2f83777c4f  \
--enable-serving-output --output-dir ./outputtest --batch-size 0
```

**Error***
```bash
Generating synthetic data using 'full' pipeline, '/home/itzuser/.cache/instructlab/models/mixtral-8x7b-instruct-v0.1.Q3_K_M.gguf' model, '/home/itzuser/.local/share/instructlab/taxonomy' taxonomy, against http://127.0.0.1:11434 server
Traceback (most recent call last):
  File "/home/itzuser/instructlab/venv/lib64/python3.11/site-packages/instructlab/sdg/pipeline.py", line 196, in _generate_single
    block = block_type(self.ctx, self, block_name, **block_config)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/itzuser/instructlab/venv/lib64/python3.11/site-packages/instructlab/sdg/llmblock.py", line 88, in __init__
    self.server_supports_batched = server_supports_batched(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/itzuser/instructlab/venv/lib64/python3.11/site-packages/instructlab/sdg/llmblock.py", line 43, in server_supports_batched
    response = client.completions.create(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/itzuser/instructlab/venv/lib64/python3.11/site-packages/openai/_utils/_utils.py", line 274, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2839, opened 2024-12-26
add a default value for endpoint url option

reidliu41:
Some users might not know the default endpoint url is `url` + `v1` = `http://127.0.0.1:8000/v1`, so add it in description.

**Checklist:**

### DIFF
--- a/src/instructlab/model/chat.py
+++ b/src/instructlab/model/chat.py
@@ -124,7 +124,7 @@ def is_openai_server_and_serving_model(
 @click.option(
     "--endpoint-url",
     type=click.STRING,
-    help="Custom URL endpoint for OpenAI-compatible API. Defaults to the `ilab model serve` endpoint.",
+    help="Custom URL endpoint for OpenAI-compatible API. Defaults to the `ilab model serve` endpoint (http://127.0.0.1:8000/v1).",
 )
 @click.option(
     "--api-key",


src/instructlab/model/chat.py
```diff
@@ -124,7 +124,7 @@ def is_openai_server_and_serving_model(
 @click.option(
     "--endpoint-url",
     type=click.STRING,
-    help="Custom URL endpoint for OpenAI-compatible API. Defaults to the `ilab model serve` endpoint.",
+    help="Custom URL endpoint for OpenAI-compatible API. Defaults to the `ilab model serve` endpoint (http://127.0.0.1:8000/v1).",
```
RobotSail, line 127: Is this something that gets read from the config if not passed in as a flag? If so, what happens if it's a different value?

RobotSail (commented):
I like the idea behind this, though we have some established ways of presenting defaults. It might be worth exploring those options as well.

src/instructlab/model/chat.py
reidliu41: yes, it(endpoint_url) will read config file. It based on `host` and `port`. If value like --port 1234, it need to specify --endpoint-url.

reidliu41: 1. But I check,  if the option set it read from config, it should read it from `configuration`, not config file(please correct me if I am wrong), because change the value from `config file`, it will not change value from `--help`.
e.g.
```
default:
@click.option(
    "--gpu-layers",
    type=click.INT,
    cls=clickext.ConfigOption,
    config_sections="llama_cpp",
)

  --gpu-layers INTEGER        Number of model layers to offload to GPU. -1
                              means all layers. [required; default: -1; <<<<<<---
                              config: 'serve.llama_cpp.gpu_layers']

$ ilab config show -k serve.llama_cpp -wc
gpu_layers: -1     <<<<<<---
llm_family: ''
max_ctx_size: 4096

change:
$ ilab config show -k serve.llama_cpp -wc
gpu_layers: -2  <<<<<<---
llm_family: ''
max_ctx_size: 4096
  --gpu-layers INTEGER        Number of model layers to offload to GPU. -1
                              means all layers. [required; default: -1; <<<<<<---
                              config: 'serve.llama_cpp.gpu_layers']
```

2. So based on this default value, the default endpoint_url value in `--help` should based on `host: 127.0.0.1` and `port: 8000` in the config field, unless change it. If it need to change, seems other also need to change to read it from `config file`.

reidliu41:
> I like the idea behind this, though we have some established ways of presenting defaults. It might be worth exploring those options as well.

Yeah, sometimes it might be confused for the users, seems only the developer who created this option and he/she should very clear about this option/function. So make some simple tips, user can easy to use it, even a word.

src/instructlab/model/chat.py
reidliu41: hi @RobotSail any suggestions that I need to modify? Actually, my purpose is simple just share the api example, may try to avoid the situation like https://github.com/instructlab/instructlab/issues/2287

courtneypacheco: @RobotSail any thoughts?

Issue #2777, opened 2024-12-12
Make running vLLM optional when running MMLU evaluations


RobotSail:
**What is the current experience of ilab you've found issue with?**

The instructlab/eval library supports running MMLU without a server by directly evaluating the HuggingFace model. This avoids the requirement of the user needing to rely on vLLM for doing evaluations, which may sometimes be desired if unwanted behavior arises.

**How do you think the described experience should be?**

There could be an optional flag such as `--no-vllm` or maybe `--with-hf` indicating to run without vLLM or just with the huggingface library (instead of vLLM).

This would prevent the vLLM server from being created.

**What is the benefit of the change you are proposing?**

This way users can run evaluation without vLLM if they just want to see how the model performs when the raw model is loaded.

**If applicable, how would backwards-compatability be handled?**

N/a

**Additional context**

The instructlab/eval supports this but the CLI doesn't. Not that the CLI needs to support everything just because a library does, but this seems like a nice UX feature to have.

reidliu41:
https://github.com/instructlab/instructlab/pull/2844

PR #2844, opened 2024-12-30
add skip-server option for evaluate

reidliu41:
Resolves https://github.com/instructlab/instructlab/issues/2777

1. add new option
```
$ ilab model evaluate --help
Usage: ilab model evaluate [OPTIONS]
  -ss, --skip-server              Skip launching the server and evaluate
                                  directly with the HuggingFace model. This
                                  option supports mmlu and mmlu_branch
                                  benchmarks.
```
2. refactor the launch_server request duplicate parts

**Checklist:**

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -52,6 +52,7 @@ def evaluate_model(
     tls_client_key,  # pylint: disable=unused-argument
     tls_client_passwd,  # pylint: disable=unused-argument
     enable_serving_output,
+    skip_server: bool,
 ):
     """Evaluates a trained model"""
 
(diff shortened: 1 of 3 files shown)


src/instructlab/model/evaluate.py
```diff
@@ -516,6 +516,12 @@ def launch_server(
     is_flag=True,
     help="Print serving engine logs.",
 )
+@click.option(
+    "-ss",
```
RobotSail, line 520: Does this work? Click should allow multiple flags like `-cl` vs. `-c -l`, so how does it know this is `-ss` vs. `-s -s`?
We should probably avoid using multiple character flags like this to avoid confusion.

src/instructlab/model/evaluate.py
```diff
@@ -516,6 +516,12 @@ def launch_server(
...
+    "--skip-server",
+    is_flag=True,
+    help="Skip launching the server (without using backend e.g. vLLM or llama-cpp) and evaluate directly with the HuggingFace model. This option supports mmlu and mmlu_branch benchmarks.",
```
RobotSail, line 523: MMLU isn't supported by `llama-cpp` since we can't read the logprobs, so we can get rid of this part

```suggestion
    help="Skip launching the server and evaluate directly with the HuggingFace model. This option supports mmlu and mmlu_branch benchmarks.",
```

src/instructlab/model/evaluate.py
```diff
@@ -799,25 +806,28 @@ def evaluate(
...
-                    max_workers=None,
-                    gpus=gpus,
-                    backend=backend,
-                    enable_serving_output=enable_serving_output,
-                )
-                overall_score, individual_scores = evaluator.run(api_base)
-            finally:
-                if server is not None:
-                    server.shutdown()
+            if skip_server:
```
RobotSail, line 809: Since the complexity of this portion is growing but we're duplicating the same block in multiple places, we should extract this into a separate function to make it more readable/maintainable.

reidliu41:
> Thanks for the PR @reidliu41 ! I left some comments but otherwise this looks good!

thanks, will check later

src/instructlab/model/evaluate.py
reidliu41: thanks for remind, removed `-ss`

src/instructlab/model/evaluate.py
reidliu41: updated

src/instructlab/model/evaluate.py
RobotSail: No you don't need a separate PR for this. You can just extract your the entirety of this try/catch structure that you're changing which is common to both portions and place it into a single function.

RobotSail (approved):
LGTM

nathan-weinberg (approved):

PR #2863, opened 2025-01-07
update the train test result

reidliu41:
For https://github.com/instructlab/instructlab/pull/2741#issuecomment-2573960090

**Checklist:**

- [ ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Functional tests have been added, if necessary.
- [ ] E2E Workflow tests have been added, if necessary.

PR #2865, opened 2025-01-07
chore!: Update PyTorch to 2.5

### LINKED ISSUE #2864: Allow PyTorch 2.5
The currently supported version of PyTorch is 2.5. It would be great to support it in instructlab.
Moreover, updating AMD ROCm to 6.3 and Intel Gaudi to 1.19 requires PyTorch 2.5.1.

fabiendupont:
This change increases the upper version of PyTorch to allow version 2.5. For the AMD variant, it also switches to ROCm 6.2, which is required for PyTorch 2.5.

Resolves #2864

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 * A new command `ilab model upload` has been introduced so users can now upload their trained models to [Hugging Face](https://huggingface.co/) via the `ilab` CLI
 * `ilab model serve` now has separate `--host` and `--port` options, replacing the `host_port` configuration. The default values are `127.0.0.1` for `--host` and `8000` for `--port`, allowing users to configure the server's binding address and port independently through the configuration file or command-line flags.
+* Extends PyTorch support to version 2.5
 
 ## v0.22
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ sentencepiece>=0.2.0
 tokenizers>=0.11.1
 toml>=0.10.2
 # Default version. Can be overridden in extra requirements
-torch>=2.3.0,<2.5.0
+torch>=2.3.0,<2.6.0
 tqdm>=4.66.2
 # 'optimum' for Intel Gaudi needs transformers <4.44.0,>=4.43.0
 transformers>=4.41.2
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -299,7 +299,7 @@ pip_install_with_nvidia() {
 
 pip_install_with_amd() {
     "${BASH_SOURCE[0]}" "$CLOUD_TYPE" ssh -n "$INSTANCE_NAME" "pushd instructlab && source venv/bin/activate && pip cache remove llama_cpp_python && \
-    pip install -e .[rocm] --extra-index-url https://download.pytorch.org/whl/rocm6.0 \
+    pip install -e .[rocm] --extra-index-url https://download.pytorch.org/whl/rocm6.2 \
    -C cmake.args='-DGGML_HIPBLAS=on' \
    -C cmake.args='-DAMDGPU_TARGETS=all' \
    -C cmake.args='-DCMAKE_C_COMPILER=/opt/rocm/llvm/bin/clang' \


nathan-weinberg:
@fabiendupont have you done any kind of testing around this?

fabiendupont:
@nathan-weinberg, I have used our downstream build pipeline to test instructlab v0.22.1 with PyTorch 2.5.1 manually. I was able to run chat, serve, data generate and train steps without any issue. That's why I went ahead to propose this MR, as it seems pretty safe. Do you have any specific concern?

nathan-weinberg:
@fabiendupont if you look at the existing issues (I linked them in the additional one you opened, I think that may be a dup), the reason we've been holding off is we want to make sure the Training library (@JamesKunstle) is functional with this version before we made the bump here 

@prarit curious around your thoughts on this as well

fabiendupont:
@nathan-weinberg, the training library doesn't seem to cap the PyTorch version, so I would have expected it to have already been tested with PyTorch 2.5.1 during a previous PR in instructlab/training.

@JamesKunstle, is there somewhere where the PyTorch version is controlled besides requirements file?

JamesKunstle:
@fabiendupont No it isn't capped elsewhere but we don't have an independent test quite yet to run through everything w/ a higher torch version- that'll be ready in a bit. The 'instructlab/instructlab' tests should be good enough to roughly confirm it though.

nathan-weinberg:
I'm going to trigger a couple E2E jobs on this (Large and XLarge) just as a sanity check

prarit:
@nathan-weinberg LGTM

nathan-weinberg:
Okay, if the CI jobs pass I'm fine to approve this - @JamesKunstle can you approve as well if you are signing off, which it sounds like you are?

JamesKunstle (approved):
Since all our CI succeeded w/ torch<2.6 I think we're okay to bump it.

nathan-weinberg:
Yeah let me rerun it here

nathan-weinberg:
Weird, these E2E failures should have been fixed by https://github.com/instructlab/instructlab/pull/2863

@bbrowning @cdoern @reidliu41 did we miss something?

cdoern:
no we didn't because my e2e runs are getting past this point. Has this PR been rebased properly?

bbrowning:
I believe @cdoern is right - this PR will need to rebase on top of the latest main to pick up the fixes for those E2E tests, since the fix was also in this repository. Since the launched jobs are not using a `pull_request_target` dispatch, when they are manually triggered they are running directly off the contents of this PR as opposed to using the workflow files from the target branch.

nathan-weinberg:
@Mergifyio rebase

reidliu41:
Seems still not update? The `grep` still old.
```
ᕦ(òᴗóˇ)ᕤ Accelerated model training completed successfully! ᕦ(òᴗóˇ)ᕤ
Best final checkpoint: /tmp/tmp.uAmOJhmRbo/.local/share/instructlab/skills-only/phase2/checkpoints/hf_format/samples_662 with score: 7.894736842105263
Journal: /tmp/tmp.uAmOJhmRbo/.local/share/instructlab/skills-only/journalfile.yaml
+ grep -o '/[^ ]*'
+ grep 'Training finished! Best final checkpoint: ' <<<<<<<<<<<<-------- /tmp/tmp.uAmOJhmRbo/skills_only_training.log
+ rm -rf /tmp/tmp.uAmOJhmRbo
```

bbrowning:
It seems we ran out of disk space on the e2e-xlarge test kicked off? From its logs at https://github.com/instructlab/instructlab/actions/runs/12659779418/job/35279654906 - 
```2025-01-08T00:55:49.8129077Z ##[warning]You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 9 MB
```

However, we did get past the final checkpoint grep issue. So, the question is did PyTorch 2.5 increase our disk space usage? Or is this just a flake of that e2e test setup?

bbrowning:
Looking at the e2e-xlarge-test history, I see it has never passed on main so the fact that it failed may not be surprising. It did get past the point of training a model at least, so that gives some indication that PyTorch 2.5 is working properly on that setup.

nathan-weinberg (approved):
Given the large job passed with no issue and the various other approvals/signoffs, I am going to go ahead and approve this.

bbrowning:
So, looking at the logs, I'm not sure this actually ran any CI run with Torch 2.5.x. CI was green with this change, but I'm still seeing multiple references to torch 2.4.x in the CI logs when installing ilab. Are we sure we have confidence torch 2.5.x works properly if CI isn't actually using torch 2.5.x in our tests?

PR #2885, opened 2025-01-08
fix: Change temporary write directory in all e2e CI jobs from `tmpfs` to `/home/tmp`

### LINKED ISSUE #2866: CI Enhancement: Change default write location for temporary files for cloud cost savings (and potentially reduce EBS size)
Medium e2e CI jobs were occasionally exiting prematurely due to insufficient disk space:

Image

We can see that the temporary file in the output above was created using `SpooledTemporaryFile` class through `gguf_writer`: https://github.com/ggerganov/llama.cpp/blob/a0ec17b32ec6077f5ca22fe833ebdc9b86795a4d/gguf-py/gguf/gguf_writer.py#L65C24-L65C45 

The `SpooledTemporaryFile` class referenced in the above link buffers the data in memory before writing, but the class doesn’t appear to have any hard, pre-defined limits, so the size of the file being created doesn’t appear to be an issue.

The insufficient memory issue is likely related to the `tmpfs` default configurations, which state that `tmpfs` can use up to 50% of your available physical memory: https://docs.kernel.org/filesystems/tmpfs.html 

We currently allocate 800G of physical space for our medium jobs, but when we use the `tmpfs` defaults, we realistically can only use up to 400G of that available memory. At the same time, the README states that we need a minimum of 250G of available space, even though it actually recommends 500G: https://github.com/instructlab/instructlab/blob/92099a58869a416548240a4f3abaced5a0958937/README.md?plain=1#L98

So if we can only use 400G out of 800G, and some of that memory is already consumed by systemd and other processes, we realistically probably don't even have the full 400G available to us. (When I manually replicated the medium runner instance with the same AMI + launch configs and SSH'd into it, 17G was already taken up under `tmpfs`.) This could possibly explain the intermittent disk space failures.

Because the `tmpfs` default configurations are likely the root cause of the issue, we should change the default write location from `tmpfs` (i.e., `/tmp/<some-hash>`) to another directory that doesn't have these restrictions.

Alternative solutions that are _not_ ideal and why:
  1. Increase the the available EBS size from 800G to at least 1TB
     - This is not ideal because it will increase cloud storage costs and circumvents the issue

  2. Override `tmpfs` defaults to increase the usable memory
     - If we allow `tmpfs` to utilize more than 50% of the available memory, we could run into deadlock issues:  "If you oversize your tmpfs instances the machine will deadlock since the OOM handler will not be able to free that memory." (Source:  https://docs.kernel.org/filesystems/tmpfs.html)

**Additional context**
We may also want to reduce the EBS size. We only need 500G of the 800G we have available to us. I presume the EBS size was increased to 800G to avoid OOM issues, but we may be able to work with 550G or 600G.

courtneypacheco:
Resolves #2866

**Checklist:**

Keep in mind that while PR _does_ resolve #2866, which only mentions the medium e2e job having OOM issues, we have started to see OOM issues in the large e2e jobs this morning under certain conditions. So as a preventative measure, I've updated all default temporary write directories from `/tmp` to `/home/tmp`.

Also note: In order to successfully test these workflow changes in a pull request, I had to temporarily modify the trigger conditions in the workflow files. I've since reverted those trigger conditions, so you don't see the changes in this PR reflected in the current logging.

Original line of code that I have since removed:
https://github.com/instructlab/instructlab/commit/ae08713cbb44bd70786b5e8f2719413bc8719701#diff-f84f57a944419bb3920eea214c71890696e3d711cf15fcdcde1bb0e3bcbe5737R11

Here is the previous run from the same medium e2e job when I had the temporary trigger conditions in place: https://github.com/instructlab/instructlab/actions/runs/12677138707/job/35331915188

You can confirm by viewing the "Install Packages" and "Run e2e test" steps under the `e2e-medium-test` job linked above. You will see things like:
Screenshot 2025-01-08 at 4 13 04 PM
Screenshot 2025-01-08 at 4 13 15 PM
Screenshot 2025-01-08 at 4 13 31 PM

^ which includes my `/home/tmp` changes.

See rest of logs for more details

### DIFF
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -29,6 +29,7 @@ concurrency:
 
 env:
   LC_ALL: en_US.UTF-8
+  TMPDIR: /home/tmp
 
 defaults:
   run:
@@ -99,6 +100,7 @@ jobs:
       - name: Install Packages
         run: |
           cat /etc/os-release
+          mkdir -p /home/tmp
           sudo dnf install -y gcc gcc-c++ make git python3.11 python3.11-devel
 
       - name: Checkout
--- a/.github/workflows/e2e-nvidia-l40s-x4.yml
+++ b/.github/workflows/e2e-nvidia-l40s-x4.yml
@@ -12,6 +12,9 @@ on:
         required: true
         default: 'main'
 
+env:
+  TMPDIR: /home/tmp
+
 jobs:
   start-large-ec2-runner:
     runs-on: ubuntu-latest
@@ -57,6 +60,7 @@ jobs:
       - name: Install Packages
         run: |
           cat /etc/os-release
+          mkdir -p /home/tmp
           sudo dnf install -y gcc gcc-c++ make git python3.11 python3.11-devel
 
       - name: Checkout
--- a/.github/workflows/e2e-nvidia-l40s-x8.yml
+++ b/.github/workflows/e2e-nvidia-l40s-x8.yml
@@ -12,6 +12,9 @@ on:
         required: true
         default: 'main'
 
+env:
+  TMPDIR: /home/tmp
+
 jobs:
   start-xlarge-ec2-runner:
     runs-on: ubuntu-latest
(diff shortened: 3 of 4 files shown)


nathan-weinberg (approved):

n1hility:
LGTM the one thing to look out for is if this adds a time delay of any significance. I assume not but there is just the potential for lots safe of smaller file activity to incur a delay from moving from ram to ebs iops. Once we move to local ssd I think the risk is completely gone so even if there is a temporary delay it’s probably acceptable.

n1hility (approved):

PR #2892, opened 2025-01-09
Update vLLM to 0.6.6.post1

### LINKED ISSUE #2702: outdated (0.62.) vllm causes dependency hell
Please, put vllm 0.64 instead of current 0.62, as 0.62 causes dependency hell. Also change from the requirements to build 0.62 yourself to existing pip version

fabiendupont:
This PR is a follow-up to #2865 that relaxed the PyTorch version range. Even with that range extension, we realized that PyTorch 2.4 is still used when installing `instructlab[vllm-cuda]`, because vLLM 0.6.2 has a requirement on PyTorch 2.4.

This new PR updates the version of vLLM to 0.6.6.post1, which is the latest available in the Open Data Hub fork of vLLM. The vLLM changelog doesn't highlight much risk in this version bump.

Resolves #2702

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## v0.24
+
+### Breaking Changes
+
+### Features
+
+- Update vLLM to version 0.6.6.post1. As a requirement for this new vLLM version, PyTorch is updated to 2.5.1
+
 ## v0.23
 
 ### Breaking Changes
--- a/requirements-vllm-cuda.txt
+++ b/requirements-vllm-cuda.txt
@@ -2,4 +2,4 @@
 # Dependencies for installing vLLM on CUDA
 
 # vLLM only supports Linux platform (including WSL)
-vllm @git+https://github.com/opendatahub-io/vllm@v0.6.4.post1 ; sys_platform == 'linux' and platform_machine == 'x86_64'
+vllm @git+https://github.com/opendatahub-io/vllm@v0.6.6.post1 ; sys_platform == 'linux' and platform_machine == 'x86_64'


nathan-weinberg:
We'll need a new SDG z-stream that allows for a higher Torch version to resolve the dependency issue that will allow us to test this - that backport is being worked on here: https://github.com/instructlab/sdg/pull/469

bbrowning:
SDG v0.6.3 has been released - the only change from v0.6.2 (used by instructlab currently) is the raising of this PyTorch upper version limit, which should unblock this PR.

nathan-weinberg:
@fabiendupont can you rebase this? we'll see if CI is able to complete the install and then we can test if the new vLLM package will work

bbrowning:
Do we need to manually kick off a large or xlarge CI test for this so that vllm actually gets used? I think both the small and medium e2e tests use llama_cpp, right?

nathan-weinberg:
> Do we need to manually kick off a large or xlarge CI test for this so that vllm actually gets used? I think both the small and medium e2e tests use llama_cpp, right?

Small is using only llama_cpp, Medium actually uses both: https://github.com/instructlab/instructlab/blob/main/docs/ci.md/#e2e-test-coverage-matrix

Now `ilab` was successfully installed here with the Medium job - that's good! But to be safe, we should definitely kick off at least a Large job for some functional confidence - I will do that right now.

nathan-weinberg (approved):
CI suggest no functional regressions from this vLLM bump, so I am approving from my side

n1hility:
We have some users using a combo with 0.6.4.post1, would love to see a CI test run pass on that.

nathan-weinberg:
> We have some users using a combo with 0.6.4.post1, would love to see a CI test run pass on that.

Do you think we should be updating to that version instead? If not, can we run that in a separate PR so we can get this in as early before the 0.23.0 release as we can?

fabiendupont:
@n1hility, @nathan-weinberg, I have created https://github.com/instructlab/instructlab/pull/2910 that updates vLLM to 0.6.4.post1. This will allow us to run the CI tests and be more comfortable with the bigger leap forward to 0.6.6.post1.

alinaryan (approved):

PR #2899, opened 2025-01-09
feat: add S3 support to `ilab model upload`

### LINKED ISSUE #2671: Implement command to push Models to S3

nathan-weinberg:
Resolves #2671

**Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1047,9 +1047,25 @@ argument to specify your new model:
 
 If you are happy with your newly-trained model and wish to upload it to a supported endpoint, you can do so with the `ilab model upload` command.
 
-InstructLab currently supports uploading to [Hugging Face](https://huggingface.co/) and OCI registires such as [Quay.io](https://quay.io/), with plans to support S3 uploads as well.
+InstructLab currently supports uploading to [Hugging Face](https://huggingface.co/), OCI registries such as [Quay.io](https://quay.io/), and [AWS S3](https://aws.amazon.com/s3/) buckets.
 
 You can upload safetensors, GGUFs, and/or OCI-compliant models to any supported destination you have access to - see `ilab model upload --help` for more info.
+Some examples are listed below:
+
+```shell
+# upload model to Hugging Face
+ilab model upload --dest-type hf --model granite-7b-lab --destination instructlab/granite-7b-lab --release main
+
+# upload model to OCI registry
+ilab model upload --dest-type oci --model granite-7b-lab --destination docker://instructlab/granite-7b-lab --release latest
+
+# upload model to S3 bucket
+ilab model upload --dest-type s3 --model granite-7b-lab --destination my-aws-bucket
+
+```
+
+> [!NOTE]
+> You do not need to use the `--release` flag if you are uploading to AWS S3
 
 ## 🎁 Submit your new knowledge or skills
 
--- a/README.md
+++ b/README.md
@@ -1065,7 +1065,7 @@ ilab model upload --dest-type s3 --model granite-7b-lab --destination my-aws-buc
 ```
 
 > [!NOTE]
-> You do not need to use the `--release` flag if you are uploading to AWS S3
+> You do not need to use the `--release` flag if you are uploading to AWS S3, and you should not prepend your bucket name with `s3://`
 
 ## 🎁 Submit your new knowledge or skills
 
--- a/src/instructlab/model/upload.py
+++ b/src/instructlab/model/upload.py
@@ -10,6 +10,7 @@ import subprocess
 
 # Third Party
 from huggingface_hub import HfApi
+import boto3
 
 # First Party
 from instructlab.configuration import DEFAULTS
@@ -19,6 +20,7 @@ from instructlab.utils import (
     is_model_gguf,
     is_model_safetensors,
     is_oci_repo,
+    is_s3_repo,
 )
 
 logger = logging.getLogger(__name__)
@@ -45,14 +47,19 @@ class InvalidDestinationForUpload(Exception):
 
     Attributes
         destination        destination of upload
-        release            release of upload
+        release            release of upload (if applicable)
 
     """
 
-    def __init__(self, destination, release) -> None:
+    def __init__(self, destination, release=None) -> None:
         self.destination = destination
         self.release = release
-        super().__init__(f"Cannot upload model to {destination}@{release}")
+        self.msg = (
+            f"Cannot upload model to {destination}@{release}"
+            if release
+            else f"Cannot upload model to {destination}"
+        )
+        super().__init__(self.msg)
 
 
 class HFUploadError(Exception):
@@ -91,6 +98,23 @@ class OCIUploadError(Exception):
         )
 
 
+class S3UploadError(Exception):
+    """
+    Error raised when an InstructLab call to S3 for model uploading fails
+
+    Attributes
+        path        filepath of model location
+        s3_exc      S3 client exception
+    """
+
+    def __init__(self, path, s3_exc) -> None:
+        self.path = path
+        self.s3_exc = s3_exc
+        super().__init__(
+            f"Uploading model at {self.path} failed with the following S3 client error:\n{self.s3_exc}"
+        )
+
+
 class ModelUploader(abc.ABC):
     """Base class for a model uploading backend"""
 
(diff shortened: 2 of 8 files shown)


nathan-weinberg:
cc @mbestavros would value your input on this one as well!

src/instructlab/model/upload.py
```diff
@@ -276,3 +298,91 @@ def upload_oci(self) -> None:
...
+        # AWS cred check
+        try:
+            sts_client = boto3.client("sts")
+            sts_client.get_caller_identity()
+        except Exception as exc:
+            raise ValueError(
+                """AWS credentials need to be set in your environment to upload to S3.
+                You can set your credentials using the AWS CLI
+                For alternative methods to set credentials, please review the boto3 documentation"""
+            ) from exc
```
courtneypacheco, line 330: So... there are many reasons one may not be able to authenticate with AWS, one of them being unset credentials, but another (for example) being invalid/expired credentials. (Some people have credentials automatically rotated, sometimes as often as every 30 mins.)

Also, the recommended approach is to use an AWS credentials file to make the credentials automatically load.

```suggestion
            raise ValueError(
                """AWS credentials need to be set in your environment to upload to S3. 
                If you have not already set your credentials, you can set them using the 
                AWS CLI or through an AWS credentials file. 
                If you have set your credentials, please ensure they are valid and not expired.
                For alternative methods to set credentials, please review the boto3 documentation."""
            ) from exc
```

README.md
```diff
@@ -1047,10 +1047,13 @@ argument to specify your new model:
 
 If you are happy with your newly-trained model and wish to upload it to a supported endpoint, you can do so with the `ilab model upload` command.
 
-InstructLab currently supports uploading to [Hugging Face](https://huggingface.co/) and OCI registires such as [Quay.io](https://quay.io/), with plans to support S3 uploads as well.
+InstructLab currently supports uploading to [Hugging Face](https://huggingface.co/), OCI registries such as [Quay.io](https://quay.io/), and [AWS S3](https://aws.amazon.com/s3/) buckets.
```
courtneypacheco, line 1050: Can you add examples later on in the README showing how to upload to S3?

I only mention this because S3 URLs commonly take on one of two different syntaxes:

 1. Path style: `s3.amazonaws.com/<bucket>/<key>`
 2. Virtual hosted style: `<bucket>.s3.amazonaws.com/<key>`

I've also used global s3 URLs (basically ones that start with `s3://` instead of `http://`) and access points.

If we don't want `s3://` being used, then we should definitely prohibit it.

src/instructlab/utils.py
```diff
@@ -602,6 +602,17 @@ def ensure_legacy_dataset(
...
 
+def is_s3_repo(repo_url: str) -> bool:
+    """
+    Checks if a provided repository follows the S3 bucket URL syntax
+    """
+    s3_url_regex = r"^[a-zA-Z0-9.\-_]{1,255}$"
+    if not re.match(s3_url_regex, repo_url):
+        logger.warning(f"Invalid S3 bucket URL: {repo_url}")
+        return False
+    return True
```
courtneypacheco, line 613: Do we want to allow all formats? Specifically `s3://` formats, etc.? If not, let's block the ones we won't allow by returning `False`

src/instructlab/model/upload.py
```diff
@@ -276,3 +298,91 @@ def upload_oci(self) -> None:
...
+            f"Uploading model to S3 endpoint:\n{DEFAULT_INDENT}Model: {self.model}\n{DEFAULT_INDENT}Destination: {self.destination}\n"
+        )
+
+        # determine if model is path or name - look in checkpoints dir if the latter
+        self.is_model_path_or_name()
+
+        # raise exception if user did not provide valid S3 bucket endpoint
+        if not is_s3_repo(self.destination):
+            logger.error(
+                f"\nInvalid destination supplied:\n{DEFAULT_INDENT}Please specify valid S3 bucket URL syntax via --destination",
```
courtneypacheco, line 342: minor nitpick for clarity. (sometimes when there are a lot of logs printed out, it can be difficult for users to debug issues.)

```suggestion
                f"\nInvalid S3 destination supplied:\n{DEFAULT_INDENT}Please specify valid S3 bucket URL syntax via --destination",
```

courtneypacheco (commented):
Had a few questions

src/instructlab/utils.py
nathan-weinberg: No, we only want to allow for the exact bucket name (at least for now)

README.md
nathan-weinberg: Yes, added some examples!

src/instructlab/model/upload.py
nathan-weinberg: Done!

src/instructlab/model/upload.py
nathan-weinberg: Done!

src/instructlab/utils.py
courtneypacheco: ok, so if users have to pass in the actual bucket name and not a URL, then the warning message could get a little confusing.

i'd suggest instead:

```suggestion
    s3_bucket_name_regex = r"^[a-zA-Z0-9.\-_]{1,255}$"
    if not re.match(s3_bucket_name_regex, bucket_name):
        logger.warning(f"Invalid S3 bucket name: {bucket_name}")
        return False
    return True
```

README.md
```diff
@@ -1047,9 +1047,25 @@ argument to specify your new model:
...
+# upload model to Hugging Face
+ilab model upload --dest-type hf --model granite-7b-lab --destination instructlab/granite-7b-lab --release main
+
+# upload model to OCI registry
+ilab model upload --dest-type oci --model granite-7b-lab --destination docker://instructlab/granite-7b-lab --release latest
+
+# upload model to S3 bucket
+ilab model upload --dest-type s3 --model granite-7b-lab --destination my-aws-bucket
+
+```
```
courtneypacheco, line 1065: this is good!

I would add a note below telling people to use the exact bucket name as it appears in AWS, to not pass in any kind of URL (like `s3://` or `http://`).

it could get confusing when people see the above example:

```
--destination docker://instructlab/granite-7b-lab
```

and they may think to use `s3://` or something

mbestavros (approved):
Looks good to me! :heavy_check_mark:

src/instructlab/utils.py
nathan-weinberg: Sure, done

README.md
nathan-weinberg: Sure, done

alinaryan (approved):
I saw this live so LGTM

src/instructlab/model/upload.py
```diff
@@ -276,3 +298,93 @@ def upload_oci(self) -> None:
...
+                Key=self.model,
+            )
+            logger.debug(f"S3 Client response:\n{resp}")
+        except Exception as exc:
+            raise S3UploadError(self.local_model_path, exc) from exc
+        logger.info(
+            f"\nUploading model at {self.local_model_path} succeeded!",
+        )
+
+    def upload_folder(self) -> None:
```
cdoern, line 371: so no docstring?

src/instructlab/model/upload.py
```diff
@@ -276,3 +298,93 @@ def upload_oci(self) -> None:
...
+
+        # upload either file or folder
+        if self.local_model_path.is_file():
+            self.upload_file()
+        elif self.local_model_path.is_dir():
+            self.upload_folder()
+        else:
+            raise InvalidModelForUpload(self.local_model_path)
+
+    def upload_file(self) -> None:
```
cdoern, line 356: docstring here too please

cdoern (approved):
just a few docstring changes, but won't block

src/instructlab/model/upload.py
nathan-weinberg: added sir

src/instructlab/model/upload.py
nathan-weinberg: done

PR #2902, opened 2025-01-10
feat(rag): Add lab rag convert CLI command

### LINKED ISSUE #2890: [RAG][Dev] Implement ilab rag convert
As per [this dev doc](https://github.com/instructlab/dev-docs/pull/161), `ilab rag convert` will process documents or taxonomy via docling to prepare them for ingestion.

Acceptance Criteria
* `ilab rag convert` command group is defined and has the options specified in the dev doc
* Unit tests

jwm4:
Adds new `ilab rag convert` command which takes in input documents from a taxonomy or a directory of original document files (e.g., PDFs) and outputs Docling JSON for each input document.

Resolves #2890

**Dev Docs related to this Pull Request:**
Link to Dev Doc or PR: https://github.com/instructlab/dev-docs/pull/161

**Checklist:**


src/instructlab/rag/process_docs.py
```diff
@@ -0,0 +1,108 @@
...
+    Latest version of docling schema is used (currently, v2).
+    """
+    logger.info(f"Processing {input_dir} to {output_dir}")
+
+    clear_directory(Path(output_dir))
+
+    source_files = _load_source_files(input_dir=input_dir)
+    logger.info(f"Transforming source files {[p.name for p in source_files]}")
+
+    doc_converter = DocumentConverter()
```
bbrowning, line 50: When we use docling in SDG, we have some logic to automatically detect which OCR engine to use and configure Docling appropriately. You will likely need similar steps here, as Docling by default attempts to use EasyOCR which downloads model weights from GitHub behind the scenes but some downstream use-cases use alternative OCR implementations and should not reach out to download blobs from GitHub while converting documents.

See https://github.com/instructlab/sdg/blob/v0.6.3/src/instructlab/sdg/utils/chunkers.py#L261-L270 for how we do that today.

jwm4: Sounds good.  I will make that change for OCR.

I see earlier in the same code it also uses `docling_model_path`.  @bbrowning or @khaledsulayman do either of you know what's going on with `docling_model_path`?  From what I can tell, it is currently [hard-coded to None](https://github.com/instructlab/sdg/blob/725e5ab810900ce83b53d63ca9b2ebda42c2bf56/src/instructlab/sdg/generate_data.py#L211C1-L211C30) in `_sdg_init`, and I can't find any code path that sets it to anything else.  So I guess it is just a placeholder for an actual configuration option to come in the future.  Is that right or am I missing something?

src/instructlab/rag/process_docs.py
dmartinol, line 1: should we rename the module to `convert.py`, for consistency with other commands?

src/instructlab/cli/rag/convert.py
```diff
@@ -0,0 +1,85 @@
...
+
+@click.command()
+@click.option(
+    "--input_dir",
+    required=False,
+    type=click.Path(exists=True, file_okay=False, dir_okay=True, readable=True),
+    default=None,
+    help="The folder with user documents to process. In case it's missing, the knowledge taxonomy files will be processed instead.",
+)
+@click.option(
```
dmartinol, line 29: shouldn't we also allow to configure the taxonomy flags with `clickext.ConfigOption`? of course the best would be to reuse the settings under the `generate` section, but we know this is not doable ATM.

src/instructlab/cli/rag/convert.py
```diff
@@ -0,0 +1,85 @@
...
+                taxonomy_path = DEFAULTS.TAXONOMY_DIR
+        if taxonomy_base is None:
+            taxonomy_base = ctx.obj.config.generate.taxonomy_base
+            if taxonomy_base is None:
+                taxonomy_base = DEFAULTS.TAXONOMY_BASE
+
+        logger.info(
+            f"Pre-processing latest taxonomy changes at {taxonomy_path}@{taxonomy_base}"
+        )
+        process_docs_from_taxonomy(
```
dmartinol, line 75: since we renamed the command from `process` to `convert`, shouldn't we also rename the processing functions accordingly?

src/instructlab/rag/process_docs.py
```diff
@@ -0,0 +1,156 @@
...
+from docling.models.tesseract_ocr_model import TesseractOcrModel  # type: ignore
+
+# First Party
+from instructlab.rag.taxonomy_utils import lookup_knowledge_files
+from instructlab.utils import clear_directory
+
+logger = logging.getLogger(__name__)
+
+
+def process_docs_from_taxonomy(taxonomy_path, taxonomy_base, output_dir):
```
dmartinol, line 30: I imagine that @cdoern will ask for more doc strings in all these functions (I agree tbh)

src/instructlab/rag/process_docs.py
```diff
@@ -0,0 +1,156 @@
...
+        knowledge_files = lookup_knowledge_files(taxonomy_path, taxonomy_base, temp_dir)
+        logger.info(f"Found {len(knowledge_files)} knowledge files")
+        logger.info(f"{knowledge_files}")
+
+        process_docs_from_folder(temp_dir, output_dir)
+
+
+# Copied from sdg.utils.chunkers because that code is being refactored so we want to avoid importing anything from it.
+# TODO: Once the code base has settled down, we should make sure this code exists only in one place.
+def resolve_ocr_options() -> OcrOptions:
```
dmartinol, line 48: seems this is a "private" function, maybe we can rename to `_resolve_ocr_options`?

src/instructlab/rag/process_docs.py
```diff
@@ -0,0 +1,156 @@
...
+
+def _load_source_files(input_dir) -> list[Path]:
+    return [
+        Path(os.path.join(input_dir, f))
+        for f in os.listdir(input_dir)
+        if os.path.isfile(os.path.join(input_dir, f))
+    ]
+
+
+def _export_documents(
```
dmartinol, line 123: In case anyone asks the same, this function was taken from the Docling [Batch conversion](https://ds4sd.github.io/docling/examples/batch_convert/) example

src/instructlab/rag/process_docs.py
bbrowning: It's not hardcoded to None - that's just the default value before it goes into the logic a bit further down in that file to locate the docling models. This is how we find the docling models to use, in scenarios (like downstream) where we cannot download these on the fly from the network and instead they need to be shipped on-disk. It's probably worth syncing with @khaledsulayman on how we handle these docling nuances for downstream needs, where the environments are more restricted than upstream and we should not be reaching out to the internet.

jwm4: Oh, I see now; sorry I missed that in the first review.  I will DM @khaledsulayman to try to get more clarity.

tests/test_lab_rag_convert.py
```diff
@@ -0,0 +1,154 @@
...
+            yield conv_res
+
+
+def run_rag_convert_test(
+    params: List[str],
+    expected_strings: List[str],  # pylint: disable=unused-argument; noqa: ARG002
+    expected_output_file: Union[Path, None],  # pylint: disable=unused-argument; noqa: ARG002
+    should_succeed: bool,
+):
+    """
```
cdoern, line 54: can you add some info about each arg?

tests/test_lab_rag_convert.py
```diff
@@ -0,0 +1,154 @@
...
+    Core logic for testing conversion using the CLI runner and checking for expected output string and expected output file.
+    """
+    runner = CliRunner()
+    with patch("instructlab.rag.convert.DocumentConverter", MockDocumentConverter):
+        result = runner.invoke(lab.ilab, ["rag", "convert"] + params)
+        if should_succeed:
+            assert (
+                result.exit_code == 0
+            ), f"Unexpected failure for parameters {params}: {result.output}"
+            assert params[1] in result.output in result.output
```
cdoern, line 64: not sure I am understanding what this is meant to do. should it be 

`assert params[1] in result.output`?

tests/test_lab_rag_convert.py
```diff
@@ -0,0 +1,154 @@
...
+    test_input_dir = "tests/testdata/documents/md"
+    test_output_dir = tmp_path / "convert-outputs"
+    params = ["--input-dir", str(test_input_dir), "--output-dir", str(test_output_dir)]
+    expected_strings = ["Transforming source files ['hello.md']"]
+    expected_output_file = test_output_dir / "hello.json"
+    run_rag_convert_test(params, expected_strings, expected_output_file, True)
+
+
+# Note: This test uses a local taxonomy that references a file in a remote git repository.  For that reason, this
+# test won't pass when run on a machine with no connection to the internet.  It wil also fail if the repository
```
cdoern, line 105: this is fine, I think most of our tests fail without internet connection as of now

cdoern (commented):
just reviewing the tests. These align with the team's expectations of basic unit tests for a new implementation, no major concerns from me

tests/test_lab_rag_convert.py
```diff
@@ -7,8 +7,14 @@
 
 # Third Party
 from click.testing import CliRunner
-from docling.datamodel.base_models import DocumentStream, InputFormat  # type: ignore
+from docling.backend.docling_parse_v2_backend import DoclingParseV2DocumentBackend
```
dmartinol, line 10: pls add the `# type: ignore` annotation to skip this error

tests/test_lab_rag_convert.py
dmartinol: my fault, you are right

tests/test_lab_rag_convert.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    runner = CliRunner()
+    with patch("instructlab.rag.convert.DocumentConverter", MockDocumentConverter):
+        result = runner.invoke(lab.ilab, ["rag", "convert"] + params)
+        if should_succeed:
+            assert (
+                result.exit_code == 0
+            ), f"Unexpected failure for parameters {params}: {result.output}"
+            assert params[1] in result.output in result.output
+            assert expected_output_file.exists()
```
dmartinol, line 65: pls add `assert expected_output_file is not None` before this one

src/instructlab/configuration.py
```diff
@@ -268,6 +268,23 @@ def api_base(self):
...
+        default_factory=lambda: DEFAULTS.CONVERTED_DOCUMENTS_DIR,
+        description="Directory where converted documents are stored.",
+    )
+    taxonomy_path: StrictStr = Field(
+        default_factory=lambda: DEFAULTS.TAXONOMY_DIR,
+        description="Directory where taxonomy is stored and accessed from.",
+    )
+    taxonomy_base: StrictStr = Field(
+        default=DEFAULTS.TAXONOMY_BASE,
+        description="Branch of taxonomy used to calculate diff against.",
```
nathan-weinberg, line 284: Why are the first two options using `default_factory` and `lambda` expressions but the third is not?

Not saying this is incorrect, just trying to understand the inconsistency here

src/instructlab/rag/convert.py
```diff
@@ -0,0 +1,226 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# This is the implementation code for the ilab rag convert command, which converts documents from their
+# native format (e.g., PDF) to Docling JSON for use by ilab rag ingest.  See also the code in clu/rag/convert.py
```
nathan-weinberg, line 4:
```suggestion
# native format (e.g., PDF) to Docling JSON for use by ilab rag ingest.  See also the code in cli/rag/convert.py
```

src/instructlab/rag/convert.py
```diff
@@ -0,0 +1,226 @@
...
+        f"of which {failure_count} failed "
+        f"and {partial_success_count} were partially converted."
+    )
+    return success_count, partial_success_count, failure_count
+
+
+# Adapted from part of sdg/generate_data.py:_sdg_init and sdg/utils/chunkers.py:DocumentChunker
+# because that code is being refactored so we want to avoid importing anything from it.
+# TODO: Once the code base has settled down, we should make sure this code exists only in one place.
+# TODO: Also, it is very messy to be pulling the docling_model_path from
```
nathan-weinberg, line 149: Big +1 to both of these points

cdoern (commented):
small review just to clarify discussions around config usage. I am working on some changes to the clickext class now that will alleviate this.

tests/testdata/leanimports.py
```diff
@@ -5,7 +5,7 @@
 import sys
 
 # block slow imports
-for unwanted in ["deepspeed", "llama_cpp", "torch", "vllm"]:
+for unwanted in ["deepspeed", "llama_cpp", "vllm"]:
```
nathan-weinberg, line 8: Why the change here?

jwm4: Good catch!  We needed to drop torch here before we had a mock for Docling; I believe it was interfering with the Docling testing.  Now that we have a mock though I think we can put it back.  I will do that.

src/instructlab/configuration.py
jwm4: It is this way for here because we followed what is being done in `_generate` for pulling the taxonomy_path and taxonomy_base.  More broadly, I think lambda is consistently used throughout configuration.py for anything involving directories where things are stored and pulled from -- and often not used for other things.  That probably makes sense abstractly because directory structure seems like the sort of thing that could be dynamic in the future even if it is not now.

dmartinol: IIUC, `default_factory` is especially useful when we want to mock the current data in test functions. For instance this is used to generate an envinronment-agnostic `default_config` with the `make regenerate-testdata` target. 
```py
@patch.multiple(
    config.DEFAULTS,
    _cache_home="/cache/instructlab",
    _data_dir="/data/instructlab",
)
```

nathan-weinberg: Makes sense - thanks for the explanation!

pyproject.toml
```diff
@@ -104,14 +108,16 @@ include = [
   "instructlab.generator",
   "instructlab.cli",
   "instructlab.cli.model",
-  "instructlab.model",
   "instructlab.cli.process",
+  "instructlab.cli.rag",
+  "instructlab.llamacpp",
```
dmartinol, line 113: some changes are not related to your PR L113-115

jwm4: Just re-alphabetizing the list -- it has been alphabetical and then got out of order in a recent PR so I am restoring that.

tests/testdata/leanimports.py
jwm4: Looks like I was wrong -- keeping "torch" on the list causes the test to fail even without the mock because it turns out we need torch even to be able to import the data structures the mock uses.  So I have reverted that change.  Sorry!

bbrowning: I don't think we should drop torch here, as it's one of the things we want to make sure is not loaded until it's needed versus loaded eagerly any time instructlab gets imported or the ilab CLI invoked. The way we fixed this in SDG, which is in some of that copied over code from _resolve_ocr_options, is to not do top-level imports of Docling but instead import it lazily when needed to keep from eagerly causing torch to get imported.

bbrowning: An example of how we lazily load pieces of Docling in SDG to avoid eagerly importing torch - https://github.com/instructlab/sdg/blob/v0.6.3/src/instructlab/sdg/utils/chunkers.py#L43-L45 . I see you didn't follow this example when porting the code here, but that's why we do it - so that we don't import torch until someone actually tries to use an SDG feature that requires torch.

jwm4: Oh, that makes sense -- I didn't bring that part over because I didn't understand it.  I will do that now in my copy of resolve_ocr_options because it seems like a good idea.  With that said, I don't see how that would help with the unit testing issue because we do need it eventually for the unit tests that mock Docling so we can use their data structures, right?

bbrowning: @jwm4 So, for the unit tests, `leanimports.py` gets used from `tests/test_lab.py` in the `test_ilab_cli_imports` method, which I'm copying here for context:

```python
def test_ilab_cli_imports(testdata_path: pathlib.Path):
    script = testdata_path / "leanimports.py"
    subprocess.check_call([sys.executable, str(script)], text=True)
```

What this test does is start a new Python process that runs `leanimports.py`. All that spawned process does is go through and load each command/subcommand in the `ilab` CLI. It doesn't actually run any commands - just loads them. And, it ensures that slow imports - such as `torch` - don't load when running any of the commands because if they do, it slows down the entire `ilab` CLI even for simple operations that for example would not need to use torch.

This means that it's totally fine for your unit tests to import and use Docling. The only thing that will cause the leanimports test to fail is if something eagerly causes torch (or deepspeed, vllm, etc in that list) to load just from loading the `ilab` commands. Practically, that means you can't have any top-level imports within the new `ilab rag` commands (or anything imported by them) that cause torch to get loaded, and instead need to lazily import any Docling dependencies that load torch. By lazily import I mean defer importing the Docling bits that load torch until you're actually executing the commands that need them instead of importing them by default at the top of the Python files like we do in typical scenarios.

jwm4: Ah, that does make sense.  With this clarification, I was able to make some more adjustments and get torch back on the list.  I also updated the doc string at the top of leanimports.py with information from your message above.

bbrowning: Awesome - thanks for making that clearer for us all in the future by adding the explanation into the code!

nathan-weinberg (approved):
I have one question for Charlie about the testdata but otherwise this is looking good to me - approving from side

src/instructlab/rag/convert.py
```diff
@@ -0,0 +1,250 @@
...
+#  {xdg_data_home()}/instructlab/sdg/config.yaml.  This seems like a non-standard way to configure
+#  InstructLab functionality.  As part of unifying the code base, we should reconsider how this
+#  is configured.
+def _initialize_docling():
```
cdoern, line 149: yeah, I did not realize SDG had this sort of nested cfg +1 on moving it.

src/instructlab/rag/convert.py
```diff
@@ -0,0 +1,250 @@
...
+        return ocr_options
+    except ImportError:
+        # no easyocr either, so don't use any OCR
+        logger.error(
+            "Failed to load Tesseract and EasyOCR - disabling optical character recognition in PDF documents"
+        )
+        return None
+
+
+def _load_converter_and_format_options():
```
cdoern, line 240: what is this function actually doing? just imports and global initializations?

cdoern (commented):
nothing major, I appreciate using the new `config_class` and the added comments to some of the existing assertions, doc strings, etc.

tests/testdata/default_config.yaml
```diff
@@ -226,6 +226,19 @@ metadata:
   # Specific SKU related information about the given GPU, ex: PCIe, NVL
   # Default: None
   gpu_sku:
+# RAG configuration section.
+rag:
+  # RAG convert configuration section.
+  convert:
+    # Directory where converted documents are stored.
+    # Default: /data/instructlab/converted_documents
+    output_dir: /data/instructlab/converted_documents
```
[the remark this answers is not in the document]
cdoern: the defaults are all broken, so yes, but not in this PR. Something is wrong with `make regenerate-testdata`. But it doesn't quite matter since this is just used for tests.

src/instructlab/rag/convert.py
jwm4: Yes, so that leanimports.py can exclude torch.   See the doc string I added to leanimports.py for more context.  I can add more commenting here too if that would help.

src/instructlab/cli/rag/convert.py
```diff
@@ -0,0 +1,80 @@
...
+@click.option(
+    "--input-dir",
+    required=False,
+    type=click.Path(exists=True, file_okay=False, dir_okay=True, readable=True),
+    default=None,
+    help="A folder with user documents to process. If it set, the taxonomy is ignored.  If it's missing, the knowledge taxonomy files will be processed instead.",
+)
+@click.option(
+    "--taxonomy-path",
+    required=False,
```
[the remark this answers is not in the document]
dmartinol: better to avoid the `required` option it entirely, which seems the standard practice.

tests/testdata/default_config.yaml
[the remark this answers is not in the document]
nathan-weinberg: Cool, won't block on this then

src/instructlab/cli/rag/convert.py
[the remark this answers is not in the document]
jwm4: This seems like odd semantics for required=True -- in most contexts, if users can omit it and expect a default to be filled in, that is the definition of *not required*.  However, if that's how we do things here, I will get on board and make this change.

cdoern (approved):
lgtm. @jwm4 can you squash the commits and then I will remove the hold?

jwm4:
Commits are squashed now.

PR #2906, opened 2025-01-10
LLMaaJ Evaluation

alimaredia:
`ilab` interface for Ragas Evaluator (https://github.com/instructlab/eval/blob/main/src/instructlab/eval/ragas.py).

More information to come as this PR gets moved away from draft status.

### DIFF
--- a/src/instructlab/model/evaluate.py
+++ b/src/instructlab/model/evaluate.py
@@ -49,6 +49,7 @@ def validate_options(
     batch_size,
     tasks_dir,
     input_questions,
+    llmaaj_job_file,
 ):
     """takes in arguments from the CLI and uses 'benchmark' to validate other arguments
     if all needed configuration is present, raises an exception for the missing values
@@ -118,14 +119,20 @@ def validate_options(
             validate_model(base_model, "--base-model", allow_gguf=False)
 
     if benchmark == Benchmark.LLMAAJ:
-        required_args = [input_questions]
-        required_arg_names = ["--input-questions"]
-        if None in required_args:
-            click.secho(
-                f"Benchmark {benchmark} requires the following flags to be set: {required_arg_names}",
-                fg="red",
-            )
-            raise click.exceptions.Exit(1)
+        if llmaaj_job_file is None:
+            required_args = [input_questions]
+            required_arg_names = ["--input-questions"]
+            if None in required_args:
+                click.secho(
+                    f"Benchmark {benchmark} requires the following flags to be set: {required_arg_names}",
+                    fg="red",
+                )
+                raise click.exceptions.Exit(1)
+        else:
+                click.secho(
+                    f"Job file passed in at {llmaaj_job_file} for benchmark {benchmark}. All other flags ignored",
+                    fg="green",
+                )
 
 
 def validate_model(model: str, model_arg: str = "--model", allow_gguf: bool = True):
@@ -531,7 +538,7 @@ def launch_server(
 )
 @click.option(
     "--input-questions",
-    type=click.Path(path_type=pathlib.Path),
+    type=click.Path(exists=True, path_type=pathlib.Path),
     default=None,
     help="Path to the questions and reference answers the model will be evaluating in the LLMaaJ evaluation",
 )
@@ -561,16 +568,16 @@ def launch_server(
     help="Model to evaluate for the LLMaaJ evaluation",
 )
 @click.option(
-    "--model-name",
-    type=click.STRING,
-    default=None,
-    help="Prompt for the model when getting responses in the LLMaaJ evaluation",
-)
-@click.option(
     "--judge-model-name",
     type=click.STRING,
     default=ragas_eval.DEFAULT_JUDGE_MODEL,
     help="Prompt for the model when getting responses in the LLMaaJ evaluation",
+)
+@click.option(
+    "--llmaaj-job-file",
+    type=click.Path(exists=True,path_type=pathlib.Path),
+    default=None,
+    help="Path to the job file for the LLMaaJ evaluation",
 )
 @click.pass_context
 @clickext.display_params
@@ -602,8 +609,8 @@ def evaluate(
     model_prompt,
     temperature,
     llmaaj_model,
-    model_name,
     judge_model_name,
+    llmaaj_job_file
 ):
     """Evaluates a trained model"""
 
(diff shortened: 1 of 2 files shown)


src/instructlab/model/evaluate.py
```diff
@@ -31,7 +33,7 @@ class Benchmark(str, enum.Enum):
     MMLU_BRANCH = "mmlu_branch"
     MT_BENCH = "mt_bench"
     MT_BENCH_BRANCH = "mt_bench_branch"
-
+    LLMAAJ = "llmaaj"
```
RobotSail, line 36: LLMAAJ is a very broad term. We are specifically doing open response QA eval. So I would either call this QA grading or RAG eval.

src/instructlab/model/evaluate.py
```diff
@@ -516,6 +518,43 @@ def launch_server(
...
+)
+@click.option(
+    "--temperature",
+    type=click.FLOAT,
+    default=0.0,
+    help="Temperature for the model when getting responses in the LLMaaJ evaluation",
+)
+@click.option(
+    "--model-name",
+    type=click.STRING,
```
RobotSail, line 548: Should we call this `--model` instead of `--model-name`? I believe in other evaluations we just refer to it as `--model`.

Same for the judge model - should it just be `--judge-model` instead of `--judge-model-name`?

RobotSail: ^ It looks the above already exist as flags, we should just reuse those

src/instructlab/model/evaluate.py
```diff
@@ -552,6 +597,23 @@ def evaluate(
     with contextlib.suppress(ValueError):
         batch_size = int(batch_size)
 
+    if benchmark == Benchmark.LLMAAJ:
+        from instructlab.model.llmaaj import run_llmaaj
+
+        run_llmaaj(ctx,
```
RobotSail, line 603: Nothing here should be called llmaaj. 

LLMaaJ is just a way of evaluating answers. MT-Bench is also LLMaaJ, but we don't call it that.


I.e., MT-Bench is to LLMaaJ as MMLU is to multiple-choice.

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+
+    output_path = pathlib.Path(output_dir)
+    if not output_path.exists() or not output_path.is_dir():
+        output_dir = DEFAULTS.EVAL_DATA_DIR
+        logger.info(f"Output dir provided is not a directory or does not exist, writing results to output directory {output_dir}")
+
+    return f"{output_dir}/responses-{timestamp}.{file_type}"
+
+def run_llmaaj(ctx, model, max_workers, gpus, backend, enable_serving_output, input_questions, output_file_formats, output_dir, model_prompt, temperature, model_name, judge_model_name):
+    logger.info(f"Input questions path is {input_questions}")
```
RobotSail, line 84: We should avoid logging by interpolating into the string. We want to instead pass it in as an argument. This is because the logger needs to format everything internally, so we have to pass it in like the classic C style.  

```suggestion
    logger.info("Input questions path is %s", input_questions)
```

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+
+    try:
+        student_api_key = None
+        if model_is_local(model):
+            logger.info(f"Model is local")
+            if model_name == None:
```
RobotSail, line 90: You can just do:

```suggestion
            if not model_name:
```

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+
+    return f"{output_dir}/responses-{timestamp}.{file_type}"
+
+def run_llmaaj(ctx, model, max_workers, gpus, backend, enable_serving_output, input_questions, output_file_formats, output_dir, model_prompt, temperature, model_name, judge_model_name):
+    logger.info(f"Input questions path is {input_questions}")
+
+    try:
+        student_api_key = None
+        if model_is_local(model):
+            logger.info(f"Model is local")
```
RobotSail, line 89: No need to interpolate here, and also you shouldn't anyway. 

```suggestion
            logger.info("Model is local")
```

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+            if model_name == None:
+                model_name = get_model_name(model)
+            server, api_base, effective_gpus = launch_server(
+                eval_serve=ctx.obj.config.serve,
+                tls_client_cert=ctx.params["tls_client_cert"],
```
RobotSail, line 94: Use `.get("tls_client_cert", <enter default here>)` on these so you can handle the default case when it's missing instead of getting an unexpected KeyError

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+def run_llmaaj(ctx, model, max_workers, gpus, backend, enable_serving_output, input_questions, output_file_formats, output_dir, model_prompt, temperature, model_name, judge_model_name):
+    logger.info(f"Input questions path is {input_questions}")
+
+    try:
+        student_api_key = None
+        if model_is_local(model):
+            logger.info(f"Model is local")
+            if model_name == None:
+                model_name = get_model_name(model)
+            server, api_base, effective_gpus = launch_server(
```
RobotSail, line 92: It might also be worth checking to see if an OpenAI-compatible server is already running at this endpoint and if it's serving the model we are expecting to evaluate.

RobotSail: This way we can avoid relaunching the server a second time.

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+                model=model,
+                model_name=model_name,
+                max_workers=max_workers,
+                gpus=gpus,
+                backend=backend,
+                enable_serving_output=enable_serving_output,
+            )
+        else:
+            logger.info(f"Model is an endpoint")
+            student_api_key = os.environ.get("STUDENT_ENDPOINT_API_KEY", None)
```
RobotSail, line 107: Does MT-Bench use this value too? If not, I recommend dropping it. Otherwise it might create inconsistent behavior.

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+                    "tls_insecure": ctx.params["tls_insecure"],
+                }
+            )
+
+            if model_name == None:
+                model_name = get_endpoint_model_name(model, student_api_key, http_params)
+            api_base = model
+
+        openai_client = get_openai_client(model_api_base=api_base, api_key=student_api_key)
+        logger.info(f"Model name is {model_name}")
```
RobotSail, line 122: More interpolations that we should replace with log formats

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+            
+        response_df = result.dataset.to_pandas()
+        scores = [score["domain_specific_rubrics"] for score in result.scores]
+        response_df['scores'] = scores
+        output_formats = output_file_formats.split(",")
+
+        if IOFileType.JSONL.value in output_formats:
+            results_file = create_results_file_name(IOFileType.JSONL.value, output_dir)
+            response_df.to_json(f"{results_file}", orient="records", lines=True)
+            print(f"Responses and scores written to {results_file}")
```
RobotSail, line 149: Why are we printing here but logging further up?

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+            api_key=api_key,
+            timeout=DEFAULTS.CONNECTION_TIMEOUT,
+            http_client=http_client,
+        )
+        models = client.models.list()
+    except OpenAIError as exc:
+        raise client_utils.ClientException(f"Connection Error {exc}") from exc
+
+    if len(models.data) != 1:
+        raise client_utils.ClientException(f"More than one model at endpoint, pass in model name manually")
```
RobotSail, line 62: Seems like this will still throw even if the server has 0 models. Maybe we should update the message?

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+logger = logging.getLogger(__name__)
+
+class IOFileType(enum.Enum):
+    CSV: str = "csv"
+    JSONL: str = "jsonl"
+    XLSX: str = "xlsx"
+    QNA_YAML: str = "qna.yaml"
+
+def model_is_local(model: str) -> bool:
+    if os.path.exists(model):
```
RobotSail, line 38: Since the majority of your complex logic exists in a single condition, you can simplify this function by checking the opposite condition here and returning early. That way, the rest of your function can be on a single indent level:

```py
def is_local_model(model):
  if not os.path.exists(model):
    return False

  # continue here at the top-level indent
  if is_safetensors(model):
     ...
```

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,164 @@
...
+
+    return models.data[0].id
+
+def create_results_file_name(file_format: IOFileType, output_dir: str) -> str:
+    file_type = "jsonl"
+    if IOFileType.CSV.value == file_format:
+        file_type = "csv"
+    if IOFileType.XLSX.value == file_format:
+        file_type = "xlsx"
+
```
RobotSail, line 72: What happens if the type is QNA.yaml?

RobotSail (commented):
Great progress so far, the big picture is there now we just need to refine the little details.

src/instructlab/model/llmaaj.py
```diff
@@ -0,0 +1,388 @@
...
+logger = logging.getLogger(__name__)
+
+class llmaaj_judge(BaseModel):
+    name: str = Field(
+        default='gpt-4o',
+        description="Judge model name for judge doing evaluation in the LLMaaJ evaluation. Model name must be an Open AI GPT model.",
+    )
+
+class llmaaj_model(BaseModel):
```
cdoern, line 44: so instead of having this as a separate config class in this pkg, I think this class should live in `configuration.py` and the flags in `evaluate.py` should inherit the defaults from the `config.yaml` as all other commands do. Introducing it this way would be a pretty big departure from what other commands are doing.

we just introduced `config_class=` which you can specify in your command line options like:

```
click.option(
"--llamaaj",
config_class="llamaaj",
...
```

meaning your flags can inherit defaults from other parts of the config besides `evaluate`

alimaredia:
Closed for: https://github.com/instructlab/instructlab/pull/2940

PR #2929, opened 2025-01-16
Allow functional tests to run on MacOS

anastasds:
The functional test script requires `timeout`, which is available as `gtimeout` from `coreutils` on MacOS. Adds an alias from `timeout` to `gtimeout` if MacOS is detected, or prints error message with install instructions if `gtimeout` is not present.


nathan-weinberg (approved):
Thanks for this!

alinaryan (approved):

alinaryan (approved):

PR #2930, opened 2025-01-16
add support for `config_class` in clickext

cdoern:
currently, a click option can only inherit its value from a field within the config class matching the command name this adds `config_class` which can be used in combination with `config_section` allowing users to specify a different parent class and subsections to inherit the default from.

this is a good idea since it avoid people directly reading from `ctx.obj.config` and also removes the need to duplicate fields

**Checklist:**


jwm4 (approved):
This seems like it will be very useful.

nathan-weinberg (changes requested):
LGTM, but requesting we add a test or two for this

nathan-weinberg (approved):
@cdoern Yes! Thank for the explanation!

PR #2933, opened 2025-01-16
fix: process listing, do not modify list len

cdoern:
currently, we pop items off of the process registry while listing it causing an error. Remove all processes at the end of listing and mark them for removal properly

**Checklist:**


src/instructlab/process/process.py
```diff
@@ -310,21 +311,28 @@ def list_processes():
         return
 
     list_of_processes = []
+
+    to_remove = []
```
nathan-weinberg, line 315:
```suggestion
    processes_to_remove = []
```

src/instructlab/process/process.py
```diff
@@ -39,6 +39,7 @@ def add_process(
                 start_time, "%Y-%m-%d %H:%M:%S"
             ).isoformat(),
             "status": status,
+            "removal": False,
```
nathan-weinberg, line 42: Why add a field here? Couldn't we just have a status for processes flagged for removal?

cdoern: if I mark a process as `Removal`. user will see this. The point of this is so a user can run 

`ilab process list` after a process finishes and they will see `Done` in their list a single time. the next time, the process will be removed. This allows users to understand a process has finished before we kick it off the registry. using `Remove` will probably confused people

alinaryan (approved):

src/instructlab/process/process.py
cdoern: How about this: 

Done is the state we use to indicate "lets kick this off the registry" but I can amend my code to only set a process to "done" in the saved registry once it has been added to the list as such. That means if we every query a processes status as "Done", its time to kick of off the registry?

nathan-weinberg: Up to you - your explanation makes sense. I will approve but if you want to modify the logic slightly feel free :+1:

alinaryan (approved):
This makes more sense to separate listing the processes vs. managing process state. LGTM!

nathan-weinberg (approved):

PR #2934, opened 2025-01-17
[Draft]: Allow config_sections to refer to a leaf field in clickext module

dmartinol:
This change allows to use an argument name in the click option that is different from the actual field name.
Until now, the `config_sections` value was used to match the parent sections of the target field, but the field name and the argument name had to match exactly.

For example, using the following option configuration we can have an argument called `store_uri`:
```py
@click.option(
    "--document-store-uri",
    "store_uri",
    type=click.STRING,
    cls=clickext.ConfigOption,
    config_class="rag",
    config_sections="document_store.uri",
)
def cmd( ...., store_uri,...):
```
Match a configuration field called `uri`:
```yaml
document_store:
  collection_name: ilab
  uri: embeddings.db
```

This change may be useful to reduce the verbosity of the configuration model (e.g. `document_store.uri` instead of `document_store.document_store_uri`) and at the same time having consistent argument names in the function (e.g. `document_store_uri`)

More UT may be added to validate this change, even if, tbh, there aren't many tests validating the `clickext` module.

Will create an issue if needed

**Checklist:**

### DIFF
--- a/src/instructlab/clickext.py
+++ b/src/instructlab/clickext.py
@@ -224,7 +224,10 @@ class ConfigOption(click.Option):
             # we need to overwrite the value
             for secname in self.config_sections:
                 section = section.get(secname, {})
-            value = section.get(self.name)
+            if isinstance(section, dict):
+                value = section.get(self.name)
+            else:
+                value = section
         # fix parameter source for config section that are mis-reported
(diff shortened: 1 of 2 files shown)


dmartinol:
Need some further design and team commitment

PR #2903, opened 2025-01-10
feat: Ingest document embeddings for Retrieval augmented generation

dmartinol:
**Issue resolved by this Pull Request:**
Resolves #2875 
Addresses #2957
Depends on https://github.com/instructlab/instructlab/pull/2832

**Dev Docs related to this Pull Request:**
Link to Dev Doc or PR: https://github.com/instructlab/dev-docs/pull/161

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [x] Unit tests have been added, if necessary.
- [ ] Functional tests have been added, if necessary.
- [ ] E2E Workflow tests have been added, if necessary.

**Unit test code in the next commit**

PR #2946, opened 2025-01-21
Feature gates

### LINKED ISSUE #2888: [RAG][Dev] Put all RAG related functionality behind feature flag
As described in this [dev doc](https://github.com/instructlab/dev-docs/blob/a9184fe4bcae8eed8cd1f1c7425df4edb9aaef99/docs/rag/ilab-rag-retrieval.md), RAG related functionality should be hidden behind a feature flag called `ILAB_ENABLE_EXPERIMENTAL`.

Acceptance Criteria
* `ilab rag` command group exits with error message if feature flag is off
* passing any RAG-related options to `ilab model chat` exits with error message if feature flag is off
* config YAML has comments by each of the RAG related options indicating that they are experimental
* Using the `/r` command in the chat has the same envvar-specific behavior

anastasds:
Related ADR: https://github.com/instructlab/dev-docs/pull/178
Depends on #2886 , #2903 
Resolves #2888


src/instructlab/feature_flags.py
```diff
@@ -0,0 +1,32 @@
...
+        GatedFeatures.RAG: FeatureScopes.DevPreviewNoUpgrade
+    }
+
+    @staticmethod
+    def available_scopes():
+        return [f.value for f in FeatureScopes]
+
+    @staticmethod
+    def feature_available(feature: GatedFeatures):
+        if FeatureGating._feature_scopes.get(feature, FeatureScopes.Default) == os.environ.get('_env_var_name', FeatureScopes.Default):
```
dmartinol, line 27: maybe w/o quotes, e.g.:  `os.environ.get(_env_var_name,...` ?

src/instructlab/cli/rag/convert.py
```diff
@@ -59,6 +60,10 @@ def convert(
     input_dir,
     output_dir,
 ):
+    if not FeatureGating.feature_available(GatedFeatures.RAG):
```
dmartinol, line 63: what about a `FeatureGating.check_feature_available(GatedFeatures.RAG)` to raise a consistent error from any client, instead of having each consumer to repeat the same check? (this would also let you keep the `env_var_name` internal to the module)

src/instructlab/feature_flags.py
```diff
@@ -0,0 +1,32 @@
...
+    env_var_name = 'ILAB_FEATURE_SCOPE'
+    _feature_scopes = {
+        GatedFeatures.RAG: FeatureScopes.DevPreviewNoUpgrade
+    }
+
+    @staticmethod
+    def available_scopes():
+        return [f.value for f in FeatureScopes]
+
+    @staticmethod
```
dmartinol, line 25: just out of curiosity: why a static method instead of a function, here and in the rest of this module? (I'm not against it, but I see it's not widely adopted pattern in the project)

dmartinol (commented):
lgtm, apart from some minor comments

src/instructlab/cli/rag/convert.py
anastasds: I chose this way so that different error messages could be output in different contexts, allow for different degrees of logging, etc.

Not a strong opinion either way.

src/instructlab/feature_flags.py
anastasds: In order to avoid having in instantiate a `FeatureGates` object.

dmartinol: if we only use functions, the class it's not even needed. anyway, it's fine (I like it, it's just rarely used in the codebase)

src/instructlab/feature_gates.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @staticmethod
+    def available_scopes():
+        return [f.value for f in FeatureScopes]
+
+    @staticmethod
+    def feature_available(feature: GatedFeatures):
+        return FeatureGating._feature_scopes.get(feature, None) == os.environ.get(
+            FeatureGating.env_var_name, FeatureScopes.Default
+        )
```
bbrowning, line 29: It's not a big deal for now, since there's only one gated feature, but if we want this feature gating to be generic and usable for multiple features we'll have to consider how someone enables multiple feature gates or whether there's a hierarchy. As a concrete example, should enabling `DevPreviewNoUpgrade` also enable all `TechPreviewNoUpgrade` features? If not, how would I enable both? The same goes for `CustomNoUpgrade` - in effect, is there a hierarchy here where enabling one less-stable gate automatically enables all more stable gates above it? As written today a user can only enable a single level.

src/instructlab/model/chat.py
```diff
@@ -586,6 +587,12 @@ def chat_model(
     visible_overflow,
 ):
     """Runs a chat using the modified model"""
+    if not FeatureGating.feature_available(GatedFeatures.RAG):
+        raise ValueError(
+            f"In order to use experimental features, set {FeatureGating.env_var_name}"
+            f" to one of: {FeatureGating.available_scopes()}"
+        )
+
```
bbrowning, line 595: Same comment about the error message as above - this could be more specific and helpful to the user, such as telling them which exact feature scope to use and let them which experimental feature they tried to use that we're blocking unless they enable the gate.

bbrowning (commented):
Leaving a couple of comments here - nothing that warranted requesting changes on GitHub, but things that could be improved if you're so inclined as well as a typo that we should fix (CustomNoUpgrade value), but that is not in a code path anything would hit yet so is not critical.

src/instructlab/feature_gates.py
anastasds: That's a great question that I think needs to be figured out down the line, but that would be premature to model now. I don't see anything that prevents us from choosing any particular model.

src/instructlab/feature_gates.py
```diff
@@ -0,0 +1,29 @@
+# Standard
+from enum import StrEnum
+import os
+
+
+class FeatureScopes(StrEnum):
+    Default = "Default"
+    DevPreviewNoUpgrade = "DevPreviewNoUpgrade"
+    TechPreviewNoUpgrade = "TechPreviewNoUpgrade"
+    CustomNoUpgrade = "TechPreviewNoUpgrade"
```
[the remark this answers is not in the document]
anastasds: Thanks, fixed.

src/instructlab/model/chat.py
anastasds: That's a good point. There is a risk of those error messages going out of date by being forgotten to be changed - but that can be addressed with unit tests, which I'll add now.

anastasds:
I was running just the new feature gating tests and only then realized that some other RAG-related unit tests need to be modified to run properly, so putting this back into draft mode.

src/instructlab/feature_gates.py
cdoern: If a feature is DevPreview, it is assumed it might really break  your environment. But if a feature is TechPreview it's assumed that what the user is running is a bit closer to production. If someone _just_ wants the new TechPreview features I don't think they should get the DevPreview ones that might break their environment  more than they intended

src/instructlab/cli/rag/convert.py
anastasds: The risk there is that those error messages can become outdated - which can be addressed with unit testing, so will do.

anastasds: Done.

src/instructlab/model/chat.py
anastasds: Done.

src/instructlab/feature_gates.py
anastasds: I'm sure that this will come up in the short to medium term future; I don't think this PR is the place to decide it.

src/instructlab/cli/rag/convert.py
```diff
@@ -55,6 +56,13 @@ def convert(
...
+        print(
+            f"This functionality is experimental; set {FeatureGating.env_var_name}"
+            f' to "{FeatureScopes.DevPreviewNoUpgrade}" to enable.'
+        )
```
nathan-weinberg, line 63: Can we make this `click.echo`?

src/instructlab/model/chat.py
```diff
@@ -586,6 +587,13 @@ def chat_model(
     visible_overflow,
 ):
     """Runs a chat using the modified model"""
+    if not FeatureGating.feature_available(GatedFeatures.RAG):
+        print(
+            f"This functionality is experimental; set {FeatureGating.env_var_name}"
+            f' to "{FeatureScopes.DevPreviewNoUpgrade}" to enable.'
+        )
```
nathan-weinberg, line 594: Can we use the logger here?

nathan-weinberg (commented):
Overall looking good to me @anastasds - couple of comments - I'd also like if we can mention this somewhere in the Core `CONTRIBUTING.md` for future development - I know @jwm4 is planning a revamp of that doc so if he wants to include documenting Feature Gates as part of that work instead of doing so in the PR that's fine with me

cdoern:
@anastasds looks like you'll need to enable the FG to get the unit tests for rag convert to pass

src/instructlab/cli/rag/convert.py
anastasds: Done.

src/instructlab/model/chat.py
anastasds: Done.

anastasds:
Actually it was because I redid the changes instead of rebasing, and forgot the `rag_enabled` check for `ilab model chat` so the logic was broken - fixed and tested locally before pushing.

tests/test_utils.py
```diff
@@ -1,5 +1,4 @@
 # SPDX-License-Identifier: Apache-2.0
-
```
nathan-weinberg, line 2: This is very nit but I'd prefer we kept this line for readability (if CI demanded this change, feel free to disregard)

nathan-weinberg (approved):
one nit otherwise LGTM

cdoern (approved):

tests/test_utils.py
anastasds: `make fix` did this

anastasds: It did not actually some sort of other autoformatting did. I'll restore it, it would have failed CI before if it were a problem.

Issue #2947, opened 2025-01-21
[RAG][Dev] End to end tests


anastasds:
Acceptance criteria:
* e2e test that covers `ilab rag {convert,ingest}` and `ilab chat --rag`
* Follows e2e testing patterns
* Does not block regular feature testing - e2e tests for experimental features should not be able to block PRs for non-experimental work
* CI job added and able to be triggered on PRs containing work on experimental features

PR #2948, opened 2025-01-21
feat: Add end to end tests for RAG-enabled chat using user-supplied documents

anastasds:
#2947 
Depends on #2903, #2886

Run as `e2e-custom.sh -dm`

* Adds a PDF document into a new `test-data/raw_documents` repo
* Adds end to end test of converting and ingesting a document, then RAG-enabled chat
* Expands `e2e-custom.sh` to accommodate dev preview functionality


scripts/e2e-experimental.sh
```diff
@@ -0,0 +1,284 @@
...
+#   DATA_HOME
+#   CACHE_HOME
+#   CONFIG_HOME
+#   E2E_TEST_DIR
+#   HOME
+# Outputs:
+#   None
+########################################
+
+init_e2e_tests() {
```
dmartinol, line 59: shouldn't we also init the `ILAB_FEATURE_SCOPE` env var introduced in the feature-gates PR?

scripts/e2e-experimental.sh
```diff
@@ -0,0 +1,284 @@
...
+    mkdir -p "$DATA_HOME"/instructlab/taxonomy/knowledge/phoenix/overview/e2e-phoenix
+    cp "$SCRIPTDIR"/test-data/knowledge/e2e-qna-knowledge-phoenix.yaml "$DATA_HOME"/instructlab/taxonomy/knowledge/phoenix/overview/e2e-phoenix/qna.yaml
+
+    step Verify the taxonomy
+    ilab taxonomy diff
+
+    task Taxonomy updates and verification Complete
+}
+
+test_convert() {
```
dmartinol, line 142: if you want to convert taxonomy knowledge files:
`ilab rag convert`
if you want to convert user docs:
`ilab rag convert --input-dir "$SCRIPTDIR"/test-data/raw_documents`

Probably we can test both flows, since we have the taxonomy knowledge files already in. Just put the outputs in different folders with `--output-dir ....` option

scripts/e2e-experimental.sh
```diff
@@ -0,0 +1,284 @@
...
+    step Convert document for ingestion
+    ilab rag convert "$SCRIPTDIR"/test-data/raw_documents
+    task Document converstion Complete
+}
+
+test_ingest() {
```
dmartinol, line 148: again:
`ilab rag ingest` to ingest preprocessed JSONs from taxonomy knowledge
`ilab rag ingest --input-dir ...` to ingest JSONs from `ilab rag convert `

dmartinol (commented):
> How to set this up as an optional CI check?

What do you mean?

scripts/e2e-experimental.sh
anastasds: Good call, done.

anastasds:
>>How to set this up as an optional CI check?
>
>What do you mean?

The goal is for this set of tests to be able to be triggered on a PR involving changes to experimental features and then be required for merge - but to not run automatically for PRs so that experimental feature work never blocks other work. I'm not clear on what needs to be done to set that up.

nathan-weinberg:
There's an existing `e2e-custom.sh` script - rather than adding a third script, why not just use that?

anastasds:
@nathan-weinberg I wasn't sure what all uses it and wanted to keep experimental features totally cordoned off, at least to start with, in order to avoid issues with experimental features from potentially blocking anything else. Happy to merge them in; I was being risk avoidant since this whole experimental features pattern seems to be net new.

nathan-weinberg:
@anastasds yes please use that script, it should mesh well with the mission of us keeping it: https://github.com/instructlab/instructlab/blob/main/docs/ci.md#end-to-end-e2e-tests

Only `e2e-ci.sh` is used for the regular CI, so no worries about breaking anything by modifying `e2e-custom.sh` - it's also already hooked up to a [custom job](https://github.com/instructlab/instructlab/blob/main/.github/workflows/e2e-aws-custom.yml) you can use

anastasds:
@nathan-weinberg will do today

nathan-weinberg (commented):
Looking good to me overall - when this is out of draft are you able to trigger a test custom job @anastasds so we can see it in action?

scripts/e2e-custom.sh
```diff
@@ -71,13 +72,26 @@ function init_e2e_tests() {
...
     # ensure that our mock e2e dirs exist
     for dir in "${CONFIG_HOME}" "${DATA_HOME}" "${STATE_HOME}" "${CACHE_HOME}"; do
     	mkdir -p "${dir}"
     done
 
     E2E_LOG_DIR="${HOME}/log"
     mkdir -p "${E2E_LOG_DIR}"
+
+    # set appropriate feature gating flag
+    if [ "$DEV_PREVIEW_FEATURE_GATING" -eq 1 ]; then
```
dmartinol, line 85: Note: you probably have these env vars in your shell, so you're not seeing the `DEV_PREVIEW_FEATURE_GATING: unbound variable` error.
Since we have `set -u` at the beginning, we have to ensure that the env variables are defined.
We can use either `-n` or provide a default as in:
```
    if [ ${DEV_PREVIEW_FEATURE_GATING:-} -eq 1 ]; then
```

Explanation of `set -u`:
```
Treat unset variables and parameters other than the special parameters ‘@’ or ‘*’, or array variables subscripted with ‘@’ or ‘*’, as an error when performing parameter expansion. An error message will be written to the standard error, and a non-interactive shell will exit.
```

scripts/e2e-custom.sh
```diff
@@ -257,6 +288,31 @@ test_taxonomy() {
...
+test_convert_taxonomy_for_rag() {
+    step Convert user taxonomy for ingestion into vector store
+    ilab rag convert
+    task Taxonomy document conversion Complete
+}
+
+test_convert_user_document_for_rag() {
+    step Convert documents for ingestion into vector store
+    mkdir -p "${DATA_HOME}/instructlab/converted_documents"
+    ilab rag convert --input-dir "${DATA_HOME}/instructlab/raw_documents" --output-dir="${DATA_HOME}/instructlab/converted_documents"
```
dmartinol, line 300: I can't find the `raw_documents` folder in the temp dir, maybe we missed to copy it from the `test-data` folder?

scripts/e2e-custom.sh
dmartinol, line 1: Pls be aware that some of the files referenced in this script and used by the `test_taxonomy` function are not present in the repo, like:
```
$SCRIPTDIR/test-data/e2e-qna-freeform-skill.yaml 
$SCRIPTDIR/test-data/e2e-qna-grounded-skill.yaml
```
They got moved to a different folder and renamed in this [commit ](https://github.com/instructlab/instructlab/commit/318b6856f55815255f87da104ae90266c6386877)
Just to anticipate to you that the `test_taxonomy 1/2` will probably fail.

scripts/e2e-custom.sh
```diff
@@ -435,6 +491,17 @@ test_exec() {
 
     test_chat
 
+    # Test RAG capabilities
+    if [ "$DEV_PREVIEW_FEATURE_GATING" -eq 1 ]; then
+      test_convert_taxonomy_for_rag
```
dmartinol, line 496: we probably need to run `test_taxonomy 3` before these steps, so the taxonomy is initialized with knowledge data (tests 1 and 2 have compositional skills, instead)

scripts/e2e-custom.sh
```diff
@@ -257,6 +288,31 @@ test_taxonomy() {
     ilab taxonomy diff "${DIFF_ARGS[@]}"
 }
 
+test_convert_taxonomy_for_rag() {
+    step Convert user taxonomy for ingestion into vector store
+    ilab rag convert
```
dmartinol, line 293: This should be configured to locate the path where `test_taxonomy` puts the taxonomy repo, e.g. `./taxonomy`:
```bash
    ilab rag convert --taxonomy-path taxonomy
```

cdoern:
is there a reason we aren't adding these tests to `e2e-ci.sh`. The tests are passing on this PR but I think the S and M jobs use `e2e-ci.sh`.

nathan-weinberg:
> is there a reason we aren't adding these tests to `e2e-ci.sh`. The tests are passing on this PR but I think the S and M jobs use `e2e-ci.sh`.

Yes, @anastasds and I already discussed this - we can possibly move them over later but for now I told him to use `custom`

anastasds:
@cdoern yes, and I am working with @nathan-weinberg to trigger the new custom job on this PR. 

There should be a more rigorous build verification process for experimental features than "remember to run the custom job and make sure you pass it the right parameters" but this is at least a reasonable start.

scripts/e2e-custom.sh
alinaryan: Isn't there a default set in the config? https://github.com/instructlab/instructlab/blob/main/src/instructlab/configuration.py#L301

dmartinol: there is, but the script clones taxonomy repo under the current folder, so we need to override it (of course we could update the script to use the default but it looks beyond the scope)

scripts/e2e-custom.sh
```diff
@@ -156,6 +172,11 @@ test_download() {
         ilab model download --repository instructlab/merlinite-7b-lab
     fi
 
+    if [ "$DEV_PREVIEW_FEATURE_GATING" -eq 1 ]; then
+        step Downloading embedding model
+        ilab model download --repository "${GRANITE_EMBEDDING_MODEL}"
```
dmartinol, line 177: probably needs `--hf-token "${HF_TOKEN}`

nathan-weinberg: Will be handled by https://github.com/instructlab/instructlab/pull/2997 - `ilab` automatically pulls from the env var if it is present

cdoern:
to be safe ^

scripts/e2e-custom.sh
```diff
@@ -31,6 +31,8 @@ MIXTRAL_GGUF_MODEL="mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf"
 GRANITE_GGUF_MODEL="granite-7b-lab-Q4_K_M.gguf"
 MERLINITE_GGUF_MODEL="merlinite-7b-lab-Q4_K_M.gguf"
 GRANITE_SAFETENSOR_REPO="instructlab/granite-7b-lab"
+GRANITE_EMBEDDING_MODEL_NAME="granite-embedding-125m-english"
```
dmartinol, line 34: we need to initialize GRANITE var, ~ line 15:
```bash
GRANITE=0
```

PR #2950, opened 2025-01-21
rewrite ilab model convert

### LINKED ISSUE #2351: Redesign `ilab model convert`
Currently, `ilab model convert` is a mac specific command that does more than what the name might suggest. It expects to be supplied with a MLX adapter file, which it fuses with the base model that was used to perform `simple` training on a mac environment. It also creates and cleans up a bunch of intermediate folders in the process of conversion. It is currently just an extension to the macOS workflow, and has a dependency on other commands being executed before it (such as `train` and `generate`)

`ilab model convert` should be decoupled from the rest of the worklow, and should be presented as a standalone command that can be used to convert any model from one format to another. This would fit in well with our vision of the ilab CLI being a true "model engine"
Users should be able to supply a source model, target destination and a target format as input, and receive a converted model as output 
To begin with we can support the following formats: `.gguf`, `.safetensors`, `.bin`, mlx safetensors

- Provides a new entry point for users into the `ilab` workflow
- Decouples model conversion from the rest of the workflow
- Expands access to this functionality to other platforms other than MacOS
- Better testability of the command

Currently, the macOS workflow works as follows:
- train a model and receive an adapter file as output
- run `ilab model test` to try out the adapter as a manual smoke test
- run `ilab model convert` - which fuses the supplied adapter to the base model and then converts it

By making this change we will be removing the logic that fuses an adapter to a model before conversion for mac users. By extension we would be removing users' ability to smoke test an adapter and decide to explicitly fuse that one into their model.
Instead, the workflow will become aligned with how `simple` train works on linux today - which is to automatically pick a checkpoint and fuse it for the users

An alternative for this would be to introduce a new `ilab model fuse` if we want to retain the ability to test and fuse adapters explicitly 

**Additional context**
related issues:
#2110 

**Long term stretch goal**
Support conversion between different model architectures

jaideepr97:
This PR redesigns `ilab model convert` to work as a completely independent command. Previously it was only usable on macOS and performed additional steps such as adapter fusing etc. In this rewrite there are no assumptions made, and the command strictly performs only conversion operations, which allows us to use it on any platform

at present only safetensors -> GGUF conversions are supported but the conversion mechanism has been made such that it is easily extensible for any new conversions to be added in the future

NOTE: the `llamacpp/convert_to_gguf.py` script has been borrowed from llamacpp, and it replaces the old conversion script. As such, no code changes should be done there 

Resolves #2351 

**Checklist:**


src/instructlab/cli/model/train.py
```diff
@@ -515,6 +515,7 @@ def train(
         from instructlab.model import simple_train
 
         try:
+            print("here!!")
```
nathan-weinberg, line 518: Needs to be removed

src/instructlab/model/convert.py
```diff
@@ -1,127 +1,115 @@
...
-    help="Whether to skip quantization while converting to GGUF.",
-)
-@click.option(
-    "--model-name",
-    help="Name of the model being trained/converted. Informs the naming of the final trained model file",
-    default=None,
-    show_default=True,
+class SUPPORTED_FORMATS(Enum):
+    GGUF = "gguf"
+    SAFETENSORS = "safetensors"
```
nathan-weinberg, line 20: Does this include `.bin` as well?

tests/test_lab_model_convert.py
```diff
@@ -1,98 +0,0 @@
-# Standard
```
nathan-weinberg, line 1: Any reason all of the convert tests are being removed here?

src/instructlab/llamacpp/convert_to_gguf.py
```diff
@@ -0,0 +1,1441 @@
...
+
+    # only used when converting a torch.Tensor to a np.ndarray
+    _dtype_map: dict[torch.dtype, type] = {
+        torch.float16: np.float16,
+        torch.float32: np.float32,
+    }
+
+    # used for safetensors slices
+    # ref: https://github.com/huggingface/safetensors/blob/079781fd0dc455ba0fe851e2b4507c33d0c0d407/bindings/python/src/lib.rs#L1046
+    # TODO: uncomment U64, U32, and U16, ref: https://github.com/pytorch/pytorch/issues/58734
```
courtneypacheco, line 1271: Can we file a GitHub issue to track this?

src/instructlab/llamacpp/convert_to_gguf.py
```diff
@@ -0,0 +1,1441 @@
...
+        tokenizer_path = Path(sys.path[0]) / "models" / f"ggml-vocab-{model_name}.gguf"
+        logger.warning(
+            f"Using tokenizer from '{os.path.relpath(tokenizer_path, os.getcwd())}'"
+        )
+        vocab_reader = gguf.GGUFReader(tokenizer_path, "r")
+
+        default_pre = "mpt" if model_name == "gpt-neox" else "default"
+
+        field = vocab_reader.get_field(gguf.Keys.Tokenizer.MODEL)
+        assert field  # tokenizer model
```
courtneypacheco, line 995: What is this assert for? Was this added during your testing or is this intended to catch an error? If it's the latter, can we raise an error with the appropriate message?

src/instructlab/llamacpp/convert_to_gguf.py
```diff
@@ -0,0 +1,1441 @@
...
+        tokens = []
+        scores = []
+        toktypes = []
+
+        for text, score, toktype in vocab.all_tokens():
+            tokens.append(text)
+            scores.append(score)
+            toktypes.append(toktype)
+
+        assert len(tokens) == vocab.vocab_size
```
courtneypacheco, line 972: Can we replace this `assert` with an exception like

```suggestion
        if len(tokens) != vocab.vocab_size:
            # Raise some error and provide a message letting the user know what happened
```

src/instructlab/model/convert.py
```diff
@@ -1,127 +1,115 @@
...
+
+    @classmethod
+    def register_format_converter(
+        cls, src_fmt: SUPPORTED_FORMATS, tgt_fmt: SUPPORTED_FORMATS
+    ) -> Callable[[AnyConverter], AnyConverter]:
+        def func(convertercls: AnyConverter) -> AnyConverter:
+            cls.converter_dict[(src_fmt, tgt_fmt)] = convertercls
+            return convertercls
+
+        return func
```
courtneypacheco, line 53: Can we add unit tests for this function?

src/instructlab/model/convert.py
```diff
@@ -1,127 +1,115 @@
...
+    @classmethod
+    def get_fmt_converter(
+        cls, src_fmt: SUPPORTED_FORMATS, tgt_fmt: SUPPORTED_FORMATS
+    ) -> type[Converter]:
+        try:
+            return cls.converter_dict[(src_fmt, tgt_fmt)]
+        except KeyError:
+            raise ValueError(
+                f"Conversion from {src_fmt.value} to {tgt_fmt.value} is not supported"
+            ) from None
```
courtneypacheco, line 64: Can we add unit tests for this function too?

src/instructlab/model/convert.py
```diff
@@ -1,127 +1,115 @@
...
-    logger.info(f"deleting {source_model_dir}...")
-    shutil.rmtree(source_model_dir)
-    click.echo("ᕦ(òᴗóˇ)ᕤ Model convert completed successfully! ᕦ(òᴗóˇ)ᕤ")
+        src_model_fmt = get_model_format(model)
+        converter_cls = Converter.get_fmt_converter(src_model_fmt, target_format)
+        converter_instance = converter_cls(model, destination, outtype)
+        converter_instance.convert()
+        logger.info("ᕦ(òᴗóˇ)ᕤ Model convert completed successfully! ᕦ(òᴗóˇ)ᕤ")
+    except ValueError as e:
+        raise ValueError(f"Model conversion failed: {e}") from e
```
courtneypacheco, line 103: Do we have unit tests for this function? Unless I'm missing them

src/instructlab/model/convert.py
```diff
@@ -1,127 +1,115 @@
...
+def get_model_format(model: Path) -> SUPPORTED_FORMATS:
+    if is_model_safetensors(model):
+        return SUPPORTED_FORMATS.SAFETENSORS
+
+    if is_model_gguf(model):
+        return SUPPORTED_FORMATS.GGUF
+
+    raise ValueError(
+        f"Source model {model} is in an unsupported format. Supported formats are {SUPPORTED_FORMATS._member_names_}"
+    ) from None
```
courtneypacheco, line 115: Can we add unit testing for this function as well?

courtneypacheco (commented):
Had some questions about unit testing.

src/instructlab/llamacpp/convert_to_gguf.py
jaideepr97: this is coming from llamacpp's script so I dont think there would be much value in us tracking it

src/instructlab/llamacpp/convert_to_gguf.py
jaideepr97: this is coming from llamacpp's script that we have borrowed so im inclined to leave it as is

src/instructlab/llamacpp/convert_to_gguf.py
jaideepr97: this is coming from llamacpp's script that we have borrowed so im inclined to leave it as is

src/instructlab/model/convert.py
jaideepr97: havent tested with .bin so im going to say not at the moment

tests/test_lab_model_convert.py
jaideepr97: i needed to rewrite them, will add in a follow up commit

src/instructlab/model/convert.py
jaideepr97: added

src/instructlab/model/convert.py
```diff
@@ -1,127 +1,115 @@
...
+        src_model_fmt = get_model_format(model)
+        converter_cls = Converter.get_fmt_converter(src_model_fmt, target_format)
+        converter_instance = converter_cls(model, destination, outtype)
+        converter_instance.convert()
+        logger.info("ᕦ(òᴗóˇ)ᕤ Model convert completed successfully! ᕦ(òᴗóˇ)ᕤ")
+    except ValueError as e:
+        raise ValueError(f"Model conversion failed: {e}") from e
+
+
+def get_model_format(model: Path) -> SUPPORTED_FORMATS:
```
cdoern, line 106: this is a nit, but noticing none of these class functions have docstrings, please add them if you have time

src/instructlab/train/lora_mlx/utils.py
```diff
@@ -161,7 +161,10 @@ def save_model(save_dir: str, weights, tokenizer, config):
 def load(path_or_hf_repo: str):
     # If the path exists, it will try to load model form it
     # otherwise download and cache from the hf_repo and cache
+    print("????")
```
cdoern, line 164: assuming these are for debugging 😉

cdoern (commented):
code looks clean, will approve once unit is passing and debugging prints are removed. This is a nice approach, thanks for taking the time on this

src/instructlab/model/convert.py
```diff
@@ -1,127 +1,115 @@
...
-    os.remove(os.path.join(model_dir_fused_pt, f"{model_name}.gguf"))
-
-    logger.info(f"deleting {source_model_dir}...")
-    shutil.rmtree(source_model_dir)
-    click.echo("ᕦ(òᴗóˇ)ᕤ Model convert completed successfully! ᕦ(òᴗóˇ)ᕤ")
+        src_model_fmt = get_model_format(model)
+        converter_cls = Converter.get_fmt_converter(src_model_fmt, target_format)
+        converter_instance = converter_cls(model, destination, outtype)
+        converter_instance.convert()
+        logger.info("ᕦ(òᴗóˇ)ᕤ Model convert completed successfully! ᕦ(òᴗóˇ)ᕤ")
```
reidliu41, line 101: seems it would be better to share the `destination ` at the end. and also default is "", maybe set a default path would be better.

jaideepr97: updated, thanks!

src/instructlab/cli/model/convert.py
```diff
@@ -0,0 +1,54 @@
...
+    type=click.Choice(["f32", "f16", "bf16", "q8_0", "tq1_0", "tq2_0", "auto"]),
+    default="f16",
+)
+@click.pass_context
+@clickext.display_params
+def convert(
+    ctx,
+    model: Path,
+    destination: Path = Path(""),
+    outtype: str = "f16",
```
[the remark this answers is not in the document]
jaideepr97: updated

cdoern (approved):

src/instructlab/model/simple_train.py
```diff
@@ -156,9 +157,20 @@ def get_files(directory: str, pattern: str) -> list[str]:
             save_every=10,
             steps_per_eval=10,
         )
+
+        source_model_dir = dest_model_dir
+        model_dir_fused = f"{source_model_dir}-fused"
+        # source_model_dir = os.path.expandvars(os.path.expanduser(source_model_dir)).rstrip("/")
```
nathan-weinberg, line 163: Should be removed

cdoern: removed, thanks

src/instructlab/cli/model/convert.py
```diff
@@ -0,0 +1,52 @@
...
+def convert(
+    ctx,
+    model: Path,
+    destination: Path,
+    outtype: str,
+    target_format: str = SUPPORTED_FORMATS.GGUF.value,
+):
+    convert_model(
+        Path(model),
+        Path(destination),
```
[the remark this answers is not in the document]
jaideepr97: yeah, for whatever reason click was still accepting it as a str even after i set the flag type as path 
so i had to add this to make sure it actually was a path at all times

src/instructlab/llamacpp/convert_to_gguf.py
```diff
@@ -0,0 +1,1381 @@
...
+        hparams: dict[str, Any] | None = None,
+    ):
+        if type(self) is Model:
+            raise TypeError(
+                f"{type(self).__name__!r} should not be directly instantiated"
+            )
+
+        self.dir_model = dir_model
+        self.ftype = ftype
+        self.fname_out = fname_out
```
mchoma, line 109:
```suggestion
        self.fname_out = Path(fname_out)
```
To make code more robust and prevent similar error like described in https://github.com/instructlab/instructlab/pull/2950/files#r1975005755

booxter:
AFAIU this work won't move forward, so closing.

PR #2970, opened 2025-01-23
feat: read hf_token from file

reidliu41:
Hugging face CLI login will save the token in default `~/.cache/huggingface/token` which is read it from file.
https://huggingface.co/docs/hub/en/datasets-polars-auth

For security reason or UX(e.g. user might not want to show the token in the command line), it might be better for ilab can read hf token from the file.

```
$ ilab model download --help
Usage: ilab model download [OPTIONS]

Options:
  --hf-token TEXT         User access token for connecting to the Hugging Face
                          Hub (can use environment variable 'HF_TOKEN' or set
                          token in '~/.cache/huggingface/token').
```

**Checklist:**

### DIFF
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -13,7 +13,7 @@ import subprocess
 from huggingface_hub import hf_hub_download, list_repo_files
 from huggingface_hub import logging as hf_logging
 from huggingface_hub import snapshot_download
-from huggingface_hub.errors import GatedRepoError
+from huggingface_hub.errors import GatedRepoError, RepositoryNotFoundError
 
 # First Party
 from instructlab.configuration import DEFAULTS
@@ -27,6 +27,10 @@ from instructlab.utils import (
 )
 
 logger = logging.getLogger(__name__)
+
+HF_TOKEN_MESSAGE = """The HF_TOKEN environment variable needs to be set in your environment to download this Hugging Face Model.
+Alternatively, the token can be passed with --hf-token flag or set it in ~/.cache/huggingface/token.
+The Hugging Face token is used to authenticate your identity to the Hugging Face Hub."""
 
 
 class ModelDownloader(abc.ABC):
@@ -47,6 +51,10 @@ class ModelDownloader(abc.ABC):
     @abc.abstractmethod
     def download(self) -> None:
         """Downloads model from specified repo/release and stores it into download_dest"""
+
+
+class DownloaderAuthError(Exception):
+    """Raised when authentication fails during the download process"""
 
 
 class HFDownloader(ModelDownloader):
@@ -87,12 +95,14 @@ class HFDownloader(ModelDownloader):
             else:
                 self.download_gguf()
 
-        except GatedRepoError as exc:
-            raise ValueError(
-                """The HF_TOKEN environment variable needs to be set in your environment to download this Hugging Face Model.
-                Alternatively, the token can be passed with --hf-token flag.
-                The Hugging Face token is used to authenticate your identity to the Hugging Face Hub."""
-            ) from exc
+        except (GatedRepoError, RepositoryNotFoundError) as exc:
+            raise DownloaderAuthError(HF_TOKEN_MESSAGE) from exc
+        except RuntimeError as exc:
+            if isinstance(
+                exc, RuntimeError
+            ) and "Cannot access gated repo for url https://huggingface.co" in str(exc):
+                raise DownloaderAuthError(HF_TOKEN_MESSAGE) from exc
+            raise RuntimeError from exc
         except Exception as exc:
             raise RuntimeError(
                 f"\nDownloading model failed with the following Hugging Face Hub error:\n{DEFAULT_INDENT}{exc}"
@@ -289,15 +299,14 @@ def download_models(
                 f"\nᕦ(òᴗóˇ)ᕤ {downloader.repository} model download completed successfully! ᕦ(òᴗóˇ)ᕤ\n"
             )
         # pylint: disable=broad-exception-caught
-        except (ValueError, Exception) as exc:
-            if isinstance(exc, ValueError) and "HF_TOKEN" in str(exc):
-                logger.warning(
-                    f"\n{downloader.repository} requires a HF Token to be set.\nPlease use '--hf-token' or 'export HF_TOKEN' to download all necessary models."
-                )
-            else:
-                raise ValueError(
-                    f"\nDownloading model failed with the following error: {exc}"
-                ) from exc
+        except DownloaderAuthError as auth_exc:
+            logger.warning(
+                f"Authentication issue while downloading {downloader.repository}: {auth_exc}"
+            )
+        except Exception as exc:
+            raise ValueError(
+                f"\nDownloading model failed with the following error: {exc}"
+            ) from exc
 
     logger.info("Available models (`ilab model list`):")
     list_and_print_models([Path(model_dir)], False)


src/instructlab/model/download.py
```diff
@@ -304,3 +304,11 @@ def download_models(
     data = list_models([Path(model_dir)], False)
     data_as_lists = [list(item) for item in data]
     print_table(["Model Name", "Last Modified", "Size"], data_as_lists)
+
+
+def read_hf_token_from_file():
+    token_file = os.path.expanduser("~/.huggingface/token")
```
nathan-weinberg, line 310: This is a different path than in the above text

src/instructlab/model/download.py
```diff
@@ -293,7 +293,7 @@ def download_models(
         except (ValueError, Exception) as exc:
             if isinstance(exc, ValueError) and "HF_TOKEN" in str(exc):
                 logger.warning(
-                    f"\n{downloader.repository} requires a HF Token to be set.\nPlease use '--hf-token' or 'export HF_TOKEN' to download all necessary models."
+                    f"\n{downloader.repository} requires a HF Token to be set.\nPlease use '--hf-token' or 'export HF_TOKEN' or set token in '~/.cache/huggingface/token' to download all necessary models."
```
nathan-weinberg, line 296: This is a different path than the below text

src/instructlab/cli/model/download.py
```diff
@@ -57,9 +57,9 @@
 )
 @click.option(
     "--hf-token",
-    default="",
+    default=read_hf_token_from_file(),
```
nathan-weinberg, line 60: @cdoern will this work with our config scheme?

nathan-weinberg (changes requested):

src/instructlab/model/download.py
reidliu41: sorry, use to test and forget to correct it

src/instructlab/cli/model/download.py
reidliu41: seems not a good way to use in `default`, I updated it to `hf_token=read_hf_token_from_file(hf_token)`.

src/instructlab/model/download.py
```diff
@@ -304,3 +304,17 @@ def download_models(
     data = list_models([Path(model_dir)], False)
     data_as_lists = [list(item) for item in data]
     print_table(["Model Name", "Last Modified", "Size"], data_as_lists)
+
+
+def read_hf_token_from_file(hf_token: str) -> str:
```
RobotSail, line 309: If the purpose of the function is to read from the file then we shouldn't be accepting a token as an argument. We should instead have a separate function called something like `get_hf_token` which just retrieves the token in the order of priority as defined by HF. For example, `--hf-token` > `HF_TOKEN` > `~/.cache/huggingface/token` as one possibility. 

And if you go with this pattern then I would just get rid of this function and lump its definition into the `get_hf_token` function.

RobotSail (approved):
One comment that I would like addressed but otherwise LGTM

src/instructlab/model/download.py
reidliu41: sorry, seems a little confused, My first thought just want to add a new way `read from file`, and it only one arg pass, so that function will cover all the ways(cli, envvar, file), so the definition of read_hf_token_from_file, should be not that correct.
cli will handle, --hf-token , HF_TOKEN , and default
so : --hf-token , HF_TOKEN >   ~/.cache/huggingface/token > default

bbrowning:
So, after a HuggingFace API change yesterday that forced our hand today, we had to adjust to not pass invalid or empty Hugging Face tokens into these methods. Specifically, #3074 was merged to adapt to that, and I think that may have incidentally fixed this? The huggingface-hub library should automatically pick up authentication from `~/.cache/huggingface/token` without us actually needing to read that file and pass it in now, since if `HF_TOKEN` is unset we pass `None` into the huggingface-hub library APIs and they'll automatically fallback to the filesystem path. So, it needs confirmation, but I believe the underlying issue of this PR is fixed but this PR itself should not be merged, as we leave the reading of the token to huggingface libraries instead of doing it ourselves.

bbrowning:
Added a hold label as this needs additional investigation to determine if it's still applicable as outlined in my comment above.

reidliu41:
Hi, thank you so much for your feedback.
But as I check, it should just save and read from `~/.cache/huggingface/token`  by `huggingface-cli`, we directly call `hf_hub_download ` and `snapshot_download `, it should `cannot` read from the file. Please correct me if I am wrong.
e.g.
```
>>> huggingface-cli login --token $HF_TOKEN --add-to-git-credential
Token is valid (permission: write).
The token `token_name` has been saved to /home/wauplin/.cache/huggingface/stored_tokens
Your token has been saved in your configured git credential helpers (store).
Your token has been saved to /home/wauplin/.cache/huggingface/token
Login successful
The current active token is: `token_name`
```

bbrowning:
So, if you look at the huggingface code, you'll see that [`hf_hub_download` calls `build_hf_headers`](https://github.com/huggingface/huggingface_hub/blob/v0.28.1/src/huggingface_hub/file_download.py#L822-L828) and `snapshot_download` also ultimately calls into the same `build_hf_headers`, although via a few more steps in-between.

The [code comments for `build_hf_headers`](https://github.com/huggingface/huggingface_hub/blob/v0.28.1/src/huggingface_hub/utils/_headers.py#L71-L76) explain the various token values allowed, but basically if you pass it `None` then it will automatically attempt to read the token off the disk.

So, in short, now that #3074 has merged, I believe the HuggingFace libraries will automatically read and use the token from the user's local disk as long as the user is not explicitly passing in a `--hf-token` or setting the `HF_TOKEN` environment variable. It's worth confirming that, but we should not need to do any file parsing or loading ourselves of that token.

reidliu41:
Thank you so much for your remind, re-read the code again, can find the function to read from the file. And the pr merged is will required the token.. so this pr no need to do it again.
is it ok to add it to the message in this pr?
e.g.
```
                """A Hugging Face token is required to download models from the Hugging Face Hub. 
                If no token is provided or if the provided token is invalid, the following methods can be used to supply it:

                - **Environment Variable**: Set the `HF_TOKEN` environment variable to your token.
                - **Command Line**: Pass the token directly using the `--hf-token` flag.
                - **Local Cache**: If no token is explicitly provided, it can be automatically read from the local cache 
                e.g., by default, the token is stored in `~/.cache/huggingface/token` if no other settings are provided.  

                The Hugging Face token is used to authenticate your identity to the Hugging Face Hub."""
```

bbrowning:
I think expanding upon the logged message to make it more actionable for the user or more clear about what their options are is very reasonable to do. We don't actually test that the token from the user's disk works - is this something you've been able to verify? It should theoretically but it would be good to know for sure.

reidliu41:
> I think expanding upon the logged message to make it more actionable for the user or more clear about what their options are is very reasonable to do. We don't actually test that the token from the user's disk works - is this something you've been able to verify? It should theoretically but it would be good to know for sure.

==>  Yes， I just found one --> `meta-llama/Llama-2-7b-hf` to test(it really need a valid token).
```
$ ls /Users/xx/.cache/huggingface/token
ls: /Users/xx/.cache/huggingface/token: No such file or directory

 ilab model download -rp meta-llama/Llama-2-7b-hf
INFO 2025-02-04 11:36:30,625 instructlab.model.download:77: Downloading model from Hugging Face:
    Model: meta-llama/Llama-2-7b-hf@main
    Destination: /Users/xx/.cache/instructlab/models
Downloading failed with the following exception:
Downloading model failed with the following error:
Downloading model failed with the following Hugging Face Hub error:

Downloading safetensors model failed with the following Hugging Face Hub error:
    401 Client Error. (Request ID: Root=1-67a18b42-291b61c0095916d71f78277a;28157fff-78de-47c1-a6b2-16d1b2c97f68)

Cannot access gated repo for url https://huggingface.co/meta-llama/Llama-2-7b-hf/resolve/01c7f73d771dfac7d292323805ebc428287df4f9/Responsible-Use-Guide.pdf.
Access to model meta-llama/Llama-2-7b-hf is restricted. You must have access to it and be authenticated to access it. Please log in.

Monitor the file
$ sudo fs_usage -w -f filesys | grep "/Users/xx/.cache/huggingface/token"
11:36:30.625209  open                   [  2] (R__________X)  /Users/xx/.cache/huggingface/token              <<---                                                                                                              0.000003   Python.156296
11:36:31.606580  open                   [  2] (R__________X)  /Users/xx/.cache/huggingface/token        <<---   even not exist


$ ilab model download -rp meta-llama/Llama-2-7b-hf --hf-token xxxx    # <<--- wrong token, and will not read the file, but  still error
INFO 2025-02-04 11:38:30,747 instructlab.model.download:77: Downloading model from Hugging Face:
    Model: meta-llama/Llama-2-7b-hf@main
    Destination: /Users/xxx/.cache/instructlab/models
Downloading failed with the following exception:
Downloading model failed with the following error:
Downloading model failed with the following Hugging Face Hub error:
    401 Client Error. (Request ID: Root=1-67a18bb9-471f7e36721dfdfc02d6a1c6;810bdabc-afdb-4aac-8be6-bebd1af5ac56)

Repository Not Found for url: https://huggingface.co/api/models/meta-llama/Llama-2-7b-hf/tree/main?recursive=True&expand=False.
Please make sure you specified the correct `repo_id` and `repo_type`.
If you are trying to access a private or gated repo, make sure you are authenticated.
Invalid credentials in Authorization header


Correct token in file and not pass via --hf-token
$ ls ~/.cache/huggingface/token
/Users/xx/.cache/huggingface/token

$ ilab model download -rp meta-llama/Llama-2-7b-hf      ## read from the file, it will download
INFO 2025-02-04 11:41:24,006 instructlab.model.download:77: Downloading model from Hugging Face:
    Model: meta-llama/Llama-2-7b-hf@main
    Destination: /Users/xx/.cache/instructlab/models
Removing incomplete file '/Users/xx/.cache/instructlab/models/meta-llama/Llama-2-7b-hf/.cache/huggingface/download/1x0Bpn8Ok_AelmMUF3e8JJ_7TYw=.525dc349d71fe257fce4098c146446df6fef4247174f351381e4c3214af126f0.incomplete' (hf_transfer=True)
INFO 2025-02-04 11:41:25,728 huggingface_hub.file_download:1528: Removing incomplete file '/Users/xx/.cache/instructlab/models/meta-llama/Llama-2-7b-hf/.cache/huggingface/download/1x0Bpn8Ok_AelmMUF3e8JJ_7TYw=.525dc349d71fe257fce4098c146446df6fef4247174f351381e4c3214af126f0.incomplete' (hf_transfer=True)
Downloading 'Responsible-Use-Guide.pdf' to '/Users/xx/.cache/instructlab/models/meta-llama/Llama-2-7b-hf/.cache/huggingface/download/1x0Bpn8Ok_AelmMUF3e8JJ_7TYw=.525dc349d71fe257fce4098c146446df6fef4247174f351381e4c3214af126f0.incomplete'
INFO 2025-02-04 11:41:25,729 huggingface_hub.file_download:1536: Downloading 'Responsible-Use-Guide.pdf' to '/Users/xx/.cache/instructlab/models/meta-llama/Llama-2-7b-hf/.cache/huggingface/download/1x0Bpn8Ok_AelmMUF3e8JJ_7TYw=.525dc349d71fe257fce4098c146446df6fef4247174f351381e4c3214af126f0.incomplete'
Responsible-Use-Guide.pdf: 100%|███████████████████████████████| 1.25M/1.25M [00:00<00:00, 1.55MB/s]
```

==>  And another thing is that `GatedRepoError`, seems cannot cover the `token` error:
```
$ ilab model download -rp meta-llama/Llama-2-7b-hf
INFO 2025-02-04 11:49:41,810 instructlab.model.download:77: Downloading model from Hugging Face:
    Model: meta-llama/Llama-2-7b-hf@main
    Destination: /Users/xx/.cache/instructlab/models
--------> type exc <class 'RuntimeError'>                ## <<<<<<<<<<<<<<<< if not provided token, should be RuntimeError 
Downloading failed with the following exception:
Downloading model failed with the following error:
Downloading model failed with the following Hugging Face Hub error:

Downloading safetensors model failed with the following Hugging Face Hub error:
    401 Client Error. (Request ID: Root=1-67a18e59-5a8d964e0b193afe367e89dc;5d227fcf-836c-402f-9dcd-2107f477fdd1)

Cannot access gated repo for url https://huggingface.co/meta-llama/Llama-2-7b-hf/resolve/01c7f73d771dfac7d292323805ebc428287df4f9/generation_config.json.
Access to model meta-llama/Llama-2-7b-hf is restricted. You must have access to it and be authenticated to access it. Please log in.


$ ilab model download -rp meta-llama/Llama-2-7b-hf --hf-token xxxx # <<<<
INFO 2025-02-04 11:47:48,518 instructlab.model.download:77: Downloading model from Hugging Face:
    Model: meta-llama/Llama-2-7b-hf@main
    Destination: /Users/xxx/.cache/instructlab/models
--------> type exc <class 'huggingface_hub.errors.RepositoryNotFoundError'>   ## <<<<<<<<<<<<<<<< if provided wrong token, should be RepositoryNotFoundError

Downloading failed with the following exception:
Downloading model failed with the following error:
Downloading model failed with the following Hugging Face Hub error:
    401 Client Error. (Request ID: Root=1-67a18de7-408a9111328df7756f28b721;25d58fc5-c941-4e36-aa33-e28b8411804b)

Repository Not Found for url: https://huggingface.co/api/models/meta-llama/Llama-2-7b-hf/tree/main?recursive=True&expand=False.
Please make sure you specified the correct `repo_id` and `repo_type`.
If you are trying to access a private or gated repo, make sure you are authenticated.
Invalid credentials in Authorization header
```

src/instructlab/model/download.py
```diff
@@ -289,10 +289,13 @@ def download_models(
                 f"\nᕦ(òᴗóˇ)ᕤ {downloader.repository} model download completed successfully! ᕦ(òᴗóˇ)ᕤ\n"
             )
         # pylint: disable=broad-exception-caught
-        except (ValueError, Exception) as exc:
-            if isinstance(exc, ValueError) and "HF_TOKEN" in str(exc):
+        except (ValueError, RuntimeError, Exception) as exc:
```
booxter, line 292: I know this pattern of matching against the exception string existed before this patch, but it's a bit unfortunate we proliferate it further here. I see that `RuntimeError` can be raised by OCI Downloader, and if by chance it is ever updated to return "You must have access to it and be authenticated" in its message, the code will start logging about HF_TOKEN that has nothing to do about OCI. :-| I know this is kinda theoretical since we control both sides of the call chain (the code that raises and this code that catches it), but this indicates a problem in how Downloader classes return errors. I think the proper fix that would avoid this problem would be to make the downloader classes raise appropriate instructlab specific exceptions (of specific types like DownloaderAuthError) that are then caught by the caller *as a type*. What do you think?

PR #2368, opened 2024-10-02
chore: bump llama_cpp_python to 0.3.1

Ian321:
Instructlab had been using a outdated version of llama_cpp_python that did not support models such as Mixtral NeMo. This PR simply bumps the version of that dependency to the latest one.

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [x] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [x] Documentation has been updated, if necessary.
- [x] Unit tests have been added, if necessary.
- [x] Functional tests have been added, if necessary.
- [x] E2E Workflow tests have been added, if necessary.

PR #2972, opened 2025-01-23
Update llama_cpp_python to 0.3.6

fabiendupont:
While building instructlab 0.23.0a0 from source on aarch64 with CUDA, I noticed compilation errors. Empirically testing with llama_cpp_python 0.3.6, which is the current latest version, the issues disappeared.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Breaking Changes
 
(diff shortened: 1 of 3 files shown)


cdoern:
This might work, 3.5 has a bunch of issues causing us to only bump to 3.2, but 3.6 seems to have possibly solved them all?

nathan-weinberg:
There is already a PR for this: https://github.com/instructlab/instructlab/pull/2368

Issue #2861, opened 2025-01-06
CI enhancement: Use GitHub variables to store AWS subnet and security group IDs


courtneypacheco:
**Is your feature request related to a problem? Please describe.**

The CI workflow currently utilizes AWS resources to launch its e2e, unit test, etc. jobs, and references some of those resources via their exact IDs (e.g., subnet ID, security group ID, etc.) See example:

https://github.com/instructlab/instructlab/blob/5085f63f03d62959609a5041a48b55429eecb334/.github/workflows/e2e-nvidia-l4-x1.yml#L77L78

However, given that AWS resource IDs tend to change over time (due to resources being destroyed, recreated with new configs, etc.), we should ideally use GitHub variable names to replace those IDs and update the variables' values as needed.

**Describe the solution you'd like**

Replace all AWS resource IDs in the CI files with GitHub variable names for ease of maintenance and clarity.

**Additional context**

N/A

PR #2975, opened 2025-01-23
fix: Add fallback logic to select a new AZ if AWS has insufficient capacity for our desired XL e2e instance

### LINKED ISSUE #2974: XL e2e nightly job fails due to insufficient AWS capacity for `g6e.48xlarge` instances in `us-east-2`
## Overview

Our XL e2e nightly job failed last night because Amazon did not have any `g6e.48xlarge` instances available at the time:

Image

## Recommended Solution

We should keep our existing workflow logic to launch a `g6e.48xlarge` instance in our desired subnet on `us-east-2`. However, if that instance fails to launch due to capacity constraints, we should then fall back and select another subnet.

We appear to have 3 subnets available to us within the `us-east-2` region, and we can try the remaining two regions we have not touched.

Worst case scenario, we can create a follow-up issue to investigate the possibility of trying other regions too (as fallbacks). We just must be aware that different regions tend to have different pricing models.

courtneypacheco:
Resolves #2974

**Checklist:**

This workflow update enables us to fall back on a different AWS availability zone when provisioning an EC2 instance fails due to insufficient capacity in a specific availability zone. See example of this PR working as intended:

Screenshot 2025-01-23 at 10 37 06 AM

Logs: https://github.com/instructlab/instructlab/actions/runs/12933687455/job/36072889297

### DIFF
--- a/.github/workflows/e2e-nvidia-l40s-x8.yml
+++ b/.github/workflows/e2e-nvidia-l40s-x8.yml
@@ -18,9 +18,11 @@ env:
 jobs:
   start-xlarge-ec2-runner:
     runs-on: ubuntu-latest
+    env:
+      INSTANCE_TYPE: "g6e.48xlarge"
     outputs:
-      label: ${{ steps.start-ec2-runner.outputs.label }}
-      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+      label: ${{ steps.selected-availability-zone.outputs.label }}
+      ec2-instance-id: ${{ steps.selected-availability-zone.outputs.ec2-instance-id }}
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -39,14 +41,39 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ${{ vars.AWS_REGION }}
 
-      - name: Start EC2 runner
-        id: start-ec2-runner
+      # Attempt to launch the EC2 runner on us-east-2a. If it fails to launch due to insufficient capacity
+      # for this instance, we will continue and try another availability zone within this region.
+      - name: Attempt to start EC2 runner on us-east-2a
+        id: start-ec2-runner-us-east-2a
         uses: machulav/ec2-github-runner@fcfb31a5760dad1314a64a0e172b78ec6fc8a17e # v2.3.6
+        continue-on-error: true
         with:
           mode: start
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
           ec2-image-id: ${{ vars.AWS_EC2_AMI }}
-          ec2-instance-type: g6e.48xlarge
+          ec2-instance-type: ${{ env.INSTANCE_TYPE }}
+          subnet-id: subnet-02d230cffd9385bd4
+          security-group-id: sg-06300447c4a5fbef3
+          iam-role-name: instructlab-ci-runner
+          aws-resource-tags: >
+            [
+              {"Key": "Name", "Value": "instructlab-ci-github-xlarge-runner"},
+              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"},
+              {"Key": "GitHubRef", "Value": "${{ github.ref }}"},
+              {"Key": "GitHubPR", "Value": "${{ github.event.number }}"}
+            ]
+
+      # Fallback in case of failure
+      - name: Attempt to start EC2 runner on us-east-2b (if us-east-2a failed)
+        if: steps.start-ec2-runner-us-east-2a.outcome == 'failure'
+        id: start-ec2-runner-us-east-2b
+        uses: machulav/ec2-github-runner@fcfb31a5760dad1314a64a0e172b78ec6fc8a17e # v2.3.6
+        continue-on-error: true
+        with:
+          mode: start
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id: ${{ vars.AWS_EC2_AMI }}
+          ec2-instance-type: ${{ env.INSTANCE_TYPE }}
           subnet-id: subnet-024298cefa3bedd61
           security-group-id: sg-06300447c4a5fbef3
           iam-role-name: instructlab-ci-runner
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-nvidia-l40s-x8.yml
```diff
@@ -39,14 +41,17 @@ jobs:
...
+        id: start-ec2-runner-us-east-2b
         uses: machulav/ec2-github-runner@fcfb31a5760dad1314a64a0e172b78ec6fc8a17e # v2.3.6
+        continue-on-error: true
         with:
           mode: start
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
           ec2-image-id: ${{ vars.AWS_EC2_AMI }}
-          ec2-instance-type: g6e.48xlarge
+          ec2-instance-type: ${{ env.INSTANCE_TYPE }}
           subnet-id: subnet-024298cefa3bedd61
```
kami619, line 55: @courtneypacheco would we want to handle these subnet-id and security-group-id values without exposing them to public, as a general security best practice ? it could be a env variable or a secret if we want to keep it more tightened, WDYT?

nathan-weinberg: A good idea but I think outside of the scope of what this PR is trying to achieve

nathan-weinberg: In fact there is already an issue for this: https://github.com/instructlab/instructlab/issues/2861

kami619: fair point @nathan-weinberg. Thanks for pointing to the original ticket.

courtneypacheco:
> Good idea/addition - is there a reason we go B->A->C? Wouldn't A->B->C be more intuitive?

@nathan-weinberg No. I agree A->B->C would be more intuitive, but I was just using "B" by default because that's what we were using by default before. I wasn't sure if there was a particular motive/reason behind choosing B over the other availability zones, so I just kept it as the default place to start.

I can update to use A->B->C though if you prefer that

.github/workflows/e2e-nvidia-l40s-x8.yml
```diff
@@ -58,6 +63,66 @@ jobs:
...
+              {"Key": "GitHubPR", "Value": "${{ github.event.number }}"}
+            ]
+
+      - name: Determine EC2 availability zone used
+        id: selected-availability-zone
+        run: |
+          # shellcheck disable=SC2086 # False positive error from shellcheck about word splitting/gobbling
+          if [ "${{ steps.start-ec2-runner-us-east-2b.outcome }}" = "success" ]; then
+            echo "label=${{ steps.start-ec2-runner-us-east-2b.outputs.label }}" >> $GITHUB_OUTPUT
+            echo "ec2-instance-id=\"${{ steps.start-ec2-runner-us-east-2b.outputs.ec2-instance-id }}" >> $GITHUB_OUTPUT
```
bbrowning, line 117: This line is different from the others echoing ec2-instance-id - `ec2-instance-id=\"${{ ...` vs `ec2-instance-id=${{ ...`

They should all be the same, I think? But it's hard for me to guess just by looking whether we need the escaped quote at the beginning or not. However, if we do need the escaped quote at the beginning, we'd also need one at the end? Is this perhaps what the shellcheck warning was erroring about, and maybe it wasn't a false positive after all?

bbrowning (dismissed):
Requesting one change where I see some inconsistent quoting in the selected-availability-zone script.

.github/workflows/e2e-nvidia-l40s-x8.yml
courtneypacheco: ohhh, forgot to clean that up during testing. *doh* thanks! will remove the escaped chars

.github/workflows/e2e-nvidia-l40s-x8.yml
courtneypacheco: I'd prefer to assign all of our AWS "configs" (subnet IDs, SG IDs, instance types, etc.) to variables, except in the few cases where it might not make sense.

nathan-weinberg:
> > Good idea/addition - is there a reason we go B->A->C? Wouldn't A->B->C be more intuitive?
> 
> @nathan-weinberg No. I agree A->B->C would be more intuitive, but I was just using "B" by default because that's what we were using by default before. I wasn't sure if there was a particular motive/reason behind choosing B over the other availability zones, so I just kept it as the default place to start.
> 
> I can update to use A->B->C though if you prefer that

Yes, let's do that please

danmcp:
> > > Good idea/addition - is there a reason we go B->A->C? Wouldn't A->B->C be more intuitive?
> > 
> > 
> > @nathan-weinberg No. I agree A->B->C would be more intuitive, but I was just using "B" by default because that's what we were using by default before. I wasn't sure if there was a particular motive/reason behind choosing B over the other availability zones, so I just kept it as the default place to start.
> > I can update to use A->B->C though if you prefer that
> 
> Yes, let's do that please

Note that not all the zones have historically had the instance types needed (specifically g6e instances).  So there might be some logic to use a particular order if we know some zones will almost certainly fail.

courtneypacheco:
> > > > Good idea/addition - is there a reason we go B->A->C? Wouldn't A->B->C be more intuitive?
> > > 
> > > 
> > > @nathan-weinberg No. I agree A->B->C would be more intuitive, but I was just using "B" by default because that's what we were using by default before. I wasn't sure if there was a particular motive/reason behind choosing B over the other availability zones, so I just kept it as the default place to start.
> > > I can update to use A->B->C though if you prefer that
> > 
> > 
> > Yes, let's do that please
> 
> Note that not all the zones have historically had the instance types needed (specifically g6e instances). So there might be some logic to use a particular order if we know some zones will almost certainly fail.

Makes sense. Though in this case, I confirmed the instance exists us-east-2a, as evidenced in the screenshot.

bbrowning:
My requested change was fixed, but I dismissed my review instead of toggling it to approval because I don't want to approve this and cause it to auto-merge on behalf of the other reviewers without their input. Thanks for fixing the quoting issue!

nathan-weinberg:
> My requested change was fixed, but I dismissed my review instead of toggling it to approval because I don't want to approve this and cause it to auto-merge on behalf of the other reviewers without their input. Thanks for fixing the quoting issue!

You can use the `hold` label in these circumstances :+1:

nathan-weinberg (approved):

danmcp (approved):
Cool change!

PR #2977, opened 2025-01-23
feat: add ability to generate OCI artifact from local models 

jaideepr97:
This PR adds a new converter function to convert safetensors models to OCI artifact format before it can be pushed via `ilab model upload`

This PR builds on top of #2950 

**Checklist:**

### DIFF
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -201,7 +201,9 @@ class _serve_vllm(BaseModel):
     gpus: Optional[int] = Field(default=None, description="Number of GPUs to use.")
     # arguments to pass into vLLM process
     vllm_args: list[str] | None = Field(
-        default_factory=list,
+        # adding type: ignore here to workaround this issue: https://github.com/python/mypy/issues/18191
+        # above bug is triggered by https://github.com/pydantic/pydantic/issues/10950
+        default_factory=list,  # type: ignore
         description="vLLM specific arguments. All settings can be passed as a list of strings, see: https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html",
         examples=[
             ["--dtype", "auto"],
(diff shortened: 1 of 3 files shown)


booxter:
OK, about your mypy failures. tl;dr we'll have to ignore these (add `type: ignore` to affected lines).

---

1. `src/instructlab/configuration.py:204: error: Argument "default_factory" to "Field" has incompatible type "type[list[Any]]"; expected "Callable[[], Never] | Callable[[dict[str, Any]], Never]"  [arg-type]`

This is happening because of a [bug](https://github.com/python/mypy/issues/18191) in `mypy` that is [triggered](https://github.com/pydantic/pydantic/issues/10950) by update of `pydantic` to 2.10.x branch.

Your patch adds `olot` dependency that requires at least `pydantic>=2.10.3`, so we cannot consider capping `pydantic` to `<2.10`. We'll have to `type: ignore` the violation.

Please add the links to the two issues above when adding the `ignore`s so we can track when we will be able to remove them.

2. `src/instructlab/configuration.py:1187: error: Too few arguments  [call-arg]`

This is happening because of a [bug](https://github.com/pydantic/pydantic/issues/10945) in `pydantic`. The bug is fixed in main branch but so far not released. We'll have to add `type: ignore` here too.

Please link to the issue and mention that `pydantic>=2.10.7` will solve the issue.

---

I've also noticed that we don't enable `mypy` plugin for `pydantic` in the repo. I will send a patch to do this. It won't solve your problems though, so it's not related.

src/instructlab/configuration.py
```diff
@@ -201,7 +201,7 @@ class _serve_vllm(BaseModel):
     gpus: Optional[int] = Field(default=None, description="Number of GPUs to use.")
     # arguments to pass into vLLM process
     vllm_args: list[str] | None = Field(
-        default_factory=list,
+        default_factory=list,  # type: ignore
```
booxter, line 204: yes but please add comments with links to issues / patches that will allow us to remove these ignores

jaideepr97: yes, added now!
Thank you for the investigative work!

tarilabs:
In my view, being able to push to OCI with Olot is a required capability, can't speak for the other refactor/update in this pr. Commenting for the Stale.

PR #2988, opened 2025-01-24
create a symlink for ilab after installation

reidliu41:
create a symlink for ilab after installation, e.g.
```
ln -s /path/venv/bin/ilab /usr/local/bin/ilab
```
then, no need to source to the venv every time.

**Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
    ```
 
 > [!IMPORTANT]
-> Every `ilab` command needs to be run from within your Python virtual environment. You can enter the Python environment by running the `source venv/bin/activate` command.
+> Every `ilab` command needs to be run from within your Python virtual environment. You can enter the Python environment by running the `source venv/bin/activate` command. Or you can simply create a symbolic link to a directory that is included in your system’s `$PATH`, for example in Linux: `mkdir -p ~/bin/ && ln -s /path/venv/bin/ilab ~/bin/ilab`, or use an alias: `alias ilab='/path/venv/bin/ilab'` (add it to `~/.bashrc` or `~/.zshrc` to persist).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -225,7 +225,7 @@ For an overview of the full workflow, see the [workflow diagram](./docs/workflow
    ```
 
 > [!IMPORTANT]
-> Every `ilab` command needs to be run from within your Python virtual environment. You can enter the Python environment by running the `source venv/bin/activate` command.
+> Every `ilab` command needs to be run from within your Python virtual environment. You can enter the Python environment by running the `source venv/bin/activate` command. Or you can simply create a symbolic link to point to the system's executable directory, making it easy to use, e.g. `ln -s /path/venv/bin/ilab /usr/local/bin/ilab`.
```
booxter, line 228: FYI This won't work for non-root user (or whoever doesn't have access to this directory). The directory may not even exist. Another option that wouldn't require root access would be defining a shell alias. (I am not convinced we should document all the different ways platforms may work though.)

reidliu41: Thank you so much for your feedback and shell alias suggestion.
yes, it cannot cover all the situation, just try to provide the better `UX`,  actually, I think I just want to mention:
1.  system's executable directory, like linux user 
2.  and just example
how about??
```
Or you can simply create a symbolic link to a directory that is included in your system’s `$PATH`, for example in Linux: `mkdir -p ~/bin/ && ln -s /path/venv/bin/xx ~/bin/xx`, or use an alias: `alias ilab='/path/venv/bin/ilab'` (add it to `~/.bashrc` or `~/.zshrc` to persist).
```

booxter (approved):
I guess it's better than non-existing ideal world where we have a solution that would apply to all situations.

nathan-weinberg (approved):

PR #2993, opened 2025-01-24
infra: When extracting instance-id, ignore terminated instances

booxter:
Otherwise, multiple ids are extracted, then parsed as a single id. This
makes the script complain about InvalidInstanceID.Malformed:

An error occurred (InvalidInstanceID.Malformed) when calling the
DescribeInstances operation: Invalid id: "instance-id-1 instance-id-2"

**Checklist:**

### DIFF
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -89,14 +89,19 @@ ec2__terminate() {
 
 ec2_calculate_instance_id() {
     if [ -z "$INSTANCE_ID" ]; then
+        # shellcheck disable=SC2016
         INSTANCE_ID="$(aws ec2 describe-instances \
             --filters "Name=tag:Name,Values=$INSTANCE_NAME" \
             --region "$EC2_REGION" \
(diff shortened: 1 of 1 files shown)


booxter:
Yeah sorry that's a bit sloppy indeed. I haven't actually consulted the full list of states before sending this. :) Let me see if query language supports negatives.

bbrowning (approved):
This looks good to me. I had no idea you could do conditionals on the `--query` there, and that's cleaner than having to list all non-terminated states in the `--filters`.

I pulled this PR to test locally and was able to confirm it correctly identifies my non-terminated instance as the desired instance when I have two with the same name. It finds that instance correctly even when it's in `pending` or `stopping` states. And, when I created multiple non-terminated instances with the same name, it printed the new helpful message about setting `$INSTANCE_ID` and I confirmed setting `$INSTANCE_ID` was able to choose which of my non-terminated instances the shell script selects.

Since I was fighting with this yesterday, able to easily reproduce it today, and everything looks good from my manual testing giving this one approval.

danmcp (approved):
Thanks for the fix!

PR #2994, opened 2025-01-24
fix: Set default K for RAG to 3

jwm4:
We've found the following issue with RAG in InstructLab:

- By default, the RAG runtime requests 20 search results (configurable via `--retriever-top-k` parameter with default 20).
- Then it puts all of those search results into a single message in the chat session history
- Then if the chat session history is too large to fit into the context window of the generator model, the chat system discards messages from the chat context (oldest to latest) until the chat session history fits into the context window.

However, in practice 20 results is often bigger than the context window of the generator model.   So the effect is that we throw out *all* search results for all queries -- effectively disabling RAG completely (and also obliterating the rest of the chat context which all came before the search results).

In the future, I think we need to be smarter about not flooding the chat session history with too much content.  For now though, just setting the value to 3 provides a work-around that seems like it works out OK with the default models.  The problem could still arise even with 3 search results if the user overrides the default embedding model and replaces it with one that handles very large chunks of text, especially if the response model is one that has a very small context window.  For now, having a default that works with the default models is a lot better than the alternative.

**Checklist:**

### DIFF
--- a/src/instructlab/defaults.py
+++ b/src/instructlab/defaults.py
@@ -88,7 +88,7 @@ class _InstructlabDefaults:
     GRANITE_EMBEDDING_MODEL_NAME = "ibm-granite/granite-embedding-125m-english"
     DOCUMENT_STORE_NAME = "embeddings.db"
     DOCUMENT_STORE_COLLECTION_NAME = "ilab"
-    RETRIEVER_TOP_K = 20
+    RETRIEVER_TOP_K = 3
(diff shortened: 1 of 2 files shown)


anastasds (approved):

cdoern (approved):

alinaryan (approved):

cdoern:
testdata was not regenerated, pushing that on top

jwm4:
> testdata was not regenerated, pushing that on top

Ah, I was working on that but you got to it first.  Thx!

nathan-weinberg (approved):

PR #3012, opened 2025-01-27
mypy: Enable pydantic plugin

booxter:
This may help us avoid some spurious mypy warnings when it gets confused by pydantic magick.

https://docs.pydantic.dev/latest/integrations/mypy/

Noticed when looking at mypy violations under https://github.com/instructlab/instructlab/pull/2977

**Checklist:**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -181,6 +181,7 @@ from-first = true
 known-local-folder = ["tuning"]
 
 [tool.mypy]
+plugins = ['pydantic.mypy']
 mypy_path = ["src"]
 packages = ["instructlab", "tests"]
 disable_error_code = []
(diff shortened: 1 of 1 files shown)


nathan-weinberg:
What is the functional expectation as a result of this change?

booxter:
@nathan-weinberg sorry should have included the URL to plugin docs in PR description (it was in commit message only). Here it is: https://docs.pydantic.dev/latest/integrations/mypy/

tl;dr it may help mypy not fail on some tricky pydantic code. One example of this would be https://github.com/instructlab/instructlab/pull/2977#issuecomment-2616415717 (point 2) except because there's a bug in pydantic itself and the fix for it is not, yet, released, we have to `type: ignore` it atm. But once we adopt a new pydantic version, we will need the plugin to not fail on that line.

---

On second thought, there are probably more options to tighten checks with the plugin, so I'll explore these first and include in this PR. I will move this to draft for now.

nathan-weinberg (approved):

booxter:
On one hand, it's really a low importance change. On the other hand, I am surprised we can't find a second reviewer to just let it in. :)

RobotSail (approved):

PR #3015, opened 2025-01-27
deps: cap dependencies for 0.23.0 release

cdoern:
**Checklist:**

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,11 +7,11 @@ datasets>=2.18.0
 gguf>=0.6.0
 GitPython>=3.1.42
 httpx>=0.25.0
-instructlab-eval>=0.5.1
+instructlab-eval>=0.5.1,<0.6.0
 instructlab-quantize>=0.1.0
 instructlab-schema>=0.4.2
-instructlab-sdg>=0.7.0
-instructlab-training>=0.7.0
+instructlab-sdg>=0.7.0,<0.8.0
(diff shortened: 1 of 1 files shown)


alinaryan (approved):

booxter:
@cdoern just for my understanding: why do we cap some but not all instructlab-* dependencies?

cdoern:
@booxter we only cap the "main" libraries (SDG, Train, Eval) typically.

booxter:
@cdoern ack but whatever the reasoning behind the cap, it probably applies to other instructlab- libraries, no? (The reasoning I assume is - we don't want to be broken by a later library release. I am not necessarily convinced this is good, but I'm not questioning it here.)

nathan-weinberg:
That reasoning is correct - I think typically, the reason we haven't capped the other libraries is that they don't get bumped at all (or ever, in the case of `instructlab-quantize`) - that being said it's probably worth fleshing out in `release-strategy.md` - I think @courtneypacheco is taking over https://github.com/instructlab/instructlab/pull/1758 from @tiran, maybe we could include more guidance around this there?

booxter:
The way I see it: either these other libraries get bumped, and then we probably want to be guarded against them too; or they don't get bumped, and then a probable cap would not be an issue anyway. So while capping more active libraries is more pressing, capping the rest would be at worst a no-op.

booxter (approved):

PR #3054, opened 2025-01-29
fea: Injected hf_token in test job

dmartinol:
Required to fix https://github.com/instructlab/instructlab/pull/3051

**Checklist:**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,6 +153,8 @@ jobs:
           echo "not a directory" > ~/.local/share/instructlab
 
       - name: Run unit and functional tests with tox
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
           tox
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):

courtneypacheco (approved):

nathan-weinberg:
> The test job will NOT have access to this secret on pull requests, but it will on pushes. Was that the intended change here? In practice it means when the tests run against a PR the hf_token will be an empty string instead of the actual secret value, because we're using a `pull_request` workflow dispatch and that does not get access to the actual secrets from the InstructLab org. Since it won't have access to secrets, would it be better to just hard-code the hf_token to an empty string in the workflow? To make it clear that we don't actually need the secret value, and that any value works?
> 
> Or, if this actually needs the secret value, this test will fail on all PRs opened.

Discussed on Slack

PR #3064, opened 2025-01-30
fix: Skipping docs with same ID by default

### LINKED ISSUE #3062: [RAG][Dev] Fix ID already exists bug in rag ingest
Reported by Jorge Catalan:

Step 1: Converted pdfs. This completed without issue. 

```
(venv) jocatala-mac:converteddocs jcats$ ilab rag convert --input-dir docstoconvert/ --output-dir converteddocs/
```

Step 2: Started ingestion and aborted with ctrl+c

```
(venv) jocatala-mac:converteddocs jcats$ ilab rag ingest --input-dir=../converteddocs/
```

Step 3: Started ingestion again. Errors out 

```
ERROR 2025-01-28 20:07:37,760 instructlab.rag.haystack.document_store_ingestor:74: Ingestion attempt failed: ID 'a661c8e7ed3425460d732c52291e83a012325398747d9091bdb92d0c5c002452' already exists.
```

@dmartinol suggests the following fix:

> Since we’re using these in memory implementation that does not support the drop_old flag, I’d suggest to delete the file before the ingest runs again. Hopefully this can fix the issue (together with the other workaround related to this known issue in Haystack)

dmartinol:
With in-memory implementation of document store, the document ID is generated by hashing the document content (of each chunk). Because of that, if two chunks have the same content, the duplicate ID issue is raised.

This fix overrides the embeddings in the document store every time this happens. 
Note: we could also use the `OVERWRITE` option, which should provide the same result in terms of store content, but we assume that this would have worst performance since it puts unneeded data in the document store.

Resolves #3062 

**Checklist:**

### DIFF
--- a/src/instructlab/rag/haystack/component_factory.py
+++ b/src/instructlab/rag/haystack/component_factory.py
@@ -10,6 +10,7 @@ from haystack.components.embedders import (  # type: ignore
 from haystack.components.retrievers import InMemoryEmbeddingRetriever  # type: ignore
 from haystack.components.writers import DocumentWriter  # type: ignore
 from haystack.document_stores.in_memory import InMemoryDocumentStore  # type: ignore
+from haystack.document_stores.types import DuplicatePolicy  # type: ignore
 
 # First Party
 from instructlab.rag.haystack.components.document_splitter import (
@@ -27,7 +28,8 @@ def create_document_writer(
             document_store_uri=document_store_uri,
             document_store_collection_name=document_store_collection_name,
             drop_old=True,
-        )
+        ),
+        policy=DuplicatePolicy.SKIP,
     )
 
 


dmartinol:
@jwm4

src/instructlab/rag/haystack/component_factory.py
```diff
@@ -27,7 +28,8 @@ def create_document_writer(
             document_store_uri=document_store_uri,
             document_store_collection_name=document_store_collection_name,
             drop_old=True,
-        )
+        ),
+        policy=DuplicatePolicy.SKIP,
```
jwm4, line 32: This seems like a good work-around for now.  In the future, I think we'll want something more sophisticated for handling cases where you have identical text but different metadata (e.g., document title), but that would depend on having a complete strategy around metadata, which we don't really have yet.

jwm4 (approved):
This is a good fix for this bug.

src/instructlab/rag/haystack/component_factory.py
dmartinol: agree, metadata is another topic we need to discuss. not here, unfortunately

nathan-weinberg (approved):

cdoern (approved):

src/instructlab/rag/haystack/component_factory.py
karthikbabuks: I was also hitting this issue last week and was testing changes locally. Do you think including this in configuration for the RAG component makes sense rather than hard-coding it?

dmartinol: Do you think that someone might deliberately configure an option to fail the whole ingestion because of that?
Until we use the in memory store, which calculates IDs by hashing the content of the chunks, I’d say that it doesn’t make sense. Chunks can be identical for different reasons, even from the same document, and the user could use only the SKIP option to complete the process. 
Once we introduce metadata and can distinguish between chunks in different documents, we can review the decision and allow to FAIL or OVERRIDE. In case of FAIL, we also need to provide diagnostic messages to identify the conflicting chunks, otherwise the user will never be able to fix his dataset.

karthikbabuks: Okay, Thanks. This is sufficient for now.

courtneypacheco:
@Mergifyio backport release-v0.23

PR #3065, opened 2025-01-30
fix: Reverting previous change to test job

dmartinol:
Removed injection of HF token.

**Checklist:**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,8 +153,6 @@ jobs:
           echo "not a directory" > ~/.local/share/instructlab
 
       - name: Run unit and functional tests with tox
-        env:
-          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
           tox
 


nathan-weinberg (approved):

nathan-weinberg:
Big credit to @bbrowning for realizing this is not the approach we want to take: https://github.com/instructlab/instructlab/pull/3054#issuecomment-2624955050

bbrowning (approved):
Thanks for the quick fix! If you need a test job that runs against PRs that does require a valid HuggingFace token, we can discuss separately how to do that as it would need to happen in a different test suite run from a different workflow.

dmartinol:
> Thanks for the quick fix! If you need a test job that runs against PRs that does require a valid HuggingFace token, we can discuss separately how to do that as it would need to happen in a different test suite run from a different workflow.

@anastasds do we want to insist with this functional test? wait for quantized and public embedding model instead? other options?

PR #3072, opened 2025-01-31
fix: test_lab_download mocks list_repo_files

jwm4:
We're seeing test_lab_download failing to connect to HF, presumably due to stricter checking of authentication.  It was already using a mock to to the downloading but not to do the listing of files.  So I added mocks for that too.

**Issue resolved by this Pull Request:**
Resolves #3066 

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Functional tests have been added, if necessary.
- [ ] E2E Workflow tests have been added, if necessary.

PR #3074, opened 2025-01-31
fix: Using None for hf_token instead of ""

dmartinol:
This is aimed to fix some unexpected issues with model downloads that we faced recently:
```
Repository Not Found for url: https://huggingface.co/instructlab/granite-7b-lab-GGUF/resolve/main/granite-7b-lab-Q4_K_M.gguf.
Please make sure you specified the correct `repo_id` and `repo_type`.
If you are trying to access a private or gated repo, make sure you are authenticated.
Invalid credentials in Authorization header)
```

Changed the logic to initialize `hf_token` according to HF specification:
```py
   token (Union[bool, str, None], optional): 
```

- Instead of `""` we pass `None`.
- Added a parametrized test
- Anticipating changes from #3072 to unblock broken tests
- Including one linting issue in `convert.py` (not clear why it's not blocking other PRs)

I'll create an issue if needed

**Checklist:**

### DIFF
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -67,7 +67,7 @@ class HFDownloader(ModelDownloader):
             download_dest=download_dest,
         )
         self.filename = filename
-        self.hf_token = hf_token
+        self.hf_token = hf_token or None
 
     def download(self):
         """
@@ -77,7 +77,7 @@ class HFDownloader(ModelDownloader):
             f"Downloading model from Hugging Face:\n{DEFAULT_INDENT}Model: {self.repository}@{self.release}\n{DEFAULT_INDENT}Destination: {self.download_dest}"
         )
 
-        if self.hf_token == "" and "instructlab" not in self.repository:
+        if self.hf_token is None and "instructlab" not in self.repository:
             raise ValueError(
                 """HF_TOKEN var needs to be set in your environment to download HF Model.
                 Alternatively, the token can be passed with --hf-token flag.
--- a/src/instructlab/cli/rag/convert.py
+++ b/src/instructlab/cli/rag/convert.py
@@ -56,6 +56,8 @@ def convert(
     input_dir,
     output_dir,
 ):
+    """Pipeline to convert documents from their original format (e.g., PDF) into Docling JSON format for use by ilab rag ingest"""
+
     if not FeatureGating.feature_available(GatedFeatures.RAG):
         click.echo(
             f"This functionality is experimental; set {FeatureGating.env_var_name}"
@@ -63,7 +65,6 @@ def convert(
         )
         return
 
-    """Pipeline to convert documents from their original format (e.g., PDF) into Docling JSON format for use by ilab rag ingest"""
     # Local
     from ...rag.convert import (
         convert_documents_from_folder,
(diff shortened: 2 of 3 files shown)


src/instructlab/model/download.py
```diff
@@ -67,7 +71,9 @@ def __init__(
             download_dest=download_dest,
         )
         self.filename = filename
-        self.hf_token = hf_token
+        self.hf_token = (
+            False if hf_token == "" and _can_have_empty_token(repository) else hf_token
```
bbrowning, line 75: I don't think we want `False` here, but instead `None`. See the docs at https://huggingface.co/docs/huggingface_hub/v0.28.1/package_reference/hf_api#huggingface_hub.HfApi.list_repo_files - a value of `False` explicitly disables authentication. A value of `None` will default to the user's locally saved token (if any). In the case of a repo that requires no auth they would function the same, but in the case of a repo that requires authentication, we'd be explicitly disabling authentication instead of letting HuggingFace fall back to checking its typical on-disk token locations that get written by commands like `huggingface-cli login
`.

We could probably simplify this to `self.hf_token = hf_token or None`? I don't think we need the `can_have_empty_token` check here - if user passes an empty string we'd just treat that as None. Any other value, we'd use as the token.

bbrowning:
While looking into this, I opened #3075 to track fixing our existing overly-eager `HF_TOKEN` checks that prompt users to provide a token even when the repository doesn't actually need one.

src/instructlab/model/download.py
dmartinol: Thanks. Changed to `None `and added one more test to check failure with `ValueType` exception

bbrowning: The new logic looks reasonable to me - thanks!

bbrowning:
@dmartinol If you rebase this on top of the latest main, the failing Ubuntu tests where it fails to install a package should be fixed. After that rebase, this will hopefully pass all tests and we can get it reviewed/merged?

cdoern:
@Mergifyio rebase

dmartinol:
> @Mergifyio rebase

Someone knows the secret tricks! 😉 Thanks a lot!

bbrowning:
I believe every existing install of InstructLab will be unable to download any models from HuggingFace unless the user has a valid hf_token set. We need to confirm that and determine if this should get backported and released as a z-stream to fix current users.

bbrowning (approved):
The updated token logic looks good to me as well as the new tests to validate the various scenarios of token handling. Giving this one thumbs-up to expedite merging and unblocking our CI.

bbrowning:
Sigh - the macos test failed with the error below, but it's not clear to me if this was some transient network error or something else.
```
  Downloading failed with the following exception: 
  Downloading model failed with the following error: 
  Downloading model failed with the following Hugging Face Hub error:
      
  Downloading GGUF model failed with the following Hugging Face Hub error:
      An error occurred while downloading using `hf_transfer`. Consider disabling HF_HUB_ENABLE_HF_TRANSFER for better error handling.
```

cdoern (approved):

bbrowning:
That macos failure was a transient network issue - something to keep an eye on, but when I reran the macos test job it passed this time.

bbrowning:

PR #3099, opened 2025-01-31
ci: run yamllint on all files

### LINKED ISSUE #3041: CI: Enforce YAML lint checks to run on the entire repo
## Background / Context

Currently, we perform YAML lint checks on some of the files within our repo. See our current `yamllint` definition in `tox.ini`:

https://github.com/instructlab/instructlab/blob/a4402817fc2e506ebb9727c84dafa761ac51c45e/tox.ini#L108L116

...however, we have YAML files elsewhere in the repo that would be nice to run these lint checks on as well.

## Motivation and Value

If we enforce proper YAML formatting at the beginning of a PR build, we are more likely to catch errors earlier on in the pipeline. Catching errors early on can increase developer productivity.

## Desired Outcomes

Enforce all YAML files in the repo to undergo the `yamllint` checks before any E2E job gets kicked off.

ktdreyer:
Prior to this change, `tox yamllint` only linted certain files under `test-data`. Expand this to lint all files in the tree.

Some YAML files are more readable with lines greater than 120 characters. For these files, I've inserted per-file magic override comments. These files get copy-pasted a bit, so I've included the comments within each file.

Resolves #3041

### DIFF
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -83,7 +83,7 @@ runs:
           sudo dnf clean all
           rm -rf /var/cache/dnf*
         else
-          echo "Unrecognized OS '${os_id}'. Skipping large package cleanup, as this logic has not been implemented for ${os_id}."
+          echo "Skipping large package cleanup for OS '${os_id}' (not implemented)."
         fi
       shell: bash
     - name: Free Disk Space MacOS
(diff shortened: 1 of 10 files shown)


nathan-weinberg (approved):
Thanks!

courtneypacheco (approved):
Nice! Expanding this check to all YAML files should be really helpful as our project continues to grow.

PR #3100, opened 2025-01-31
fix: Only require HF_TOKEN when actually needed

### LINKED ISSUE #3075: Don't require me to set HF_TOKEN unless authentication actually fails to HuggingFace
**What is the current experience of ilab you've found issue with?**
Commands like `ilab model download` don't even attempt to download anything from a non-InstructLab HuggingFace repository unless I pass in a `--hf-token` or set the `HF_TOKEN` environment variable. However, there are plenty of non-InstructLab repositories that do not require a token so it's onerous to generate and provide a token when one is not required. Adding to this is that HuggingFace recently made a server-side change where empty string tokens, invalid tokens, etc actually get checked and rejected so I can't just do `--hf-token foo` anymore. Additionally, there are other ways to authentication to HuggingFace such as token files on disk that the HuggingFace libraries we use respect. So, in that case, I can authenticate but still not have to pass a token explicitly here.

**How do you think the described experience should be?**
InstructLab should attempt to download the model I've asked for and only let me know that I need to set a token if HuggingFace actually returns an authentication error. See https://huggingface.co/docs/huggingface_hub/package_reference/utilities#handle-http-errors for their documentation on the types of errors returned that we'd need to handle.

**What is the benefit of the change you are proposing?**
Users will no longer be prompted to set a HuggingFace token unless it's actually needed. We have a bad habit of passing empty or invalid tokens around to workaround this, but recent HuggingFace server-side changes have made it so in some cases that actually fails with a HTTP 401 error and instead we're expected to pass either no token or a valid token. Requiring me to generate and pass a valid token to download an ungated model is not a great experience.

This would be backwards compatible as proposed.

bbrowning:
Instead of assuming everything that doesn't have "instructlab" in it needs a HuggingFace access token, we actually try to download whatever model they have and only warn them to set a token if they really need one.

Resolves #3075

**Checklist:**

### DIFF
--- a/src/instructlab/model/download.py
+++ b/src/instructlab/model/download.py
@@ -13,6 +13,7 @@ import subprocess
 from huggingface_hub import hf_hub_download, list_repo_files
 from huggingface_hub import logging as hf_logging
 from huggingface_hub import snapshot_download
+from huggingface_hub.errors import GatedRepoError
 
 # First Party
 from instructlab.configuration import DEFAULTS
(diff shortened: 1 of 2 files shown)


danmcp (approved):
Great improvement!

bbrowning:
Manually merging this one as its a prerequisite to #3102.

bbrowning:

PR #3101, opened 2025-01-31
Flip to python 3.12

### LINKED ISSUE #1463: Support Python 12

Support for Python 12 is being withdrawn in https://github.com/instructlab/instructlab/pull/1461 for reasons.

This will need to be resolved so py12 support can be brought back.  It's the only real active Python version right now and will be popular for the next few years.

**Additional context**

booxter:
Let's check the blast map!

Closes #1463

TODO:
- docs update
- determine if we want to drop older python releases
- determine which should be the default version for non-targeted jobs (e2e? spellchecks? etc.)

**Checklist:**


booxter:
@mergify rebase

booxter:
CUDA job failure here is because of https://github.com/instructlab/instructlab/issues/3111 which is not related to python 3.12

booxter:
@RobotSail so the NVIDIA L40S x4 job passed. What else is missing / broken that we could validate?

fabiendupont:
@booxter, probably worth rebasing to use prebuilt vLLM wheel and pass the failed test.

RobotSail:
@booxter This test didn't test the codepath of training with `use_dolomite: true`, which is the part that blocked us from migrating last time.

booxter:
@RobotSail thanks for this. Would that be enough to trigger it?

```
diff --git a/src/instructlab/defaults.py b/src/instructlab/defaults.py
index 67f4695e..7afb9eda 100644
--- a/src/instructlab/defaults.py
+++ b/src/instructlab/defaults.py
@@ -123,7 +123,7 @@ class _InstructlabDefaults:
         "lora_alpha": 32,
         "lora_dropout": 0.1,
-        "use_dolomite": False,
+        "use_dolomite": True,
     }
     SUPPORTED_CONTENT_FORMATS = ["json"]


```

RobotSail:
Yeah that should work

RobotSail:
@booxter If you are already running on 4xL40s then make sure to disable lora to do a full fine tune

booxter:
@RobotSail probably a stupid question but! :) how do I do this in CI? Isn't it already disabled for 4xL40s job?

booxter:
We can revive when there's a request and staffing to make this happen.

RobotSail:
@booxter If you want to handle this upgrade, can you not also just try to upgrade or deprecate GPTDolomite? That would be a really useful thing for us to do

booxter:
FYI: functional failure in the latest run is because of infra issue in Quay returning 502. Unrelated.

booxter:
Sigh. I hit the bug with symbols because the workflow files from the Cortney's branch for #2975 didn't include it. I will have to wait for #2975 to land before retrying.

PR #3102, opened 2025-01-31
ci: Don't require secrets in medium e2e test

bbrowning:
This removes the need to have any secrets in the medium e2e test itself, simplifying the setup across repos as multiple repositories run this job and it's tedious to have to sync those secrets in multiple places.

Note that the workflow itself still uses secrets to spin up the EC2 runners, but the actual e2e test job no longer uses any secrets.

### DIFF
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -137,9 +137,6 @@ jobs:
           df -h
 
       - name: Run e2e test
-        env:
-          HF_TOKEN: ${{ secrets.HF_TOKEN }}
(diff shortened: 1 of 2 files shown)


bbrowning:
This should only get merged once #3100 is in, as it depends on that.

danmcp (approved):

danmcp:
Should also be backported to 0.23 with #3100

bbrowning:
Rebased on top of #3100, removing hold label. I expect e2e-medium to be green now without any secrets, but will monitor it.

courtneypacheco (approved):

bbrowning:

PR #3112, opened 2025-02-01
feat: Add new in-house GitHub action to detect secrets exposure

courtneypacheco:
**Checklist:**

## Overview

This GitHub action is used to:
  1. Detect when GitHub secrets would be accidentally exposed through GitHub workflows that auto-trigger on PR builds, and
  2. Notify the PR author and reviewers about the security concerns before the PR gets merged into `main`.

**NOTE:** When we talk about "auto-triggered workflows", we're essentially talking about the small E2E job, medium E2E job, unit tests, etc. because they automatically run any time a given contributor creates a pull request and/or pushes a new commit to an existing pull request. In the cases of the large E2E and x-large E2E jobs, neither of them auto-trigger on PR builds, which means if a PR author or reviewer wants the large/x-large E2E test to run on a PR, a trusted repo maintainer must review the PR's proposed changes and manually trigger the desired E2E job if they do not see any security concerns.

## What is Specifically Analyzed

This action looks to see if any secrets are loaded into Git workflow environment variables, like in this case with top-level environment secrets:

```yaml
name: "Some E2E Job"

 # <<< Exposed secrets at the top level (global) >>>
env:
  SECRET_TOKEN: ${{ secrets.SECRET_TOKEN }}
  SECRET_KEY: ${{ secrets.SECRET_KEY }}

jobs:
  some-job: {}
  some-other-job: {}
```

or secrets embedded into individual steps:

```yaml
name: "Some E2E Job"

jobs:

  some-job:
    steps:
      # <<< Exposed secrets at the step level (local to the job, but still an issue) >>>
      - name: First step
        env:
          SECRET_TOKEN: ${{ secrets.SECRET_TOKEN }}
          SECRET_KEY: ${{ secrets.SECRET_KEY }}
        run: some-cmd

  some-other-job:
    steps:
      # <<< Exposed secrets at the step level (local to the job, but still an issue) >>>
      - name: First step
        env:
          SECRET_TOKEN: ${{ secrets.SECRET_TOKEN }}
          SECRET_KEY: ${{ secrets.SECRET_KEY }}
        uses: some-instructlab-gh-action
```

Since GitHub secrets are defined using the form `${{ secrets.SECRET_NAME }}`, we can iterate through the env blocks in the file and use regular expressions to detect GitHub secrets.

## What is Not Analyzed

This check does not look to see if a user outright copy-pastes a token or a password. In other words, it does NOT detect something like:

```yaml
name: "Some E2E Job"

jobs:
  some-job:
    steps:

      # Password strings are not detected
      - name: Store my password to the GitHub env
        run: echo "MY_PASSWORD=my-super-secret-password" >> $GITHUB_ENV

      # Neither are tokens or API keys, like git tokens
      - name: Store my git token to the GitHub env
        run: echo "git_token=ghp_1234567890" >> $GITHUB_ENV
```

## How it Works

This new in-house GitHub action allows the repository maintainers to either provide a valid path to a Git workflow file (via the `--file` flag) OR provide a directory to read Git workflow files from (via the `--dir` flag). Providing both flags simultaneously is not allowed or supported, and is blocked by the CLI. (Just call the CLI tool each time you want to check a new location.)

The tool then validates that: (1.) it can find the desired YAML file(s), and (2.) that the YAML files are indeed Git workflow files, not any random YAML file. If any valid Git workflow files are found, the tool then parses those workflow files to determine if any GitHub secrets are loaded into environment variables, either at the top-level of the workflow file OR through any one of the jobs' steps.

Example output from this PR's build:

Screenshot 2025-02-01 at 4 21 42 PM

## Additional Info

I named this check `security-lint` because the tool that powers it technically falls under the category of "linters". However, I don't think it falls 100% under the existing "lint" category, as this linter is dedicated more towards security and not formatting etc.. I figure we can put future security lint steps in the same section of the `lint.yml` file later on as well, if we need to add more.

### DIFF
--- a/.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/cli.py
+++ b/.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/cli.py
@@ -1,0 +1,179 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+import argparse
+import sys
+
+# Local
+from .custom_exceptions import (
+    GitWorkflowFilesNotFoundError,
+    ExposedSecretsError,
+)
+from .utils import (
+    find_all_yaml_files_in_directory,
+    find_exposed_github_secrets_in_env_block,
+    find_exposed_github_secrets_in_job,
+    get_workflow_trigger_conditions,
+    is_git_workflow_file,
+    load_workflow_file,
+    workflow_auto_triggers_on_pull_request,
+)
+
+
+def create_parser() -> argparse.ArgumentParser:
+    """
+    Generates a parser for the CLI.
+
+    Returns
+    -------
+    parser: argparse.ArgumentParser
+    """
+    parser = argparse.ArgumentParser(
+        prog="Exposed-Secrets-Detection",
+        description="Attempts to find any potentially exposed secrets, and reports them",
+    )
+
+    # User can pass in a directory or file, hence they both have "required=False", but one must be chosen
+    parser.add_argument(
+        "-d",
+        "--dir",
+        type=str,
+        help="Directory to read Git workflow files from",
+        required=False,
+    )
+    parser.add_argument(
+        "-f",
+        "--file",
+        type=str,
+        help="Path to the Git workflow file to read",
+        required=False,
+    )
+
+    return parser
+
+
+def run_validation(args: dict):
+    """
+    Runs validation steps to detect potentially-exposed secrets. In the process, this function also validates
+    that the user has provided at least "--dir" or "--file" as inputs
+
+    Inputs
+    ------
+    args: dict
+        Dictionary of the user's arguments
+    """
+    read_dir = args.get("dir")
+    file_path = args.get("file")
+
+    if not read_dir and not file_path:
+        print(
+            "This tool requires either: a path to a valid Git workflow file or a directory that contains Git workflow files. Detected neither."
+        )
+        sys.exit(1)
+
+    if read_dir and file_path:
+        raise NotImplementedError(
+            "This tool currently does not support both --dir and --file being passed in simultaneously."
+        )
+
+    if read_dir:
+        print(
+            f"Finding GitHub workflow files under {read_dir} and its sub-directories ..."
+        )
+        git_workflow_files = {}
+        yaml_files = find_all_yaml_files_in_directory(read_dir)
+
+        # For each YAML file we've found, store it in a dictionary for later parsing. We need the yaml file *path* as the
+        # key so we can provide actionable error messages if an exception is caught.
+        workflow_files_found = 0
+        for yaml_file in yaml_files:
+            loaded_file = load_workflow_file(yaml_file)
+
+            # Only parse the file if we've confirmed it's a valid Git workflow file and not a random YAML
+            if is_git_workflow_file(loaded_file):
+                git_workflow_files.update({yaml_file: loaded_file})
+                workflow_files_found += 1
+                print(f"  - Found and loaded workflow file: {yaml_file}")
+
+        if workflow_files_found == 0:
+            raise GitWorkflowFilesNotFoundError(
+                f"YAML files were found under directory '{read_dir}', but none were identified as Git workflow files"
+            )
+
+    if file_path:
+        print(f"Reading Git workflow file {file_path}...")
+        loaded_file = load_workflow_file(file_path)
+
+        # Only parse the file if we've confirmed it's a valid Git workflow file and not a random YAML
+        if not is_git_workflow_file(loaded_file):
+            raise GitWorkflowFilesNotFoundError(
+                f"YAML file '{file_path}' is not recognized as a Git workflow file."
+            )
+        print(f"  - Successfully found and loaded workflow file: {file_path}")
+        git_workflow_files = {file_path: loaded_file}
+
+    # For each workflow file, detect if there are exposed secrets
+    exposed_secrets_by_file = {}
+    print("Reviewing workflow file(s) to find any exposed GitHub secrets...")
+
+    for file_path, file_contents in git_workflow_files.items():
+        # This will find any workflow trigger conditions. If none exist, an error is raised.
+        trigger_conditions = get_workflow_trigger_conditions(file_path, file_contents)
+
+        # If the workflow doesn't automatically trigger on pull requests, then org/repo maintainers don't have to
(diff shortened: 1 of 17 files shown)


.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/cli.py
```diff
@@ -0,0 +1,176 @@
...
+        help="Path to the Git workflow file to read",
+        required=False,
+    )
+
+    return parser
+
+
+def run_validation(args: dict):
+    """
+    Runs validation steps to detect potentially-exposed secrets. In the process, this function also alidates
```
danmcp, line 57: Typo: validates

.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/cli.py
```diff
@@ -0,0 +1,176 @@
...
+
+    Inputs
+    ------
+    args: dict
+        Dictionary of the user's arguments
+    """
+    read_dir = args.get("dir")
+    file_path = args.get("file")
+
+    if not read_dir and not file_path:
```
danmcp, line 68: Nit: Since nothing is catching this error, it might be better to print a message end exit(1)

danmcp: Similar comment for the below errors

.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/cli.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+    if read_dir and file_path:
+        raise NotImplementedError(
+            "This tool currently does not support both --dir and --file being passed in simultaneously."
+        )
+
+    if read_dir:
+        print(
+            f"Finding Git workflow files under {read_dir} and its sub-directories ..."
```
danmcp, line 80: GitHub workflow... ?

danmcp: Similar comment for usage below.

.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/utils.py
```diff
@@ -0,0 +1,248 @@
...
+        The `env` block to parse
+
+    Returns
+    -------
+    exposed_secrets: dict
+        A dictionary which contains the env var names that have secrets, but does
+        not contain the secrets themselves.
+    """
+    # GitHub secrets follow the form: "${{ secrets.SOME_NAME }}"
+    secrets_regexp = re.compile(r"\$\{\{.*secrets.*\}\}")
```
bbrowning, line 115: Note that secrets can also be referenced such as `${{ secrets['SOME_NAME'] }}` instead of the dot-syntax.

See this snippet from https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#about-contexts :

```
As part of an expression, you can access context information using one of two syntaxes.

Index syntax: github['sha']
Property dereference syntax: github.sha
```

bbrowning (commented):
This looks like a great initial pass at preventing us from accidentally exposing secrets in our workflow files, thanks! A couple of comments mostly around the scope of what we want to allow/check for.

There are also ways the `permissions` block can get used to inadvertently expose some things, but I think the nuances of that are different enough that attempting to address it would be quite a different scope and/or a different action entirely.

.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/utils.py
courtneypacheco: Hey Ben! Yep, that's definitely a valid secrets syntax. In regular expressions, `.*` is how we define wildcards, so the `.*` you see is for the wildcard in specific, not for capturing `.` :

```bash
# Captures: ${{ secrets['SOME_NAME'] }}  and ${{ secrets.SOME_NAME }} 
\$\{\{.*secrets.*\}\}

# Captures only ${{ secrets.SOME_NAME }}  because of the `\.`
\$\{\{.*secrets\..*\}\}
```

I can definitely update the unit tests to confirm that `${{ secrets['SOME_NAME'] }}` would also be captured

.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/cli.py
courtneypacheco: Fixed typo

.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/cli.py
courtneypacheco: Changed from exception to: `print()` + `sys.exit(1)`

.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/utils.py
courtneypacheco: @bbrowning  can you look at this test?

https://github.com/instructlab/instructlab/pull/3112/files#diff-d5b1e42c71983c3d621801b7d35c94230047644435e7505c67428c5a95840d15R150-R166

ktdreyer:
I've reviewed this and run `pytest` locally to independently verify it. Great job.

.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/utils.py
```diff
@@ -0,0 +1,248 @@
...
+        List of workflow triggers. For example: ["schedule", "workflow_dispatch", "pull_request"]
+
+    Returns
+    -------
+    True/False
+    """
+    acceptable_manual_triggers = {
+        "schedule",
+        "workflow_dispatch",
+    }
```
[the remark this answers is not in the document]
courtneypacheco: I updated it from an allow-list to a deny-list.

Allow list is technically safer, but Dan did also mention that it's easier to maintain a deny-list. Can you have a look to see if I missed any deny cases?

.github/actions/detect-exposed-secrets/exposed_secrets_detection/src/utils.py
bbrowning: Ahh, perfect! Thanks for being so thorough here :smile:

RobotSail (approved):

booxter (approved):
Impressive work. I love how this is structured and tested.

PR #3131, opened 2025-02-07
fix: Update vLLM to 0.7.2

fabiendupont:
vLLM 0.6.6.post1 has an important CVE:
[CVE-2025-24357](https://nvd.nist.gov/vuln/detail/CVE-2025-24357). The fix is available is vLLM 0.7.0 or higher.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Features
 
 - Update NVIDIA CUDA to version 12.8, in order to support NVIDIA Blackwell GPUs.
-- Update vLLM to version 0.6.6.post1. As a requirement for this new vLLM version, PyTorch is updated to 2.5.1
+- Update vLLM to version 0.7.2. As a requirement for this new vLLM version, PyTorch is updated to 2.5.1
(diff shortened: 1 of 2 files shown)


booxter:
@fabiendupont remind me please what's the rationale of pinning vllm instead of letting it update as new versions are released. (What I wonder is why we have to pin it like this if it looks like it's so easy / supported to change versions. I thought there's something very fragile about vllm that requires adjustments on our side and testing, but looks like you only have to touch requirements file?)

fabiendupont:
@booxter, vLLM has been more brittle in the past, so we pinned the version.
We also pin PyTorch, so limiting vLLM versions sounds reasonable.
But that's definitely something we can challenge.

fabiendupont:
Test failed because machine ran out of space. @nathan-weinberg , @booxter , would you mind relaunching the test?

nathan-weinberg:
> Test failed because machine ran out of space. @nathan-weinberg , @booxter , would you mind relaunching the test?



> Test failed because machine ran out of space. @nathan-weinberg , @booxter , would you mind relaunching the test?

Done - kicked off a large job as well

PR #3132, opened 2025-02-07
vllm: unpin version (bumping to 0.7.0+)

booxter:
Looks like we can bump versions without significant issues. If so, we
should consider relaxing the dependency for the ease of integration in
distributions etc.

The 0.7.0 bump is to reflect the rationale in:
https://github.com/instructlab/instructlab/pull/3131

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Features
 
 - Update NVIDIA CUDA to version 12.8, in order to support NVIDIA Blackwell GPUs.
-- Update vLLM to version 0.7.2. As a requirement for this new vLLM version, PyTorch is updated to 2.5.1
+- Update vLLM to version 0.7.0. As a requirement for this new vLLM version, PyTorch is updated to 2.5.1
 - Use vLLM wheel from PyPI for NVIDIA CUDA to speed up installation of InstructLab.
 - `ilab model chat` now has `--no-decoration` option to display chat responses without decoration.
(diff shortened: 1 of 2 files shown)


booxter:
@fabiendupont would you mind checking what should be done for this to happen? Can it even happen (for product reasons or whatever)?

fabiendupont:
Instead of unpinning, we could also use something like Dependabot to update the versions progressively with 2e2 tests to gate the process.

booxter:
@fabiendupont This would work for our own sake, but I also wonder about how this project is supposed to be packaged downstream where such pins may get in the way of figuring out a single vllm version to package to satisfy all projects that depend on vllm. In Python world, site-packages are often global, so distributors have to pick a version. If ilab depends on version1 and some other project depends on version2, and they don't have overlapping ranges, it makes it painful.

booxter:
@courtneypacheco I think @fabiendupont points out the fact that our requirements define what we CI against. Which is unsafe for uncapped dependencies. AFAIU the proper way to deal with it is to keep dependencies uncapped / with wider ranges, while using a lock / pin file for CI purposes (that is constantly updated by dependabot or similar solution).

Do you think this is something we should have to enable us broadening supported ranges for dependencies?

booxter:
Similar attempts to move the dependency / discussion over caps / pins happening here: https://github.com/instructlab/instructlab/pull/3283 Closing.

PR #3157, opened 2025-02-17
Make SDG batch size configurable via system profile

npalaska:
Currently, the batch size for SDG is only configurable via the CLI, but a single batch size across all hardware profiles is not optimal. Different hardware configurations have varying capabilities, and using a fixed batch size can lead to under-utilization or over-utilization of resources during the SDG process.

To ensure efficient performance across different hardware, we should set the batch sizes independently in each system profile.

### DIFF
--- a/src/instructlab/cli/data/generate.py
+++ b/src/instructlab/cli/data/generate.py
@@ -143,8 +143,8 @@ logger = logging.getLogger(__name__)
 @click.option(
     "--batch-size",
     type=click.IntRange(min=0),
-    default=None,
-    help="Number of elements to process in each batch through the SDG pipeline. Enabled by default for the vLLM serving backend, with a batch size of 8 chosen based on experiments to optimize for throughput. Use 0 to disable.",
+    cls=clickext.ConfigOption,
+    default=DEFAULTS.BATCH_SIZE,
 )
 @click.option(
     "--enable-serving-output",
--- a/src/instructlab/configuration.py
+++ b/src/instructlab/configuration.py
@@ -373,6 +373,10 @@ class _generate(BaseModel):
         default=DEFAULTS.NUM_CPUS,
         description="Number of CPU cores to use for generation.",
     )
+    batch_size: PositiveInt = Field(
+        default=DEFAULTS.BATCH_SIZE,
+        description="Number of Batches to send for generation on each core.",
+    )
     chunk_word_count: PositiveInt = Field(
         default=DEFAULTS.CHUNK_WORD_COUNT,
         description="Maximum number of words per chunk.",
--- a/src/instructlab/defaults.py
+++ b/src/instructlab/defaults.py
@@ -105,6 +105,8 @@ class _InstructlabDefaults:
     MAX_CONTEXT_SIZE = 4096
     # TODO: these constants should be removed, they should not leak out
     NUM_CPUS = 10
+    # Number of batches to send on each core. Tune the batch size to optimize the vLLM performance
+    BATCH_SIZE = 8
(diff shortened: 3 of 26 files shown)


src/instructlab/cli/data/generate.py
```diff
@@ -143,8 +143,8 @@
 @click.option(
     "--batch-size",
     type=click.IntRange(min=0),
-    default=None,
-    help="Number of elements to process in each batch through the SDG pipeline. Enabled by default for the vLLM serving backend, with a batch size of 8 chosen based on experiments to optimize for throughput. Use 0 to disable.",
+    cls=clickext.ConfigOption,
+    default=DEFAULTS.BATCH_SIZE,
```
cdoern, line 147: the default comes from the config when you set `cls=...` so no default is necessary here.

cdoern (commented):
looks good! Technically, if you are using the default value you don't need to add this to all of the profiles manually. One comment about the default in the click flag.

src/instructlab/cli/data/generate.py
npalaska: Thanks @cdoern, hmm I think I am missing something. If I run `ilab data generate --debug-params`. I get all the values of the parameter correct except for the batch size for which I get 
```batch_size: None     	[type: None, src: default]``` (if I remove the default)

Comment (cdoern) [inline src/instructlab/cli/data/generate.py:147]:
```suggestion

```

removing this line and re-installing I get the following:

```
ilab data generate --debug-params
Parameters:
             model_path: '/Users/charliedoern/.cache/instructlab/models/mistral-7b-instruct-v0.2.Q4_K_M.gguf' 	[type: str, src: default_map]
               num_cpus: 10       	[type: int, src: default_map]
       chunk_word_count: 1000     	[type: int, src: default_map]
       num_instructions: -1       	[type: int, src: default]
       sdg_scale_factor: 30       	[type: int, src: default_map]
          taxonomy_path: '/Users/charliedoern/.local/share/instructlab/taxonomy' 	[type: str, src: default_map]
          taxonomy_base: 'origin/main' 	[type: str, src: default_map]
             output_dir: '/Users/charliedoern/.local/share/instructlab/datasets' 	[type: str, src: default_map]
                  quiet: False    	[type: bool, src: default]
           endpoint_url: None     	[type: None, src: default]
                api_key: 'no_api_key' 	[type: str, src: default]
             yaml_rules: None     	[type: None, src: default]
        server_ctx_size: 4096     	[type: int, src: default]
           tls_insecure: False    	[type: bool, src: default]
        tls_client_cert: ''       	[type: str, src: default]
         tls_client_key: ''       	[type: str, src: default]
      tls_client_passwd: ''       	[type: str, src: default]
           model_family: None     	[type: None, src: default]
               pipeline: 'full'   	[type: str, src: default_map]
             batch_size: 8        	[type: int, src: default_map]
  enable_serving_output: False    	[type: bool, src: default]
                   gpus: None     	[type: None, src: default_map]
         max_num_tokens: 4096     	[type: int, src: default_map]
               detached: False    	[type: bool, src: default]
```

npalaska: Thanks it indeed worked after I reinstall it

cdoern:
@mergifyio rebase

cdoern (approved):

eshwarprasadS (approved):
LGTM 🚀

cdoern (approved):

nathan-weinberg (approved):

ktdreyer:
@mergify backport release-v0.24

ktdreyer:
@mergify backport release-v0.23

Issue #2999, opened 2025-01-24
The nightly XL e2e job terminates after exactly 6 hours due to GitHub runner limitations


courtneypacheco:
## Overview

Our nightly XL e2e job terminates after exactly 6 hours of runtime due to GitHub runner limitations around job execution time. For example:

<img width="869" alt="Image" src="https://github.com/user-attachments/assets/fa77ee81-4c63-4dae-97a4-ddf89e5d38f2" />

[Logs here](https://github.com/instructlab/instructlab/actions/runs/12944223752/job/36104955488)

If we look at the GitHub Actions documentation, it states the following about runner execution time:

> Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete.

Source: https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration

## "Root Cause"

Essentially, the total runtime of our XL e2e job has recently increased beyond 6 hours due to us needing to test more features, etc.. Because we need more than 6 hours to complete the job, we need to find a way to allow the XL e2e job to run for more than 6 hours;

## Possible Solutions

Here are some possible solutions, but there may be other ideas that could work better:

1. Self-host our runners so we can keep them running >6 hours
2. Use a GitHub runner trigger the nightly XL E2E job, shut down the GitHub runner after the XL E2E job starts,  then spin up a new GitHub runner every 30 mins to check the status of the job. (For reference: We never launch the XL e2e job during a PR anyway, so if someone needed to debug the XL e2e job, they could technically edit the workflow logic to keep the GitHub runner alive.)

booxter:
Can the job be split into multiple pieces, each taking fewer than 6h?

PR #3166, opened 2025-02-18
Pause the e2e-nvidia-l40s-x8.yml workflow

kami619:
Until we fix the underlying root cause for this https://github.com/instructlab/instructlab/issues/2999 and/or add more time if that becomes the inevitable path.

The change in this PR comments the schedule block of the workflow yaml and left a comment to reflect the change. This becomes easier to turn it back on, than removing the schedule block altogether.

**Checklist:**


courtneypacheco (approved):

booxter (approved):
I hope you have a plan to fix it because we can't let it disabled for a long time. (Can the job be split into pieces that are shorter than 6h?)

PR #3186, opened 2025-02-25
infra: fix cloud-instance ec2 wait loop

### LINKED ISSUE #3185: `cloud-instance.sh ec2 launch` exits before displaying information about running instance
**Describe the bug**
When  I create an AWS instance with`scripts/infra/cloud-instance.sh`, the script creates the instance, but it exits with an error before it can tell me anything about its state or how to access it.

**To Reproduce**
1. Clone `instructlab` from GitHub, `cd scripts/infra`
2. Set up `~/.instructlab/cloud-config`.
3. Ensure I'm logged into AWS (`aws sts get-caller-identity`)
4. Run `./cloud-instance.sh ec2 launch`

**Actual behavior**
The script prints `Waiting for instance to be in the 'running' state...`, then exits with a non-zero exit code

**Expected behavior**
The script shows the user how to log in or what to do next

ktdreyer:
Prior to this change, `./cloud-instance.sh ec2 launch` would fail to print information about the host it was launching.

Bash with `set -e` will bail out on `(( $i++ ))` when `$i` is zero. Start `i` on `1` to avoid this. This also clarifies that this is the "first" logical time the script has polled `describe-instances` before sleeping.

Resolves #3185

### DIFF
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -34,7 +34,7 @@ ec2__launch() {
         echo "Failed to launch instance"
         exit 1
     else
-        local i=0
+        local i=1
         echo "Waiting for instance to be in the 'running' state..."
         while [[ $i -lt $MAX_ITERS ]]; do
             local state


scripts/infra/cloud-instance.sh
```diff
@@ -34,7 +34,7 @@ ec2__launch() {
         echo "Failed to launch instance"
         exit 1
     else
-        local i=0
+        local i=1
```
booxter, line 37: (joking) Ah but now this code waits for 55 seconds, not 12*5! Horrible!

booxter: (can't stop myself) Ah and the last sleep is useless because it will not retry again before claiming defeat! Full 5 seconds wasted before telling the user their VM is not up!

booxter (approved):

scripts/infra/cloud-instance.sh
ktdreyer: I cannot disagree with these comments. My only defense is that the PR makes it not worse :laughing:

courtneypacheco (approved):

ktdreyer:
@mergify backport release-v0.24

PR #3190, opened 2025-02-26
fix(infra): fix nvidia drivers installation script

### LINKED ISSUE #3016: cloud-instance.sh script fails to run install_rh_nvidia_drivers function when there are multiple kernel core releases/versions available
**Describe the bug**

The [cloud-instance.sh](https://github.com/instructlab/instructlab/blob/main/scripts/infra/cloud-instance.sh) utility script fails to run `install_rh_nvidia_drivers` function when there are multiple kernel core releases/versions available.

**To Reproduce**
1. Provision any Cloud/Bare metal server and install a Linux OS distribution.
2. Run through the steps mentioned [here](https://github.com/instructlab/instructlab/tree/main/scripts/infra#cloud-instancesh).
3. Make sure, you install multiple `kernel-core` versions on the host node, before you run the `install_rh_nvidia_drivers` step.
4. Notice the error in the console logs.

**Expected behavior**
We would expect and desire the `cloud-instance.sh` and the `nvidia-setup.sh` scripts to handle the situation when there are multiple kernel-core versions available. It kind of does in this [step](https://github.com/instructlab/instructlab/blob/main/scripts/infra/nvidia-setup.sh#L150), but it could fail even before that over [here](https://github.com/instructlab/instructlab/blob/main/scripts/infra/nvidia-setup.sh#L51).

**Logs**

```
Dependencies resolved.
Nothing to do.
Complete!
+ '[' '' == '' ']'
++ dnf info --installed kernel-core
++ awk -F: '/^Release/{print $2}'
++ tr -d '[:blank:]'
+ RELEASE='412.el9
547.el9'
++ dnf info --installed kernel-core
++ awk -F: '/^Version/{print $2}'
++ tr -d '[:blank:]'
+ VERSION='5.14.0
5.14.0'
+ export 'KERNEL_VERSION=5.14.0
5.14.0-412.el9
547.el9'
+ KERNEL_VERSION='5.14.0
5.14.0-412.el9
547.el9'
+ dnf install -y 'kernel-devel-5.14.0
5.14.0-412.el9
547.el9'
Last metadata expiration check: 0:23:44 ago on Mon 27 Jan 2025 03:54:29 PM EST.
No match for argument: kernel-devel-5.14.0
5.14.0-412.el9
547.el9
Error: Unable to find a match: kernel-devel-5.14.0
5.14.0-412.el9
547.el9
+ CUDA_REPO_ARCH=
+ '[' '' == aarch64 ']'
+ cp -a /etc/dnf/dnf.conf /etc/dnf/dnf.conf.tmp
+ mv /etc/dnf/dnf.conf.tmp /etc/dnf/dnf.conf
+ dnf config-manager --best --nodocs --setopt=install_weak_deps=False --save
+ dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel//cuda-rhel.repo
Adding repo from: https://developer.download.nvidia.com/compute/cuda/repos/rhel//cuda-rhel.repo
Status code: 404 for https://developer.download.nvidia.com/compute/cuda/repos/rhel//cuda-rhel.repo (IP: 23.205.107.71)
Error: Configuration of repo failed
+ systemctl daemon-reload
+ systemctl enable --now nvidia-toolkit-setup.service
Job for nvidia-toolkit-setup.service failed because the control process exited with error code.
See "systemctl status nvidia-toolkit-setup.service" and "journalctl -xeu nvidia-toolkit-setup.service" for details.
```

 - Hardware Specs: IBM Cloud - gx3-24x120x1l40s
 - OS Version: CentOS Stream release 9 (Red Hat Enterprise Linux 9)
 - Python Version: Python 3.11.11
 - InstructLab Version: NA

**Additional context**
We would have to handle the situation where the right kernel-core version is initially used for the login and then handle the deletion of the additional kernel-core versions and then re-install the nvidia drivers or use a pre-loaded image with everything installed.

booxter:
- **fix(infra): honor 2+ kernels when installing nvidia drivers**
- **fix(infra): use -dkms nvidia module**
- **fix(infra): fix nvidia installation due to missing package**

The script was broken due to some changes on the nvidia rpm repo side.

In addition, I'm tackling the problem with the script not handling multiple
kernel versions installed at the same time. In this case, the script will pick
the latest version installed.

Resolves #3016

**Checklist:**

### DIFF
--- a/scripts/infra/nvidia-setup.sh
+++ b/scripts/infra/nvidia-setup.sh
@@ -48,8 +48,8 @@ FILEEOF
 rm -rf yum-packaging-precompiled-kmod
 dnf install libicu podman skopeo git rpm-build make openssl elfutils-libelf-devel python3.11 python3.11-devel -y
 if [ "${KERNEL_VERSION}" == "" ]; then \
-      RELEASE=$(dnf info --installed kernel-core | awk -F: '/^Release/{print $2}' | tr -d '[:blank:]') \
-      && VERSION=$(dnf info --installed kernel-core | awk -F: '/^Version/{print $2}' | tr -d '[:blank:]') \
+      RELEASE=$(dnf info --installed kernel-core | sort | awk -F: '/^Release/{print $2}' | tr -d '[:blank:]' | tail -n 1) \
+      && VERSION=$(dnf info --installed kernel-core | sort | awk -F: '/^Version/{print $2}' | tr -d '[:blank:]' | tail -n 1) \
       && export KERNEL_VERSION="${VERSION}-${RELEASE}" ;\
 fi \
     && dnf install -y "kernel-devel-${KERNEL_VERSION}" \
--- a/scripts/infra/nvidia-setup.sh
+++ b/scripts/infra/nvidia-setup.sh
@@ -121,7 +121,7 @@ fi \
         tmux \
         nvidia-driver-cuda-${DRIVER_VERSION} \
         nvidia-driver-libs-${DRIVER_VERSION} \
-        nvidia-driver-NVML-${DRIVER_VERSION} \
+        libnvidia-ml-${DRIVER_VERSION} \
         cuda-compat-${CUDA_DASHED_VERSION} \
         cuda-cudart-${CUDA_DASHED_VERSION} \
         cuda-cudart-devel-${CUDA_DASHED_VERSION} \
--- a/scripts/infra/nvidia-setup.sh
+++ b/scripts/infra/nvidia-setup.sh
@@ -113,7 +113,7 @@ fi \
     && cp -a /etc/dnf/dnf.conf{,.tmp} && mv /etc/dnf/dnf.conf{.tmp,} \
     && dnf config-manager --best --nodocs --setopt=install_weak_deps=False --save \
     && dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel${OS_VERSION_MAJOR}/${CUDA_REPO_ARCH}/cuda-rhel${OS_VERSION_MAJOR}.repo \
-    && dnf -y module enable nvidia-driver:${DRIVER_STREAM}/default \
+    && dnf -y module enable nvidia-driver:${DRIVER_STREAM}-dkms/default \
     && export NCCL_PACKAGE=$(dnf search libnccl --showduplicates 2>/dev/null | grep ${CUDA_MAJOR_MINOR} | awk '{print $1}' | grep libnccl-2 | tail -1) \
     && dnf install -y \
         cloud-init \
@@ -154,7 +154,7 @@ fi \
     && if [ "$DRIVER_TYPE" != "vgpu" ] && [ "$TARGET_ARCH" != "arm64" ]; then \
         versionArray=(${DRIVER_VERSION//./ }); \
         DRIVER_BRANCH=${versionArray[0]}; \
-        dnf module enable -y nvidia-driver:${DRIVER_BRANCH} && \
+        dnf module enable -y nvidia-driver:${DRIVER_BRANCH}-dkms && \
         dnf install -y nvidia-fabric-manager-${DRIVER_VERSION} libnvidia-nscq-${DRIVER_BRANCH}-${DRIVER_VERSION} ; \
     fi \
     && . /etc/os-release && if [ "${ID}" == "rhel" ]; then \


booxter:
@ktdreyer FYI

(Sorry for stepping on your toes, I needed this script fixed myself, so I couldn't wait for your fix...)

scripts/infra/nvidia-setup.sh
```diff
@@ -113,15 +113,15 @@ fi \
...
+    && dnf -y module enable nvidia-driver:${DRIVER_STREAM}-dkms/default \
     && export NCCL_PACKAGE=$(dnf search libnccl --showduplicates 2>/dev/null | grep ${CUDA_MAJOR_MINOR} | awk '{print $1}' | grep libnccl-2 | tail -1) \
     && dnf install -y \
         cloud-init \
         pciutils \
         tmux \
         nvidia-driver-cuda-${DRIVER_VERSION} \
         nvidia-driver-libs-${DRIVER_VERSION} \
-        nvidia-driver-NVML-${DRIVER_VERSION} \
+        libnvidia-ml-${DRIVER_VERSION} \
```
courtneypacheco, line 124: I understand the DKMS changes since you need the DKMS versions to support multiple kernels, but I am not quite understanding why we removing the NVML driver package and replacing it with `libnvidia-ml`. Do we not need  `nvidia-driver-NVML` anymore? I know that `libnvidia-ml` will install NVML, which is necessary for certain `nvidia-smi` functionality. Were we simply pulling in `nvidia-driver-NVML` this whole time when we should have been pulling in `livnvidia-ml` specifically? Just making sure we don’t “need” this particular driver package.

booxter: Also perhaps helpful if you haven't seen it in the commit message: https://github.com/NVIDIA/yum-packaging-nvidia-driver/blob/d06f50a507eb6e053e2e7dd4fcd7781060f8ff06/nvidia-driver.spec#L117

scripts/infra/nvidia-setup.sh
```diff
@@ -113,15 +113,15 @@ fi \
     && cp -a /etc/dnf/dnf.conf{,.tmp} && mv /etc/dnf/dnf.conf{.tmp,} \
     && dnf config-manager --best --nodocs --setopt=install_weak_deps=False --save \
     && dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel${OS_VERSION_MAJOR}/${CUDA_REPO_ARCH}/cuda-rhel${OS_VERSION_MAJOR}.repo \
-    && dnf -y module enable nvidia-driver:${DRIVER_STREAM}/default \
+    && dnf -y module enable nvidia-driver:${DRIVER_STREAM}-dkms/default \
```
danmcp, line 116: Is this the right change or should we be using a different version?

danmcp:
@fabiendupont FYI

scripts/infra/nvidia-setup.sh
booxter: I assume `-open` is the new driver: https://github.com/NVIDIA/open-gpu-kernel-modules and `-dkms` is the legacy / old / previous one. Correct? I assume we want to stick to the previous version we used which - I assume - is the `-dkms` one, and if `-open` is a better alternative now, then a separate change would be sent to do the switch. I don't know much about the nvidia drivers to make the decision either way, so sticking to the status quo here.

Does it make sense?

ktdreyer:
I haven't used dkms in a decade, so I'm learning.

I've tested this manually on [CentOS's](https://www.centos.org/download/aws-images/) `ami-09fc0e32ec75bfe3f`.
 - Running this code with single `kernel-5.14.0-565.el9` installed: works
 - Running this code with single `kernel-5.14.0-565.el9` installed, then: `dnf update && reboot` into `5.14.0-570.el9`. Driver does not load. I thought dkms dynamically recompiled for us on kernel updates?

scripts/infra/nvidia-setup.sh
danmcp: AFAIU, the previous working one was 550 (not dkms or open):

```
sudo dnf module list nvidia-driver
Last metadata expiration check: 0:58:29 ago on Wed 26 Feb 2025 09:06:08 PM UTC.
cuda-rhel9-x86_64
Name                            Stream                         Profiles                                  Summary                                              
nvidia-driver                   latest                         default [d], fm, ks, src                  Nvidia driver for latest branch                      
nvidia-driver                   latest-dkms                    default [d], fm, ks                       Nvidia driver for latest-dkms branch                 
nvidia-driver                   open-dkms [d]                  default [d], fm, ks, src                  Nvidia driver for open-dkms branch                   
nvidia-driver                   515                            default [d], fm, ks, src                  Nvidia driver for 515 branch                         
nvidia-driver                   515-dkms                       default [d], fm, ks                       Nvidia driver for 515-dkms branch                    
nvidia-driver                   515-open                       default [d], fm, ks, src                  Nvidia driver for 515-open branch                    
nvidia-driver                   520                            default [d], fm, ks, src                  Nvidia driver for 520 branch                         
nvidia-driver                   520-dkms                       default [d], fm, ks                       Nvidia driver for 520-dkms branch                    
nvidia-driver                   520-open                       default [d], fm, ks, src                  Nvidia driver for 520-open branch                    
nvidia-driver                   525                            default [d], fm, ks, src                  Nvidia driver for 525 branch                         
nvidia-driver                   525-dkms                       default [d], fm, ks                       Nvidia driver for 525-dkms branch                    
nvidia-driver                   525-open                       default [d], fm, ks, src                  Nvidia driver for 525-open branch                    
nvidia-driver                   530                            default [d], fm, ks, src                  Nvidia driver for 530 branch                         
nvidia-driver                   530-dkms                       default [d], fm, ks                       Nvidia driver for 530-dkms branch                    
nvidia-driver                   530-open                       default [d], fm, ks, src                  Nvidia driver for 530-open branch                    
nvidia-driver                   535                            default [d], fm, ks, src                  Nvidia driver for 535 branch                         
nvidia-driver                   535-dkms                       default [d], fm, ks                       Nvidia driver for 535-dkms branch                    
nvidia-driver                   535-open                       default [d], fm, ks, src                  Nvidia driver for 535-open branch                    
nvidia-driver                   545                            default [d], fm, ks, src                  Nvidia driver for 545 branch                         
nvidia-driver                   545-dkms                       default [d], fm, ks                       Nvidia driver for 545-dkms branch                    
nvidia-driver                   545-open                       default [d], fm, ks, src                  Nvidia driver for 545-open branch                    
nvidia-driver                   550 [e]                        default [d], fm, ks, src                  Nvidia driver for 550 branch                         
nvidia-driver                   550-dkms                       default [d], fm, ks                       Nvidia driver for 550-dkms branch                    
nvidia-driver                   550-open                       default [d], fm, ks, src                  Nvidia driver for 550-open branch                    
nvidia-driver                   555                            default [d], fm, ks, src                  Nvidia driver for 555 branch                         
nvidia-driver                   555-dkms                       default [d], fm, ks                       Nvidia driver for 555-dkms branch                    
nvidia-driver                   555-open                       default [d], fm, ks, src                  Nvidia driver for 555-open branch                    
nvidia-driver                   560                            default [d], fm, ks, src                  Nvidia driver for 560 branch                         
nvidia-driver                   560-dkms                       default [d], fm, ks                       Nvidia driver for 560-dkms branch                    
nvidia-driver                   560-open                       default [d], fm, ks, src                  Nvidia driver for 560-open branch                    
nvidia-driver                   565                            default [d], fm, ks, src                  Nvidia driver for 565 branch                         
nvidia-driver                   565-dkms                       default [d], fm, ks                       Nvidia driver for 565-dkms branch                    
nvidia-driver                   565-open                       default [d], fm, ks, src                  Nvidia driver for 565-open branch                    
nvidia-driver                   570-dkms                       default [d], fm, ks                       Nvidia driver for 570-dkms branch                    
nvidia-driver                   570-open                       default [d], fm, ks, src                  Nvidia driver for 570-open branch    
```

I think up to this point we have always been using the "-null" stream.

booxter:
@ktdreyer did it work before? How I understand the script, we are building a `kmod` RPM with `rpmbuild`, then install it from the RPM, not from Nvidia's repo. So I doubt it ever allowed to update the kernel after the fact without re-running the script.

Maybe now that we have dkms dnf module for nvidia, we CAN avoid building a RPM package and could instead install:

```
dnf info kmod-nvidia-latest-dkms.x86_64
Last metadata expiration check: 0:04:34 ago on Wed 26 Feb 2025 10:08:58 PM UTC.
Available Packages
Name         : kmod-nvidia-latest-dkms
Epoch        : 3
Version      : 570.86.15
Release      : 1.el9
Architecture : x86_64
Size         : 69 M
Source       : kmod-nvidia-latest-dkms-570.86.15-1.el9.src.rpm
Repository   : cuda-rhel9-x86_64
Summary      : NVIDIA display driver kernel module
URL          : http://www.nvidia.com/object/unix.html
License      : NVIDIA License
Description  : This package provides the proprietary Nvidia kernel driver modules.
             : The modules are rebuilt through the DKMS system when a new kernel or
             : modules become available.

```

booxter:
@ktdreyer @danmcp I have this commit in my private repo that switches the script to in-repo dkms package for kmod: https://github.com/instructlab/instructlab/commit/bb4ef061dd6f918a3b1323233d76fce5222a96aa but I will send it after this PR here is merged since it's a different topic. (Confirmed the change works against a fresh centos9 node.)

scripts/infra/nvidia-setup.sh
booxter: OK but it was bumped to 570 at [ecda29302](https://github.com/instructlab/instructlab/commit/ecda2930265d2e8e29a22db9aeb7560dec48c515) so do we want to revert this? not sure I understand the implication.

booxter: For the record, I have no idea why `570` doesn't have a null stream. But if we want to stick to this version, looks like we'll have to choose one of open or dkms.

danmcp:
> @danmcp do you want to use DKMS in this script?

My concern would be over consistency with downstream environments.  At a min I think we need the "-null" option to still be available.

booxter:
@danmcp what does `the "-null" option to still be available.` mean in the situation where the version doesn't seem to have `-null` available from the repo? Do you suggest we should revert to an older version in the script that would have a `-null` flavor? Alternatively, should we reach out to Nvidia to see why there's no `-null`?

I'm trying to understand what the path forward here is. Note the script is currently completely broken, so something has to be changed.

danmcp:
@booxter That would mean using an older version that still has a -null for now.  But I am unclear on the original expectation on moving to 570.

@fabiendupont Can you clarify which module you were expecting to pick up with the move to 570?

fabiendupont:
The move to driver 570 and CUDA 12.8 is motivated by the support for NVIDIA Blackwell. For the driver, the 570 branch is a long-term support version, so what NVIDIA recommends to their customers.

Which version of RHEL is used? If RHEL 9.5, the `570-open` stream will provide precompiled kernel modules. Otherwise, you'll need to install DKMS, which is not supported by RHEL. Another option remains downloading the `.run` installer and execute it in unattended mode.

booxter:
Thanks for this @fabiendupont I will then try to switch to `-open`.

fabiendupont:
Nevermind my comment about RHEL version. The script is compiling the kernel modules from source, so you shouldn't have any issue. I then definitely recommend `570-open`.

booxter:
@fabiendupont but should the script compile the driver if it's included in the -open?

booxter:
OK.

Wondering then: Should the script then detect if it's not RHEL and if so - use DKMS dnf module to spare the `rpmbuild`? (otherwise use `-open`.) It seems like both `rpmbuild` path and `dkms` path are not the way you'd do it for RHEL, so if we can remove the `rpmbuild` code, it could be beneficial.

ktdreyer (commented):
Would you please take https://github.com/booxter/instructlab/pull/1 into your branch? This will fix #3016 completely.

I've tested this a couple times now. It solves one of the bugs I was hitting (`dnf module enable` fails).

I have an improvement to the `KERNEL_VERSION` selection that I can post after we merge this.

ktdreyer (approved):

scripts/infra/nvidia-setup.sh
```diff
@@ -113,15 +113,15 @@ fi \
     && cp -a /etc/dnf/dnf.conf{,.tmp} && mv /etc/dnf/dnf.conf{.tmp,} \
     && dnf config-manager --best --nodocs --setopt=install_weak_deps=False --save \
     && dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel${OS_VERSION_MAJOR}/${CUDA_REPO_ARCH}/cuda-rhel${OS_VERSION_MAJOR}.repo \
-    && dnf -y module enable nvidia-driver:${DRIVER_STREAM}/default \
+    && dnf -y module enable nvidia-driver:${DRIVER_STREAM}-open/default \
```
danmcp, line 116: This seems like it could have performance implications and a diversion from downstream(s).  Do we not at least need the option to use the non open version?

booxter: @fabiendupont what's the version used in downstream? Is it 570? Which steam is (to be) used there?

danmcp: I remove my concern.  It wasn't clear to me from the comments that downstream had switched to -open with 570 as well.  Apologies for the confusion.

PR #3197, opened 2025-02-27
Adds support for simple global/general model configuration

RobotSail:
**Checklist:**

### DIFF
--- a/src/instructlab/cli/data/generate.py
+++ b/src/instructlab/cli/data/generate.py
@@ -167,9 +167,17 @@ logger = logging.getLogger(__name__)
 )
 @click.option(
     "--student-model-id",
-    help="ID of the custom model you want to use for the student.",
+    # help="ID of the custom model you want to use as the student model.",
     default=None,
     type=click.STRING,
+    show_default=True,
+)
+@click.option(
+    "--teacher-model-id",
+    # help="ID of the custom model you want to use as the teacher model.",
+    default=None,
+    type=click.STRING,
+    show_default=True,
 )
 @click.pass_context
 @clickext.display_params
@@ -200,6 +208,7 @@ def generate(
     max_num_tokens,
     detached,
     student_model_id: str | None,
+    teacher_model_id: str | None,
 ):
     """Generates synthetic data to enhance your example data"""
 
@@ -254,6 +263,20 @@ def generate(
         }
     )
 
+    if teacher_model_id:
+        try:
+            teacher_model_config = resolve_model_id(
+                teacher_model_id, ctx.obj.config.models
+            )
+            if not teacher_model_config:
+                raise ValueError(
+                    f"Teacher model with ID '{teacher_model_id}' not found in the configuration."
+                )
+            model_path = teacher_model_config.path
+        except ValueError as ve:
+            click.secho(f"failed to locate teacher model by ID: {ve}", fg="red")
+            raise click.exceptions.Exit(1)
+
     # determine student model arch from train section of config and pick system prompt to
     # pass to SDG appropriately
     system_prompt, legacy_pretraining_format = None, None
@@ -276,18 +299,23 @@ def generate(
             student_model_config = resolve_model_id(
                 student_model_id, ctx.obj.config.models
             )
+            if not student_model_config:
+                raise ValueError(
+                    f"Student model with ID '{student_model_id}' not found in the configuration."
+                )
         except ValueError as ve:
             click.secho(f"failed to locate student model by ID: {ve}", fg="red")
             raise click.exceptions.Exit(1)
-        else:
-            system_prompt = student_model_config.system_prompt
-            legacy_pretraining_format = (
-                True  # just set this to true for testing. We need to fix this in SDG
-            )
 
-    assert None not in (legacy_pretraining_format, system_prompt), (
-        "system_prompt and legacy_pretraining_format must have been set"
-    )
+        system_prompt = student_model_config.system_prompt
+        legacy_pretraining_format = (
+            True  # just set this to true for testing. We need to fix this in SDG
+        )
+
+    assert None not in (
+        legacy_pretraining_format,
+        system_prompt,
+    ), "system_prompt and legacy_pretraining_format must have been set"
 
     process_mode = ILAB_PROCESS_MODES.ATTACHED
     if detached:
--- a/src/instructlab/cli/data/generate.py
+++ b/src/instructlab/cli/data/generate.py
@@ -14,7 +14,7 @@ import click
 # First Party
 from instructlab import clickext
 from instructlab.client_utils import HttpClientParams
-from instructlab.configuration import DEFAULTS
+from instructlab.configuration import DEFAULTS, resolve_model_id
 from instructlab.data.generate_data import gen_data  # type: ignore
 from instructlab.defaults import ILAB_PROCESS_MODES
 from instructlab.utils import (
@@ -165,6 +165,12 @@ logger = logging.getLogger(__name__)
 @click.option(
     "-dt", "--detached", is_flag=True, help="Run ilab data generate in the background"
 )
+@click.option(
+    "--student-model-id",
+    help="ID of the custom model you want to use for the student.",
+    default=None,
+    type=click.STRING,
+)
 @click.pass_context
 @clickext.display_params
 def generate(
@@ -193,6 +199,7 @@ def generate(
     gpus,
     max_num_tokens,
     detached,
+    student_model_id: str | None,
 ):
     """Generates synthetic data to enhance your example data"""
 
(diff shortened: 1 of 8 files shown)


src/instructlab/model/chat.py
```diff
@@ -600,6 +603,7 @@ def chat_model(
     logs_dir,
     vi_mode,
     visible_overflow,
+    models: List[cfg._model_config],
```
cdoern, line 606: hmmmm, so it would be preferable to keep config class types out of the `model` pkg since its a "cli" only related structure. 

but if its not possible that's fine. Could we resolve the model before getting down into this file and just pass it in?

RobotSail: @cdoern This is just so we can do type-checking. How can we structure it so we can reuse the types later throughout the library?

src/instructlab/configuration.py
```diff
@@ -749,6 +763,10 @@ class Config(BaseModel):
         default_factory=_metadata,
         description="Metadata pertaining to the specifics of the system which the Configuration is meant to be applied to.",
     )
+    models: List[_model_config] = Field(
+        default_factory=lambda: [],
+        description="User-defined custom set of models. This allows people to use their own models for the InstructLab model customization process.",
+    )
```
reidliu41, line 769: sorry, I not sure about the purpose and may not understand correctly, based on the code, user defined `models`, and then need to specify the model id in the models?
why not just specify the model id?

RobotSail: @reidliu41 Not sure if I follow your question

src/instructlab/cli/data/generate.py
```diff
@@ -165,6 +165,12 @@
 @click.option(
     "-dt", "--detached", is_flag=True, help="Run ilab data generate in the background"
 )
+@click.option(
+    "--student-model-id",
```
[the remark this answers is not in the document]
jaideepr97: yeah i think thats a good idea 
after speaking with @RobotSail we decided to keep the first PR fairly barebones so that this logic at least gets in
I have plans to work on a follow up PR where all the model references in the config are updated to be ids instead of paths and then it would make sense to wire the click defaults up the way you suggested

src/instructlab/configuration.py
jaideepr97: @reidliu41 it would look something like
```
chat: ...
generate: ...
models:
  - id: 'ibm/granite-7b-lab'
    path: <local/hf path>
    ctx_size: 4096
  - id: ...
```
and then the long term goal would be to have model references in other sections (like in `chat.model` or `generate.model`) start pointing to these model ids instead of having the path directly  so we can centralize all the model info in one place for a given model

src/instructlab/cli/model/evaluate.py
```diff
@@ -235,9 +250,34 @@ def evaluate(
...
+                raise click.exceptions.BadParameter(
+                    f"Base model with ID '{judge_model_id}' not found in the configuration."
+                )
+
+            judge_model = judge_model_cfg.path
+            logger.debug(
+                "detected `--judge-model-id`, selecting custom model from config list."
+            )
+        else:
+            judge_model, output_dir = set_benchmark_specific_vars(
```
RobotSail, line 278: @jaideepr97 Should these be separate cases? 

It seems like `output_dir` is being set in the fallback case from `set_benchmark_specific_vars`, but when the IDs are passed it seems like we bypass the benchmark-specific settings.

src/instructlab/configuration.py
```diff
@@ -1520,6 +1546,15 @@ def map_train_to_library(ctx, params):
     train_args = TrainingArgs(**params)
     torch_args = TorchrunArgs(**params)
 
+    # let's handle the case when `--model-id` was provided
+    if params["model_id"]:
```
RobotSail, line 1550: I know that this will work for now (I think I wrote it this way originally lol), but we might want to change this to use `params.get` here in case a refactor happens down the road so we can avoid a `KeyError` randomly becoming a bug:

```suggestion
    if params.get("model_id"):
```

RobotSail (commented):
@jaideepr97 This looks good! I just have 2 very short comments, but otherwise it should be good to go.

jaideepr97 (approved):

src/instructlab/cli/data/generate.py
```diff
@@ -165,6 +165,20 @@
...
+    # help="ID of the custom model you want to use as the student model.",
+    default=None,
```
cdoern, line 171: do we want this to default to the model in the train part of the cfg or are we getting rid of those defaults?

src/instructlab/cli/data/generate.py
```diff
@@ -165,6 +165,20 @@
...
+    type=click.STRING,
+    show_default=True,
+)
+@click.option(
+    "--teacher-model-id",
```
cdoern, line 176: same here for default

src/instructlab/cli/data/generate.py
```diff
@@ -247,22 +263,60 @@ def generate(
...
+            raise click.exceptions.Exit(1)
+
     # determine student model arch from train section of config and pick system prompt to
     # pass to SDG appropriately
-    student_model_path = pathlib.Path(ctx.obj.config.train.model_path)
-    student_model_arch = get_model_arch(student_model_path)
-    system_prompt = get_sysprompt(student_model_arch)
+    system_prompt, legacy_pretraining_format = None, None
+    if not student_model_id:
+        student_model_path = pathlib.Path(ctx.obj.config.train.model_path)
```
cdoern, line 284: yeah this can be the default in the flag using `config_class`

src/instructlab/cli/model/evaluate.py
```diff
@@ -69,6 +70,12 @@ def set_benchmark_specific_vars(
     cls=clickext.ConfigOption,
     required=True,  # default from config
 )
+@click.option(
+    "--base-model-id",
```
cdoern, line 74: default from cfg?

src/instructlab/cli/model/evaluate.py
```diff
@@ -80,6 +87,12 @@ def set_benchmark_specific_vars(
     default=None,
     type=click.STRING,
 )
+@click.option(
+    "--judge-model-id",
```
cdoern, line 91: same here

src/instructlab/configuration.py
```diff
@@ -685,6 +685,29 @@ class _train(BaseModel):
...
+    """
+    Class representing the configuration for a model.
+
+    Attributes:
+        id (str): Internal ID referring to a particular model from the list.
+        path (str | None): Path to where the model can be found. Can either be a HF reference or local filepath.
+        system_prompt (str | None): The initial message used to prompt the conversation with this model.
+    """
+
+    id: str = Field(
```
cdoern, line 698: `default_factory` for these?

cdoern: and `default` for `id` is missing even if None

src/instructlab/configuration.py
```diff
@@ -740,8 +763,6 @@ class Config(BaseModel):
     general: _general = Field(
         default_factory=_general, description="General configuration section."
     )
-    # model configuration
-    model_config = ConfigDict(extra="ignore")
```
cdoern, line 744: I think this is a pydantic thing not a model thing, I could be wrong. I think other classes have it

cdoern (commented):
comments on defaults specifically, looks good so far

src/instructlab/cli/data/generate.py
RobotSail: We want to leave this blank for now.

src/instructlab/cli/data/generate.py
RobotSail: @cdoern We should leave it blank for now and figure out the best way for this later.

src/instructlab/cli/data/generate.py
RobotSail: Can we please avoid this for now? I would like to leave it as something where it's merely a lookup for models in the config, rather than something with a known default that will be referenced in the existing system. This way we can keep the logic simple and change it later as needed.

src/instructlab/cli/model/evaluate.py
RobotSail: Let's leave it blank to keep logic simple

src/instructlab/configuration.py
RobotSail: This will be a user-specified field for now, we shouldn't have a default here.

src/instructlab/configuration.py
RobotSail: Yes good catch

src/instructlab/cli/data/generate.py
jaideepr97: not quite - this default is a path, not a model id yet 
the way im imagining it, if we dont have to care about breaking changes as such we can rename all the model_path fields to either `model_id` or just `model` and have them reference model ids instead of paths. Until we do that, setting defaults in the click flags wont achieve much

cdoern (approved):

PR #3198, opened 2025-02-27
feat(infra): add cloud-instance list command; ignore shutting-down instances

booxter:
Some improvements to the `cloud-instance.sh` script:

- **feat(infra): add cloud-instance.sh list command**
- **feat(infra): ignore shutting-down machines when calculating aws ids**

**Checklist:**

### DIFF
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -64,7 +64,7 @@ _get_instances() {
     aws ec2 describe-instances \
         --filters "Name=tag:Name,Values=$INSTANCE_NAME" \
         --region "$EC2_REGION" \
-        --query 'Reservations[*].Instances[?State.Name!=`terminated`].InstanceId' \
+        --query 'Reservations[*].Instances[?State.Name!=`terminated` && State.Name!=`shutting-down`].InstanceId' \
         --output text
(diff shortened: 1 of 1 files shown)


scripts/infra/cloud-instance.sh
```diff
@@ -474,6 +482,8 @@ Commands
         -n
             Name of the instance to terminate (default provided in config)
 
+    list - List all instances
```
danmcp, line 485: Should this say list all instance with specified name?

booxter: I'm following the pattern for other command in the new revision. Lemme know if this is what you think of.

danmcp: It's a bit confusing because everything else is implicitly operating on 1 instance.  So it might not be clear what all instances you are listing still.  Would it be ok to state it as:

`List all applicable instances`

booxter: Yeah; we could have `-a` to list *all* instances; but your suggestion is probably better one for the current behavior.

danmcp (approved):

courtneypacheco (approved):
This is a good change. 👍

PR #3201, opened 2025-02-27
Add option to search multiple subnets with cloud-instance launch

danmcp:
This PR adds new optional config to cloud-instance for letting `cloud-instance.sh ec2 launch` search for an instance type across multiple subnets for an available instance.  Example of the new config:

```
EC2_SUBNET_VARS=("EC2_SUBNET_ID_1" "EC2_SUBNET_ID_2")
EC2_SUBNET_ID_1=subnet-12345678901234567
EC2_SUBNET_ID_2=subnet-12345678901234568
```

So if you were looking for a g6e.12xlarge, it would first look in subnet-12345678901234567, if an available instance wasn't found there, it would then look in subnet-12345678901234568.  If still no instance was found it would fail.

**Checklist:**


courtneypacheco (approved):

scripts/infra/instructlab-cloud-config-example
```diff
@@ -6,7 +6,11 @@ EC2_INSTANCE_TYPE="m4.large"
 # Nvidia 1x A10 GPU, 24GB VRAM
 #EC2_INSTANCE_TYPE="g5.4xlarge"
 EC2_SECURITY_GROUP_ID="sg-12345678901234567"
+# EC2_SUBNET_ID is ignored if you specify EC2_SUBNET_VARS which searches across the specified subnets for an available instance on launch
 EC2_SUBNET_ID="subnet-12345678901234567"
+EC2_SUBNET_VARS=("EC2_SUBNET_ID_1" "EC2_SUBNET_ID_2")
```
booxter, line 11: the interface seems overcomplex? Could we e.g. instead allow to list multiple values in the same variable, e.g. separated by space characters or commas?

booxter (commented):
One question about the interface.

scripts/infra/instructlab-cloud-config-example
danmcp: That's a fair critic.  I had used a similar mechanism in another place but that location was also supporting multi region.  So in that scenario I needed to be able to name the subnets so I could keep them in groups that would be paired with region, security groups, amis, and ssh keys.  I kept the same pattern here so we could potentially add multi region later in the same fashion.  If this mechanism feels too complex or confusing for the current impl here, I can certainly change it now and adjust later if we ever do add multi-region support.

booxter (approved):

PR #3222, opened 2025-03-06
fix(infra): reset nvidia-driver module before enabling -open one

booxter:
Till recently, we used "null" profile. With the switch to 570, we
switched to -open. When switching a dnf module, dnf complains as
follows:

```
The operation would result in switching of module 'nvidia-driver' stream
'XXX' to stream '570-open'

Error: It is not possible to switch enabled streams of a module unless
explicitly enabled via configuration option module_stream_switch.

It is recommended to rather remove all installed content from the
module, and reset the module using 'dnf module reset <module_name>'
command. After you reset the module, you can install the other stream.
```

To avoid this issue on upgrade, always reset the module when running the
script.

**Checklist:**


ktdreyer:
I tested this on a basic CentOS Stream image and it works.

PR #3224, opened 2025-03-06
infra: do not reinstall ibmcloud CLI

ktdreyer:
If a developer already has the `ibmcloud` CLI utility present in `$PATH`, print a message and exit rather than re-installing it.

### DIFF
--- a/scripts/infra/local-setup.sh
+++ b/scripts/infra/local-setup.sh
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 rh__ensure_ibm_system_pkgs() {
+    if command -v ibmcloud &> /dev/null; then
+        echo "ibmcloud CLI is already installed"
+        return
+    fi
     # shellcheck source=/dev/null
     if [ -f /etc/os-release ]; then
         source /etc/os-release


ktdreyer:
I have tested this manually with `./scripts/infra/local-setup.sh rh ensure-ibm-system-pkgs`

scripts/infra/local-setup.sh
```diff
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 rh__ensure_ibm_system_pkgs() {
+    if command -v ibmcloud &> /dev/null; then
+        echo "ibmcloud CLI is installed"
```
danmcp, line 6: nit: should this say "is already installed" so it's clear the script didn't just install it?

scripts/infra/local-setup.sh
```diff
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 rh__ensure_ibm_system_pkgs() {
+    if command -v ibmcloud &> /dev/null; then
```
courtneypacheco, line 5: This makes a lot of sense

one question though: if you call the installation script and `ibmcloud` is already installed, will the script reinstall `ibmcloud` or will it crash?

I guess I’m wondering about the CLI tool and how often it updates. If it has frequent updates, then there may be a situation where you want to reinstall the CLI tool so that you can use the latest features.

ktdreyer: I have `ibmcloud` installed already, so the script never re-installs it.

`ibmcloud` has its own internal update mechanism, so I occasionally run `ibmcloud update`. I can't remember if it prompts for this occasionally.

courtneypacheco: Oh neat! Good to know!!

courtneypacheco (approved):

scripts/infra/local-setup.sh
ktdreyer: I think I mis-understood your original question, apologies. I tried https://clis.cloud.ibm.com/install/linux again and it does actually re-install binaries. I'm not sure if it installs the latest?

booxter (approved):

PR #3225, opened 2025-03-06
infra: remove os version

ktdreyer:
Auto-detect the operating system identifier from `/etc/os-release`, so that users do not need to specify it manually.

The `ibmcloud` setup command applies to both RH and non-RH distros, so we don't need to check the OS there.

### DIFF
--- a/scripts/infra/local-setup.sh
+++ b/scripts/infra/local-setup.sh
@@ -1,23 +1,15 @@
 #!/usr/bin/env -S bash -e
 # SPDX-License-Identifier: Apache-2.0
 
-rh__ensure_ibm_system_pkgs() {
+ensure_ibm_system_pkgs() {
     if command -v ibmcloud &> /dev/null; then
         echo "ibmcloud CLI is already installed"
         return
     fi
-    # shellcheck source=/dev/null
-    if [ -f /etc/os-release ]; then
-        source /etc/os-release
-    fi
-    if [[ "$ID" == "fedora" ]] || [[ "$ID_LIKE" =~ "fedora" ]]; then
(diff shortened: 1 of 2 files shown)


scripts/infra/local-setup.sh
```diff
@@ -1,23 +1,15 @@
...
-    if [ -f /etc/os-release ]; then
-        source /etc/os-release
-    fi
-    if [[ "$ID" == "fedora" ]] || [[ "$ID_LIKE" =~ "fedora" ]]; then
-        curl -fsSL https://clis.cloud.ibm.com/install/linux | sh
-    else
-        echo "neither system 'ID' nor 'ID_LIKE' is 'fedora' - did you mean to specify a 'rh' system?"
-    fi
+    curl -fsSL https://clis.cloud.ibm.com/install/linux | sh
 }
```
reidliu41, line 10: seems better to add a check to make sure install success...

danmcp: Is that not handled by the exit code of the command still?

danmcp (approved):

booxter (approved):

scripts/infra/local-setup.sh
ktdreyer: I did some testing.

When the `sh` command fails, `local-setup.sh` will exit with a non-zero exit code.

When the `curl` command fails, `sh` eats the exit code of `curl`, so `local-setup.sh` still exits with `0`. That is a problem that is not directly related to the changes I'm making in this PR. We could fix that with `set -o pipefail` at the top of this script.

(This type of thing is why I prefer other languages to Bash. There are so many gotchas like this with error handling. But I understand this particular script makes sense to keep in Bash for now.)

Issue #3233, opened 2025-03-18
Users of  cloud-instance.sh should be warned about updating the kernel


derekhiggins:
Updating your kernel breaks the CUDA drivers, we should warn users against it

PR #3234, opened 2025-03-18
docs: Add MOTD to warn about updating the kernel

derekhiggins:
Closes: #3233

### DIFF
--- a/scripts/infra/cloud-instance.sh
+++ b/scripts/infra/cloud-instance.sh
@@ -367,6 +367,14 @@ install_rh_nvidia_drivers() {
     fi
     "${BASH_SOURCE[0]}" "$CLOUD_TYPE" scp -n "$INSTANCE_NAME" "${SCRIPT_DIR}"/nvidia-setup.sh
     "${BASH_SOURCE[0]}" "$CLOUD_TYPE" ssh -n "$INSTANCE_NAME" "sudo ./nvidia-setup.sh"
+
+    cat <<-EOF | "${BASH_SOURCE[0]}" "$CLOUD_TYPE" ssh -n "$INSTANCE_NAME" sudo dd of=/etc/motd.d/cuda-warning
+===========================================================================
+Important: Running \`dnf update\` may invalidate the GPU drivers. Requiring a
+           rerun of \`cloud-instance.sh $CLOUD_TYPE install-rh-nvidia-drivers\`
+===========================================================================
+EOF
+
     echo "You may want to reboot even though the install is live (${BASH_SOURCE[0]} ${CLOUD_TYPE} ssh sudo reboot)"
 }
 


scripts/infra/cloud-instance.sh
```diff
@@ -367,6 +367,14 @@ install_rh_nvidia_drivers() {
     fi
     "${BASH_SOURCE[0]}" "$CLOUD_TYPE" scp -n "$INSTANCE_NAME" "${SCRIPT_DIR}"/nvidia-setup.sh
     "${BASH_SOURCE[0]}" "$CLOUD_TYPE" ssh -n "$INSTANCE_NAME" "sudo ./nvidia-setup.sh"
+
+    cat <<-EOF | sudo dd of=/etc/motd.d/cuda-warning
+========================================================
+Important: Avoid running \`dnf update\`, As updating the
+           kernel will invalidate the GPU drivers
```
danmcp, line 374: It is possible to fix the drivers after dnf updating.  It might just work by rerunning nvidia-setup.sh.  Or cloud-instance.sh install-rh-nvidia-drivers.  Would it be appropriate to change the message to something like:

Running \`dnf update\` may invalidate the GPU drivers.  See \`cloud-instance.sh\` for more details.

derekhiggins: I'll try it out, I might have to add some more context as when I hit this problem I was using the AWS image outside of instructlab   (i.e. I used cloud-instance.sh to launch the instance but not for any of the setup)

derekhiggins: I've updated it to add more details and actually install the motd on the correct machine :wink: 

Also rerunning install_rh_nvidia_drivers on a older image isn't working (due to package dep errors)
https://github.com/instructlab/instructlab/issues/3241
https://github.com/instructlab/instructlab/pull/3242

scripts/infra/cloud-instance.sh
```diff
@@ -367,6 +367,14 @@ install_rh_nvidia_drivers() {
     fi
     "${BASH_SOURCE[0]}" "$CLOUD_TYPE" scp -n "$INSTANCE_NAME" "${SCRIPT_DIR}"/nvidia-setup.sh
     "${BASH_SOURCE[0]}" "$CLOUD_TYPE" ssh -n "$INSTANCE_NAME" "sudo ./nvidia-setup.sh"
+
+    cat <<-EOF | "${BASH_SOURCE[0]}" "$CLOUD_TYPE" ssh -n "$INSTANCE_NAME" sudo dd of=/etc/motd.d/cuda-warning
+===========================================================================
+Important: Running \`dnf update\` may invalidate the GPU drivers. Requiring a
+           rerun of \`cloud-instance.sh <cloud-type> install-rh-nvidia-drivers\`
```
danmcp, line 374: Since you have the CLOUD_TYPE in this method, should you include that in the message rather than <cloud-type>?

derekhiggins: done

danmcp (approved):

reidliu41 (approved):

PR #3238, opened 2025-03-19
Bump minimum version of SDG to 0.7.2

ktdreyer:
SDG created a new bugfix release. It is important enough that we want all instructlab users to consume it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ httpx>=0.25.0
 instructlab-eval>=0.5.1
 instructlab-quantize>=0.1.0
 instructlab-schema>=0.4.2
-instructlab-sdg>=0.7.0
+instructlab-sdg>=0.7.2
 instructlab-training>=0.7.0
 llama_cpp_python[server]==0.3.6
 mlx>=0.5.1,<0.6.0; sys_platform == 'darwin' and platform_machine == 'arm64'


courtneypacheco (approved):
Good catch! I should have added it to `main` too.

booxter:
FYI a decision for the project is to NOT bump minimal versions unless the older version won't work at all. That we continue bumping it, ignoring the decision, seems wrong to me.

ktdreyer:
@booxter , I'm not sure I understand. Are there users of SDG 0.7.0 that would like to stay on that version while also consuming the latest version of instructlab?

ktdreyer:
I'd like to move this forward because I expect SDG `v0.7.3` to come out soon (with backported https://github.com/instructlab/sdg/pull/574), and we would bump [instructlab's main `requirements.txt`](https://github.com/instructlab/instructlab/blob/main/requirements.txt) again for that.

bbrowning:
I'd like to make sure that if users upgrade their InstructLab version, they also get the newer instructlab-sdg version. And I want to make sure all new users installing InstructLab get the fixes from our latest 0.7.x release. Bumping the minimum version of instructlab-sdg in the requirements.txt would be the right way to accomplish this I believe?

booxter:
I'm referring to https://github.com/instructlab/dev-docs/commit/c4378ce432b808978a8da29a0b1234c9e2754252

Specifically, to the following wording: "Specifying the minimum value for the range (`foo>=x.y`) allows us to declare that we **need** features that only show up in or after that version of the dependency, which means we won't get bugs from users trying to use instructlab with an old dependency that **has an incompatible API** or is completely lacking a feature we **need**." and "If the new dependency **requires incompatible changes** in our code, then it becomes the new minimum version for our requirements range."

My interpretation of this wording is that we should not bump minimal version whenever we **like**, but only when we **have** to. I know this policy was not enforced so far, but this is what our downstream colleagues asked us to do, to allow for some freedom for packagers to pick a particular version off the allowed range.

danmcp:
By the rules written, it seems like this case could easily fit into the bucket of "or is completely lacking a feature we need."  In this case the feature being a usable docling version that we want to ensure all users get upgraded to.

I think the intention of the wording is that we don't want to force users to upgrade when it's not necessary.  And it's trying to balance that with we don't want users hitting known bugs/incompatibilities.  In the end that comes down to a judgment call of how likely are users to hit the issue vs how painful is the forced upgrade.  AFAIK, most of the pain from an upgrade would come from something else being broken with the new changes as the upgrade itself isn't difficult.  In this case since it's only the fixes for issues since 0.7.0 included, the risk seems very low and so I don't see any reason not to err on the side of getting the fixes out to a larger audience.

booxter:
>  In this case the feature being a usable docling version that we want to ensure all users get upgraded to.

Here's the diff between the two versions: https://github.com/instructlab/sdg/compare/v0.7.0...v0.7.2 Which docling version change do you refer to? (The only change to requirements.txt there is to exclude `tesserocr` for darwin platform, which AFAIU was done for CI sake only.)

The other changes included seem like straightforward bug fixes that while are useful for those interested in SDG, may not necessarily justify the upgrade of the package for all users (some may not even care about SDG for their use; or not hitting the issue).

---

In general, the practice of version pinning and minimal version bumps in this project, as practiced, is not packager-friendly, and the general recommendation in Python packaging world - and from our own downstream packagers - is to not bump dependencies without a clear necessity, because Python environments are shared across all other Python packages (in contrast to e.g. JavaScript).

Granted, the instructlab libraries have special status here and are probably less problematic since they are not generally used by other apps except instructlab CLI (at least at this moment); and probably the same packagers will package both CLI and libraries. If that's the argument, it's fine and I'd advise to update the policy with this decision.

FYI I'm not blocking it and you are free to merge it as-is. I only remind that we have a policy and it has an intent to not disrupt packagers' life.

danmcp:
> Which docling version change do you refer to?

Apologies.  I mixed up the change targeted for 0.7.3.  So the reward statement is different than what I stated, but the risk side of the equation still seems very low to ensure users get the targeted fixes.

I agree with your concerns in the general case and that the instructlab libraries don't currently fall into any of the pitfalls.  +1 to clarifying further in our documentation.

booxter (approved):
Anyway, let's move on with this, since we bumped in a release branch already anyway. I am open to consider instructlab libraries are somewhat different, though I'd encourage folks to clarify the policy if there are plans to continue bumping these at will.

ktdreyer:
> I expect SDG `v0.7.3` to come out soon (with backported https://github.com/instructlab/sdg/pull/574)

To close the loop here: SDG developers revisited this decision today, and we probably won't cherry-pick that to [`release-v0.7`](https://github.com/instructlab/sdg/commits/release-v0.7/) for now.

PR #3248, opened 2025-03-27
infra: default cloud-instance.sh to latest CentOS 9 AMI

ktdreyer:
If the `cloud-init.sh` user has not hard-coded an AMI ID in `~/.instructlab/cloud-config`, default to using the newest CentOS 9 AMI that the CPE team has published to the AWS marketplace.

The purpose of this change is to make it easier to launch the latest cloud images with fewer security vulnerabilities.

-------

To test this, I've run `cloud-instance.sh` with the following scenarios:

1. Setting an old `EC2_AMI_ID="ami-09fc0e32ec75bfe3f"` in `~/.instructlab/cloud-config` continues to launch the old AMI
2. Deleting `EC2_AMI_ID` from in `~/.instructlab/cloud-config` properly selects the latest CentOS image, and `dnf` shows no pending updates on the node.


scripts/infra/instructlab-cloud-config-example
```diff
@@ -11,6 +11,7 @@ EC2_SUBNET_ID="subnet-12345678901234567"
 EC2_SUBNET_VARS=("EC2_SUBNET_ID_1" "EC2_SUBNET_ID_2")
 EC2_SUBNET_ID_1=subnet-12345678901234567
 EC2_SUBNET_ID_2=subnet-12345678901234568
+# Omit EC2_AMI_ID to use the latest CentOS Stream 9 image.
```
booxter, line 14: Should this be the default? (Comment out the particular AMI selection and instruct user that they can hardcode the image if needed?)

ktdreyer: Good idea. I've amended and repushed.

booxter (approved):

danmcp (approved):

danmcp:
Nice change!

PR #3254, opened 2025-04-02
chore: update docling api references

jaideepr97:
bring changes in https://github.com/instructlab/sdg/pull/574/files to instructlab 

**Checklist:**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,6 +122,18 @@ jobs:
           rm -rf "$GITHUB_WORKSPACE"/src/github.com/containers/skopeo
           skopeo --version
 
+        # deactivate MPS acceleration on Github CI for MacOS
+        # see https://github.com/actions/runner-images/issues/9918
+      - name: Disable MPS acceleration on MacOS
+        if: startsWith(matrix.platform, 'macos')
+        run: |
+          echo 'INSTRUCTLAB_DISABLE_GPU_ACCELERATION="true"' >> "$GITHUB_ENV"
+
+      - name: Enable MPS acceleration on non-MacOS runners
+        if: ! startsWith(matrix.platform, 'macos')
+        run: |
+          echo 'INSTRUCTLAB_DISABLE_GPU_ACCELERATION=' >> "$GITHUB_ENV"
+
       - name: Install tools on MacOS
         if: startsWith(matrix.platform, 'macos')
         run: |
--- a/requirements.txt
+++ b/requirements.txt
@@ -34,7 +34,9 @@ toml>=0.10.2
 # Default version. Can be overridden in extra requirements
 torch>=2.3.0,<2.6.0
 tqdm>=4.66.2
-transformers>=4.41.2
+# temporary cap until https://github.com/instructlab/training/pull/443 is merged and consumed within instructlab
+# above PR fixes interactions with newer versions of transformers through the training library
+transformers>=4.41.2,<4.51.0
 trl>=0.12.2,<0.15.0
 wandb>=0.16.4
 xdg-base-dirs>=6.0.1
@@ -42,4 +44,5 @@ psutil>=6.0.0
 huggingface_hub[hf_transfer]>=0.1.8
 haystack-ai>=2.8
 docling-core[chunking]>=2.10.0
+docling>=2.18.0
 sentence-transformers>=3.0.0
--- a/src/instructlab/rag/convert.py
+++ b/src/instructlab/rag/convert.py
@@ -6,7 +6,7 @@
 
 # Standard
 from pathlib import Path
-from typing import Iterable
+from typing import Iterable, Optional
 import json
 import logging
 import os
@@ -21,6 +21,7 @@ from docling.datamodel.pipeline_options import EasyOcrOptions  # type: ignore
 from docling.datamodel.pipeline_options import OcrOptions  # type: ignore
 from docling.datamodel.pipeline_options import PdfPipelineOptions  # type: ignore
 from docling.datamodel.pipeline_options import TesseractOcrOptions  # type: ignore
+from docling.datamodel.pipeline_options import AcceleratorDevice, AcceleratorOptions
 from xdg_base_dirs import xdg_data_dirs, xdg_data_home
 import yaml
 
@@ -175,7 +176,12 @@ def _initialize_docling():
         artifacts_path=docling_model_path,
         do_ocr=False,
     )
-    ocr_options = _resolve_ocr_options()
+
+    if os.getenv("INSTRUCTLAB_DISABLE_GPU_ACCELERATION") == "true":
+        pipeline_options.accelerator_options = AcceleratorOptions(
+            device=AcceleratorDevice.CPU
+        )
+    ocr_options = resolve_ocr_options(docling_model_path=docling_model_path)
     if ocr_options is not None:
         pipeline_options.do_ocr = True
         pipeline_options.ocr_options = ocr_options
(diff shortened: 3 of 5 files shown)


src/instructlab/rag/convert.py
```diff
@@ -175,7 +177,13 @@ def _initialize_docling():
         artifacts_path=docling_model_path,
         do_ocr=False,
     )
-    ocr_options = _resolve_ocr_options()
+
+    # deactivate MPS acceleration on Github CI
+    if os.getenv("CI") and sys.platform == "darwin":
```
booxter, line 182: It could be cleaner if we'd have a variable with a very explicit name here, instead of `CI`. E.g. `INSTRUCTLAB_DISABLE_GPU_ACCELERATION`, and then we don't need to check for platform since we'll set the variable only for macos runs.

jaideepr97: updated

src/instructlab/rag/convert.py
```diff
@@ -175,7 +176,13 @@ def _initialize_docling():
         artifacts_path=docling_model_path,
         do_ocr=False,
     )
-    ocr_options = _resolve_ocr_options()
+
+    # deactivate MPS acceleration on Github CI for MacOS
```
booxter, line 180: This comment probably belongs, if at all, to github workflow file

tox.ini
```diff
@@ -10,6 +10,7 @@ minversion = 4.4
 description = run tests (unit, unitcov, functional)
 passenv =
     CMAKE_ARGS
+    CI
```
booxter, line 13: you will need to adjust this

.github/workflows/test.yml
```diff
@@ -122,6 +122,10 @@ jobs:
           rm -rf "$GITHUB_WORKSPACE"/src/github.com/containers/skopeo
           skopeo --version
 
+      - name: set CI value
+        run: |
+          export INSTRUCTLAB_DISABLE_GPU_ACCELERATION="true"
```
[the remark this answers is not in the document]
jaideepr97: i had added that in the previous run, but it failed at the same place 
just running it again without to see if that specific unit test passes 
will add it back after that and check again

src/instructlab/rag/convert.py
ktdreyer: Thanks Ihar for finding the reference to https://github.com/actions/runner-images/issues/9918

Jaideep would you mind adding a comment with that URL as well, so that we have information a year from now, so we can determine if/when we can drop this workaround.

.github/workflows/test.yml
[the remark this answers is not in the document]
booxter: Probably because `passenv` is not updated for the new variable name?

src/instructlab/rag/convert.py
jaideepr97: added the URL and moved the comment 👍

.github/workflows/test.yml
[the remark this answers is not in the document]
jaideepr97: updated!

tox.ini
jaideepr97: updated, thanks!

.github/workflows/test.yml
```diff
@@ -122,6 +122,13 @@ jobs:
           rm -rf "$GITHUB_WORKSPACE"/src/github.com/containers/skopeo
           skopeo --version
 
+        # deactivate MPS acceleration on Github CI for MacOS
+        # see https://github.com/actions/runner-images/issues/9918
+      - name: Disable MPS acceleration on MacOS
+        if: startsWith(matrix.platform, 'macos')
+        run: |
+          export INSTRUCTLAB_DISABLE_GPU_ACCELERATION="true"
```
courtneypacheco, line 130: So... if you want to set `INSTRUCTLAB_DISABLE_GPU_ACCELERATION=true` on macOS runners only _and_ allow `INSTRUCTLAB_DISABLE_GPU_ACCELERATION=true` to be used in different steps for macOS runners only, you would need to leverage `GITHUB_ENV` to do so: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#passing-values-between-steps-and-jobs-in-a-workflow

If you don't echo to `GITHUB_ENV` in this step, then `INSTRUCTLAB_DISABLE_GPU_ACCELERATION` is not able to be referenced outside this step.

It's the equivalent of doing this in Python:

```python3
def some_function():
    var = 1

print(var + 1) #<--- 'var' is only set in the scope of the function, so you can't reference it this way
```

Suggestion to try:

```suggestion
        # deactivate MPS acceleration on Github CI for MacOS
        # see https://github.com/actions/runner-images/issues/9918
      - name: Disable MPS acceleration on MacOS
        if: startsWith(matrix.platform, 'macos')
        run: |
          echo 'INSTRUCTLAB_DISABLE_GPU_ACCELERATION="true"' >> $GITHUB_ENV

      - name: Enable MPS acceleration on non-MacOS runners
        if: !startsWith(matrix.platform, 'macos')
        run: |
          echo 'INSTRUCTLAB_DISABLE_GPU_ACCELERATION=' >> $GITHUB_ENV
```

The reason I suggest the second step above ^ is that... once you set `INSTRUCTLAB_DISABLE_GPU_ACCELERATION="true"` for macOS, it will actually set `INSTRUCTLAB_DISABLE_GPU_ACCELERATION="true"` for all the other runners too.  We don't want that. Thus, we "unset" the variable for the other runners by setting it to an empty value.

jaideepr97: Thanks @courtneypacheco! trying this now

>once you set INSTRUCTLAB_DISABLE_GPU_ACCELERATION="true" for macOS, it will actually set INSTRUCTLAB_DISABLE_GPU_ACCELERATION="true" for all the other runners too

are you saying that `if: startsWith(matrix.platform, 'macos')` is not respected then? this seems unnecessarily complicated

.github/workflows/test.yml
```diff
@@ -122,6 +122,18 @@ jobs:
...
+        # see https://github.com/actions/runner-images/issues/9918
+      - name: Disable MPS acceleration on MacOS
+        if: startsWith(matrix.platform, 'macos')
+        run: |
+          echo 'INSTRUCTLAB_DISABLE_GPU_ACCELERATION="true"' >> "$GITHUB_ENV"
+
+      - name: Enable MPS acceleration on non-MacOS runners
+        if: ! startsWith(matrix.platform, 'macos')
+        run: |
+          echo 'INSTRUCTLAB_DISABLE_GPU_ACCELERATION=' >> "$GITHUB_ENV"
```
booxter, line 135: nit: probably not necessary to set it to empty string here since env.get() will return None if it's not set, which will still not pass the `== true` check.

booxter (approved):
Code wise looks correct. Hope CI passes too.

booxter:
FYI CI failure is because of training library "mocking" the transformers forward method in an unsafe way: https://github.com/instructlab/training/pull/443

.github/workflows/test.yml
booxter: @courtneypacheco are you saying `$GITHUB_ENV` is global to all runners in the test matrix? isn't every matrix entry getting its own isolated runner?

requirements.txt
```diff
@@ -34,12 +34,13 @@ toml>=0.10.2
 # Default version. Can be overridden in extra requirements
 torch>=2.3.0,<2.6.0
 tqdm>=4.66.2
-transformers>=4.41.2
+transformers>=4.41.2,<4.51.0
```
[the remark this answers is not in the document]
jaideepr97: added

booxter (approved):

RobotSail (approved):
LGTM

courtneypacheco:
@mergify backport release-v0.23

PR #3255, opened 2025-04-02
feat: add model family in cfg

jaideepr97:
adds the ability to pass a specific model-id to generate, chat and serve. This model id is then resolved out of the config if found, and details like the model-family, path etc are extracted and passed further 
This is required to support 3rd party models being used as students, teachers etc in an easy way. The model family value is passed into sdg and appropriate chat template is selected based on what is passed 

**Checklist:**

### DIFF
--- a/src/instructlab/cli/data/generate.py
+++ b/src/instructlab/cli/data/generate.py
@@ -241,6 +241,25 @@ def generate(
     if batch_size > 0:
         checkpoint_dir = os.path.join(output_dir, "checkpoints")
 
+    if teacher_model_id:
+        try:
+            teacher_model_config = resolve_model_id(
+                teacher_model_id, ctx.obj.config.models
+            )
+            if not teacher_model_config:
+                raise ValueError(
+                    f"Teacher model with ID '{teacher_model_id}' not found in the configuration."
+                )
+            model_path = teacher_model_config.path
+            model_family = (
+                teacher_model_config.family
+                if teacher_model_config.family
+                else model_family
+            )
+        except ValueError as ve:
+            click.secho(f"failed to locate teacher model by ID: {ve}", fg="red")
+            raise click.exceptions.Exit(1)
+
     serve_cfg = copy.deepcopy(ctx.obj.config.generate.teacher)
     serve_cfg.llama_cpp.llm_family = model_family or serve_cfg.llama_cpp.llm_family
     serve_cfg.vllm.llm_family = model_family or serve_cfg.vllm.llm_family
@@ -262,20 +281,6 @@ def generate(
             "tls_insecure": tls_insecure,
         }
     )
-
-    if teacher_model_id:
-        try:
-            teacher_model_config = resolve_model_id(
-                teacher_model_id, ctx.obj.config.models
-            )
-            if not teacher_model_config:
-                raise ValueError(
-                    f"Teacher model with ID '{teacher_model_id}' not found in the configuration."
-                )
-            model_path = teacher_model_config.path
-        except ValueError as ve:
-            click.secho(f"failed to locate teacher model by ID: {ve}", fg="red")
-            raise click.exceptions.Exit(1)
 
     # determine student model arch from train section of config and pick system prompt to
     # pass to SDG appropriately
--- a/src/instructlab/cli/data/generate.py
+++ b/src/instructlab/cli/data/generate.py
@@ -251,11 +251,7 @@ def generate(
                     f"Teacher model with ID '{teacher_model_id}' not found in the configuration."
                 )
             model_path = teacher_model_config.path
-            model_family = (
-                teacher_model_config.family
-                if teacher_model_config.family
-                else model_family
-            )
+            model_family = model_family if model_family else teacher_model_config.family
         except ValueError as ve:
             click.secho(f"failed to locate teacher model by ID: {ve}", fg="red")
             raise click.exceptions.Exit(1)
--- a/src/instructlab/cli/model/chat.py
+++ b/src/instructlab/cli/model/chat.py
@@ -10,6 +10,7 @@ import click
 # First Party
 from instructlab import clickext
 from instructlab import configuration as cfg
+from instructlab.configuration import resolve_model_id
 from instructlab.defaults import DEFAULTS
 from instructlab.model.chat import chat_model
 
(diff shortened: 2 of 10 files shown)


src/instructlab/configuration.py
```diff
@@ -1490,6 +1494,11 @@ def init(
 ) -> None:
     config_obj: Config
     error_msg: str | None = None
+
+    if hasattr(ctx, "obj") and ctx.obj:
```
booxter, line 1498: Is this change to allow to pass modified config with additional models directly through the ctx object? If it's just for tests, can we instead produce a config file on disc with the necessary adjustments, like we e.g. do with tests that use `common.setup_gpus_config`?

src/instructlab/cli/model/serve.py
```diff
@@ -127,6 +133,19 @@ def serve(
         ctx.obj.config.serve.server.current_max_ctx_size = max_ctx_size
         write_config(ctx.obj.config)
 
+    if model_id:
+        try:
+            model_config = resolve_model_id(model_id, ctx.obj.config.models)
```
booxter, line 138: this could be a common util function (`get_model_path_and_family`) since you write the code twice.

tests/test_lab_generate.py
```diff
@@ -22,3 +29,43 @@ def test_vllm_args_null(cli_runner: CliRunner):
...
+            "simple",
+            "--output-dir",
+            str(tmp_path),
+        ],
+        obj=ctx.obj,
+    )
+
+    assert result.exit_code == 0, f"Command failed with output: {result.output}"
+    assert mock_gen_data.call_count == 1
+    assert mock_storage_dirs_exist.call_count == 1
```
booxter, line 71: nit: not sure what this check tries to achieve that is important to the intent of the test case. Is it important that the implementation called storage_dirs_exist?

booxter (changes requested):

src/instructlab/cli/data/generate.py
```diff
@@ -241,6 +241,25 @@ def generate(
...
+        try:
+            teacher_model_config = resolve_model_id(
+                teacher_model_id, ctx.obj.config.models
+            )
+            if not teacher_model_config:
+                raise ValueError(
+                    f"Teacher model with ID '{teacher_model_id}' not found in the configuration."
+                )
+            model_path = teacher_model_config.path
+            model_family = (
```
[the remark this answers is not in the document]
jaideepr97: oops, updated

src/instructlab/configuration.py
jaideepr97: understood, fixed

src/instructlab/cli/model/serve.py
jaideepr97: i did think about that, but then its useful to have the entire model_config object with you so we can access any fields we add in the future as well 
and may not need to access all fields everywhere necessarily

there is some repeated code in terms of extracting what we want from it but it seems relatively acceptable to me 
there is the other issue of placing such util functions in the right place without introducing import cycles etc, since we'd have to import the model_config class from configuration.py, which already imports from utils

tests/test_lab_generate.py
jaideepr97: fixed

tests/test_model_chat.py
```diff
@@ -130,3 +132,41 @@ def mock_sys_print_output(*args, **kwargs):
...
     assert rendered_output == expected_output_without_box
+
+
+@patch("instructlab.cli.model.chat.chat_model")
+@patch("instructlab.model.model.storage_dirs_exist", return_value=True)
+def test_chat_with_model_id(
+    mock_chat_model, mock_storage_dirs_exist, cli_runner: CliRunner
+):
+    # Set up a mock configuration with a model
+    test_ctx = Config()
```
booxter, line 143: can't we use `setup_test_models_config` here too?

tests/test_model_chat.py
```diff
@@ -130,3 +132,41 @@ def mock_sys_print_output(*args, **kwargs):
...
+            "chat",
+            "--model-id=test_model",
+        ],
+        obj=ctx.obj,
+    )
+
+    # Assert that the command executed successfully
+    assert result.exit_code == 0, f"Command failed with output: {result.output}"
+    mock_chat_model.assert_called_once()
+    mock_storage_dirs_exist.assert_called_once()
```
booxter, line 172: that other nit applied here too I think

src/instructlab/configuration.py
```diff
@@ -1321,7 +1323,9 @@ def get_model_family(family, model_path):
     family = MODEL_FAMILY_MAPPINGS.get(family, family)
     if family:
         if family.lower() not in MODEL_FAMILIES:
-            raise ConfigException(f"Unknown model family: {family}")
+            # do not raise ConfigException(f"Unknown model family: {family}") anymore
```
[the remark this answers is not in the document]
booxter: where? :)

tests/test_model_chat.py
jaideepr97: apologies for missing this, fixed

RobotSail (approved):
LGTM, great work on this @jaideepr97

src/instructlab/cli/model/serve.py
```diff
@@ -127,6 +133,19 @@ def serve(
...
+            if not model_config:
+                raise ValueError(
+                    f"Model with ID '{model_id}' not found in the configuration."
+                )
+            model_path = pathlib.Path(model_config.path)
+            model_family = model_config.family if model_config.family else model_family
```
booxter, line 144: shouldn't you prefer CLI option here too?

src/instructlab/cli/model/chat.py
```diff
@@ -37,7 +37,6 @@
     "-c",
     "--context",
     cls=clickext.ConfigOption,
-    required=True,  # default from config
```
booxter, line 40: why is this change needed?

src/instructlab/cli/model/serve.py
RobotSail: @booxter The order of precedence should be like this:

`--model-id` CLI argument >  legacy `--model` flag >  XOR(global model ID config, global model)

cdoern (approved):
besides the few nits left over, LGTM

src/instructlab/cli/model/chat.py
jaideepr97: holdover from when ctx was being passed directly in the test
reverted

src/instructlab/configuration.py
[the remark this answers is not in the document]
jaideepr97: i had only udpated the log message 
updated the comment as well now

src/instructlab/cli/model/serve.py
jaideepr97: im not sure mixing the id related fields and the legacy model flag is a good idea
they do different things - for eg the legacy model flag only deals with the model path. so this precedence would only apply to the path field, not necessarily the other things like model family etc
so i think it might end up complicating the resolution logic unnecessarily

booxter (approved):
Thanks for working through the comments. This looks ready.

PR #3250, opened 2025-03-31
chore: bump the training library to 0.8.0 and refactor tokenizer usage

RobotSail:
The latest release of instructlab-training improves the usage for tokenizing datasets by removing the requirement of needing to always supply a chat template. This breaks the API for `setup_tokenizer` where in the past it needed to be passed some special objects only present in the training library. Now, it accepts only the model path and an optional template path. And given the scenario, it will automatically configure the tokenizer to the best of its abilities.

**Checklist:**

- [ ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Functional tests have been added, if necessary.
- [ ] E2E Workflow tests have been added, if necessary.

PR #3269, opened 2025-04-09
Revert "chore: bump the training library to 0.8.0 and refactor tokenizer usage"

courtneypacheco:
Reverts instructlab/instructlab#3250 because it was merged after https://github.com/instructlab/instructlab/pull/3254, which was reverted because the unit tests stopped running after that PR. There might not be anything wrong with this PR, but we need to re-validate this PR by forcing it to undergo unit + functional testing.


courtneypacheco:
Closing!

PR #3283, opened 2025-04-10
feat(deps): Bump `torch` upper bound (<2.7.0) + set allowable range for `vllm` versions (>=0.8.0,<0.9.0)

courtneypacheco:
**Checklist:**

## Changes

- In **requirements.txt**: update `vllm==0.7.3` with `vllm>=0.8.0` and utilize a new `constraints-dev.txt` file to restrict the upper bound on `vllm` in the CI only
- In **requirements.txt**:  update `torch>=2.3.0,<2.6.0` to `torch>=2.6.0,<2.7.0`, and set `torch==2.6.0` in the new `constraints-dev.txt` file for the CI only

**Note**: `vllm>=0.8.0,<0.8.0` is only compatible with `torch>=2.6.0` right now, hence we have bumped the `torch` lower bound.

### DIFF
--- a/requirements-vllm-cuda.txt
+++ b/requirements-vllm-cuda.txt
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 # Dependencies for installing vLLM on CUDA
 
-# vLLM only supports Linux platform (including WSL)
-vllm==0.7.3 ; sys_platform == 'linux' and platform_machine == 'x86_64'
+# vLLM only supports Linux platform (including WSL). Do not cap this dependency here. Cap in constraints-dev.txt
+vllm>=0.8.0 ; sys_platform == 'linux'
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,8 +31,8 @@ sentencepiece>=0.2.0
 # "old" version required for vLLM on CUDA to build
 tokenizers>=0.11.1
 toml>=0.10.2
-# Default version. Can be overridden in extra requirements
-torch>=2.3.0,<2.6.0
+# Default version. Can be overridden in extra requirements. Do not cap this dependency here. Cap in constraints-dev.txt
+torch>=2.6.0
 tqdm>=4.66.2
 # temporary cap until https://github.com/instructlab/training/pull/443 is merged and consumed within instructlab
 # above PR fixes interactions with newer versions of transformers through the training library
--- a/tox.ini
+++ b/tox.ini
@@ -21,7 +21,9 @@ package = wheel
 wheel_build_env = pkg
 # equivalent to `pip install instructlab[cpu]`
 extras = cpu
-deps = -r requirements-dev.txt
+deps =
+    -r requirements-dev.txt
+    -c constraints-dev.txt
 commands =
     ilab --version
     {envpython} -m instructlab --version
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -134,7 +134,7 @@ jobs:
           # https://github.com/instructlab/instructlab/issues/1821
           # install with Torch and build dependencies installed
           python3.11 -m pip install -v packaging wheel setuptools-scm
-          python3.11 -m pip install -v .[cuda] -r requirements-vllm-cuda.txt
+          python3.11 -m pip install -v .[cuda] -r requirements-vllm-cuda.txt -c constraints-dev.txt
 
       - name: Check disk before tests
         run: |
--- a/.github/workflows/e2e-nvidia-l40s-x4-llama.yml
+++ b/.github/workflows/e2e-nvidia-l40s-x4-llama.yml
@@ -20,28 +20,58 @@ jobs:
   start-large-ec2-runner:
     runs-on: ubuntu-latest
     outputs:
-      label: ${{ steps.start-ec2-runner.outputs.label }}
-      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+      label: ${{ steps.launch-ec2-instance-with-fallback.outputs.label }}
+      ec2-instance-id: ${{ steps.launch-ec2-instance-with-fallback.outputs.ec2-instance-id }}
+      ec2-instance-region: ${{ steps.launch-ec2-instance-with-fallback.outputs.ec2-instance-region }}
     steps:
-      - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@ececac1a45f3b08a01d2dd070d28d111c5fe6722 # v4.1.0
+      - name: Checkout "launch-ec2-runner-with-fallback" in-house CI action
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-region: ${{ vars.AWS_REGION }}
+          repository: instructlab/ci-actions
+          # clone the "ci-actions" repo to a local directory called "ci-actions", instead of overwriting the current WORKDIR contents
+          path: ci-actions
+          ref: release-v0.1
+          sparse-checkout: |
+            actions/launch-ec2-runner-with-fallback
 
-      - name: Start EC2 runner
-        id: start-ec2-runner
-        uses: machulav/ec2-github-runner@fcfb31a5760dad1314a64a0e172b78ec6fc8a17e # v2.3.6
+      - name: Launch EC2 Runner with Fallback
+        id: launch-ec2-instance-with-fallback
+        uses: ./ci-actions/actions/launch-ec2-runner-with-fallback
+        env:
+          TMPDIR: "/tmp"
         with:
-          mode: start
-          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
-          ec2-image-id: ${{ vars.AWS_EC2_AMI }}
-          ec2-instance-type: g6e.12xlarge
-          subnet-id: subnet-024298cefa3bedd61
-          security-group-id: sg-06300447c4a5fbef3
-          iam-role-name: instructlab-ci-runner
-          aws-resource-tags: >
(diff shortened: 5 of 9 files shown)


constraints.txt
```diff
@@ -0,0 +1 @@
+torch<2.6.0
```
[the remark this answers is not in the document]
courtneypacheco: Fixed

requirements-vllm-cuda.txt
```diff
@@ -2,4 +2,4 @@
 # Dependencies for installing vLLM on CUDA
 
 # vLLM only supports Linux platform (including WSL)
-vllm==0.7.3 ; sys_platform == 'linux' and platform_machine == 'x86_64'
+vllm==0.8.3 ; sys_platform == 'linux' and platform_machine == 'x86_64'
```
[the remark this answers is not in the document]
courtneypacheco: Updated.

src/instructlab/train/linux_train.py
```diff
@@ -256,6 +256,7 @@ def model_generate(user, **kwargs):
             stopping_criteria=stopping_criteria,
             do_sample=True,
             output_logits=True,
+            return_dict_in_generate=True,
```
courtneypacheco, line 259: This change is out of scope for my PR, but I'm debugging. Not a permanent change.

constraints.txt
```diff
@@ -0,0 +1 @@
+torch==2.6.0
```
tiran, line 1: Suggestion: Since this file is only used for constraining versions in tests, how about you rename the file to `constraints-dev.txt` and explain the purpose in the a comment?

requirements-vllm-cuda.txt
```diff
@@ -2,4 +2,4 @@
 # Dependencies for installing vLLM on CUDA
 
 # vLLM only supports Linux platform (including WSL)
-vllm==0.7.3 ; sys_platform == 'linux' and platform_machine == 'x86_64'
+vllm>=0.8.0,<0.9.0 ; sys_platform == 'linux' and platform_machine == 'x86_64'
```
tiran, line 5: I would prefer if you would remove the upper version cap and instead move the constraint into `constraints-dev.txt`. If you keep the upper version cap, then we'll have to request an InstructLab update every time we want to test a new version of vLLM.

requirements.txt
```diff
@@ -32,11 +32,10 @@ sentencepiece>=0.2.0
 tokenizers>=0.11.1
 toml>=0.10.2
 # Default version. Can be overridden in extra requirements
-torch>=2.3.0,<2.6.0
+torch>=2.3.0,<2.7.0
```
tiran, line 35: Same here, please reconsider the upper version cap.

Also is anybody testing that InstructLab still works with `torch==2.3.0`? If not, then you should raise the lower cap.

requirements.txt
```diff
@@ -32,11 +32,10 @@ sentencepiece>=0.2.0
...
 tqdm>=4.66.2
-# temporary cap until https://github.com/instructlab/training/pull/443 is merged and consumed within instructlab
-# above PR fixes interactions with newer versions of transformers through the training library
-transformers>=4.41.2,<4.51.0
+# Testing. Not intended to be merged. Seeing if the latest 'patch' release from transformers is causing issues
+transformers>=4.41.2,<4.51.2
```
tiran, line 38: And same here, please reconsider the upper version cap. if `transfomers 4.51.2` is not working for you, then please exclude that version with `transformers>=4.41.2,!=4.51.2`.

tox.ini
```diff
@@ -17,6 +17,7 @@ setenv =
     PIP_EXTRA_INDEX_URL=https://download.pytorch.org/whl/cpu
     CMAKE_ARGS={env:CMAKE_ARGS:-DGGML_NATIVE=off}
     ILAB_MAX_STABLE_VRAM_WAIT=0
+    PIP_CONSTRAINT=constraints.txt
```
tiran, line 20: It's better to use

```
deps =
    -r requirements-dev.txt
    -c constraints-dev.txt
```

tiran (changes requested):

ktam3:
@instructlab/core-maintainers PTAL and review

tox.ini
```diff
@@ -36,12 +37,14 @@ basepython = python3.11
 
 [testenv:py3-unit]
 basepython = {[testenv:py3]basepython}
+setenv = {[testenv]setenv}
```
booxter, line 40: isn't the default `testenv` section value already inherited here? why do we need to add these two lines?

requirements.txt
booxter: FYI this cap is being worked on at: https://github.com/instructlab/instructlab/pull/3264 There were some changes to transformers library recently that broke instructlab/training. Training library should be fixed first to be able to bump instructlab-training and remove the cap.

requirements.txt
booxter: Is it true? I don't think we ever tested lowest versions before. It would be great to have this kind of testing, but in its absence, I think it's fine to keep caps conservative and bump them only when we know of a particular incompatibility. Am I thinking about it in a wrong way?

dhellmann: We know that the version of vllm being used now only works with torch 2.6, for example. So saying that instructlab could work with 2.3 is clearly wrong.

requirements.txt
dhellmann: If instructlab-training already caps this library, why is that cap replicated here?

cdoern (commented):
nothing additional here, but I figured I'd +1 all of the other concerns raised, like to see them addressed before merging.

requirements.txt
booxter: If we *know* then sure let's bump.

requirements.txt
booxter: AFAIU no instructlab-training release currently caps the library. And the history in #3264 suggests that even the latest 0.8.1 is still not compatible with the latest transformers release.

dhellmann: OK, I misunderstood something you said. Since the code in this repository does import transformers, it makes sense to leave any cap here, even if other libraries also use transformers.

constraints.txt
courtneypacheco: Yep, great idea! I'll do that

requirements-vllm-cuda.txt
courtneypacheco: Sure, makes sense. I like that better. I'll remove that upper cap here and add to `constraints-dev.txt`

tox.ini
courtneypacheco: Ah, that's an artifact of my debugging. I agree. Let's remove it.

constraints.txt
courtneypacheco: Renamed this file to `constraints-dev.txt`!

requirements-vllm-cuda.txt
courtneypacheco: Updated.

requirements.txt
courtneypacheco: Changed

```
torch>=2.3.0,<2.7.0
```

to 

```
torch>=2.6.0,<2.7.0
```

just now

requirements.txt
courtneypacheco: I reverted this temporary change. I was just trying to figure out if that version was related to the `linux_train.py` failure I saw earlier. (It actually wasn't.)

tox.ini
courtneypacheco: Updated!

tox.ini
courtneypacheco: Removed

requirements.txt
```diff
@@ -32,9 +32,8 @@ sentencepiece>=0.2.0
 tokenizers>=0.11.1
 toml>=0.10.2
 # Default version. Can be overridden in extra requirements
-torch>=2.3.0,<2.6.0
+torch>=2.6.0,<2.7.0
 tqdm>=4.66.2
-# temporary cap until https://github.com/instructlab/training/pull/443 is merged and consumed within instructlab
 # above PR fixes interactions with newer versions of transformers through the training library
```
booxter, line 37: which "above" PR? :)

requirements.txt
```diff
@@ -32,9 +32,8 @@ sentencepiece>=0.2.0
 tokenizers>=0.11.1
 toml>=0.10.2
 # Default version. Can be overridden in extra requirements
-torch>=2.3.0,<2.6.0
+torch>=2.6.0,<2.7.0
```
booxter, line 35: does it have to be capped now that it's pinned via constraints? Do we know 2.7.0+ is broken?

requirements.txt
courtneypacheco: Whoops! good catch.

requirements.txt
courtneypacheco: Yeah, that's true. We can cap in the `constraints-dev.txt` and leave it uncapped here. I'll also leave a comment in this file letting people know to not put a cap here.

booxter (approved):
Looks fine. We'll need an automation to move the constraints forward over time.

ktdreyer (approved):

courtneypacheco:
@RobotSail when the large E2E job passes, will we be okay to merge this change and release it? Or, would you like any additional testing performed to ensure that training works as intended?

courtneypacheco:
We need to validate this change in an E2E test with `use_dolomite=True`  set. We also need to validate this against the Lama 70B model, which requires a new config.

courtneypacheco:
Please see: https://github.com/instructlab/instructlab/pull/3288

ktdreyer:
Should we merge https://github.com/instructlab/instructlab/pull/3288 instead of this?

requirements-vllm-cuda.txt
```diff
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 # Dependencies for installing vLLM on CUDA
 
-# vLLM only supports Linux platform (including WSL)
-vllm==0.7.3 ; sys_platform == 'linux' and platform_machine == 'x86_64'
+# vLLM only supports Linux platform (including WSL). Do not cap this dependency here. Cap in constraints-dev.txt
+vllm>=0.8.0 ; sys_platform == 'linux' and platform_machine == 'x86_64'
```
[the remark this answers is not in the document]
courtneypacheco: Good catch. Let me update this to remove the `x86_64` limitation.

PR #3286, opened 2025-04-11
pick up model id defaults from config yaml

jaideepr97:
allow setting a global student and teacher model id in the config, which can be overridden by cmd args 

**Checklist:**

### DIFF
--- a/src/instructlab/cli/data/generate.py
+++ b/src/instructlab/cli/data/generate.py
@@ -281,7 +281,14 @@ def generate(
     # determine student model arch from train section of config and pick system prompt to
     # pass to SDG appropriately
     system_prompt, legacy_pretraining_format = None, None
-    if not student_model_id:
+
+    student_id = (
+        student_model_id
+        if student_model_id
+        else ctx.obj.config.general.student_model_id
+    )
+
+    if not student_id:
         student_model_path = pathlib.Path(ctx.obj.config.train.model_path)
         student_model_arch = get_model_arch(student_model_path)
         system_prompt = get_sysprompt(student_model_arch)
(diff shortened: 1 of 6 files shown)


src/instructlab/cli/data/generate.py
```diff
@@ -168,16 +168,20 @@
...
+    cls=clickext.ConfigOption,
 )
 @click.option(
     "--teacher-model-id",
     # help="ID of the custom model you want to use as the teacher model.",
-    default=None,
     type=click.STRING,
-    show_default=True,
+    show_default=False,
+    config_class="general",
```
cdoern, line 182: idk if you need config_sections if its in the top level of `general`

jaideepr97: hmmm, doesnt seem to be making a difference either way

src/instructlab/configuration.py
```diff
@@ -85,6 +85,18 @@ class _general(BaseModel):
         default=False,
         description="Use legacy IBM Granite chat template (default uses 3.0 Instruct template)",
     )
+    # Global student model id
+    student_model_id: str | None = Field(
```
RobotSail, line 89: Try using `typing.NotRequired` here, we should see whether this type can allow us to avoid explicitly set this field within the config

```suggestion
    student_model_id: typing.NotRequired[str] = Field(
```

RobotSail: Specifically today all config fields will be rendered in the default on-disk config which is super confusing. I want to see if using this type will prevent this field from being listed there entirely until it's necessary.

RobotSail: @jaideepr97 I see, thank you for checking this. The way you have done it is perfect

RobotSail (approved):

cdoern (approved):

PR #3292, opened 2025-04-16
chore: Remove python 3.10 support

booxter:
**Checklist:**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,7 +62,6 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - "3.10"
           - "3.11"
         platform:
           - "ubuntu-latest"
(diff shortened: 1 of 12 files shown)


RobotSail (approved):

nathan-weinberg:
Motivation?

RobotSail:
@nathan-weinberg There are a number of features we use in Python3.11 that are not present in 3.10 so we'll often need to introduce compatibility logic to maintain it. It doesn't seem like there is a strong need to maintain support for 3.10 given that the upgrade to 3.11 is very seamless. Unless there is a pressing concern for something working on Python 3.10 that I'm not aware of, we should aim to move off of it sooner than later.

nathan-weinberg:
Makes sense to me :+1:

nathan-weinberg (approved):

PR #3303, opened 2025-04-18
Preserve storage after e2e tests

danmcp:
Updating to get a better idea over time how much storage is being used and the storage isn't needed after the test

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -143,7 +143,7 @@ jobs:
       - name: Run e2e test
         run: |
           . venv/bin/activate
-          ./scripts/e2e-ci.sh -m
+          ./scripts/e2e-ci.sh -m -p
 
       - name: Check disk after tests
         run: |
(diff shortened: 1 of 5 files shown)


ktdreyer:
@danmcp would you mind explaining a bit more about how you're testing this?

The GH Action will delete the node as soon as the e2e script completes, right? Are you doing something behind the scenes to avoid deleting the node?

danmcp:
> @danmcp would you mind explaining a bit more about how you're testing this?
> 
> The GH Action will delete the node as soon as the e2e script completes, right? Are you doing something behind the scenes to avoid deleting the node?

The change is useful because we print out the usage after the test is run.  Example:

image

Currently we delete the test results then delete the node immediately after.  That makes it so reporting the disks after the tests isn't helpful to determine how much disk was needed.

To test, I simply ran the script with the parameter to prove it works.  We won't see the actual numbers until it merges since the workflows don't take effect with pull request target.

booxter (approved):

nathan-weinberg (approved):

ktdreyer:
Oh, got it! This would be faster as well, then.

PR #3290, opened 2025-04-16
ci: temporarily disable small job on PRs

jaideepr97:
Temporarily disable the small job on PRs until https://github.com/instructlab/instructlab/issues/3289 is investigated 
this is simply to prevent blocking important changes that need to go in 

**Checklist:**

- [ ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the
  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] [Changelog](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md) updated with breaking and/or notable changes for the next minor release.
- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Functional tests have been added, if necessary.
- [ ] E2E Workflow tests have been added, if necessary.

Issue #3289, opened 2025-04-16
Logits issue with transformers >4.51.0


jaideepr97:
**Describe the bug**
There is a bug within the newer transformers releases (>4.51.0) that results in a consistent failure within `linux_train.py`
The training process fails somewhere inside of the transformers library call to generate samples and provides no stack trace, other than saying that a tuple object is generated somewhere that does not contain any logits

Updating parameters to the function call does not seem to fix the problem 
This issue only affects linux_train.py which is legacy code at this point

**To Reproduce**
1. Try to execute simple train pipeline in a linux environment or run small-test job in CI 
2. hit error

**Expected behavior**

**Screenshots**
```
LINUX_TRAIN.PY: SANITY CHECKING THE BASE MODEL
after tokenizer call
before generate call
'tuple' object has no attribute 'logits'
  0%|          | 0/5 [00:00<?, ?it/s]
  0%|          | 0/5 [00:00<?, ?it/s]
+ rm -rf /home/tmp/tmp.O2La1qkAMT
Error: Process completed with exit code 1.
```

**Device Info (please complete the following information):**
 - Hardware Specs: [e.g. Apple M2 Pro Chip, 16 GB Memory, etc.]
 - OS Version: [e.g. Mac OS 14.4.1, Fedora Linux 40]
 - Python Version: [output of `python --version`]
 - InstructLab Version: [output of `ilab system info`]

**Additional context**

jpodivin:
Instructlab version 0.26 is affected by this issue.

jpodivin:
I did some research on this. Exception is raised from [transformers utils.py](https://github.com/huggingface/transformers/blob/main/src/transformers/generation/utils.py#L3434 ), it's only obfuscated by lack of trace.

While I'm not certain what exactly is going wrong, I did find a workaround.

```
diff --git a/src/instructlab/train/linux_train.py b/src/instructlab/train/linux_train.py
index b651251f..0e0f218b 100644
--- a/src/instructlab/train/linux_train.py
+++ b/src/instructlab/train/linux_train.py
@@ -212,7 +212,7 @@ def linux_train(
     # Loading the model
     print("LINUX_TRAIN.PY: LOADING THE BASE MODEL")
     config = AutoConfig.from_pretrained(
-        model_name, torchscript=True, trust_remote_code=True
+        model_name, trust_remote_code=True
     )
 
[...]

PR #3307, opened 2025-04-21
fix: Removing torchscript from autoconfig

jpodivin:
Simple pipeline hasn't been working for some time now, due to #3289 . I've tracked the issue down in [transformers utils.py](https://github.com/huggingface/transformers/blob/main/src/transformers/generation/utils.py#L3434 ).

It appears that usage of torchscript, is changing type of output from invocations of `GenerationMixin` objects. I'm not sure about the exact reason why is this happening. But the behavior is normalized by simply removing `torchscript` parameter from config.

Since purpose of  [Torchscript](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) is to improve performance, I can guess that merging this fix may cause the pipeline to run a bit slower. But at least it will run. 

Proper fix should be implemented in transformers, but I don't have enough experience with the code base to know what to look for. 

I'm also reverting https://github.com/instructlab/instructlab/pull/3290 since things should be working now.

**Checklist:**

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -50,8 +50,26 @@ pull_request_rules:
         - -files=.github/workflows/e2e-nvidia-l4-x1.yml
 
     # small e2e workflow
-    # temporarily remove requirement for small job passing
-    # until https://github.com/instructlab/instructlab/issues/3289 is fixed
+    - or:
+      - and:
+        # note this should match the triggering criteria in 'e2e-nvidia-t4-x1.yml'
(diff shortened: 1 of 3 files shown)


jpodivin:
I'm executing the pipeline now, looking for further issues. But so far I'm progressing well, with loss dropping as expected.

RobotSail (approved):
LGTM, thank you for finding and fixing this!

courtneypacheco:

PR #3320, opened 2025-04-28
use `constraints-dev.txt` in e2e tests

ktdreyer:
For instructlab, `pip install .` does not install `vllm`, but it does install an uncapped `torch` (`2.7.0` currently).

When we install `vllm` later, we compile a binary `flash_attn` wheel against `torch 2.7.0`. `vllm 0.8.4` requires `torch==2.6.0`, so we downgrade `torch`, and then we use that with the incompatible `flash_attn` binary wheel.

To resolve this, use `constraints-dev.txt` in the first `pip install` operation. This restricts `torch` to `2.6.0` immediately when we first install instructlab, so that we will compile `flash_attn` against that `torch` version.

### DIFF
--- a/.github/workflows/e2e-aws-custom.yml
+++ b/.github/workflows/e2e-aws-custom.yml
@@ -148,7 +148,7 @@ jobs:
           nvidia-smi
           python3.11 -m pip cache remove llama_cpp_python
 
-          CMAKE_ARGS="-DGGML_CUDA=on" python3.11 -m pip install -v .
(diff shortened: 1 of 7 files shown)


ktdreyer:
This fixes the flash-attn problem. The e2e tests get further, but they still fail in NCCL timeouts. I've filed https://github.com/instructlab/instructlab/issues/3321 to track that separately.

cdoern (approved):

booxter:
Force merging since Ken confirmed this improves situation even if it doesn't make CI green yet.

courtneypacheco:

PR #3326, opened 2025-04-29
fix: Ignore unraisable exceptions during unitcov testing

### LINKED ISSUE #3324: Unit tests fail with "AttributeError: 'TesseractOcrModel' object has no attribute 'reader'"
https://github.com/instructlab/instructlab/actions/runs/14734963356/job/41358694626?pr=3323

```
 _______________________ test_convert_pdf_from_directory ________________________
  
  cls = <class '_pytest.runner.CallInfo'>
  func = <function call_and_report.<locals>.<lambda> at 0x7fbcb5ef98a0>
  when = 'call'
  reraise = (<class '_pytest.outcomes.Exit'>, <class 'KeyboardInterrupt'>)
  
      @classmethod
      def from_call(
          cls,
          func: Callable[[], TResult],
          when: Literal["collect", "setup", "call", "teardown"],
          reraise: type[BaseException] | tuple[type[BaseException], ...] | None = None,
      ) -> CallInfo[TResult]:
          """Call func, wrapping the result in a CallInfo.
      
          :param func:
              The function to call. Called without arguments.
          :type func: Callable[[], _pytest.runner.TResult]
          :param when:
              The phase in which the function is called.
          :param reraise:
              Exception or exceptions that shall propagate if raised by the
              function, instead of being wrapped in the CallInfo.
          """
          excinfo = None
          start = timing.time()
          precise_start = timing.perf_counter()
          try:
  >           result: TResult | None = func()
  
  .tox/py311-unitcov/lib/python3.11/site-packages/_pytest/runner.py:341: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  .tox/py311-unitcov/lib/python3.11/site-packages/_pytest/runner.py:242: in <lambda>
      lambda: runtest_hook(item=item, **kwds), when=when, reraise=reraise
  .tox/py311-unitcov/lib/python3.11/site-packages/pluggy/_hooks.py:513: in __call__
      return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
  .tox/py311-unitcov/lib/python3.11/site-packages/pluggy/_manager.py:120: in _hookexec
      return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  .tox/py311-unitcov/lib/python3.11/site-packages/_pytest/threadexception.py:92: in pytest_runtest_call
      yield from thread_exception_runtest_hook()
  .tox/py311-unitcov/lib/python3.11/site-packages/_pytest/threadexception.py:68: in thread_exception_runtest_hook
      yield
  .tox/py311-unitcov/lib/python3.11/site-packages/_pytest/unraisableexception.py:95: in pytest_runtest_call
      yield from unraisable_exception_runtest_hook()
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
      def unraisable_exception_runtest_hook() -> Generator[None]:
          with catch_unraisable_exception() as cm:
              try:
                  yield
              finally:
                  if cm.unraisable:
                      if cm.unraisable.err_msg is not None:
                          err_msg = cm.unraisable.err_msg
                      else:
                          err_msg = "Exception ignored in"
                      msg = f"{err_msg}: {cm.unraisable.object!r}\n\n"
                      msg += "".join(
                          traceback.format_exception(
                              cm.unraisable.exc_type,
                              cm.unraisable.exc_value,
                              cm.unraisable.exc_traceback,
                          )
                      )
  >                   warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
  E                   pytest.PytestUnraisableExceptionWarning: Exception ignored in: <function TesseractOcrModel.__del__ at 0x7fbcb5473240>
  E                   
  E                   Traceback (most recent call last):
  E                     File "/home/runner/work/instructlab/instructlab/.tox/py311-unitcov/lib/python3.11/site-packages/docling/models/tesseract_ocr_model.py", line 106, in __del__
  E                       if self.reader is not None:
  E                          ^^^^^^^^^^^
  E                   AttributeError: 'TesseractOcrModel' object has no attribute 'reader'
```

bbrowning:
Newer versions of Docling can raise an exception during custom garbage collection code of TesseractOcrModel that nothing ever catches, because exceptions during garbage collection are generally not caught by anything. The exception is harmless to any of our InstructLab use-cases, and only happens when the TesseractOcrModel was not able to construct itself anyway, which we check for and handle to fallback to EasyOCR or other implementations.

However, py3-unitcov by default fails the test suite if any of these unraisable exceptions happen. This adjusts that to not fail the test suite. Messages still get logged about the exception during regular py3-unit testing, so we can see when this gets fixed in Docling, but there's no need for it to actually fail our test suite.

Resolves #3324

### DIFF
--- a/tests/test_lab_rag_convert.py
+++ b/tests/test_lab_rag_convert.py
@@ -18,6 +18,7 @@ from docling.datamodel.base_models import (  # type: ignore  # noqa: F401
 from docling.datamodel.document import ConversionResult  # type: ignore  # noqa: F401
 from docling.datamodel.document import InputDocument  # type: ignore  # noqa: F401
 from docling.document_converter import FormatOption  # type: ignore  # noqa: F401
+import pytest
 
 # First Party
 from instructlab import lab
@@ -98,6 +99,7 @@ def run_rag_convert_test(
 
 
 @dev_preview
+@pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
 def test_convert_pdf_from_directory(tmp_path: Path):
     """
     Tests converting from the sample PDF in tests/testdata/documents/pdf.
@@ -115,6 +117,7 @@ def test_convert_pdf_from_directory(tmp_path: Path):
 
 
 @dev_preview
+@pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
 def test_convert_md_from_directory(tmp_path: Path):
     """
     Tests converting from the sample Markdown file in tests/testdata/documents/md.
@@ -133,6 +136,7 @@ def test_convert_md_from_directory(tmp_path: Path):
 # is not working or if it ever gets deleted.  That's not ideal, but we do need to test these capabilities.
 # TODO: Consider re-working this with a mock for the github server.
 @dev_preview
+@pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
 def test_convert_md_from_taxonomy(tmp_path: Path):
     """
     Tests converting from the sample Markdown file in a github repo referenced in tests/testdata/sample_taxonomy.


tox.ini
```diff
@@ -28,7 +28,7 @@ commands =
     ilab --version
     {envpython} -m instructlab --version
     unit: {envpython} -m pytest {posargs:tests}
-    unitcov: {envpython} -W error::UserWarning -m pytest --cov=instructlab --cov-report term --cov-report=html:coverage-{env_name} --cov-report=xml:coverage-{env_name}.xml --html=durations/{env_name}.html {posargs:tests -m "not (examples or slow)"}
+    unitcov: {envpython} -W error::UserWarning -m pytest --cov=instructlab --cov-report term --cov-report=html:coverage-{env_name} --cov-report=xml:coverage-{env_name}.xml --html=durations/{env_name}.html -p no:unraisableexception {posargs:tests -m "not (examples or slow)"}
```
booxter, line 31: There's also a "thread" flavor of the same warning which we may want to disable if we believe these should stay disabled even after docling fixes the problem: https://docs.pytest.org/en/7.1.x/how-to/failures.html#warning-about-unraisable-exceptions-and-unhandled-thread-exceptions

---

That said, I'm a bit uneasy to leave these warnings disabled for all code under test and I'd rather plan to revert this change later. (So maybe add a TODO comment to revert it?)

There's also [pytest.mark.filterwarnings](https://docs.pytest.org/en/7.1.x/reference/reference.html#id22)[¶](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-filterwarnings) that allows to mark just the tests that we expect to raise unhandled exception (that would be RAG tests in our case) while leaving other test cases protected. Maybe worth narrowing down the effect.

bbrowning: We could try filterwarnings - I didn't go that route in SDG itself because most of our tests hit this code path. But, that's a good point that here I can probably filter it more selectively since most of the code shouldn't be loading docling for testing.

bbrowning: Ok, pushed an update that switches to only targeted filterwarnings usage to ignore these docling errors.

booxter (approved):
Some TODO comments suggesting that this can removed after bug XXX is fixed in docling would be nice but I won't hold for it. Thanks!!!

nathan-weinberg (approved):

courtneypacheco:

cdoern:
@mergifyio backport release-v0.26

PR #3337, opened 2025-04-30
ci: add workflow_dispatch for all e2e jobs

booxter:
This allows to test workflow changes to these jobs before merging them.

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -21,7 +21,12 @@ on:
       - 'scripts/e2e-ci.sh' # Used by this workflow
       - 'scripts/test-data/**'
       - 'src/instructlab/profiles/**'
-  workflow_dispatch: {}
+  workflow_dispatch:
+    inputs:
+      pr_or_branch:
+        description: 'pull request number or branch name'
+        required: true
+        default: 'main'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.ref }}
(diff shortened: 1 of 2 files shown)


booxter:
@ktdreyer ignoring the security check, any idea how this can be tested, considering that there's no Run workflow button for these jobs? (because the `workflow_dispatch` setting is not added to them)

booxter:
I believe it's a mistake in the ci action that it targets workflows of pull_request_target type, fix here: https://github.com/instructlab/ci-actions/pull/14

booxter:
One idea to make landing this risky is:

- first, merge just bare "workflow_dispatch" addition (without inputs) - this should be relatively safe.
- then on second stage, we can add inputs - at this point we should be able to test pre-merge with the Run button.

@ktdreyer @courtneypacheco thoughts?

ktdreyer (approved):
This looks ok to me. You are right, you're just copying logic from the other jobs.

I think we should discuss the overall structure of how we do this pattern and ideas for hardening and optimizing this, but that is out of scope for this PR.

booxter:
@mergifyio rebase

booxter:
To pass gate, we'll need to disable enforcement of security-check, as I explain in: https://github.com/instructlab/instructlab/pull/3353

booxter:
Testing new functionality on this PR:  https://github.com/instructlab/instructlab/pull/3101

booxter:
So it's impossible to use `gh` without a token in CI, even for checkouts (I was hoping we could stay token-less by removing Slack / Discord / PR commenting functionality). We can use a regular `checkout` action to access a particular PR, probably. But this escalates into a territory where I'd like DevOps folks to step in and advise.

tl;dr we need a way to Run workflow for small and medium jobs that are ALSO triggered on `pull_request_target` - which makes `security-lint` job fail if we try to pull a secret like `GH_TOKEN`. This is not a problem for large jobs that are not `pull_request_target` and hence the linter rule is not executed on these other workflow files.

booxter:
I don't have capacity to complete it.

PR #3353, opened 2025-05-02
ci: Don't require security-lint for auto-merge

booxter:
The check has known false positives [1], so we cannot make it a
requirement for mergify. It's an informational signal for reviewers that
additional scrutiny may be justified.

[1]: https://github.com/instructlab/ci-actions/pull/14#issuecomment-2848052075

**Checklist:**

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -136,5 +136,4 @@ jobs:
     uses: ./.github/workflows/status-checks.yml
     with:
       job_ids: >- # Space-separated job ids to wait on for status checks
-        security-lint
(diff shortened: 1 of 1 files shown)


booxter:
Unblocking this to allow merging; I believe this is the right thing to do, knowing that the check is designed to trigger false positives, and so it cannot be used as a merge requirement and/or block running other jobs.

booxter:
Maybe I won't need after all if I get away without GH_TOKEN (that may not be needed for just checking out the current PR from git.)

ktdreyer (approved):

Issue #3299, opened 2025-04-17
Use pip constraints file for all dependencies in CI


booxter:
We are often broken by a new dependency upstream release. Instead of liberally allowing every new release into our CI, potentially breaking it, we should:

1. constrain all dependencies in constraints.txt file that we recently added.
2. make sure the constraints are applied to all our jobs.
3. uncap all dependencies as a matter of policy.
4. have a bot that would periodically propose PRs with higher constraints so that we can validate new upstream releases don't break anything and move forward over time WITHOUT BREAKING CI.

booxter:
This is related to https://github.com/instructlab/instructlab/issues/2454 that asks for policy / documentation. Which is already captured in dev-docs.

booxter:
An example of catastrophe due to upstream version bump: https://github.com/instructlab/instructlab/issues/3324 that could be avoided with constraints.

PR #3364, opened 2025-05-09
ci[cuda]: build project from local checkout, not from PyPI

### LINKED ISSUE #3339: CUDA CI job should pull instructlab from the PR, not PyPI
[As per Ken](https://github.com/instructlab/instructlab/issues/3110#issuecomment-2843202664), the job is currently pulling from PyPI. This should be fixed.

compi-migui:
This makes it possible to build the CUDA container with local changes, easing the developer workflow.

It also allows CI to actually build the container with the changes proposed in a PR, so that we can detect when a change breaks the build before it's merged rather than when it makes it into a release.

A new PIP_CONSTRAINT argument allows respecting constraints files. This is disabled by default when using `make cuda` or building the Containerfile directly, but we enable it in our CI job for more stability.

Resolves #3339
Partially addresses #3299

**Checklist:**


nathan-weinberg (approved):

compi-migui:
@nathan-weinberg sure thing, rebased it.

booxter (changes requested):
Please fix build when `CONSTRAINTS_FILE` is not passed.

compi-migui:
@booxter thanks, it should be fixed now.

booxter:
@compi-migui how do you test the build for the container? I get this when doing `make cuda` from the root of the repo:

```
STEP 58/112: COPY containers/cuda/intel-oneapi.repo /etc/yum.repos.d/oneapi.repo
Error: building at STEP "COPY containers/cuda/intel-oneapi.repo /etc/yum.repos.d/oneapi.repo": checking on sources under "/home/ec2-user/src/instructlab/containers/cuda": copier: stat: "/containers/cuda/intel-oneapi.repo": no such file or directory
make: *** [Makefile:74: cuda] Error 125
```

I don't think it's a problem with this patch, but I wonder if you were able to build to completion and I'm just doing something wrong.

compi-migui:
@booxter you can get it built from the project root like so, with and without constraints:
```
podman build --build-arg=CONSTRAINTS_FILE="constraints-dev.txt" -f containers/cuda/Containerfile .
podman build -f containers/cuda/Containerfile .
```
Not doing a full build right now because it takes forever, but I've made sure it gets past that COPY command. I've reproduced the issue running `make cuda` so it isn't just you.

Digging through git history I think https://github.com/instructlab/instructlab/pull/1961 inadvertently broke `make cuda`. Note it's setting `containers/cuda` as the build context but the COPY command uses a path that starts from the project root, `containers/cuda/intel-oneapi.repo`. That's the inconsistency breaking the make target.

The two options are:

1. Have `make cuda` use the project root as the build context, getting rid of all the `CUDA_CONTEXT_DIR` mentions in the Makefile. That way the Containerfile's attempt to `COPY containers/cuda/intel-oneapi.repo` will work.
2. Keep the smaller build context and have that COPY command use a source path relative to that context, so rewrite it to `COPY intel-oneapi.repo /etc/yum.repos.d/oneapi.repo`.

Option 2 would prevent us from copying in the whole project to build it from source, including the contraints file, so it's a non-starter.

Option 1 is what the GitHub workflow [already does](https://github.com/instructlab/instructlab/blob/main/.github/workflows/image-cuda.yml#L67) (`context: .`). I could make the changes in this PR but it's probably better as a separate one.

booxter (approved):

booxter:
FYI I removed `hold` because I believe that there's no product implications in fixing the `Containerfile` to consider.

PR #3370, opened 2025-05-12
tests: incomplete cmds have non-zero exit codes

### LINKED ISSUE #3368: CI failed due to click==8.2.0 changing return code for argument-less help call (`ilab`) from 0 to 2
https://github.com/instructlab/instructlab/actions/runs/14954396543/job/42008222190?pr=3367

```
=========================================================================================================================================== test session starts ============================================================================================================================================
platform darwin -- Python 3.11.12, pytest-8.3.5, pluggy-1.5.0
rootdir: /Users/ihrachys/src/instructlab
configfile: pyproject.toml
plugins: html-4.1.0, anyio-4.9.0, metadata-3.1.1, langsmith-0.3.42, asyncio-0.26.0, cov-6.1.1
asyncio: mode=Mode.STRICT, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 366 items / 365 deselected / 1 selected

tests/test_lab.py F                                                                                                                                                                                                                                                                                  [100%]

================================================================================================================================================= FAILURES =================================================================================================================================================
__________________________________________________________________________________________________________________________________________ test_ilab_alias_output __________________________________________________________________________________________________________________________________________

cli_runner = <click.testing.CliRunner object at 0x337527b50>

    def test_ilab_alias_output(cli_runner: CliRunner):
        expected_output = """Aliases:
      chat      model chat
      generate  data generate
      serve     model serve
      train     model train"""
        result = cli_runner.invoke(lab.ilab)
>       assert result.exit_code == 0, result.stdout
E       AssertionError:
E       assert 2 == 0
E        +  where 2 = <Result SystemExit(2)>.exit_code

/Users/ihrachys/src/instructlab/tests/test_lab.py:173: AssertionError
```

This is due to d8763b93021c416549b5f8b4b5497234619410db

```
commit d8763b93021c416549b5f8b4b5497234619410db
Author: fdavis <fdavis@users.noreply.github.com>
Date:   Sun Nov 3 11:56:41 2024 -0800

    Help shown via no_args_is_help results in exit code 2, was 0 (#1489)

    Co-authored-by: David Lord <davidism@gmail.com>
    Co-authored-by: Andreas Backx <andreas@backx.org>
```

ktdreyer:
With Click 8.2.0, incomplete commands now exit with non-zero exit codes.  This matches the behavior of other utilities like `git` or `mv`.

Update the unit tests to assert the non-zero exit code values.

This PR will:
- Fix our instructlab tests to be compatible with click 8.1 and 8.2.0
- Loosen the click version range in our own `requirements.txt` file (reverting 01458cdcfbfdfacdfcb132c91c89cb771b3053c4)

But GH Actions CI will still test with click 8.1.8. We won't be able to update `constraints-dev.txt` until the docling project removes their constraint on `click<8.2.0`.

Resolves #3368

### DIFF
--- a/tests/test_lab.py
+++ b/tests/test_lab.py
@@ -86,10 +86,21 @@ def test_ilab_cli_imports(testdata_path: pathlib.Path):
 
 
 class Command(typing.NamedTuple):
+    """
+    Represent an ilab CLI sub-command.
+
+    should_fail=False means the command should exit with a zero exit code.
+    should_fail=True means the command should exit with a non-zero exit code.
+    should_fail=None means the tests should ignore the exit code.
+
+    (Click 8.2.0+ exits with non-zero on incomplete commands, and Click
+    8.1.x does not.)
+    """
+
     args: tuple[str, ...]
     extra_args: tuple[str, ...] = ()
     needs_config: bool = True
-    should_fail: bool = True
+    should_fail: bool | None = True
 
     def get_args(self, *extra, default_config: bool = True) -> list[str]:
         args = []
@@ -109,12 +120,12 @@ class Command(typing.NamedTuple):
 subcommands: list[Command] = [
     # split first and second level for nicer pytest output
     # first, second, extra args
-    Command((), needs_config=False, should_fail=False),
-    Command(("config",), needs_config=False, should_fail=False),
+    Command((), needs_config=False, should_fail=None),
+    Command(("config",), needs_config=False, should_fail=None),
     Command(("config", "edit")),
     Command(("config", "init"), needs_config=False, should_fail=False),
     Command(("config", "show")),
-    Command(("model",), needs_config=False, should_fail=False),
+    Command(("model",), needs_config=False, should_fail=None),
     Command(("model", "chat")),
     Command(("model", "chat"), ("--rag",)),
     Command(("model", "convert"), ("--model-dir", "test")),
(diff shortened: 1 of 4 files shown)


tests/test_lab.py
```diff
@@ -160,7 +160,7 @@ def test_ilab_cli_help(command: Command, cli_runner: CliRunner):
     cmd = command.get_args("--help")
     assert "--help" in cmd
     result = cli_runner.invoke(lab.ilab, cmd)
-    assert result.exit_code == 0, result.stdout
+    assert result.exit_code == 2, result.stdout
```
booxter, line 163: no we should just not test return code at all. otherwise this change makes tests fail on older click version. the return code is of no consequence.

booxter (changes requested):

tests/test_lab.py
ktdreyer: I'm ok with removing the exit code check. This test was bordering on testing "implementation not behavior"

booxter: Right, I was wondering but wasn't sure. I think they only changed return code on "bare" call (when no "correct" option is passed - and `--help` is one of the correct options.)

booxter:
This will also require an update to e2e script that calls `ilab` (and runs with `set -e`, which makes it fail). The fix is probably to switch to `ilab --help`.

booxter:
@mergifyio rebase

booxter:
```
  =================================== FAILURES ===================================
  ____________________________ test_ilab_alias_output ____________________________
  
  cli_runner = <click.testing.CliRunner object at 0x7fabfd79e6d0>
  
      def test_ilab_alias_output(cli_runner: CliRunner):
          expected_output = """Aliases:
        chat      model chat
        generate  data generate
        serve     model serve
        train     model train"""
          result = cli_runner.invoke(lab.ilab)
      
          alias_section = False
          actual_output = []
  >       result_out = result.stdout + result.stderr
  
  /home/runner/work/instructlab/instructlab/tests/test_lab.py:178: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  self = <Result okay>
  
      @property
      def stderr(self) -> str:
          """The standard error as unicode string."""
          if self.stderr_bytes is None:
  >           raise ValueError("stderr not separately captured")
  E           ValueError: stderr not separately captured
```

ktdreyer:
We hit that `ValueError` with the old Click 8.1.8. I'll add a check for `result.stderr_bytes` prior to using `result.stderr`.

The bigger issue is that `tox` is still testing with Click 8.1.8 instead of 8.2.0 even when I remove the cap in `requirements.txt`. hmm.

booxter:
Tox will test whatever is in constraints now; the cap is of no influence (except when calculating constraints). Even if you uncap, it doesn't guarantee that the constraint will move forward since its value depends on all other dependencies caps for the same library. So if any one dependency caps it at <8.2.0, the constraint won't move. That's by design.

booxter:
@mergifyio rebase

I think your branch doesn't include the new script or smth.

cdoern (approved):

PR #3371, opened 2025-05-12
fix(deps): Constrain `click<=8.1.8` temporarily

courtneypacheco:
**Checklist:**

Temporarily constrain `click<=8.1.8` to avoid consuming the latest breaking changes in our CI introduced in `click-8.2.0`. We will simultaneously work on allowing InstructLab to consume these latest changes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -138,10 +138,12 @@ jobs:
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
           python-version: ${{ matrix.python }}
-          cache: pip
-          cache-dependency-path: |
-            **/pyproject.toml
-            **/requirements*.txt
+          ## Temporarily comment out these lines because of: https://stackoverflow.com/questions/65122957/resolving-new-pip-backtracking-runtime-issue
+          ## Undo the below commented config once this issue is resolved: https://github.com/instructlab/instructlab/issues/3368
+          # cache: pip
+          # cache-dependency-path: |
+          #   **/pyproject.toml
+          #   **/requirements*.txt
 
       - name: Remove llama-cpp-python from cache
         run: |
@@ -157,6 +159,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          ## Remove the below line once this issue is resolved: https://github.com/instructlab/instructlab/issues/3368
+          python -m pip install -r requirements-dev.txt -c constraints-dev.txt --use-deprecated=legacy-resolver
           python -m pip install tox tox-gh>=1.2
 
       # see https://github.com/instructlab/instructlab/issues/1886
--- a/constraints-dev.txt
+++ b/constraints-dev.txt
@@ -1,4 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# TODO: Remove constraint once this issue is resolved: https://github.com/instructlab/instructlab/issues/3368
+click<=8.1.8
+# Testing
+torchvision<=0.21.0
 torch<2.7.0
 vllm<0.9.0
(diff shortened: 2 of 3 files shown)


booxter (commented):
*If* we cap, we must explain why - in a comment above the cap, linking to the issue that tracks uncapping.

In this case, i don't think we should cap at all because there's no adoption work for us to do, we just need to fix tests, which is what https://github.com/instructlab/instructlab/pull/3370 is trying to do.

We also have constraints-dev.txt file in case the test fix is for some reason not expedient enough. (Again, if we do constrain, we should document in a comment why we are doing it. At least until there's a bot that constrains everything by default.)

(I think we should also be more strict with language here: the PR says it "constrains" but really it "caps". Constraints are defined in constraints file.)

requirements.txt
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 boto3>=1.35.96
-click>=8.1.7,<9.0.0
+click>=8.1.7
```
ktdreyer, line 4: 8.1.8 seems to be the last version that works

requirements.txt
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 boto3>=1.35.96
-click>=8.1.7,<9.0.0
```
booxter, line 4: This is correct change overall but probably doesn't belong here. (But I will not block on this.)

constraints-dev.txt
```diff
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
+click<=8.1.8
```
booxter, line 3: please add a comment above linking to the issue so that we know when we can move it forward.

booxter (changes requested):
Please add a comment above the "cap-constraint" and then I'm ready to approve it. Ken's fix is the ultimate patch but will take some time since e2e is not fixed there, yet.

requirements.txt
ktdreyer: I misunderstood this diff. You're just keeping the minimal version at 8.1.7, not going to 8.1.8, but removing the upper bound. Got it.

booxter (approved):

ktdreyer (approved):

.github/workflows/test.yml
```diff
@@ -157,6 +158,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          ## Temp testing -- not a permanent change
+          python -m pip install -r requirements-dev.txt -c constraints-dev.txt --use-deprecated=legacy-resolver
```
ktdreyer, line 162: This PR fixes two bugs in a single commit.

It will be difficult to revert the original click 8.1 constraint change. Can you please separate them so it will be easier to revert this?

courtneypacheco: I'm not sure I understand. Can't we just create a new PR to undo this change once we update the source code to be compatible with the newest version of click?

ktdreyer: The switch to `legacy-resolver` is unrelated to the `click` change, right? If it is unrelated, it will be clearer to have these two bug fixes as two separate commits.

booxter: We should not switch to legacy resolver, it allows to install incompatible version.

booxter: (btw I support Ken's proposal to separate changes into separate commits that we can revert with `git revert` - manually reverting these is a chore)

booxter:
functionals also fail here with

```
      Fast download using 'hf_transfer' is enabled (HF_HUB_ENABLE_HF_TRANSFER=1) but 'hf_transfer' package is not available in your environment. Try `pip install hf_transfer`.
```

even though requirements.txt pull it through optional dependencies:

```
```

This solver is broken.

booxter:
A version of this included in https://github.com/instructlab/instructlab/pull/3374

Issue #3373, opened 2025-05-13
tox takes a long time to install packages for unit tests


ktdreyer:
**Describe the bug**
GitHub Actions takes a very long time to install dependencies for our unit tests.

**To Reproduce**
1. Open a PR
2. Watch the logs for a job that runs `tox -e py3-unit`
3. Observe the timing, it takes a long time

**Expected behavior**
`tox` installs dependencies and executes in a reasonable amount of time.

ktdreyer:
Ihar's change https://github.com/instructlab/instructlab/pull/3372 provides more debugging information.

Courtney found that we can add a `--use-deprecated=legacy-resolver` argument to `pip install`. Some more investigation showed that might lead to conflicts.

More investigation needed

PR #3374, opened 2025-05-13
Mega CI-fix: tighten click cap; force pip constraints on all dependencies; improve ci log verbosity; use "fast" deps resolver for pip; uncap pydantic for mypy; ...

booxter:
- **fix: Tighten click cap until we fix tests for click 8.2.0**
- **fix: Consider constraints file for setup-python pip cache key**
-  **chore: Enable verbose output for tox**
- **chore: use --use-feature fast-deps when pip installing**
- **fix: pin all dependencies to the floor**
- **chore: don't min-cap pydantic specifically for mypy job**

Closes: #3373

**Checklist:**


booxter:
@mergifyio rebase

src/instructlab/configuration.py
```diff
@@ -209,7 +209,8 @@ class _serve_vllm(BaseModel):
     gpus: Optional[int] = Field(default=None, description="Number of GPUs to use.")
     # arguments to pass into vLLM process
     vllm_args: list[str] | None = Field(
-        default_factory=list,
+        # TODO: revisit the type check ignore
+        default_factory=list,  # type: ignore
```
cdoern, line 213: Random, but should this be `List` from `typing`. Not sure what the error is, but this seems off in general to me

booxter: `List` and `list` are identical I think; but yes it seems off. Someone should check if it's not some bug.

courtneypacheco (approved):

src/instructlab/configuration.py
fynnsu: `List` from `typing` is only needed if you're supporting python < 3.9 ([stack overflow](https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc)).

The problem here is actually the `| None`. The default factory ensures that we will always produce a `list` so it isn't necessary for this value to be optional and removing the `| None` satisfies mypy.

src/instructlab/configuration.py
```diff
@@ -1231,7 +1232,8 @@ def config_to_commented_map(
             # If the default value comes from a factory, evaluate it and use the result as the default value
             if default_value is PydanticUndefined:
                 if default_factory is not None and callable(default_factory):
-                    default_value = default_factory()
+                    # TODO: revisit the type check ignore
+                    default_value = default_factory()  # type: ignore
```
fynnsu, line 1236: This one is because there are two different valid types for default factory. One is `() -> Any` and the other is `(dict[str, Any]) -> Any`. The second type is so you can pass a dictionary of "validated values" to the factory.

Because there are two possible types for this, one of those types is incompatible with being called without any arguments, which results in
```
error: Too few arguments  [call-arg]
```
The quick fix for this is adding a typing `cast` call.
The better way to fix this would be to update this code to instead use [get_default()](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.FieldInfo.get_default) but that can be done later as needed.

booxter:
@fynnsu please post PR fixing the issues with the types since you know how to fix these. Thanks.

cdoern (approved):

PR #3453, opened 2025-05-30
Ignore mypy check for Haystack

jwm4:
This change adds and ignore directive to a line that mypy doesn't like relating to Haystack.  In general, Haystack doesn't have as much strong typing as mypy prefers, so it is important to tell mypy to accept Haystack constructs.  I tried running this locally and it seems to work fine so I think the ignore directive is fine.

**Checklist:**


booxter:
Since the PR includes lots of pin moves, I'm running large job here: https://github.com/instructlab/instructlab/actions/runs/15352377148 to confirm nothing is broken.

booxter:
I dropped the pin bumps so that we can merge it without further validation.

courtneypacheco (approved):

PR #3478, opened 2025-06-10
chore: Bump ci-actions version to v0.2 + replace commit SHA in `constraints-update.yml`

courtneypacheco:
**Checklist:**

We have a new Y-stream release: https://github.com/instructlab/ci-actions/releases

This PR bumps the minimum versions to consume the latest y-stream release's changes.

Note that we haven't made any changes to the `launch-ec2-runner-with-fallback` CI action in release v0.2, but I'm bumping the minimum version in case we do need to consume future bugfixes, etc. for that action.

### DIFF
--- a/.github/workflows/constraints-update.yml
+++ b/.github/workflows/constraints-update.yml
@@ -21,8 +21,7 @@ jobs:
         with:
           repository: instructlab/ci-actions
           path: ci-actions
-          # no tag that includes https://github.com/instructlab/ci-actions/pull/26, yet
-          ref: 88641ccaf122964eacdc1a82b18bda369b6f99bd # main
+          ref: v0.2.1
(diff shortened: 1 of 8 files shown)


booxter (commented):
These should be tags or hashes, not branches, to avoid breakage without explicit testing. In general, hashes are preferable since they are more secure but since we control the repo, it would be ok to use tags too.

courtneypacheco:
@booxter sure. Would you be good with this PR if I just revert the change in `constraints-update.yml`?

I could, alternatively, update the ref to use the v0.2.1 tag

booxter:
Yes it would be enough since then you don't break anything that is not already broken. But using hashes should be easy too.

booxter (approved):

PR #3481, opened 2025-06-12
Bump llama_cpp_python to 0.3.8

### LINKED ISSUE #3474: Fail to install llama-cpp-python with error "error: ‘uint32_t’ does not name a type uint32_t read_u32() const"
**Describe the bug**

After following the step-by-step of the official documentation to install InstructLab, it fails in the llama-cpp-python installation part with the following error:

```
Building wheels for collected packages: llama_cpp_python
  Building wheel for llama_cpp_python (pyproject.toml) ... \
error
  error: subprocess-exited-with-error
[...]
      In file included from /tmp/pip-install-r2am43ta/llama-cpp-python_d414036923ef456f9ebbaa87a39421b2/vendor/llama.cpp/src/llama-mmap.cpp:1:
      /tmp/pip-install-r2am43ta/llama-cpp-python_d414036923ef456f9ebbaa87a39421b2/vendor/llama.cpp/src/llama-mmap.h:26:5: error: ‘uint32_t’ does not name a type
         26 |     uint32_t read_u32() const;
            |     ^~~~~~~~
[...]
      *** CMake build failed
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for llama_cpp_python
Failed to build llama_cpp_python
ERROR: Failed to build installable wheels for some pyproject.toml based projects (llama_cpp_python)
```

It only works after exporting the following argument as a variable before running the install command:

```
export CXX="g++ -include cstdint"
```

**To Reproduce**
1. Follow the step-by-step of the official documentation to install InstructLab in a Linux CPU host operating system.
2. See the process failing when it gets to the llama-cpp-python installation part.

**Expected behavior**

It should install InstructLab cleanly, as per the official documentation, with no extra steps required.

**Screenshots**

 - Hardware Specs: [e.g. Apple M2 Pro Chip, 16 GB Memory, etc.]
   ThinkPad T480s, 24GiB Memory, 256GiB NVMe disk, Intel Core i7 vPro 8th Gen
 - OS Version: [e.g. Mac OS 14.4.1, Fedora Linux 40]
   Fedora Linux 42 (Workstation Edition)
 - Python Version: [output of `python --version`]
   Python 3.11.12
 - InstructLab Version: [output of `ilab system info`]
   ilab version 0.26.1

**Additional context**

Check the attached logs for more details. The build error points to `llama-mmap.h`, specifically to lines involving `uint32_t`, which are not recognized by the compiler because:

- That header file (`llama-mmap.h`) doesn't include `<cstdint>` (or `<stdint.h>`), which defines `uint32_t`.
- In older compiler versions, this might have worked “by accident” because some other headers implicitly included it - but in GCC 15 (my compiler), this no longer happens cleanly

[instructlab_issues.txt](https://github.com/user-attachments/files/20660141/instructlab_issues.txt)

[instructlab_solved.txt](https://github.com/user-attachments/files/20660155/instructlab_solved.txt)

I guess patching the source like the following, adding `#include <cstdint>` near the top of `llama-mmap.h`, would fix this:

```
#pragma once
#include <cstdint>
#include <memory>
#include <vector>
```

Meanwhile, as a workaround, I'm setting a compiler wrapper with `export CXX="g++ -include cstdint"
`, which forces every compilation unit to include `<cstdint>` first.

iurygregory:
While running the installation we can hit the following error: Fail to install llama-cpp-python with error "error: ‘uint32_t’ does not name a type uint32_t read_u32() const

To workaround this issue we can use export CXX="g++ -include cstdint"

A fix was already provided to the llama.cpp ggml-org/llama.cpp@19b392d This was doing Feburary, so by bumping 0.3.8 we should be good, since it was released in March.

Resolves #3474

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,7 @@ instructlab-quantize>=0.1.0
 instructlab-schema>=0.4.2
 instructlab-sdg>=0.7.3
 instructlab-training>=0.10.1
-llama_cpp_python[server]>=0.3.6
+llama_cpp_python[server]>=0.3.8
 mlx>=0.5.1,<0.6.0; sys_platform == 'darwin' and platform_machine == 'arm64'
 numpy>=1.26.4
 openai>=1.13.3


iurygregory:
I didn't update the CHANGELOG.md, since this doesn't seem like a breaking change or feature 😅 , please let me know if it will be required.

booxter (approved):
I think it's fair to bump here. It won't be used in CI until the next week.

PR #3522, opened 2025-08-13
fix: add Granite 3.1 architecture to in-memory template mapping

alinaryan:
This change adds SupportedModelArchitectures.GRANITE3_128K to the in-memory templates list. Without this entry, models can default to the legacy Granite 7B-style chat template, which uses <|user|>/<|assistant|> tokens instead of the expected <|start_of_role|> and <|end_of_role|> format.

Adding the explicit mapping resolves the issue and ensures the correct template is used when model_arch resolves to granite_3.1.

**Checklist:**

### DIFF
--- a/src/instructlab/model/backends/common.py
+++ b/src/instructlab/model/backends/common.py
@@ -34,6 +34,12 @@ VLLM = "vllm"
 templates = [
     {
         "family": "granite",
+        "arch": SupportedModelArchitectures.GRANITE3_128K,
+        "template": granite.CHAT_TEMPLATE,
+        "special_tokens": granite.SPECIAL_TOKENS,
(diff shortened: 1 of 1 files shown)


cdoern (approved):
looks like the right fix to me, thank you!

cdoern:
running Large Job for sanity check

courtneypacheco (approved):

RobotSail (changes requested):
The provided chat template here wouldn't work. Also, we should not be using the pre-defined chat templates anymore. All new models should be using the `.models` field so that the native chat template is read off.

alinaryan:
closing in favor of editing the models list in the config instead

